PR #1, opened 2016-12-02
Initial endpoints api_manager codes

qiwzhang:


qiwzhang:
@kyessenov,  @wlu2016, could you help to review this?

contrib/endpoints/include/version
```diff
@@ -0,0 +1 @@
+1.1.0
```
kyessenov: Why is version here and not in ESP?

kyessenov:
It's very difficult to navigate a PR of this size. Here are some other observations:

- WORKSPACE should be in the top level per Bazel expectations; also makes sense to move third_party to the top level, too.

- why didn't you flatten api_manager directory out? make it contrib/endpoints/config.cc, for example.

- maybe we should move server config out of here, too?

contrib/endpoints/include/version
qiwzhang: There are two places need version:  service_control/proto.cc and cloud_trace.  
They like to send service agent with version.

qiwzhang:
1) For now, it is easier to put WORKSPACE in contrib/endpoints.  Other folder may not use bazel.
If they use bazel,  a repo can have multiple WORKSPACE too
Or we could do the move in later phase.  Don't want to introduce too big change in one PR.

2) we are in contrib/endpoints.  In the future, it may have other endpoint components beside api_manager.

3) before we have alternative to replace server_config, we have to keep it here.

PR #6, opened 2016-12-06
Copy transcoding code

qiwzhang:


contrib/endpoints/src/grpc/transcoding/bookstore.proto
```diff
@@ -0,0 +1,77 @@
+// Copyright (C) Extensible Service Proxy Authors
```
kyessenov: Change license header.

contrib/endpoints/src/grpc/transcoding/BUILD
```diff
@@ -0,0 +1,367 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+#
+package(default_visibility = ["//visibility:private"])
```
kyessenov: Make it public or same as api_manager?

lizan:
You should keep BULD file, and change `protobuf_bzl` to `protobuf_git`, it should work.

qiwzhang:
Not easy to make it bazel buildable. Need to change many includes too. We can do that later.  For now, not to add BUILD as not to break build for other codes.

PR #12, opened 2016-12-13
Add initial travis-CI build config for bazel.build.

sebastienvas:


.travis.yml
```diff
@@ -0,0 +1,40 @@
...
+  - cd ${HOME}/bazel/install
+  - wget --no-clobber "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel_${BAZEL_VERSION}-linux-x86_64.deb" 
+  - chmod +x bazel_${BAZEL_VERSION}-linux-x86_64.deb
+  - sudo dpkg -i bazel_${BAZEL_VERSION}-linux-x86_64.deb
+  - cd ${TRAVIS_BUILD_DIR}
+  - mv .bazelrc .bazelrc.orig
+  - cat .bazelrc.travis .bazelrc.orig > .bazelrc
+
+script:
+  - bazel --output_base=${HOME}/bazel/outbase test //...
```
qiwzhang: I usually run //contrib/endpoints/src/...

I am not sure if //... works.  

It did not work in my local workspace

sebastienvas: I tried it on my workstation, it worked. I think it should work, but since it is using a cache, it does not rebuild. I ll try removing the cache

PR #14, opened 2016-12-13
Start using AuthProvider audiences.

sarvaniv:
This change consumes the new audiences field that has been added
in the AuthProvider message. The change is backward compatible
and does the following:

Check if audiences is present in AuthProvider
If present, use it. Otherwise, use the audiences in
AuthRequirement


contrib/endpoints/src/api_manager/config.cc
```diff
@@ -296,12 +296,16 @@ bool Config::LoadAuthentication(ApiManagerEnvInterface *env) {
         env->LogError(error.c_str());
         continue;
       }
-      auto issuer = utils::FindOrNull(provider_id_issuer_map, provider_id);
-      if (issuer == nullptr) {
+      auto provider = utils::FindPtrOrNull(provider_id_provider_map,
```
liminw: Indentation.

sarvaniv: the if statement is deleted :-)

PR #15, opened 2016-12-16
Extract dependencies into .bzl file

lizan:
- Easier loading from repos depending on istio proxy
- Extract third_party BUILD files into it so no need to copy around


WORKSPACE
```diff
@@ -15,172 +15,15 @@
...
-    name = "servicecontrol_client_git",
-    commit = "d739d755365c6a13d0b4164506fd593f53932f5d",
-    remote = "https://github.com/cloudendpoints/service-control-client-cxx.git",
-)
-
-bind(
-    name = "servicecontrol_client",
-    actual = "@servicecontrol_client_git//:service_control_client_lib",
-)
+load("//contrib/endpoints:repositories.bzl",
```
qiwzhang: should we move this bzl file into top level?

lizan: Putting all dependencies in top level bzl files could end up with a very big one. I put those endpoints dependencies into this level, we can extract common dependencies to top level when we have another module in this repo.

PR #20, opened 2016-12-22
Do not report latency for streaming requests.

chowchow316:


qiwzhang:
It seems that this is not correct.

My idea is: here

https://github.com/istio/proxy/blob/master/contrib/endpoints/src/api_manager/context/request_context.cc#L267

if either request_streaming or response_streaming,  not to set Latency.

chowchow316:
Revert the change in proto.cc, and add change in request_context.cc. PTAL.

chowchow316:
Yes, we have `grpc_long_streaming.t`in endpoints/esp repo.

PR #21, opened 2016-12-28
Fix t test failures in esp.

chowchow316:


chowchow316:
Changed the condition to get latency info. PTAL.

contrib/endpoints/src/api_manager/context/request_context.cc
```diff
@@ -264,7 +264,11 @@ void RequestContext::FillReportRequestInfo(
 
     // Must be after response_code and method are assigned.
     FillLogMessage(info);
-    if(method() && !method()->request_streaming() && !method()->response_streaming()) {
+    bool is_streaming = false;
+    if(method() && (method()->request_streaming() || method()->response_streaming())) {
```
qiwzhang: could you do a lint format (clang-format) for this change?

chowchow316: Done.

PR #25, opened 2017-01-03
Add mixer interface in api manager.

chowchow316:


contrib/endpoints/src/api_manager/service_control/info.h
```diff
@@ -38,6 +38,9 @@ namespace service_control {
 
 // Basic information about the API call (operation).
 struct OperationInfo {
+  // The service name
+  ::google::protobuf::StringPiece service_name;
```
qiwzhang: This will not change.  We don't want to pass in for each request.  We can store it in Mixer class, pass it into in Mixer::Create()

contrib/endpoints/src/api_manager/mixer/mixer.cc
```diff
@@ -0,0 +1,225 @@
...
+
+  ::istio::mixer::v1::ReportRequest request;
+  request.set_request_index(++request_index_);
+  CovertToPb(info, request.mutable_attribute_update());
+  env_->LogInfo(std::string("Send Report: ") + request.DebugString());
+
+  std::string request_body;
+  request.SerializeToString(&request_body);
+
+  grpc_request->set_service("istio.mixer.v1.Mixer")
```
qiwzhang: call set_server("mixer_server").

contrib/endpoints/src/api_manager/mixer/mixer.h
```diff
@@ -42,14 +43,18 @@ class Mixer : public service_control::Interface {
   virtual utils::Status Close();
 
   virtual utils::Status GetStatistics(service_control::Statistics* stat) const;
+  const std::string& service_name() const { return service_name_; }
```
qiwzhang: this getter is not needed.

contrib/endpoints/src/api_manager/mixer/mixer.cc
```diff
@@ -199,13 +202,13 @@ void Mixer::Check(
 
   ::istio::mixer::v1::CheckRequest request;
   request.set_request_index(++request_index_);
-  CovertToPb(info, request.mutable_attribute_update());
+  CovertToPb(info, service_name(), request.mutable_attribute_update());
```
qiwzhang: Just use service_name_ directly.

chowchow316: Do you mean treat `service_name_` as public variable?

qiwzhang: no.  service_name_ is a private.   Mixer::Check() should be able to use it directly.

PR #28, opened 2017-01-04
t-test prototype

lizan:
Passes with #27

### DIFF
--- a/src/envoy/prototype/BUILD
+++ b/src/envoy/prototype/BUILD
@@ -31,4 +31,5 @@ cc_binary(
         "@envoy_git//:envoy-main"
     ],
     linkstatic=1,
+    visibility=["//visibility:public"],
 )
--- a/src/envoy/prototype/api_manager_env.cc
+++ b/src/envoy/prototype/api_manager_env.cc
@@ -59,6 +59,8 @@ std::unique_ptr<google::api_manager::PeriodicTimer> Env::StartPeriodicTimer(
   return timer;
 }
 
+static const LowerCaseString kApiManagerUrl("x-api-manager-url");
+
 class HTTPRequest : public Http::Message {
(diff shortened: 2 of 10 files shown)


qiwzhang:
Could you describe how it is done?   I saw the binary is using envoy-esp,  but Nginx is loaded, and every where name is using Nginx.   and nginx config is used inside t test.    If Nginx is used,  but we don't have ESP module code for Nginx here

lizan:
I tried to make the t-test can share most part with our nginx test as well, so I didn't delete the nginx specific part from `check.t`.

Instead of `$t->run`, it use `$t->run_daemon` to start an envoy-esp, config file generated in `ApiManager.pm`.

In BUILD file nginx actually pointed to envoy-esp.

kyessenov:
This is confusing. Can you replace all "nginx" occurrences with envoy?
What's the migration plan for the t-test code? Are we moving it to this repo?

lizan:
@kyessenov This is just a prototype to review.
I will update with correct words (probably "proxy" instead of envoy or nginx).

Let's discuss migration plan later.

qiwzhang:
I see.  This is good, a big step in the right direction.  A couple suggestions
1) for easy code review, could you use a separate PR to just copy the existing codes.  Then this PR just show the diff?
2) may be we can create an abstract layer to handle Nginx/Envoy stuff, such as config and running.

such as:

$t = proxy_test;
$t->set_config(
"listen_port": xxx
"service_config_file": xxx
)
$t->run()

Ideally,  at bazel proxy.bzl level, we have flag to switch proxy for Nginx or Envoy.

mandarjog:
Are `t tests` still the preferred framework? 
I thought we were moving to a go based testing fw. @lizan ?

PR #30, opened 2017-01-05
Add bazel.rc for TSAN and ASAN

lizan:


qiwzhang:
Could you change traves.yaml to run tsan and asan test too?

lizan:
I would do a follow up PR for travis.yml, there's some test doesn't pass right now.

PR #33, opened 2017-01-06
Add Echo backend

qiwzhang:


qiwzhang:
This is for src/envoy/prototype/README.md  which need a backend to manually test it.

lizan:
I thought we have a go sample backend too, no? Node is Bazel unfriendly and I would like to avoid mixing too many languages in this repo.

qiwzhang:
I am not sure if we have go backend sample.  istio/proxy mainly is c++ language.  Even go is different.

PR #34, opened 2017-01-06
Handle HEAD request.

chowchow316:
Currently, if user send HEAD request, proxy will reject it. This PR is to fix it.
Will add t-test later.

### DIFF
--- a/contrib/endpoints/src/api_manager/context/service_context.cc
+++ b/contrib/endpoints/src/api_manager/context/service_context.cc
@@ -44,7 +44,6 @@ const int kIntermediateReportInterval = 10;
 
 const char kHTTPHeadMethod[] = "HEAD";
 const char kHTTPGetMethod[] = "GET";
-
 }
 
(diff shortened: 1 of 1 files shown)


chowchow316:
Not needed any more.

contrib/endpoints/src/api_manager/context/service_context.cc
```diff
@@ -74,7 +74,12 @@ MethodCallInfo ServiceContext::GetMethodCallInfo(
   if (config_ == nullptr) {
     return MethodCallInfo();
   }
-  return config_->GetMethodCallInfo(http_method, url, query_params);
+  MethodCallInfo method_call_info =
+      config_->GetMethodCallInfo(http_method, url, query_params);
+  if (method_call_info.method_info == nullptr && http_method == "HEAD") {
```
lizan: make `"HEAD"` and `"GET"` as constant

qiwzhang: Add comment, 
// HEAD should be treated as GET unless it is specified from service_config.

PR #35, opened 2017-01-06
Add golang echo backend sample

ayj:
Add sample Go backend server so that nodejs version and dependency can
be removed. Go is likely already present on build system due to other
istio components usage.


ayj:
2188dcf removes node.js and package.json

PR #37, opened 2017-01-06
Add pre-commit script.

chowchow316:
Borrowed from endpoints/esp.


script/pre-commit
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+#
+# Copyright (C) Extensible Service Proxy Authors
```
qiwzhang: please replace copyright header with Apache one

PR #39, opened 2017-01-07
Disable cloud trace for envoy prototype.

chowchow316:
To prevent envoy crush if removing api_manager cluster from envoy-esp.conf.

### DIFF
--- a/src/envoy/prototype/server_config.pb.txt
+++ b/src/envoy/prototype/server_config.pb.txt
@@ -1,3 +1,6 @@
+cloud_tracing_config {
+  force_disable: true
+}
 mixer_options {
   mixer_server: "mixer_server"
 }


chowchow316:
I remember Lizan wanted to retain api_manager in envoy-esp.conf. Do we still need it @lizan?

lizan:
I'd prefer keeping it but either is fine since the file is for local testing only. `api_manager` is still needed if you enable auth and/or metadata.

chowchow316:
Okay, will keep it for now, we can remove it later if needed.

qiwzhang:
Let us remove "api_manager".  If auth is added, we can add it back as "auth_pkey_server", same for meta data server

lizan:
Remove once `RunHTTPRequest` is no longer hardcoded with ["api_manager"](https://github.com/istio/proxy/blob/master/src/envoy/prototype/api_manager_env.cc#L196)?

qiwzhang:
You are right. It is still hardcoded in RunHTTPRequest.  We can use a separate PR to change RunHTTPRequest to get host name out of URL, make it as cluster name.  and remove "ap_manager" or change it to whatever it is needed.

kyessenov:
I'd prefer if we did not need api_manager for the mixer client build of ESP.
That means removing the api_manager cluster definition (presently, removing it may cause a segfault depending on the api_manager features.)
We can have another build that has api_manager.

PR #40, opened 2017-01-10
Created check security rules file and a few dummy/helper functions.

liminw:
And added it to check work flow.


qiwzhang:
The file is not formated.

Files not formatted:
contrib/endpoints/src/api_manager/check_security_rules.cc contrib/endpoints/src/api_manager/check_security_rules.h contrib/endpoints/src/api_manager/check_workflow.cc

You can run 
script/pre-submit to format it

PR #46, opened 2017-01-18
Temporary code review for the rate limiting

mangchiandjjoe:
This is not an official code review. Not to be merged.
This PR includes the new rate limiting feature. Some configurations and logic were hard coded.

@qiwzhang


mangchiandjjoe:
I signed it!

contrib/endpoints/src/api_manager/quota_control.cc
```diff
@@ -0,0 +1,90 @@
...
+    continuation(
+        Status(Code::UNAUTHENTICATED,
+               "Method doesn't allow unregistered callers (callers without "
+               "established identity). Please use API Key or other form of "
+               "API consumer identity to call this API.",
+               Status::SERVICE_CONTROL));
+    return;
+  }
+
+  service_control::QuotaRequestInfo info;
```
qiwzhang: line 36 to 70 can be removed. we always want to do quota_check.

contrib/endpoints/src/api_manager/service_control/proto.cc
```diff
@@ -993,6 +1039,28 @@ Status Proto::FillReportRequest(const ReportRequestInfo& info,
...
+    const std::string& service_name, QuotaResponseInfo* quota_response_info) {
+
+  //response.operation_id()
+
+  if (response.allocate_errors().size() == 0) {
+    return Status::OK;
+  }
+
+  const QuotaError& error = response.allocate_errors().Get(0);
+  switch (error.code()) {
```
qiwzhang: do we need a switch here?

do we need QuotaResponseInfo?

I did not find places used.

contrib/endpoints/src/api_manager/context/request_context.cc
```diff
@@ -225,6 +225,39 @@ void RequestContext::FillCheckRequestInfo(
...
+    service_control::QuotaRequestInfo *info) {
+  FillOperationInfo(info);
+
+  info->method_name = this->method_call_.method_info->name();
+
+  // TODO:: JAEBONG::
+  info->labels["servicecontrol.googleapis.com/caller_ip"] =
+      request_->GetClientIP();
+  info->labels["servicecontrol.googleapis.com/referer"] = this->http_referer_;
+  info->labels["servicecontrol.googleapis.com/user"] = "integration_test_user";
```
qiwzhang: Usually don't use map in Info struct.   Info should be just a flat struct  with all fields listed.

in the service_control/proto.c   FillQuotaRequest( info,  *pb)  that will create labels based on the fields

It will be good if we can do that same for metrics.   

But I understand quota_metics may need a vector or map since it is from service config.

contrib/endpoints/src/api_manager/service_control/aggregated.cc
```diff
@@ -85,6 +88,25 @@ CheckAggregationOptions GetCheckAggregationOptions(
                                  kCheckAggregationExpirationMs);
 }
 
+// TODO:: jaebong - need to add quota configuration
+// Generate QuotaAggregationOptions
+QuotaAggregationOptions GetQuotaAggregationOptions(
+    const ServerConfig* server_config) {
+  if (server_config && server_config->has_service_control_config() &&
+      server_config->service_control_config().has_check_aggregator_config()) {
```
qiwzhang: In the server_config proto file:

https://github.com/istio/proxy/blob/master/contrib/endpoints/src/api_manager/proto/server_config.proto

you may want to add QuotaAggregatorConfig, just like CheckAggregatorConfig

contrib/endpoints/src/api_manager/service_control/info.h
```diff
@@ -85,6 +86,73 @@ struct CheckResponseInfo {
...
+  std::string metric_name;
+  std::unordered_map<std::string, std::string> labels;
+
+  std::chrono::system_clock::time_point start_time;
+  std::chrono::system_clock::time_point end_time;
+
+  int64_t int64_value;
+  bool bool_value;
+  double double_value;
+  std::string string_value;
```
qiwzhang: We don't need such complicated MetricValueInfo.  All we need is:
string name;
int64_t  cost;

contrib/endpoints/src/api_manager/service_control/info.h
```diff
@@ -85,6 +86,73 @@ struct CheckResponseInfo {
...
+
+struct QuotaRequestInfo : public OperationInfo {
+  std::string method_name;
+
+
+  std::unordered_map<std::string, std::string> labels;
+  std::vector<MetricValueInfo> quota_metrics;
+};
+
+struct QuotaResponseInfo {
```
qiwzhang: I don't think we need this.

PR #54, opened 2017-01-26
Enable ESP to invoke Firebase Security rules.

sarvaniv:


qiwzhang:
You need to use clang-format to format following files

contrib/endpoints/src/api_manager/check_security_rules.cc contrib/endpoints/src/api_manager/check_security_rules.h contrib/endpoints/src/api_manager/context/request_context.h contrib/endpoints/src/api_manager/context/service_context.cc

contrib/endpoints/src/api_manager/context/request_context.h
```diff
@@ -163,6 +175,9 @@ class RequestContext {
   // Report().
   std::string auth_authorized_party_;
 
+  // Auth Claims: This is the decoded payload of the JWT token
+  char *auth_claims_;
```
qiwzhang: We can use std::string to store it.  so you don't need to include grpc headers in order to call gpr_free

contrib/endpoints/src/api_manager/auth.h
```diff
@@ -40,6 +40,8 @@ struct UserInfo {
   // Authorized party of the incoming JWT.
   // See http://openid.net/specs/openid-connect-core-1_0.html#IDToken
   std::string authorized_party;
+  // String of claims
+  char *claims;
```
qiwzhang: do we need this one?

contrib/endpoints/src/api_manager/check_auth.cc
```diff
@@ -213,7 +213,9 @@ void AuthChecker::LookupJwtCache() {
   if (cache_hit) {
     CheckAudience(true);
   } else {
+    env_->LogInfo("Before Parse JWT");
     ParseJwt();
+    env_->LogInfo("After Parse JWT");
```
qiwzhang: remove logging?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,271 @@
...
+    final_continuation(Status::OK);
+    return;
+  }
 
- private:
-  // Helper function to send a http GET request.
-  void HttpFetch(const std::string &url, const std::string &request_body,
-                 std::function<void(Status, std::string &&)> continuation);
+  // Fetch the Release attributes.
+  auto pChecker = GetPtr();
```
qiwzhang: Google c++ style guide, not to use camel case in variable name

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,271 @@
...
-      on_done_(continuation) {}
+              continuation(status);
+            });
+}
 
-void AuthzChecker::Check() {
-  // TODO: Check service config to see if "useSecurityRules" is specified.
-  // If so, call Firebase Rules service TestRuleset API.
+std::pair<Status, std::string> AuthzChecker::ParseReleaseResponse(
+    std::string *json_str) {
```
qiwzhang: pass in const std::string&  if you are not modifying json_str

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,271 @@
...
-      env_(context_->service_context()->env()),
-      on_done_(continuation) {}
+              continuation(status);
+            });
+}
 
-void AuthzChecker::Check() {
-  // TODO: Check service config to see if "useSecurityRules" is specified.
-  // If so, call Firebase Rules service TestRuleset API.
+std::pair<Status, std::string> AuthzChecker::ParseReleaseResponse(
```
qiwzhang: can we pass out ruleset_id in the argument. e.g.
Status Parse(...,   string* ruleset_id)

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,271 @@
...
+  } else {
+    const char *result = GetStringValue(testResults->child, "state");
+    if (result == nullptr) {
+      env_->LogInfo("Result state is empty");
+      status = invalid;
+    } else if (std::string(result) != "SUCCESS") {
+      status = Status(Code::PERMISSION_DENIED,
+              std::string("Unauthorized ")
+              + context->request()->GetRequestHTTPMethod()
+              + " access to resource " + context->request()->GetRequestPath(),
```
qiwzhang: Does response have detail info we can pass back to help users?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,271 @@
...
+    return std::string("delete");
+  }
+
+  return std::string("update");
+}
+
+std::string AuthzChecker::BuildTestRequestBody(
+    std::shared_ptr<context::RequestContext> context) {
+
+  std::shared_ptr<std::ostringstream> ss(new std::ostringstream);
```
qiwzhang: not need to use new.  just allocate it from stack.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,271 @@
...
 }
 
-}  // namespace
-
 void CheckSecurityRules(std::shared_ptr<context::RequestContext> context,
                         std::function<void(Status status)> continuation) {
-  std::shared_ptr<AuthzChecker> authzChecker =
-      std::make_shared<AuthzChecker>(context, continuation);
-  authzChecker->Check();
+  std::shared_ptr<AuthzChecker> checker(new AuthzChecker(
```
qiwzhang: why don't you use make_shared?

contrib/endpoints/src/api_manager/check_security_rules.h
```diff
@@ -15,12 +15,84 @@
...
+
+  // Parses the response for the TEST API call
+  Status ParseTestResponse(std::shared_ptr<context::RequestContext> context,
+                           std::string *json_str);
+
+  // Builds the request body for the TESP API call.
+  std::string BuildTestRequestBody(
+      std::shared_ptr<context::RequestContext> context);
+
+  void AddToBody(const std::string &key,
```
qiwzhang: can some of these function static.  

if so,  they can be in the cc file, inside anonymous name space

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,271 @@
...
     return;
   }
 
-  request->set_method("POST")
+  std::string token = GetAuthToken();
+  request->set_method(method)
       .set_url(url)
-      .set_auth_token(GetAuthToken())
-      .set_header("Content-Type", "application/json")
+      .set_auth_token(token);
```
liminw: set_auth_token(GetAuthToken()). Then you don't need "token" variable.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,271 @@
...
-
 const char kFirebaseServerStaging[] =
     "https://staging-firebaserules.sandbox.googleapis.com/";
 
-// An AuthzChecker object is created for every incoming request. It does
-// authorizaiton by calling Firebase Rules service.
-class AuthzChecker : public std::enable_shared_from_this<AuthzChecker> {
- public:
-  AuthzChecker(std::shared_ptr<context::RequestContext> context,
-               std::function<void(Status status)> continuation);
```
liminw: If AuthzChecker class is used only in check_seucurity_rules.cc/h, we prefer to put the class definition in .cc file. This way, any place that "include" check_security_rules.h will only include CheckSecurityRules() function definition (not AuthzChecker class).

contrib/endpoints/src/api_manager/auth.h
sarvaniv: Yes. We do. The UserInfo is where the decoded JWT claims are stored so it can be used in the later part of the work flow. This actually is the same flow as it is used for JWT audience and issuer fields.

contrib/endpoints/src/api_manager/check_auth.cc
sarvaniv: Done

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Hmm. This code is again similar to check_auth.cc below. Let  me know if you want me to fix it in both places.

void AuthChecker::DiscoverJwksUri(const std::string &url) {
  auto pChecker = GetPtr();
  HttpFetch(url, [pChecker](S

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: No. Unfortunatley, the Test API that we use only provided with a response that looks like this: 

{
  "testResults": [
    {
      "state": "SUCCESS"
    }
  ]
}

or 

{
  "testResults": [
    {
      "state": "FAILURE"
    }
  ]
}

The response itself is a HTTP 200 OK message.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_auth.cc
```diff
@@ -213,7 +213,9 @@ void AuthChecker::LookupJwtCache() {
   if (cache_hit) {
     CheckAudience(true);
   } else {
+    env_->LogInfo("Before Parse JWT");
```
sarvaniv: Done

contrib/endpoints/src/api_manager/check_auth.cc
sarvaniv: Done

contrib/endpoints/src/api_manager/check_security_rules.h
```diff
@@ -15,12 +15,84 @@
...
+                 std::shared_ptr<std::ostringstream> ss);
```
sarvaniv: Removed

contrib/endpoints/src/api_manager/check_security_rules.h
```diff
@@ -15,12 +15,84 @@
...
+
+  void AddToBody(const std::string &key, const std::string &value,
+                 bool end, std::shared_ptr<std::ostringstream> ss);
```
sarvaniv: REmoved

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_security_rules.h
```diff
@@ -15,12 +15,84 @@
...
+
+  // Invoke the HTTP call
+  void HttpFetch(const std::string &url, const std::string &method,
+                 const std::string &request_body,
+                 std::function<void(Status, std::string &&)> continuation);
+
+
+  // Get the auth token for Firebase service
+  const std::string &GetAuthToken() {
+    return sa_token_->GetAuthToken(
```
sarvaniv: Hmm. This is similar to The Aggregator code in cloud_trace.cc where the ServiceAccountToken is being passed in. There are no null checks there. Let me know if you think this is wrong and we can fix this at book places. However, since we don't throw exceptions, the null check should be done before passing in the value. Otherwise, we will have null checks all over the code.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done

contrib/endpoints/src/api_manager/check_security_rules.h
sarvaniv: I moved the header file into the .cc file now.

contrib/endpoints/src/api_manager/check_auth.cc
sarvaniv: Done

contrib/endpoints/src/api_manager/context/request_context.h
sarvaniv: Done.

contrib/endpoints/src/api_manager/auth.h
sarvaniv: Changed it to std::string

contrib/endpoints/src/api_manager/context/request_context.cc
```diff
@@ -20,6 +20,14 @@
...
 #include <sstream>
 
+extern "C" {
+#include "grpc/grpc.h"
+#include "grpc/support/alloc.h"
+#include "grpc/support/log.h"
+#include "grpc/support/string_util.h"
+#include "grpc/support/sync.h"
+}
+
```
sarvaniv: Removed them. After changing to std::string based on Wayne's comments, I did not need them.

contrib/endpoints/src/api_manager/auth.h
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_security_rules.h
```diff
@@ -15,12 +15,84 @@
...
+               auth::ServiceAccountToken *sa_token);
+
+  // Check for Authorization success or failure
+  void Check(std::shared_ptr<context::RequestContext> context,
+             std::function<void(Status status)> continuation);
+
+ private:
+
+  // Helper method that invokes the test firebase service api.
+  void CheckAuth(std::string ruleset_id,
```
sarvaniv: Changed to FirebaseCheck

contrib/endpoints/src/api_manager/check_security_rules.h
```diff
@@ -15,12 +15,84 @@
...
+    return sa_token_->GetAuthToken(
+        auth::ServiceAccountToken::JWT_TOKEN_FOR_FIREBASE);
+  }
+
+  // Get Firebase specific operation Id based on the http Method.
+  std::string GetOperation(std::string httpMethod);
+
+  const std::string Stringify(const char *s) {
+    return std::string("\"") + s + "\"";
+  }
```
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_auth.cc
sarvaniv: Looks like the merge put them back in. I will delete them before the next commit.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+  grpc_json *json = grpc_json_parse_string_with_len(
+      const_cast<char *>(json_str.data()), json_str.length());
+
+  if (!json) {
+    return Status(Code::INVALID_ARGUMENT, kInvalidResponse);
+  }
+
+  Status status = Status::OK;
+  const char *id = GetStringValue(json, "rulesetName");
+  (*ruleset_id) = (id == nullptr) ? "" : GetStringValue(json, "rulesetName");
```
qiwzhang: last portion is id?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+              + " access to resource " + context->request()->GetRequestPath(),
+              Status::AUTH);
+    }
+  }
+
+  grpc_json_destroy(json);
+  return status;
+}
+
+std::string AuthzChecker::GetOperation(std::string httpMethod) {
```
qiwzhang: need to be a class member function?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+    return std::string("delete");
+  }
+
+  return std::string("update");
+}
+
+std::string AuthzChecker::BuildTestRequestBody(
+    std::shared_ptr<context::RequestContext> context) {
+
+  std::ostringstream ss;
```
qiwzhang: maybe use a proto, build a proto message first, then convert it to json.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+  while(openCount-- > 0) {
+    ss << "} ";
+  }
+
+  ss << "] }";
+  return ss.str();
+}
+
+void AuthzChecker::AddToBody(const std::string &key,
+                       std::ostringstream &ss) {
```
qiwzhang: no need to be a class member function

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+}
+void AuthzChecker::AddToBody(const std::string &key, const std::string &value,
+                             bool end, std::ostringstream &ss) {
+  ss << Stringify(key.c_str()) + ": " + Stringify(value.c_str());
+  if (!end) {
+    ss << ", ";
+  }
+}
+
+const std::string AuthzChecker::GetReleaseName(
```
qiwzhang: no need to be a class member function

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: How I wish I could do it! The protos are not open sourced.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: The reason I left it inside the class is because the usage of the function is only applicable to objects of this class.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Same reason as above. There is no meaning to this method outside the context of the class.

contrib/endpoints/src/api_manager/check_security_rules.cc
qiwzhang: it should be static if the function is not using any class members.

contrib/endpoints/src/api_manager/check_security_rules.cc
qiwzhang: You can copy the proto file to our ESP code base.  Still prefer to use proto to convert to JSON.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Ah. Sure. I will make them static.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: There are multiple proto files and this will come from the firebase codebase. I am not sure if we can just pick up the code and put it in our codebase. I will need to talk to Limin / Wencheng and may be even Firebase guys before doing this.

qiwzhang: If it is too much trouble to copy their proto files,   you just create your own simple proto message which only fields you used.

sarvaniv: I see what you mean. I think I an do that. It is actually a better idea than dealing with strings.

contrib/endpoints/src/api_manager/check_security_rules.h
```diff
@@ -15,8 +15,13 @@
...
 #define API_MANAGER_CHECK_SECURITY_RULES_H_
 
-#include "contrib/endpoints/include/api_manager/utils/status.h"
+#include "contrib/endpoints/include/api_manager/api_manager.h"
+#include "contrib/endpoints/src/api_manager/auth/service_account_token.h"
 #include "contrib/endpoints/src/api_manager/context/request_context.h"
+#include "contrib/endpoints/include/api_manager/utils/status.h"
+
+#include <string>
+using ::google::api_manager::utils::Status;
```
liminw: No change is needed for this file. The additional "include" and "using" here should be removed.

contrib/endpoints/src/api_manager/context/request_context.h
```diff
@@ -112,6 +112,16 @@ class RequestContext {
     last_report_time_ = tp;
   }
 
+  void set_auth_claims(std::string &claims) {
```
liminw: const std::string &

contrib/endpoints/src/api_manager/context/request_context.h
```diff
@@ -112,6 +112,16 @@ class RequestContext {
...
+    auth_claims_ = claims;
+  }
+
+  std::string &auth_claims() {
+    return auth_claims_;
+  }
+
+  const std::string GetAuthToken();
```
liminw: Why do you need GetAuthToken() here?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+using ::google::api_manager::auth::GetProperty;
+using ::google::api_manager::auth::GetStringValue;
 using ::google::api_manager::utils::Status;
+using ::google::protobuf::util::error::Code;
 
 namespace google {
 namespace api_manager {
 
-namespace {
-
```
liminw: The whole class should be an anonymous namespace. You should not remove it.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
 const char kFirebaseServerStaging[] =
     "https://staging-firebaserules.sandbox.googleapis.com/";
 
-// An AuthzChecker object is created for every incoming request. It does
-// authorizaiton by calling Firebase Rules service.
```
liminw: This comment applies to the class. Why do you remove it?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+  // Check for Authorization success or failure
+  void Check(std::shared_ptr<context::RequestContext> context,
+             std::function<void(Status status)> continuation);
 
  private:
-  // Helper function to send a http GET request.
-  void HttpFetch(const std::string &url, const std::string &request_body,
+
+  // Helper method that invokes the test firebase service api.
+  void FirebaseCheck(std::string ruleset_id,
```
liminw: const std::string &ruleset_id

contrib/endpoints/src/api_manager/context/request_context.h
liminw: const std::string &claims

contrib/endpoints/src/api_manager/context/request_context.h
```diff
@@ -112,6 +112,16 @@ class RequestContext {
     last_report_time_ = tp;
   }
 
+  void set_auth_claims(std::string &claims) {
+    auth_claims_ = claims;
+  }
+
+  std::string &auth_claims() {
```
liminw: I think this function can simply return std::string.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Added it back. Initially I had code that was creating the object in the service context but this did not work due to circular dependences. I removed it as a part of that change and forgot to add it back.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Function is deleted after adding the new protobuf to capture this.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Changed this to static

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+  // Get the auth token for Firebase service
+  const std::string &GetAuthToken() {
+    return sa_token_->GetAuthToken(
+        auth::ServiceAccountToken::JWT_TOKEN_FOR_FIREBASE);
+  }
 
-  // Pointer to access ESP running environment.
-  ApiManagerEnvInterface *env_;
+  // Get Firebase specific operation Id based on the http Method.
+  std::string GetOperation(std::string httpMethod);
```
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.h
sarvaniv: done.

contrib/endpoints/src/api_manager/context/request_context.h
sarvaniv: done.

sarvaniv: done.

contrib/endpoints/src/api_manager/context/request_context.h
sarvaniv: Why? We just use this string to create a proto object and this string can be potentially big.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+
+  const std::string Stringify(const char *s) {
```
sarvaniv: This is removed with the new protobuf change.s

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+      context->method() == nullptr || !context->method()->auth()) {
+    env_->LogDebug(
+        std::string("Autherization and JWT validation was not performed")
+        + " skipping authorization");
+    final_continuation(Status::OK);
+    return;
+  }
+
+  // Fetch the Release attributes.
+  auto pchecker = GetPtr();
```
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Not sure what you mean ...

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Made it static function.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Removed this after proto change.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,339 @@
...
+  HttpFetch(GetReleaseUrl(context), std::string("GET"), std::string(""),
```
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Made it static.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+const char kFirebaseService[] =
+    "/google.firebase.rules.v1.FirebaseRulesService";
+
+const char kFailedFirebaseReleaseFetch[] = "Failed to fetch Firebase Release";
+const char kFailedFirebaseTest[] = "Failed to execute Firebase Test";
+const char kInvalidResponse[] = "Invalid JSON response from Firebase Service";
+
 // An AuthzChecker object is created for every incoming request. It does
 // authorizaiton by calling Firebase Rules service.
 class AuthzChecker : public std::enable_shared_from_this<AuthzChecker> {
```
qiwzhang: Since the file name is called check_security_rules,  should we call the class as SecurityRuleChecker

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+  // Check for Authorization success or failure
+  void Check(std::shared_ptr<context::RequestContext> context,
+             std::function<void(Status status)> continuation);
 
  private:
-  // Helper function to send a http GET request.
-  void HttpFetch(const std::string &url, const std::string &request_body,
+  // Helper method that invokes the test firebase service api.
+  void FirebaseCheck(std::string &ruleset_id,
+                     std::shared_ptr<context::RequestContext> context,
```
qiwzhang: how about "CallTest"?

contrib/endpoints/src/api_manager/check_security_rules.cc
qiwzhang: *ruleset_id = (id) : id ? "";

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+
+  Status status = Status::OK;
+  const char *id = GetStringValue(json, "rulesetName");
+  (*ruleset_id) = (id == nullptr) ? "" : GetStringValue(json, "rulesetName");
+
+  if (ruleset_id->empty()) {
+    env_->LogError("Empty ruleset Id received from firebase service");
+    status = Status(Code::INTERNAL, kInvalidResponse);
+  } else {
+    env_->LogInfo(std::string("Received ruleset Id: ") + *ruleset_id);
```
qiwzhang: LogDebug

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
 }
 
-const std::string &AuthzChecker::GetAuthToken() {
-  // TODO: Get Auth token for accessing Firebase Rules service.
-  static std::string empty;
-  return empty;
+void AuthzChecker::FirebaseCheck(
+    std::string &ruleset_id, std::shared_ptr<context::RequestContext> context,
+    std::function<void(Status status)> continuation) {
+  auto checker = GetPtr();
```
qiwzhang: move this line to 163

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+  }
+
+  return std::string("update");
+}
+
+Status AuthzChecker::BuildTestRequestBody(
+    std::shared_ptr<context::RequestContext> context,
+    std::string &result_string) {
+  proto::TestRulesetRequest request;
+  proto::TestRulesetRequest::TestCase *test_case = request.add_test_cases();
```
qiwzhang: you can use "auto"

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+  if (httpMethod == "DELETE") {
+    return std::string("delete");
+  }
+
+  return std::string("update");
+}
+
+Status AuthzChecker::BuildTestRequestBody(
+    std::shared_ptr<context::RequestContext> context,
+    std::string &result_string) {
```
qiwzhang: Per style guide,  you should use std::string*

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+    env_->LogError(std::string("Error creating TestRulesetRequest") +
+                   status.ToString());
+  }
+
+  return status;
+}
+
+void AuthzChecker::SetProtoValue(::google::protobuf::Value &head,
+                                 std::string key,
+                                 ::google::protobuf::Value &value) {
```
qiwzhang: per style guide, the prototype should be
(const string& key, const Value& value, Value* head)

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+  }
 
-  // Request context.
-  std::shared_ptr<context::RequestContext> context_;
+  // Get Firebase specific operation Id based on the http Method.
+  static std::string GetOperation(std::string &httpMethod);
 
-  // Pointer to access ESP running environment.
-  ApiManagerEnvInterface *env_;
+  void SetProtoValue(::google::protobuf::Value &head, std::string key,
```
qiwzhang: this can be static.

I think all these 4 static functions can be out of the class.

contrib/endpoints/src/api_manager/auth/lib/auth_jwt_validator.cc
```diff
@@ -699,12 +699,20 @@ grpc_jwt_verifier_status JwtValidatorImpl::FillUserInfoAndSetExp(
 
   // Optional field.
   const grpc_json *grpc_json = grpc_jwt_claims_json(claims_);
+
+  /*for (grpc_json *cur = json->child; cur != nullptr; cur = cur->next) {
+    user_info->claims.emplace(std::string(cur->key), std::string(cur->value));
+  } */
```
sarvaniv: Removed.

contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -143,6 +146,12 @@ message ApiAuthenticationConfig {
   bool force_disable = 1;
 }
 
+// Server config for API Authorization via Firebase Rules
+message ApiCheckSecurityRulesConfig {
+  // Allows to disable the API authorization
+  bool force_disable = 1;
+}
```
sarvaniv: This can be enabled via service config. But the service config changes are not done yet. Tao is working on those changes.

contrib/endpoints/src/api_manager/context/service_context.cc
```diff
@@ -56,7 +60,11 @@ ServiceContext::ServiceContext(std::unique_ptr<ApiManagerEnvInterface> env,
       is_auth_force_disabled_(config_->server_config() &&
                               config_->server_config()
                                   ->api_authentication_config()
-                                  .force_disable()) {
+                                  .force_disable()),
+      is_check_security_rules_disabled_(config_->server_config() &&
```
sarvaniv: Currently enabled by default. Once the service config changes are done, we will add those checks too.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: This was named by Limin and so was the file name provided by her. Let me see what she thinks too :-)

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Yup. Thank you.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+void AuthzChecker::Check(
+    std::shared_ptr<context::RequestContext> context,
+    std::function<void(Status status)> final_continuation) {
   // TODO: Check service config to see if "useSecurityRules" is specified.
   // If so, call Firebase Rules service TestRuleset API.
+
+  if (!context->service_context()->RequireRulesCheck() ||
+      context->method() == nullptr || !context->method()->auth()) {
+    env_->LogDebug(
+        std::string("Autherization and JWT validation was not performed") +
```
lizan: nit: Authorization, no need of `std::string`, just use literals without `+`

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+  void FirebaseCheck(std::string &ruleset_id,
+                     std::shared_ptr<context::RequestContext> context,
+                     std::function<void(Status status)> continuation);
+
+  // Parse the respose for GET RELEASE API call
+  Status ParseReleaseResponse(const std::string &json_str,
+                              std::string *ruleset_id);
+
+  // Parses the response for the TEST API call
+  Status ParseTestResponse(std::shared_ptr<context::RequestContext> context,
```
lizan: output parameter should be pointers, per our C++ style guide

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+                           const std::string &json_str);
+
+  // Builds the request body for the TESP API call.
+  Status BuildTestRequestBody(std::shared_ptr<context::RequestContext> context,
```
lizan: ditto

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+  const grpc_json *testResults = GetProperty(json, "testResults");
+  if (testResults == nullptr) {
+    env_->LogError("TestResults are null");
+    status = invalid;
+  } else {
+    const char *result = GetStringValue(testResults->child, "state");
+    if (result == nullptr) {
+      env_->LogInfo("Result state is empty");
+      status = invalid;
+    } else if (std::string(result) != "SUCCESS") {
```
lizan: `"SUCCESS"` as constant, and make the constant as std::string.

contrib/endpoints/src/api_manager/context/request_context.h
```diff
@@ -112,6 +112,10 @@ class RequestContext {
     last_report_time_ = tp;
   }
 
+  void set_auth_claims(const std::string &claims) { auth_claims_ = claims; }
+
+  std::string &auth_claims() { return auth_claims_; }
```
lizan: const

contrib/endpoints/src/api_manager/proto/server_config.proto
qiwzhang: If that is the case,   it will be easier for you to integrate with ESP if you default to be disabled,  only enable it from server_config.

contrib/endpoints/src/api_manager/auth.h
```diff
@@ -15,6 +15,7 @@
 #ifndef API_MANAGER_AUTH_H_
 #define API_MANAGER_AUTH_H_
 
+#include <map>
```
liminw: Why do you include <map>?

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Yes. Wayne gave me this comment and I have already fixed this. in a later commit.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: same. already fixed in later commit.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: I have declared a const char[] kTestSuccess with the rest of the cont strings. Let me know if this is not what you had in mind.

contrib/endpoints/src/api_manager/context/request_context.h
sarvaniv: done.

contrib/endpoints/src/api_manager/proto/server_config.proto
sarvaniv: I don't want to disable this by default via server config and there is a reason for this. By not disabling the feature by default I am able to apply some existing auth based t tests. For example, we have checks to ensure that if auth is disabled, then we don't do any firebase rules checks too. For all those tests, since the code is not disabled by default, we are able to exercise the existing t tests without any more tests that do the same thing.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,332 @@
...
-  void Check();
+  // Check for Authorization success or failure
+  void Check(std::shared_ptr<context::RequestContext> context,
+             std::function<void(Status status)> continuation);
 
  private:
-  // Helper function to send a http GET request.
-  void HttpFetch(const std::string &url, const std::string &request_body,
+  // Helper method that invokes the test firebase service api.
+  void CallTest(std::string &ruleset_id,
```
liminw: const std::string &ruleset_id;

contrib/endpoints/src/api_manager/check_security_rules.h
```diff
@@ -15,6 +15,7 @@
 #ifndef API_MANAGER_CHECK_SECURITY_RULES_H_
 #define API_MANAGER_CHECK_SECURITY_RULES_H_
 
+#include <string>
```
liminw: remove "include".

contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -143,6 +146,15 @@ message ApiAuthenticationConfig {
   bool force_disable = 1;
 }
 
+// Server config for API Authorization via Firebase Rules
+message ApiCheckSecurityRulesConfig {
+  // Firebase server to use.
+  string firebase_server = 1;
```
liminw: When do we use this configuration?

contrib/endpoints/src/api_manager/auth.h
sarvaniv: removed.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.h
sarvaniv: done.

contrib/endpoints/src/api_manager/proto/server_config.proto
sarvaniv: We will use this configuration with t tests. I will update this in another change once I understand t tests better. For now this is only a place holder.

contrib/endpoints/src/api_manager/auth/lib/auth_jwt_validator.cc
```diff
@@ -699,12 +699,16 @@ grpc_jwt_verifier_status JwtValidatorImpl::FillUserInfoAndSetExp(
 
   // Optional field.
   const grpc_json *grpc_json = grpc_jwt_claims_json(claims_);
+
+  char *json_str =
```
lizan: `json_str` is leaked

sarvaniv: Thank you. Missed this while addressing review comments. I was freeing this in the first commit :-)

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,333 @@
...
 
+const char kFirebaseService[] =
+    "/google.firebase.rules.v1.FirebaseRulesService";
+
+const char kFailedFirebaseReleaseFetch[] = "Failed to fetch Firebase Release";
+const char kFailedFirebaseTest[] = "Failed to execute Firebase Test";
+const char kInvalidResponse[] = "Invalid JSON response from Firebase Service";
+const char kTestSuccess[] = "SUCCESS";
+
+void SetProtoValue(std::string key, ::google::protobuf::Value &value,
```
qiwzhang: const std::string& 
const Value&

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,333 @@
...
+                   ::google::protobuf::Value *head) {
+  ::google::protobuf::Struct *s = head->mutable_struct_value();
+  Map<std::string, google::protobuf::Value> *fields = s->mutable_fields();
+  (*fields)[key] = value;
+}
+
+const std::string GetReleaseName(
+    std::shared_ptr<context::RequestContext> request_context) {
```
qiwzhang: Not to use shared_ptr if you are not keep a ref_count.
use 
const context::RequestContext&

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,333 @@
...
+         request_context->service_context()->service().apis(0).version();
+}
+
+const std::string GetRulesetTestUri(const std::string &ruleset_id) {
+  return std::string(kFirebaseServerStaging) + "v1/" + ruleset_id +
+         ":test?alt=json";
+}
+
+const std::string GetReleaseUrl(
+    std::shared_ptr<context::RequestContext> request_context) {
```
qiwzhang: not to use shared_prt

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,333 @@
...
+void AuthzChecker::Check(
+    std::shared_ptr<context::RequestContext> context,
+    std::function<void(Status status)> final_continuation) {
   // TODO: Check service config to see if "useSecurityRules" is specified.
   // If so, call Firebase Rules service TestRuleset API.
+
+  if (!context->service_context()->RequireRulesCheck() ||
+      context->method() == nullptr || !context->method()->auth()) {
+    env_->LogDebug(
+        "Autherization and JWT validation was not performed"
```
qiwzhang: is logging message correct?

qiwzhang: should it be firebase ruleset validation?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,333 @@
...
+      const_cast<char *>(json_str.data()), json_str.length());
+
+  if (!json) {
+    return Status(Code::INVALID_ARGUMENT, kInvalidResponse);
+  }
+
+  Status status = Status::OK;
+  const char *id = GetStringValue(json, "rulesetName");
+  (*ruleset_id) = (id == nullptr) ? "" : id;
+
```
qiwzhang: why do we need () in (*ruleset_id)?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,333 @@
...
+  if (!status.ok()) {
+    env_->LogError(std::string("Error creating Protobuf from claims") +
+                   status.ToString());
+    return status;
+  }
+
+  SetProtoValue("token", claims, &token);
+  SetProtoValue("auth", token, &auth);
+
+  Map<std::string, google::protobuf::Value> *variables =
```
qiwzhang: use auto?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,333 @@
...
+  return status;
 }
 
 void AuthzChecker::HttpFetch(
-    const std::string &url, const std::string &request_body,
+    const std::string &url, const std::string &method,
+    const std::string &request_body,
     std::function<void(Status, std::string &&)> continuation) {
+  env_->LogInfo(std::string("Issue HTTP Request to url :") + url +
+                " method : " + method + " body: " + request_body);
```
qiwzhang: LogDebug

contrib/endpoints/src/api_manager/context/service_context.cc
qiwzhang: it is better to be disabled by default for now.  and only enable it when server_config has api_check_securty_rules_config and its firebase_server is specified. 

we can change it when service config is ready.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: The CheckSecurityRules method is the entry point and it receives a share_ptr. Do you mean that I should extract the pointer from the shared pointer? I thought it is considered bad to extract a pointer out of a smart pointer. Lets discuss more.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Changed

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: done.

contrib/endpoints/src/api_manager/check_security_rules.cc
qiwzhang: Pass a shared_ptr is bad for performance.   it needs to lock in order to increase ref_count.

If your function don't need to keep a ref_count,  pass in the object.

contrib/endpoints/src/api_manager/auth/lib/auth_jwt_validator.cc
```diff
@@ -699,12 +699,17 @@ grpc_jwt_verifier_status JwtValidatorImpl::FillUserInfoAndSetExp(
...
+      grpc_json_dump_to_string(const_cast<::grpc_json *>(grpc_json), 0);
+  user_info->claims = json_str == nullptr ? "" : json_str;
+  gpr_free(json_str);
```
lizan: you probably want to check if `json_str` is nullptr

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,335 @@
...
 
 namespace google {
 namespace api_manager {
-
 namespace {
 
 const char kFirebaseServerStaging[] =
     "https://staging-firebaserules.sandbox.googleapis.com/";
 
+const char kFirebaseService[] =
```
lizan: where are these two constants used?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,335 @@
...
+
+const std::string GetReleaseUrl(context::RequestContext &context) {
+  return GetFirebaseServer(context) + "v1/projects/" +
+         context.service_context()->project_id() + "/releases/" +
+         GetReleaseName(context);
+}
+
+std::string GetOperation(std::string &httpMethod) {
+  if (httpMethod == "POST") {
+    return std::string("create");
```
lizan: make these constant

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,335 @@
...
 
-  request->set_method("POST")
-      .set_url(url)
-      .set_auth_token(GetAuthToken())
-      .set_header("Content-Type", "application/json")
-      .set_body(request_body);
+  request->set_method(method).set_url(url).set_auth_token(GetAuthToken());
+
+  if (method != "GET") {
+    request->set_header("Content-Type", "application/json")
```
lizan: make these constant too

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -41,29 +41,36 @@ const char kFailedFirebaseTest[] = "Failed to execute Firebase Test";
...
   ::google::protobuf::Struct *s = head->mutable_struct_value();
   Map<std::string, google::protobuf::Value> *fields = s->mutable_fields();
   (*fields)[key] = value;
 }
 
-const std::string GetReleaseName(
-    std::shared_ptr<context::RequestContext> request_context) {
-  return request_context->service_context()->service_name() + ":" +
-         request_context->service_context()->service().apis(0).version();
+const std::string GetReleaseName(context::RequestContext &context) {
```
sarvaniv: I already tried that and I get the following compilation error. I think this is because the service_context() method is not marked as const. Let me know if you are OK with marking that method as const - I think it can be made into a const but I don't know why it was not marked const. 

contrib/endpoints/src/api_manager/check_security_rules.cc: In function 'const string google::api_manager::{anonymous}::GetReleaseName(const google::api_manager::context::RequestContext&)':
contrib/endpoints/src/api_manager/check_security_rules.cc:61:34: error: passing 'const google::api_manager::context::RequestContext' as 'this' argument of 'google::api_manager::context::ServiceContext* google::api_manager::context::RequestContext::service_context()' discards qualifiers [-fpermissive]
   return context.service_context()->service_name() + ":" +
                                  ^
contrib/endpoints/src/api_manager/check_security_rules.cc:62:34: error: passing 'const google::api_manager::context::RequestContext' as 'this' argument of 'google::api_manager::context::ServiceContext* google::api_manager::context::RequestContext::service_context()' discards qualifiers [-fpermissive]
          context.service_context()->service().apis(0).version();

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -139,18 +146,16 @@ void AuthzChecker::Check(
...
+    env_->LogDebug("Skipping Firebase Rules checks since it is disabled.");
     final_continuation(Status::OK);
     return;
   }
 
   // Fetch the Release attributes.
   auto checker = GetPtr();
-  HttpFetch(GetReleaseUrl(context), "GET", "",
+  HttpFetch(GetReleaseUrl(*context.get()), "GET", "",
             [context, final_continuation, checker](Status status,
```
sarvaniv: Done.

contrib/endpoints/src/api_manager/context/service_context.h
```diff
@@ -65,8 +65,12 @@ class ServiceContext {
     return !is_auth_force_disabled_ && config_->HasAuth();
   }
 
-  bool RequireRulesCheck() const {
-    return RequireAuth() && !is_check_security_rules_disabled_;
+  bool IsRulesCheckEnabled() const {
+    return RequireAuth() && config_->server_config() &&
+           !config_->server_config()
+                ->api_check_security_rules_config()
+                .firebase_server()
```
sarvaniv: Added it. For some reason this is not being done in other parts of the code. I am adding this check in one place I know.

contrib/endpoints/src/api_manager/auth/lib/auth_jwt_validator.cc
sarvaniv: Oops.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: I will delete them once i have done more testing.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: Done. Some parts of the code were using them as I have done. I did not know this was a strict rule.

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: I have removed as many string constants from code as I could. Let me know if you think there are more that need to be changed.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+const char kState[] = "state";
+const char kToken[] = "token";
+const char kAuth[] = "auth";
+const char kRequest[] = "request";
+const char kContentType[] = "Content-Type";
+const char kApplication[] = "application/json";
+
+const std::string GetFirebaseServer(const context::RequestContext &context) {
+  return context.service_context()
+      ->config()
```
qiwzhang: make a GetFireBaseServer() function in Config class.  Later when service_config is ready, get it from there. Only need to change that function.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+const char kRulesetName[] = "rulesetName";
+const char kTestResults[] = "testResults";
+const char kState[] = "state";
+const char kToken[] = "token";
+const char kAuth[] = "auth";
+const char kRequest[] = "request";
+const char kContentType[] = "Content-Type";
+const char kApplication[] = "application/json";
+
+const std::string GetFirebaseServer(const context::RequestContext &context) {
```
qiwzhang: I still don't understand the purpose of "const" in

const std::string foo()

what is this const for?

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -14,92 +14,350 @@
...
+                   const ::google::protobuf::Value &value,
+                   ::google::protobuf::Value *head) {
+  ::google::protobuf::Struct *s = head->mutable_struct_value();
+  Map<std::string, google::protobuf::Value> *fields = s->mutable_fields();
+  (*fields)[key] = value;
+}
+
+const std::string GetReleaseName(const context::RequestContext &context) {
+  return context.service_context()->service_name() + ":" +
+         context.service_context()->service().apis(0).version();
```
qiwzhang: when enabling firebase, need to make sure apis.size() == 1.

contrib/endpoints/src/api_manager/auth/lib/auth_jwt_validator.cc
```diff
@@ -699,12 +699,19 @@ grpc_jwt_verifier_status JwtValidatorImpl::FillUserInfoAndSetExp(
 
   // Optional field.
   const grpc_json *grpc_json = grpc_jwt_claims_json(claims_);
+
+  char *json_str =
+      grpc_json_dump_to_string(const_cast<::grpc_json *>(grpc_json), 0);
+  user_info->claims = json_str == nullptr ? "" : json_str;
+  if (json_str != nullptr) {
+    gpr_free(json_str);
```
lizan: move the line above in here? member std::string is default to "", so you just need to assign once.

contrib/endpoints/src/api_manager/context/service_context.h
```diff
@@ -66,7 +66,8 @@ class ServiceContext {
   }
 
   bool IsRulesCheckEnabled() const {
-    return RequireAuth() && config_->server_config() &&
+    return RequireAuth() && service().apis_size() > 0 &&
+           config_->server_config() &&
```
qiwzhang: can we call config->GEtFirebaseServer()?

contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -870,6 +870,23 @@ TEST(Config, TestCorsDisabled) {
...
+}
+)";
+
+  std::unique_ptr<Config> config =
+      Config::Create(&env, kServiceNameConfig, server_config);
+  ASSERT_TRUE(config);
+
+  auto server = config->GetFirebaseServer();
+  ASSERT_EQ(server, "https://myfirebaseserver.com/");
+}
```
qiwzhang: could you add a test case with a non-empty server_config, but no api_check_security_rules_config.

PR #55, opened 2017-01-27
Integrate with mixer client.

qiwzhang:


src/envoy/prototype/server_config.pb.txt
```diff
@@ -2,5 +2,5 @@ cloud_tracing_config {
   force_disable: true
 }
 mixer_options {
-  mixer_server: "mixer_server"
-}
\ No newline at end of file
+  mixer_server: "localhost:9091"
```
kyessenov: So is cluster definition not needed? What about DNS?

WORKSPACE
```diff
@@ -16,29 +16,42 @@
...
+    name = "protobuf_bzl",
+    # Injecting an empty BUILD file to prevent using any build target
+    build_file_content = "",
+    commit = "05090726144b6e632c50f47720ff51049bfcbef6",
+    remote = "https://github.com/google/protobuf.git",
+)
 
 load(
-    "//contrib/endpoints:repositories.bzl",
+    "@mixerclient_git//:repositories.bzl",
```
lizan: load as little as possible from `@mixerclient_git//:repositories.bzl`

BoringSSL has to be compatible with the version required from envoy, mixerclient has little thing to do with version compatibilities. same for grpc/protobuf/googletest/googleapis.

contrib/endpoints/repositories.bzl
```diff
@@ -14,314 +14,8 @@
 #
 ################################################################################
 #
-def zlib_repositories(bind=True):
```
lizan: don't remove those, per reason stated above.

src/envoy/prototype/server_config.pb.txt
qiwzhang: Yes, we don't need cluster for Mixer server.  Proxy is using grpc Channel, could  not use Envoy cluster manager. 

I think we can specify DNS to grpc channel. it should do DNS lookup.

lizan: the approach to use grpc lib doesn't require cluster definitions, full stack are handled in grpc lib.

WORKSPACE
qiwzhang: Done

contrib/endpoints/repositories.bzl
qiwzhang: Done

src/envoy/prototype/server_config.pb.txt
kyessenov: Ok, might want to explain whether GRPC does DNS resolution on every request or if it is blocking.

PR #57, opened 2017-01-28
Extract originIp and OriginHost.

qiwzhang:


src/envoy/prototype/api_manager_filter.cc
```diff
@@ -80,8 +86,32 @@ class Request : public google::api_manager::Request {
...
+    std::vector<std::string> xff_address_list =
+        StringUtil::split(header_map_.ForwardedFor()->value().c_str(), ',');
+    if (xff_address_list.empty()) {
+      return downstream_address_;
+    }
+    return xff_address_list.front();
+  }
+
+  virtual std::string GetClientHost() override {
+    LowerCaseString key("x-forwarded-host", false);
```
lizan: make this constant.

qiwzhang: Done

qiwzhang:
@kyessenov,  No,  this PR will cause service A proxy to send an "originHost" attribute with value as "Service B" to Mixer.  This PR will not add or modify any HTTP header.

kyessenov:
The server side proxy sends to localhost and the host name is the local service instance hostname.  @mandarjog wants this proxy, not client side, to send Mixer "check". In my example, it would be repeated twice: "service B, service B". It is more useful to know that it is "service A, service B" hop.

PR #58, opened 2017-01-30
Update buckets for UI.

chowchow316:


contrib/endpoints/src/api_manager/service_control/proto.cc
```diff
@@ -96,8 +96,8 @@ struct DistributionHelperOptions {
   double scale;
 };
 
-const DistributionHelperOptions time_distribution = {8, 10.0, 1e-6};
```
qiwzhang: Only update for time_distribution.

PR #60, opened 2017-02-01
Use envoy new access_log handler for sending Report.

qiwzhang:


src/envoy/prototype/api_manager_filter.cc
```diff
@@ -275,31 +283,35 @@ class Instance : public Http::StreamFilter,
...
+                   const HeaderMap* response_headers,
+                   const AccessLog::RequestInfo& request_info) override {
+    Log().debug("Called ApiManager::Instance : {}", __func__);
+    std::unique_ptr<google::api_manager::Response> response(
+        new Response(request_info));
+    request_handler_->Report(std::move(response), []() {});
   }
+  // There are two log() functions. Use this Log() to redirect to base class
+  // log().
+  spdlog::logger& Log() { return Logger::Loggable<Logger::Id::http>::log(); }
```
lizan: hmm... you don't need this since `log()` signature is different, or if you want `Log()`, then probably we don't even need to inherit `Logger::Loggable`, just get the logger object from `Logger::Registry`

qiwzhang: The function signature will not help,  compiler still complains about the ambiguity  of two log functions.

kyessenov: I think this doesn't help with the name confusion. We probably just need to separate debug log away from the filter.

lizan: so just not inherit `Logger::Loggable`, use `Logger::Registry::getLog` and completely different name, `debug_log` or `error_log`?

qiwzhang: Done, not to use Loggable base class

PR #66, opened 2017-02-03
Integrate mixer client directly with envoy.

qiwzhang:


src/envoy/mixer/BUILD
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+#
+
+package(default_visibility = ["//visibility:public"])
+
+cc_binary(
+    name = "envoy_esp",
```
lizan: not suffix _esp?

it would be nice to export a cc_library, and a cc_binary so it can be linked together.

src/envoy/mixer/http_filter.cc
```diff
@@ -0,0 +1,266 @@
...
+
+namespace Http {
+namespace Mixer {
+namespace {
+
+// Define lower case string for X-Forwarded-Host.
+const LowerCaseString kHeaderNameXFH("x-forwarded-host", false);
+
+// Convert Status::code to HTTP code
+int HttpCode(int code) {
```
lizan: where is this mapping from? isn't there a function to do this in protobuf lib?

qiwzhang: no, this mapping is copied from endpoint status.cc

src/envoy/mixer/envoy-esp.conf
```diff
@@ -0,0 +1,71 @@
...
+            },
+            "access_log": [
+              {
+                "path": "/dev/stdout"
+              }
+            ],
+            "filters": [
+              {
+                "type": "both",
+                "name": "esp",
```
kyessenov: rename to "mixerclient"

src/envoy/mixer/http_control.cc
```diff
@@ -0,0 +1,162 @@
...
+const std::string kAttrNamePeerId = "peerId";
+const std::string kAttrNameURL = "url";
+const std::string kAttrNameHttpMethod = "httpMethod";
+const std::string kAttrNameRequestSize = "requestSize";
+const std::string kAttrNameResponseSize = "responseSize";
+const std::string kAttrNameLogMessage = "logMessage";
+const std::string kAttrNameResponseTime = "responseTime";
+const std::string kAttrNameOriginIp = "originIp";
+const std::string kAttrNameOriginHost = "originHost";
+const std::string kAttrNameTargetService = "targetService";
```
kyessenov: rename to source and destination?

src/envoy/mixer/http_control.cc
```diff
@@ -0,0 +1,162 @@
...
+const std::string kProxyPeerID = "Istio/Proxy";
+const std::string kEnvNameTargetService = "TARGET_SERVICE";
+
+// Define lower case string for X-Forwarded-Host.
+const LowerCaseString kHeaderNameXFH("x-forwarded-host", false);
+
+// Define attribute names
+const std::string kAttrNamePeerId = "peerId";
+const std::string kAttrNameURL = "url";
+const std::string kAttrNameHttpMethod = "httpMethod";
```
kyessenov: add x-request-id as an attribute since it's very useful for understanding dataflow.

src/envoy/mixer/http_filter.cc
```diff
@@ -0,0 +1,266 @@
...
+ private:
+  std::shared_ptr<HttpControl> http_control_;
+  Upstream::ClusterManager& cm_;
+
+ public:
+  Config(const Json::Object& config, Server::Instance& server)
+      : cm_(server.clusterManager()) {
+    std::string mixer_server;
+    if (config.hasObject("mixer_server")) {
+      mixer_server = config.getString("mixer_server");
```
kyessenov: can we use cluster_id for mixer? it would be nice to use SDS for mixer and bypass DNS.

qiwzhang: you mean to add mixer_server in the cluster config, and in the code to extract IP from cluster_manager to create grpc channel?

src/envoy/mixer/http_control.cc
qiwzhang: I added all http headers.  It looks like these

HTTP-request-header::authority: (STRING): localhost:9090
HTTP-request-header::method: (STRING): POST
HTTP-request-header::path: (STRING): /echo?key=API-KEY
HTTP-request-header:accept: (STRING): */*
HTTP-request-header:content-length: (STRING): 11
HTTP-request-header:content-type: (STRING): application/x-www-form-urlencoded
HTTP-request-header:user-agent: (STRING): curl/7.35.0
HTTP-request-header:x-forwarded-proto: (STRING): http
HTTP-request-header:x-request-id: (STRING): fe6cd0cf-31b6-40c0-b352-3b5186daec2c
HTTP-request-param:key: (STRING): API-KEY
HTTP-response-header::status: (STRING): 200
HTTP-response-header:content-length: (STRING): 11
HTTP-response-header:content-type: (STRING): application/x-www-form-urlencoded
HTTP-response-header:date: (STRING): Sat, 04 Feb 2017 00:22:44 GMT
HTTP-response-header:server: (STRING): envoy
HTTP-response-header:x-envoy-upstream-service-time: (STRING): 0
peerId: (STRING): Istio/Proxy
requestSize: (INT64): 11
responseSize: (INT64): 11
responseTime: (INT64): 3
targetService: (STRING): abdcecee

src/envoy/mixer/envoy-esp.conf
qiwzhang: changed to mixer.

src/envoy/mixer/http_filter.cc
kyessenov: Yeah, we might want to use to envoy's DNS resolution for mixer. I don't think it's important in the short term. That would be useful for having two versions of Mixer running and pointing different proxies to different Mixers.

src/envoy/mixer/http_control.cc
qiwzhang: let us keep it as target for now.

PR #67, opened 2017-02-06
Update the auth check to use service.experimental.authorization.provider

wattli:


contrib/endpoints/src/api_manager/context/service_context.h
```diff
@@ -67,7 +67,10 @@ class ServiceContext {
 
   bool IsRulesCheckEnabled() const {
     return RequireAuth() && service().apis_size() > 0 &&
-           !config_->GetFirebaseServer().empty();
+        (!config_->GetFirebaseServer().empty() ||
```
qiwzhang: Put this logic inside config::GetFirebaseServer() function.

wattli: I think the intention is to separate the logic of server config and service config. 

liminw@, what do you think?

liminw: I agree with Wayne that the logic should be in config::GetFirebaseServer(). This function should be a wrapper, and determine whether the server is set from server config or service config.

wattli: Discussed offline with limin, the code is updated.

contrib/endpoints/src/api_manager/config.cc
```diff
@@ -515,9 +515,17 @@ void Config::SetJwksUri(const string &issuer, const string &jwks_uri,
...
+        service_.experimental().has_authorization() &&
+        service_.experimental().authorization().has_provider()) {
+      // In real environment, server_config_ should be null. And we rely on the
+      // field in service.
+      return service_.experimental().authorization().provider();
+    }
     return "";
   }
 
+  // It is the testing environment, use server config.
```
qiwzhang: By design, server_config over-ride service_config.  Please check server_config first.  

Server_config may have other fields,  only if it doesn't have firebase server field, check service_config.

wattli: Done

contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -93,13 +93,13 @@ TEST(Config, ServerConfigProto) {
 }
 
 static const char kInvalidServerConfig[] = R"(
```
wattli: This is not a new config introduced in the PR.

contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -93,13 +93,13 @@ TEST(Config, ServerConfigProto) {
...
-service_control_config {
-  type: 1
-  config {
-    cache_entries: 1020
-    flush_interval_ms: 15
+  service_control_config {
+    type: 1
```
wattli: To follow google's dev guide: https://screenshot.googleplex.com/eLDuLGnpUNd

contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -818,22 +817,22 @@ TEST(Config, TestHttpOptionsSelector) {
...
-     get: "/shelves/{shelf}"
-   }
- }
-)";
+    name: "Service.Name"
+    endpoints {
+      name: "Service.Name"
+      allow_cors: true
+    }
+    http {
```
wattli: https://screenshot.googleplex.com/eLDuLGnpUNd

contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -751,41 +750,41 @@ TEST(Config, TestHttpOptions) {
...
+          kind: "OPTIONS"
+          path: "/shelves"
+        }
+      }
+      rules {
+        selector: "CreateShelf"
+        post: "/shelves"
+      }
+      rules {
+        selector: "GetShelf"
```
wattli: https://screenshot.googleplex.com/eLDuLGnpUNd

contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -850,18 +849,18 @@ TEST(Config, TestCorsDisabled) {
...
-   }
-   rules {
-     selector: "CORS.1"
-     get: "/shelves/{shelf}"
-   }
- }
-)";
+    name: "Service.Name"
+    http {
+      rules {
```
wattli: The style is recommended by Google dev guide
https://screenshot.googleplex.com/eLDuLGnpUNd

contrib/endpoints/src/api_manager/config_test.cc
sarvaniv: Actually, the dev guide only shows the advantage of using raw strings. There is no rules on indentation and here is why: various parsers can treat whitespaces differently. Some may ignore them and some might use them as delimiters. Here is a hint about this in the document:
"
Note that indentation rules, combined with the fact that raw string literals may contain newlines, leave you with an awkward choice on how to indent the first line of your raw string block. Because text protobufs ignore whitespace, this problem can be avoided by throwing in a leading newline (ignored by the parser) in that case, but not all uses of raw strings are so forgiving."

Here is the document link:

https://g3doc.corp.google.com/devtools/library_club/g3doc/totw/64.md?cl=head

contrib/endpoints/src/api_manager/config_test.cc
sarvaniv: The document does not say anything about required indentation. Please read my previous comment.

contrib/endpoints/src/api_manager/config_test.cc
sarvaniv: I don't see any such recommendation. It is only an example of a string that can be converted to a proto.

contrib/endpoints/src/api_manager/config_test.cc
sarvaniv: The safest thing is to not have any leading or trailing white spaces. That way the strings will work with all parsers.

wattli: Actually the raw string style is kinda new to me. And I realized the style even in this file is inconsistent. So I did some search at istio/proxy repo and google3 (https://cs.corp.google.com/search/?q=%22R%5C%22(%22+lang:cc+case:yes) and tried to follow what most people did. 

I just kept the existing style as it is. No big deal :)

sarvaniv: I suppose that is because most of google code uses protobufs and they work well with whitespaces.

Issue #63, opened 2017-02-03
Publish Ubuntu Xenial binary tagged by SHA


kyessenov:
We want to use Proxy binaries without recompiling Envoy in docker xenial-based image.

sebastienvas:
It does it for PR and I need to set it up for Postsubmit as well. It creates a SHA256 that we should be able to use in other module bazel WORKSPACE.

PR #69, opened 2017-02-06
Push tar.gz to GCS

kyessenov:


kyessenov:
closes #63

src/envoy/mixer/BUILD
```diff
@@ -32,10 +32,18 @@ cc_library(
 )
 
 cc_binary(
-    name = "envoy_esp",
+    name = "envoy",
```
chowchow316: Why are we changing the binary name to envoy?

kyessenov: The binary time is preserved in the build output. ESP is no longer included.

PR #70, opened 2017-02-07
Update grpc and protobuf

lizan:


lizan:
@qiwzhang PTAL at the auth-lib code.

sebastienvas:
You need to rebase to have Jenkins test running.

PR #71, opened 2017-02-07
Adds Jenkinsfile and updates release-binary to create a SHA.

sebastienvas:


Jenkinsfile
```diff
@@ -46,7 +46,7 @@ def presubmit(gitUtils, bazel) {
       bazel.fetch('-k //...')
     }
     stage('Bazel Build') {
-      bazel.build('//...')
+      bazel.build('--strategy=CppCompile=standalone //...')
```
lizan: Make sure this is working, the BUILD file for envoy requires bazel sandbox at the time I wrote them.

Issue #84, opened 2017-02-10
Read attributes for envoy.conf and pass them to mixer


qiwzhang:
Now the mixer filer config in the Envoy config has:

"config": {
                  "mixer_server": "localhost:9091"
          }

We can extend it as:

"config": {
   "mixer_server": "localhost:9091",
   "attributes": {
              "target.service": "foo",
             "source.name": "bar"
         }
 }

Proxy just read all of them and pass them to Mixer.   When deployment manager generating this config, it can put whatever attributes they want to pass to mixer.

With this change, we can remove code of reading from environment

qiwzhang:
@kyessenov Will this work?

qiwzhang:
@chowchow316 could you help with this one?

chowchow316:
I'm starting to work on this one.

chowchow316:
#87

PR #87, opened 2017-02-13
Get attributes from envoy config.

chowchow316:
#84

### DIFF
--- a/src/envoy/mixer/http_control.cc
+++ b/src/envoy/mixer/http_control.cc
@@ -134,22 +134,23 @@ void FillRequestInfoAttributes(const AccessLog::RequestInfo& info,
 
 }  // namespace
 
-HttpControl::HttpControl(const std::string& mixer_server,
-                         const std::map<std::string, std::string>& attributes) {
+HttpControl::HttpControl(
+    const std::string& mixer_server,
+    const std::map<std::string, std::string>&& attributes) {
   ::istio::mixer_client::MixerClientOptions options;
   options.mixer_server = mixer_server;
   mixer_client_ = ::istio::mixer_client::CreateMixerClient(options);
 
-  source_service_ = attributes.at(kAttrNameTargetService);
-  target_service_ = attributes.at(kAttrNameSourceService);
+  config_attributes_ = std::move(attributes);
 }
 
 void HttpControl::FillCheckAttributes(const HeaderMap& header_map,
                                       Attributes* attr) {
   FillRequestHeaderAttributes(header_map, attr);
 
-  SetStringAttribute(kAttrNameSourceService, source_service_, attr);
-  SetStringAttribute(kAttrNameTargetService, target_service_, attr);
+  for (auto attribute : config_attributes_) {
+    SetStringAttribute(attribute.first, attribute.second, attr);
+  }
 }
 
 void HttpControl::Check(HttpRequestDataPtr request_data, HeaderMap& headers,
--- a/src/envoy/mixer/http_control.cc
+++ b/src/envoy/mixer/http_control.cc
@@ -38,12 +38,6 @@ const std::string kAttrNameResponseTime = "response.time";
 const std::string kAttrNameOriginIp = "origin.ip";
 const std::string kAttrNameOriginHost = "origin.host";
 
-const std::string kEnvNameSourceService = "SOURCE_SERVICE";
-const std::string kEnvNameTargetService = "TARGET_SERVICE";
-
-const std::string kAttrNameSourceService = "source.service";
-const std::string kAttrNameTargetService = "target.service";
-
 Attributes::Value StringValue(const std::string& str) {
   Attributes::Value v;
   v.type = Attributes::Value::STRING;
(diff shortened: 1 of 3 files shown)


src/envoy/mixer/http_control.h
```diff
@@ -37,7 +37,8 @@ typedef std::shared_ptr<HttpRequestData> HttpRequestDataPtr;
 class HttpControl final : public Logger::Loggable<Logger::Id::http> {
  public:
   // The constructor.
-  HttpControl(const std::string& mixer_server);
+  HttpControl(const std::string& mixer_server,
+              const std::map<std::string, std::string>& attributes);
```
qiwzhang: use 

std::map<std::string, std::string>&& attributes

call it by std::move()

to reduce a copy

src/envoy/mixer/http_control.cc
```diff
@@ -134,19 +134,14 @@ void FillRequestInfoAttributes(const AccessLog::RequestInfo& info,
...
 
-  auto source_service = getenv(kEnvNameSourceService.c_str());
-  if (source_service) {
-    source_service_ = source_service;
-  }
-  auto target_service = getenv(kEnvNameTargetService.c_str());
-  if (target_service) {
-    target_service_ = target_service;
-  }
+  source_service_ = attributes.at(kAttrNameTargetService);
```
qiwzhang: keep the map in the class member.

// the attributes read from the config file.
std::map<std::string, std::string>& config_attributes_;

In the Check() call,  set all of them to the attributes data structure.

src/envoy/mixer/http_control.cc
```diff
@@ -134,27 +134,23 @@ void FillRequestInfoAttributes(const AccessLog::RequestInfo& info,
...
+  config_attributes_ = std::move(attributes);
 }
 
 void HttpControl::FillCheckAttributes(const HeaderMap& header_map,
                                       Attributes* attr) {
   FillRequestHeaderAttributes(header_map, attr);
 
-  SetStringAttribute(kAttrNameSourceService, source_service_, attr);
-  SetStringAttribute(kAttrNameTargetService, target_service_, attr);
+  for (auto attribute : config_attributes_) {
```
qiwzhang: const auto &

src/envoy/mixer/http_control.cc
```diff
@@ -134,27 +134,23 @@ void FillRequestInfoAttributes(const AccessLog::RequestInfo& info,
 
 }  // namespace
 
-HttpControl::HttpControl(const std::string& mixer_server) {
+HttpControl::HttpControl(
+    const std::string& mixer_server,
+    const std::map<std::string, std::string>&& attributes) {
```
qiwzhang: remove "const"

src/envoy/mixer/http_control.cc
```diff
@@ -134,27 +134,23 @@ void FillRequestInfoAttributes(const AccessLog::RequestInfo& info,
...
 
-  auto source_service = getenv(kEnvNameSourceService.c_str());
-  if (source_service) {
-    source_service_ = source_service;
-  }
-  auto target_service = getenv(kEnvNameTargetService.c_str());
-  if (target_service) {
-    target_service_ = target_service;
-  }
+  config_attributes_ = std::move(attributes);
```
qiwzhang: do this in ":" line

src/envoy/mixer/http_control.cc
```diff
@@ -134,27 +134,23 @@ void FillRequestInfoAttributes(const AccessLog::RequestInfo& info,
...
   mixer_client_ = ::istio::mixer_client::CreateMixerClient(options);
 
-  auto source_service = getenv(kEnvNameSourceService.c_str());
-  if (source_service) {
-    source_service_ = source_service;
-  }
-  auto target_service = getenv(kEnvNameTargetService.c_str());
-  if (target_service) {
-    target_service_ = target_service;
-  }
```
qiwzhang: remove these const string definitions. such as kEnvNameTargetService, and kAttribName

src/envoy/mixer/http_filter.cc
```diff
@@ -96,7 +96,17 @@ class Config : public Logger::Loggable<Logger::Id::http> {
...
     }
 
-    http_control_ = std::make_shared<HttpControl>(mixer_server);
+    std::map<std::string, std::string> attributes;
+    if (config.hasObject("attributes")) {
+      for (const std::string& attr : config.getStringArray("attributes")) {
+        attributes[attr] = config.getString(attr);
+      }
+    } else {
+      log().warn("No attributes is specified in the config: {}", __func__);
```
qiwzhang: Not need to log a warn,  just remove the whole else {}.

kyessenov: This needs to be extended to include `upstream_cluster`. @rshriram did something similar for the fault injection filter. Basically, the problem is that if you have two upstream clusters "service_a" and "service_b" in the same http connection manager (say via weighted cluster block), you need to send two different sets of attributes about "a" and "b". Maybe make "attributes" a list of JSON objects that have "upstream_cluster" and "attributes" list.

You do not have to do this in the same PR, but send a follow-up PR that allows this customization.

kyessenov: I'm also OK with restricting mixer attributes to be same for all service versions. The drawback is that we cannot do gradual config change for the mixer attributes. The whole idea of service versions is that we can customize what set of attributes to send for a subset of the service destination pods.

kyessenov:
@rshriram question about filter priority - we want mixer filter to be the first in the chain since it decides whether to proceed or not based on the Check call. I see that you made fault filter to be first one too. Is it really necessary for the fault filter to be first?

qiwzhang:
@kyessenov This PR has nothing to do with the filter ordering.  We can address your concerns of multiple upstream in separate PR.

src/envoy/mixer/http_filter.cc
qiwzhang: Open https://github.com/istio/proxy/issues/88 for first issue.  Did not quite understand the second issue yet.  Will open a separate issue once I understand it.

PR #100, opened 2017-02-16
Add metric_cost in config.

chowchow316:
First try, not sure how to use selector, and will add QuotaRule later.

### DIFF
--- a/contrib/endpoints/src/api_manager/config.cc
+++ b/contrib/endpoints/src/api_manager/config.cc
@@ -105,13 +105,49 @@ MethodInfoImpl *Config::GetOrCreateMethodInfoImpl(const string &name,
   auto i = method_map_.find(selector);
   if (i == std::end(method_map_)) {
     auto info =
+
         MethodInfoImplPtr(new MethodInfoImpl(name, api_name, api_version));
+
     info->set_selector(selector);
     info->set_rpc_method_full_name(path);
     i = method_map_.emplace(selector, std::move(info)).first;
   }
   return i->second.get();
 }
+
+ bool Config::LoadQuotaRule(ApiManagerEnvInterface *env) {
+    for (const auto &rule : service_.quota().metric_rules()) {
+      auto method = utils::FindOrNull(method_map_,
+                                      rule.selector());
+      if (method) {
+        for (auto &metric_cost : rule.metric_costs()) {
+          (*method)->add_metric_cost(metric_cost.first,
+                                  metric_cost.second);
+        }
+      } else {
+        env->LogDebug("Method not Found.");
+      }
+    }
+
+      for (const auto &rule : service_.quota().rules()) {
+        auto method = utils::FindOrNull(method_map_, rule.selector());
+        if (method) {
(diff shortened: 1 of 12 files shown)


contrib/endpoints/src/api_manager/config.cc
```diff
@@ -398,6 +398,14 @@ bool Config::LoadService(ApiManagerEnvInterface *env,
     string tf;
     ::google::protobuf::TextFormat::PrintToString(service_, &tf);
     env->LogDebug(tf.c_str());
+
+    auto quota_info = service_.quota();
+    for (auto &metric_rule : quota_info.metric_rules()) {
+      for (auto &metric_cost : metric_rule.metric_costs()) {
```
qiwzhang: not here. you need to create a separate function like

bool Config::LoadQuotaRule(ApiManagerEnvInterface *env) {
  for (auto &rule : service_.quota().metric_rules()) {
    auto method = utils::FindOrNull(method_map_, rule.selector());
    if (method) {
             for (auto &metric_cost : rule.metric_costs()) {
                    method->add_quota_metric(metrick_cost.first, metric_cost.second;
            }
    } else {
       log error
    }
  }
  

    for (auto &rule : service_.quota().quota_rules()) {
          auto method = utils::FindOrNull(method_map_, rule.selector());
           if (method) {
               for (const auto& rule: rule.groups) {
                         // find out the group that rule.group point
                         for (const auto& limit : group.limits()) {
                              method->add_metric(limit.metric, rule.cost);

contrib/endpoints/src/api_manager/config.h
```diff
@@ -129,6 +129,8 @@ class Config {
   // jwksUri for the issuer. It is set to true if jwksUri is not provided in
   // service config and we have not tried openId discovery to fetch jwksUri.
   std::map<std::string, std::pair<std::string, bool>> issuer_jwks_uri_map_;
+
+  std::map<std::string, int64_t> metric_cost_;
```
qiwzhang: no here

contrib/endpoints/src/api_manager/method_impl.h
```diff
@@ -62,6 +62,10 @@ class MethodInfoImpl : public MethodInfo {
 
   const std::string &backend_address() const { return backend_address_; }
 
+  const std::map<std::string, int64_t> && metric_cost() {
```
qiwzhang: const std::map<std::string, int64_t>& metric_cost_map() const { return metric_cost_map_;  }

contrib/endpoints/src/api_manager/method_impl.h
```diff
@@ -96,6 +100,10 @@ class MethodInfoImpl : public MethodInfo {
 
   void set_selector(const std::string &selector) { selector_ = selector; }
 
+  void set_metric_cost(const std::map<std::string, int64_t> &&metric_cost) {
```
qiwzhang: void add_metric_cost(string metric_name, int64 cost) { }

chowchow316:
Duplicated.

PR #101, opened 2017-02-16
Extract quota config from service config.

chowchow316:


contrib/endpoints/src/api_manager/config.cc
```diff
@@ -113,6 +113,36 @@ MethodInfoImpl *Config::GetOrCreateMethodInfoImpl(const string &name,
...
 
+bool Config::LoadQuotaRule(ApiManagerEnvInterface *env) {
+  for (const auto &rule : service_.quota().metric_rules()) {
+    auto method = utils::FindOrNull(method_map_, rule.selector());
+    if (method) {
+      for (auto &metric_cost : rule.metric_costs()) {
+        (*method)->add_metric_cost(metric_cost.first, metric_cost.second);
+      }
+    } else {
+      env->LogDebug("Method not Found.");
```
qiwzhang: LogError("Metric rule with selector %s is mismatched")

contrib/endpoints/src/api_manager/config.cc
```diff
@@ -113,6 +113,36 @@ MethodInfoImpl *Config::GetOrCreateMethodInfoImpl(const string &name,
...
+    }
+  }
+
+  for (const auto &rule : service_.quota().rules()) {
```
qiwzhang: quota().grouprules()

contrib/endpoints/src/api_manager/config.cc
```diff
@@ -113,6 +113,36 @@ MethodInfoImpl *Config::GetOrCreateMethodInfoImpl(const string &name,
...
+    auto method = utils::FindOrNull(method_map_, rule.selector());
+    if (method) {
+      for (const auto &group_name : rule.groups()) {
+        for (const auto &group : service_.quota().groups()) {
+          if (group.name() == group_name.group()) {
```
qiwzhang: we can speed up this code by creating a map of group_name to group proto.

contrib/endpoints/src/api_manager/method_impl.h
```diff
@@ -175,6 +183,9 @@ class MethodInfoImpl : public MethodInfo {
 
   // Whether the response is streaming or not.
   bool response_streaming_;
+
+  // map of metric and its cost
+  std::map<std::string, int64_t> metric_cost_map_;
```
mangchiandjjoe: one minor thing, could consider using std::unorderded_map or std::vector<std::pair<std::string, int>>. 
I recommend using vector, since api_manager would not call find, it just iterates the list.

qiwzhang: we can use std::vector in methodinfo.

contrib/endpoints/src/api_manager/config.cc
```diff
@@ -113,6 +113,36 @@ MethodInfoImpl *Config::GetOrCreateMethodInfoImpl(const string &name,
   return i->second.get();
 }
 
+bool Config::LoadQuotaRule(ApiManagerEnvInterface *env) {
```
mangchiandjjoe: https://cs.corp.google.com/piper///depot/google3/tech/internal/env/framework/wrappers/chemist/chemist_wrapper.cc?rcl=145718473&l=725

According to the source code
- quota_metrics conversion from rules and metric_rules are different. 
- if there is rules, then metric_rules will be ignored. 
- cost field can be empty, it should be 1

mangchiandjjoe: rules {
    selector: "apiserving.servicecontrol.testing.v1.QuotaTest.User"
    groups {
      group: "ReadGroupUser"
      cost: 3
    }
  }

will be converted to

quota_metrics: {  # came from rules setting
    name: "serviceruntime.googleapis.com/api/consumer/quota_used_count"
    metric_values: {
	  Labels: {
	      key: "/quota_name"
	      value: "ReadGroupUser"
	    }
	  value: {
	      Int64_value: 3
	    }
      }
  }



metric_rules {
    selector: "apiserving.servicecontrol.testing.v1.QuotaTest.User"
    metric_costs {
      key: "chemisttest.googleapis.com/operation/read_book"
      value: 100
    }
  }

will be converted to

quota_metrics: { # came from metric_rules
    name: "chemisttest.googleapis.com/operation/read_book"
    metric_values: {
	value: {
	    int64_value: 100
        }
  }

contrib/endpoints/src/api_manager/config.cc
```diff
@@ -113,6 +113,36 @@ MethodInfoImpl *Config::GetOrCreateMethodInfoImpl(const string &name,
...
+      }
+    } else {
+      env->LogDebug("Method not Found.");
+    }
+  }
+
+  for (const auto &rule : service_.quota().rules()) {
+    auto method = utils::FindOrNull(method_map_, rule.selector());
+    if (method) {
+      for (const auto &group_name : rule.groups()) {
```
qiwzhang: call it group_map

contrib/endpoints/src/api_manager/config.cc
chowchow316: There is no grouprules field in quota: https://github.com/cloudendpoints/service-control-client-cxx/blob/rate_limiting/proto/google/api/quota.proto

contrib/endpoints/src/api_manager/method_impl.h
chowchow316: Just wondering why should we use vector instead of map here?

qiwzhang: if you don't need to find by key, it is better to use vector.

contrib/endpoints/src/api_manager/config.cc
qiwzhang: sorry, it is "rules" then

contrib/endpoints/src/api_manager/config.cc
qiwzhang: we can add this check in a separate CL

for a selector,  if there is a group_rule , don't use metric_rule

chowchow316: Thanks for your info, and let me make it clear: if there is rules, we do not care about anything in metric_rules, right? And we are always saving the quota info into a <metric, cost> pair, what should we use for the metric names for rules & metric_rules? It seems they have different structures.

mangchiandjjoe: We need to know the rule is came from rules or metric_rules. We can use the same data structure but additional information is required.

contrib/endpoints/src/api_manager/method_impl.h
mangchiandjjoe: std::map using balanced tree to maintain order and lookup. if we don't need an item order, unordered_map is more efficient. and if we don't lookup, vector is better for iteration.

qiwzhang:
Just talked to Chengliang,  for ESP, we don't need to worry about group rules in "rules" field.  we only need to add metric_rules()

contrib/endpoints/src/api_manager/config.cc
chowchow316: `env->LogError()` only accepts one input variable.

chowchow316:
Will add unit test in another PR.

contrib/endpoints/src/api_manager/config.cc
```diff
@@ -113,6 +113,21 @@ MethodInfoImpl *Config::GetOrCreateMethodInfoImpl(const string &name,
...
 
+bool Config::LoadQuotaRule(ApiManagerEnvInterface *env) {
+  for (const auto &rule : service_.quota().metric_rules()) {
+    auto method = utils::FindOrNull(method_map_, rule.selector());
+    if (method) {
+      for (auto &metric_cost : rule.metric_costs()) {
+        (*method)->add_metric_cost(metric_cost.first, metric_cost.second);
+      }
+    } else {
+      env->LogError("Metric rule with selector is mismatched.");
```
qiwzhang: could you log rule.selector() too?

qiwzhang: in this error case, return false.

contrib/endpoints/src/api_manager/config.cc
```diff
@@ -113,6 +113,21 @@ MethodInfoImpl *Config::GetOrCreateMethodInfoImpl(const string &name,
...
+    }
+  }
+
+  return true;
+}
+
```
qiwzhang: you need to call this function in ::Create() function.  it can be the last one.

contrib/endpoints/src/api_manager/config.cc
chowchow316: Done.

contrib/endpoints/src/api_manager/config.cc
chowchow316: Done.

PR #108, opened 2017-02-18
Add unit test for quota config.

chowchow316:


contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -870,6 +870,86 @@ TEST(Config, TestCorsDisabled) {
   ASSERT_EQ(nullptr, method1);
 }
 
+TEST(Config, TestInvalidMetricRules) {
+  MockApiManagerEnvironmentWithLog env;
+  static const char config_text[] = R"(
```
qiwzhang: Please add comment to state why this is invalid
// There is no http.rule or api.method to match the selector.

contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -870,6 +870,86 @@ TEST(Config, TestCorsDisabled) {
...
+  std::unique_ptr<Config> config = Config::Create(&env, config_text, "");
+  ASSERT_TRUE(config);
+
+  auto method1 = config->GetMethodInfo("GET", "/shelves");
+  ASSERT_NE(nullptr, method1);
+
+  auto metric_cost_vector = method1->metric_cost_vector();
+  ASSERT_EQ(2, metric_cost_vector.size());
+  ASSERT_EQ("test.googleapis.com/operation/request",
+            metric_cost_vector[0].first);
```
qiwzhang: the order is un-determined. this test may fail in some platforms.

contrib/endpoints/src/api_manager/config_test.cc
chowchow316: Done.

contrib/endpoints/src/api_manager/config_test.cc
chowchow316: Removed the second <key, value> pair.

contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -932,13 +929,10 @@ quota {
   ASSERT_NE(nullptr, method1);
 
   auto metric_cost_vector = method1->metric_cost_vector();
-  ASSERT_EQ(2, metric_cost_vector.size());
-  ASSERT_EQ("test.googleapis.com/operation/request",
-            metric_cost_vector[0].first);
-  ASSERT_EQ(10, metric_cost_vector[0].second);
+  ASSERT_EQ(1, metric_cost_vector.size());
```
qiwzhang: but we DO want to test 2 metrics case.   You can either convert the vector to a "set", then compare the "set".  or sort the vector first,

contrib/endpoints/src/api_manager/config_test.cc
```diff
@@ -870,6 +870,80 @@ TEST(Config, TestCorsDisabled) {
...
+         value: 200
+      }
+   }
+}
+)";
+
+  std::unique_ptr<Config> config = Config::Create(&env, config_text, "");
+  ASSERT_TRUE(config);
+
+  auto method1 = config->GetMethodInfo("GET", "/shelves");
```
rshriram: To keep the code consistent with Envoy, I suggest avoiding over usage of auto method. In the Envoy codebase, we use explicit types most of the time

PR #110, opened 2017-02-21
Send StatusCode Attributes to Mixer.

chowchow316:


src/envoy/mixer/http_control.cc
```diff
@@ -37,6 +37,7 @@ const std::string kAttrNameResponseSize = "response.size";
 const std::string kAttrNameResponseTime = "response.time";
 const std::string kAttrNameOriginIp = "origin.ip";
 const std::string kAttrNameOriginHost = "origin.host";
+const std::string kStatusCode = "status.code";
```
qiwzhang: maybe "response.status.code"?

src/envoy/mixer/http_control.cc
```diff
@@ -124,6 +125,12 @@ void FillRequestInfoAttributes(const AccessLog::RequestInfo& info,
     attr->attributes[kAttrNameResponseTime] =
         Int64Value(info.duration().count());
   }
+  if (info.responseCode().valid()) {
+    attr->attributes[kStatusCode] =
+        StringValue(std::to_string(info.responseCode().value()));
+  } else {
+    attr->attributes[kStatusCode] = StringValue("500");
```
qiwzhang: Not sure if responseCode() already covers ones rejected by Envoy, such as rate limiting.

If no, we may need to check getResponseFlag.

chowchow316: `responseCode()` may not cover all the cases (it is `Optional`). Do we need to check all the possible [ResponseFlags](https://github.com/lyft/envoy/blob/453b51138c5e23e2072297b6a2cb85d74c4d39c7/include/envoy/http/access_log.h#L13)? And what should be the corresponding status code?

qiwzhang: I just want to understand if a request is rejected by our "Check" call,  what is its responseCode().

If its responseCode() is not valid,  then, we have to set proper status.code for it.  

Actually, when Check() fails,  we did not set proper ResponseFlag.  So we may have to use our own flag in Instance class to remember it.

PR #113, opened 2017-02-22
If QuotaControl service is not available, return utils::Status::OK

mangchiandjjoe:
@qiwzhang

### DIFF
--- a/contrib/endpoints/src/api_manager/quota_control.cc
+++ b/contrib/endpoints/src/api_manager/quota_control.cc
@@ -46,7 +46,10 @@ void QuotaControl(std::shared_ptr<context::RequestContext> context,
         TRACE(trace_span) << "Quota service control request returned with "
                           << "status " << status.ToString();
 
-        continuation(status);
+        // if the quota control service API is not available, then pass client
+        // requests
+        continuation((status.code() == Code::UNAVAILABLE) ? utils::Status::OK
+                                                          : status);
       });
 }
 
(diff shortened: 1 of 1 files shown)


contrib/endpoints/src/api_manager/quota_control.cc
```diff
@@ -46,7 +46,10 @@ void QuotaControl(std::shared_ptr<context::RequestContext> context,
         TRACE(trace_span) << "Quota service control request returned with "
                           << "status " << status.ToString();
 
-        continuation(status);
+        // if the quota control service API is not available, then pass client
+        // requests
+        continuation((status.code() == Code::UNAVAILABLE) ? utils::Status::OK
```
qiwzhang: Should we do this in ConvertQuotaResponse() function in proto.cc?

mangchiandjjoe: In Aggregated::Quota, we call Proto::ConvertAllocateQuotaResponse when the response of the Call is OK.

This is to handle Code::UNAVAILABLE from Aggregated::Call

mangchiandjjoe: Aggregated::Call returns Code::UNAVAILABLE when the backend api returns non 200 OK response.

contrib/endpoints/src/api_manager/quota_control.cc
```diff
@@ -46,7 +46,10 @@ void QuotaControl(std::shared_ptr<context::RequestContext> context,
         TRACE(trace_span) << "Quota service control request returned with "
                           << "status " << status.ToString();
 
-        continuation(status);
+        // if the quota control service API is not available, then pass client
+        // requests
```
qiwzhang: I see. could you update the comment
// quota control is using "failed open" policy.  If the server is not available, allow the request to go.

PR #115, opened 2017-02-22
Add send_attribute filter.

qiwzhang:
This filter can send some attributes to the next hop istio/proxy.  It can be used as outbound istio/proxy.


kyessenov:
I feel like this is a good candidate to upstream to envoy. We can add "request_headers_to_add" that appends requests on the client proxy in the decoding phase output, similarly how there is "response_headers_to_add" for the encoding phase: https://lyft.github.io/envoy/docs/configuration/http_conn_man/route_config/route_config.html#config-http-conn-man-route-table

src/envoy/mixer/envoy.conf.template
```diff
@@ -35,7 +35,62 @@
                 "type": "both",
                 "name": "mixer",
                 "config": {
-                  "mixer_server": "${MIXER_SERVER}"
+                  "mixer_server": "${MIXER_SERVER}",
```
rshriram: If this is environment variable, it should absolutely be killed! ESP in a JSON file.

src/envoy/mixer/envoy.conf.template
```diff
@@ -65,6 +120,17 @@
             "url": "tcp://${BACKEND}"
           }
         ]
```
rshriram: Same to the backend bar above.

src/envoy/mixer/envoy.conf.template
```diff
@@ -65,6 +120,17 @@
...
+      {
+        "name": "service2",
+        "connect_timeout_ms": 5000,
+        "type": "strict_dns",
+        "lb_type": "round_robin",
+        "hosts": [
+          {
+            "url": "tcp://localhost:9090"
+          }
+        ]
```
rshriram: Finally, I suggest using the JSON schema approach that Envoy uses. It's relatively easy to extend and cuts down a ton of unnecessary validation code.

src/envoy/mixer/http_control.cc
```diff
@@ -13,8 +13,12 @@
  */
 
 #include "src/envoy/mixer/http_control.h"
+
+#include "common/common/base64.h"
 #include "common/common/utility.h"
 #include "common/http/utility.h"
+#include "src/envoy/mixer/string_map.pb.h"
+#include "src/envoy/mixer/utils.h"
```
rshriram: Nit. Please sort includes and group them.

src/envoy/mixer/http_control.cc
```diff
@@ -144,8 +148,19 @@ HttpControl::HttpControl(const std::string& mixer_server,
   mixer_client_ = ::istio::mixer_client::CreateMixerClient(options);
 }
 
-void HttpControl::FillCheckAttributes(const HeaderMap& header_map,
-                                      Attributes* attr) {
+void HttpControl::FillCheckAttributes(HeaderMap& header_map, Attributes* attr) {
+  // Extract attributes from x-istio-attributes header
```
rshriram: What does Fill stand for here?

src/envoy/mixer/http_control.cc
```diff
@@ -144,8 +148,19 @@ HttpControl::HttpControl(const std::string& mixer_server,
...
+  const HeaderEntry* entry = header_map.get(Utils::kHeaderNameIstioAttributes);
+  if (entry) {
```
rshriram: Suggest renaming the kHeaderName... to IstioAttributesHeader

src/envoy/mixer/http_control.cc
```diff
@@ -144,8 +148,19 @@ HttpControl::HttpControl(const std::string& mixer_server,
...
+    ::istio::proxy::mixer::StringMap pb;
+    std::string str(entry->value().c_str(), entry->value().size());
+    pb.ParseFromString(Base64::decode(str));
+    for (const auto& it : pb.map()) {
```
rshriram: Is it possible to eliminate the auto here? There are very few places in Envoy code where auto is used. It kills readability most of the time.

src/envoy/mixer/send_attribute_filter.cc
```diff
@@ -0,0 +1,116 @@
+/* Copyright 2016 Google Inc. All Rights Reserved.
```
rshriram: Please change copy right to 2017 Istio Authors. That's what we are doing in manager.

src/envoy/mixer/send_attribute_filter.cc
```diff
@@ -0,0 +1,116 @@
...
+#include "envoy/server/instance.h"
+#include "server/config/network/http_connection_manager.h"
+#include "src/envoy/mixer/utils.h"
+
+namespace Http {
+namespace AddHeader {
+namespace {
+
+// The Json object name to specify attributes to pass to
+// next hop istio proxy.
```
rshriram: reword. Use upstream instead of next hop

src/envoy/mixer/send_attribute_filter.cc
```diff
@@ -0,0 +1,116 @@
...
+          Base64::encode(serialized_str.c_str(), serialized_str.size());
+    }
+  }
+
+  const std::string& attributes() const { return attributes_; }
+};
+
+typedef std::shared_ptr<Config> ConfigPtr;
+
+class Instance : public Http::StreamDecoderFilter {
```
rshriram: Rename class to ForwardAttributeFilter?

src/envoy/mixer/send_attribute_filter.cc
```diff
@@ -0,0 +1,116 @@
...
+// The Json object name to specify attributes to pass to
+// next hop istio proxy.
+const std::string kJsonNameAttributes("attributes");
+
+}  // namespace
+
+class Config : public Logger::Loggable<Logger::Id::http> {
+ private:
+  Upstream::ClusterManager& cm_;
+  std::string attributes_;
```
rshriram: Why do you need the cluster manager here? It's not even being used.

src/envoy/mixer/utils.cc
```diff
@@ -0,0 +1,49 @@
+/*
```
rshriram: Missing copyright header

src/envoy/mixer/string_map.proto
```diff
@@ -0,0 +1,22 @@
+// Copyright 2016 Google Inc.
```
rshriram: Please change copyright header

src/envoy/mixer/utils.h
```diff
@@ -0,0 +1,39 @@
+/* Copyright 2016 Google Inc. All Rights Reserved.
+ *
```
rshriram: Change copyright

rshriram:
With regard to adding to envoy, it might not serve any useful purpose except ours. Unless we come up with more General use cases.

kyessenov:
Yeah, it seems that nginx's "proxy_set_header" is mostly used for host, x-real-ip, x-forwarder-* header, which have special configs in envoy. So this is pretty specific to us.

src/envoy/mixer/README.md
```diff
@@ -49,3 +49,51 @@ This Proxy will use Envoy and talk to Mixer server.
...
+### *mixer* filter:
+
+This filter will intercept all HTTP requests and call Mixer. Here is its config:
+
+```
+   "filters": [
+      "type": "both",
+      "name": "mixer",
+      "config": {
+         "mixer_server": "${MIXER_SERVER}",
```
kyessenov: It's just a template value in a standalone prototype proxy + mixer. The reason for this PR was that I wanted to have explicit values in the envoy config rather than some secret environment variables that the filter picks up.

rshriram: So it gets overriden during config gen? Then it sounds fine. In that case I would say use some non colliding variables like __Mixer__server__

src/envoy/mixer/http_control.cc
qiwzhang: I have removed some auto.   

But in this line 158, the type is too long for map iterator.  it is better to use auto for readability.

src/envoy/mixer/send_attribute_filter.cc
qiwzhang: Done

src/envoy/mixer/send_attribute_filter.cc
qiwzhang: Removed

src/envoy/mixer/utils.cc
qiwzhang: Done

src/envoy/mixer/send_attribute_filter.cc
qiwzhang: Done

src/envoy/mixer/string_map.proto
qiwzhang: Done

src/envoy/mixer/utils.h
qiwzhang: Done

src/envoy/mixer/send_attribute_filter.cc
qiwzhang: Done

src/envoy/mixer/http_control.cc
qiwzhang: Done

src/envoy/mixer/http_control.cc
qiwzhang: stand for "add".   similar to fill protobuf.   to add fields in a protobuf.

src/envoy/mixer/http_control.cc
qiwzhang: Done

src/envoy/mixer/envoy.conf.template
qiwzhang: this is the actual config in JSON. not a JSON schema template.  the reason it is called "template" since it has ${MIXER_SERVER} variables,  config gen will replace them with real mixer_server address.

src/envoy/mixer/README.md
```diff
@@ -49,3 +49,51 @@ This Proxy will use Envoy and talk to Mixer server.
   curl http://localhost:9090/echo -d "hello world"
 ```
 
+## How to configurate HTTP filters
+
+This module has two HTTP filters:
+1. mixer filter: intercept all HTTP requests, call the mixer.
+2. send_attribute filter: Send some attributes to next hop istio/proxy with mixer filter.
```
qiwzhang: Done. Reworded.

src/envoy/mixer/README.md
qiwzhang: I will leave it as ${MIXER_SERFVER}

src/envoy/mixer/envoy.conf.template
qiwzhang: src/envoy/mixer/start_env script or config-gen script will replace these variables with real values.

rshriram:
@kyessenov we should really get rid of this script and possibly use either go templates (not ideal) or our existing go based config gen to generate the Envoy filters. the args can be provided as params to the manager.

PR #116, opened 2017-02-22
Fail request if api_key is not valid.

chowchow316:
Will add t test later in esp repo.

### DIFF
--- a/contrib/endpoints/src/api_manager/service_control/aggregated.cc
+++ b/contrib/endpoints/src/api_manager/service_control/aggregated.cc
@@ -193,10 +193,10 @@ Status Aggregated::Init() {
   options.periodic_timer = [this](int interval_ms,
                                   std::function<void()> callback)
       -> std::unique_ptr<::google::service_control_client::PeriodicTimer> {
-        return std::unique_ptr<::google::service_control_client::PeriodicTimer>(
-            new ApiManagerPeriodicTimer(env_->StartPeriodicTimer(
-                std::chrono::milliseconds(interval_ms), callback)));
-      };
+    return std::unique_ptr<::google::service_control_client::PeriodicTimer>(
+        new ApiManagerPeriodicTimer(env_->StartPeriodicTimer(
+            std::chrono::milliseconds(interval_ms), callback)));
+  };
   client_ = ::google::service_control_client::CreateServiceControlClient(
       service_->name(), service_->id(), options);
   return Status::OK;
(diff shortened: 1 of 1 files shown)


contrib/endpoints/src/api_manager/service_control/aggregated.cc
```diff
@@ -289,7 +289,10 @@ void Aggregated::Check(
           *response, service_control_proto_.service_name(), &response_info);
       // If allow_unregistered_calls is true, it is always OK to proceed.
       if (allow_unregistered_calls) {
-        on_done(Status::OK, response_info);
+        if (response_info.is_api_key_valid &&
```
qiwzhang: please combine two "if" into one

contrib/endpoints/src/api_manager/service_control/aggregated.cc
```diff
@@ -287,8 +287,10 @@ void Aggregated::Check(
     if (status.ok()) {
       Status status = Proto::ConvertCheckResponse(
           *response, service_control_proto_.service_name(), &response_info);
-      // If allow_unregistered_calls is true, it is always OK to proceed.
-      if (allow_unregistered_calls) {
+      // When allow_unregistered_calls is true, it is always OK to proceed if
```
qiwzhang: If server replied with either invalid api_key or not actived service, the request is failed even allow_unregistered_calls is true. 
Most likely, users provide a wrong api key.  By failing the request, the users will be notified with the error and have chance to correct it.
Otherwise, the Report call will fail.  It is very hard to notice and debug the Report failure.

qiwzhang: please add above comment

contrib/endpoints/src/api_manager/service_control/aggregated.cc
```diff
@@ -287,8 +287,12 @@ void Aggregated::Check(
     if (status.ok()) {
       Status status = Proto::ConvertCheckResponse(
           *response, service_control_proto_.service_name(), &response_info);
-      // When allow_unregistered_calls is true, it is always OK to proceed if
-      // the api_key is valid and service is activated.
+      // If server replied with either invalid api_key or not actived service,
```
qiwzhang: activated

contrib/endpoints/src/api_manager/service_control/aggregated.cc
```diff
@@ -287,8 +287,12 @@ void Aggregated::Check(
...
+      // the request is failed even allow_unregistered_calls is true. Most
```
qiwzhang: s/failed/rejected/

PR #129, opened 2017-02-24
Send headers as string map.

qiwzhang:
Changes are:
* Add request.time and response.time as TimeStamp
* Rename response.time to response.latency as Duration.
* Change request.headers and response.headers as StringMap.

### DIFF
--- a/src/envoy/mixer/http_control.cc
+++ b/src/envoy/mixer/http_control.cc
@@ -33,17 +33,21 @@ namespace {
 // Define lower case string for X-Forwarded-Host.
 const LowerCaseString kHeaderNameXFH("x-forwarded-host", false);
 
-const std::string kRequestHeaderPrefix = "request.headers.";
-const std::string kResponseHeaderPrefix = "response.headers.";
-
 // Define attribute names
-const std::string kAttrNameRequestPath = "request.path";
-const std::string kAttrNameRequestSize = "request.size";
-const std::string kAttrNameResponseSize = "response.size";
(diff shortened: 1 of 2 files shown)


src/envoy/mixer/http_control.cc
```diff
@@ -91,47 +118,50 @@ std::string GetLastForwardedHost(const HeaderMap& header_map) {
...
-                     attr);
-  SetStringAttribute(kAttrNameOriginIp, GetFirstForwardedFor(header_map), attr);
-  SetStringAttribute(kAttrNameOriginHost, GetLastForwardedHost(header_map),
-                     attr);
+void FillRequestHeaderAttributes(const HeaderMap& header_map,
+                                 Attributes* attr) {
+  SetStringAttribute(kRequestPath, header_map.Path()->value().c_str(), attr);
+  SetStringAttribute(kRequestHost, header_map.Host()->value().c_str(), attr);
+  attr->attributes[kRequestTime] = TimeValue(std::chrono::system_clock::now());
+  SetStringAttribute(kOriginIp, GetFirstForwardedFor(header_map), attr);
```
geeknoid: IP addresses should be encoded as byte arrays over the wire, 4 bytes for ipv4 and 16 bytes for ipv6

qiwzhang: They are removed. It was extracting from x-forwarded-for header.   I think it is wrong. We need better way to extract origin.ip

PR #132, opened 2017-02-28
Update postsubmit to create docker images

sebastienvas:


sebastienvas:
it seems that libevent cannot be downloaded.

PR #136, opened 2017-03-01
Fix mixer client crash.

qiwzhang:
For GET request, response_header is nullptr.  Need to check it when sending Report.

### DIFF
--- a/src/envoy/mixer/http_control.cc
+++ b/src/envoy/mixer/http_control.cc
@@ -108,10 +108,12 @@ void FillRequestHeaderAttributes(const HeaderMap& header_map,
       StringMapValue(ExtractHeaders(header_map));
 }
 
-void FillResponseHeaderAttributes(const HeaderMap& header_map,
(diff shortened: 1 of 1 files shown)


src/envoy/mixer/http_control.cc
```diff
@@ -180,7 +182,7 @@ void HttpControl::Report(HttpRequestDataPtr request_data,
                          int check_status, DoneFunc on_done) {
   // Use all Check attributes for Report.
   // Add additional Report attributes.
-  FillResponseHeaderAttributes(*response_headers, &request_data->attributes);
+  FillResponseHeaderAttributes(response_headers, &request_data->attributes);
```
lizan: just curious, in what situation response_headers will be `nullptr`?

qiwzhang: GET

lizan: hm, weird. GET should have response_headers as well..

PR #142, opened 2017-03-02
Populate origin.user attribute from the SAN field of client cert

wattli:


src/envoy/mixer/http_control.cc
```diff
@@ -169,9 +171,12 @@ void HttpControl::FillCheckAttributes(HeaderMap& header_map, Attributes* attr) {
   }
 }
 
-void HttpControl::Check(HttpRequestDataPtr request_data, HeaderMap& headers,
+void HttpControl::Check(HttpRequestDataPtr request_data, HeaderMap& headers, std::string origin_user,
                         DoneFunc on_done) {
   FillCheckAttributes(headers, &request_data->attributes);
+  if (origin_user != "") {
```
qiwzhang: you can call SetStringAttribute

src/envoy/mixer/http_filter.cc
```diff
@@ -151,8 +152,15 @@ class Instance : public Http::StreamFilter, public Http::AccessLog::Instance {
     state_ = Calling;
     initiating_call_ = true;
     request_data_ = std::make_shared<HttpRequestData>();
+
+    std::string origin_user;
+    Ssl::Connection* ssl = const_cast<Ssl::Connection *>(decoder_callbacks_->ssl());
+    if (ssl != nullptr) {
+      origin_user = ssl->uriSanPeerCertificate();
```
qiwzhang: can we use const for this function?  or why do we need const_cast?

src/envoy/mixer/http_control.cc
wattli: Done

src/envoy/mixer/http_filter.cc
wattli: Yes, it's something i can add to envoy.

src/envoy/mixer/http_control.cc
```diff
@@ -169,9 +171,10 @@ void HttpControl::FillCheckAttributes(HeaderMap& header_map, Attributes* attr) {
   }
 }
 
-void HttpControl::Check(HttpRequestDataPtr request_data, HeaderMap& headers,
+void HttpControl::Check(HttpRequestDataPtr request_data, HeaderMap& headers, std::string origin_user,
```
qiwzhang: it seems that format is wrong.

wattli: Done

PR #150, opened 2017-03-07
Not call report if decodeHeaders is not called.

qiwzhang:
Somehow there are cases where decode() is not called, only encode() and log() are called.  Here is the log

[2017-03-07 18:19:59.283][5][info][main] [C7] new connection
[2017-03-07 18:19:59.296][5][debug][http] Called Mixer::Instance : Instance
[2017-03-07 18:19:59.296][5][debug][http] Called Mixer::Instance : setDecoderFilterCallbacks
[2017-03-07 18:19:59.296][5][debug][http] Called Mixer::Instance : setEncoderFilterCallbacks
[2017-03-07 18:19:59.296][5][debug][http] Called Mixer::Instance : encodeHeaders
[2017-03-07 18:19:59.296][5][debug][http] Called Mixer::Instance : log

### DIFF
--- a/src/envoy/mixer/http_filter.cc
+++ b/src/envoy/mixer/http_filter.cc
@@ -242,6 +242,8 @@ class Instance : public Http::StreamFilter, public Http::AccessLog::Instance {
                    const HeaderMap* response_headers,
                    const AccessLog::RequestInfo& request_info) override {
     Log().debug("Called Mixer::Instance : {}", __func__);
+    // If decodeHaeders() is not called, not to call Mixer report.
+    if (!request_data_) return;
     // Make sure not to use any class members at the callback.
     // The class may be gone when it is called.
     // Log() is a static function so it is OK.


src/envoy/mixer/http_filter.cc
```diff
@@ -242,6 +242,8 @@ class Instance : public Http::StreamFilter, public Http::AccessLog::Instance {
                    const HeaderMap* response_headers,
                    const AccessLog::RequestInfo& request_info) override {
     Log().debug("Called Mixer::Instance : {}", __func__);
+    // If decodeHaeders() is not called, not to call Mixer report.
+    if (!request_data_) return;
```
mandarjog: When is decode headers not called, but the filter still gets control?

qiwzhang: at log() function which is called after the response is send

qiwzhang: our filter register all these callback functions,  it is up to envoy to decide which one to call.

qiwzhang: Our check() is done at decodeHeader(), and report() is done at log().

lizan: any specific config and/or request to reproduce this case?

mandarjog: aah, ok. So if there is a connection reset or other transport level error, there will not be any headers to decode, but `Log` will still be called.

qiwzhang: Can one of you approve this PR?

PR #154, opened 2017-03-08
Update to the mixerclient with sync-ed write.

qiwzhang:


kyessenov:
Can we wait until after we see if `fail_fast` works?

PR #155, opened 2017-03-09
Update mixerclient with sync-ed grpc write and fail-fast.

qiwzhang:


qiwzhang:
Hi @kyessenov , you want proxy to based on a PR (not submitted) code?  I don't think that is a good idea.  You can try it locally.

kyessenov:
I want to do extensive integration test for the PR. Matt doesn't want to merge it until we confirm it fixes the issue.

lizan:
@kyessenov let's do that in another PR but not this.

qiwzhang:
@kyessenov I will submit this change.  Then, in your local repo, you can rebase proxy to that Envoy PR and build your docker image for test.
Could you approve this PR?

PR #177, opened 2017-03-15
Add mixer client end to end integration test.

qiwzhang:


src/envoy/mixer/BUILD
```diff
@@ -94,6 +90,16 @@ proxy_docker_build(
...
+            "base": "@docker_ubuntu//:xenial",
+        },
+        {
+            "name": "proxy_debug",
+            "base": "@ubuntu_xenial_debug//file",
+        },
+    ],
     ports = ["9090"],
     repository = "istio",
     tags = ["manual"],
```
kyessenov: You probably also want to use envoy debug build for debug image.

kyessenov:
I can't really review mixer fake server.

src/envoy/mixer/BUILD
qiwzhang: This PR just re-format the BUILD file.  It did not change this part.  
It should be done by separate PR.

qiwzhang:
I need golang expert approval.  You are our golang expert.

WORKSPACE
```diff
@@ -96,3 +96,143 @@ http_file(
     url = "https://storage.googleapis.com/istio-build/manager/ubuntu_xenial_debug-" + DEBUG_BASE_IMAGE_SHA + ".tar.gz",
     sha256 = DEBUG_BASE_IMAGE_SHA,
 )
+
```
lizan: push these changes in to a separate .bzl file

src/envoy/mixer/BUILD
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Google Inc. All Rights Reserved.
+# Copyright 2017 Google Inc. All Rights Reserved.
```
lizan: don't change this, copyright year should be the year the work first published (so, oldest).

qiwzhang:
It seems to me it will be more work to re-use the mixer code than copying the code and simplifying it for my case.

geeknoid:
Let's say we switch to a unary API: You'll need to reimplement a lot of this stuff.
If we add a new attribute type: You'll need to do that too.
If we add attribute return values to all methods: You'll need to do that.

I think it might seem shorter and more expedient for right now, but it's definitely not more efficient in the medium to long term. I don't think we want to do this. Maybe in the future when the end-to-end story is more stable, but for now things are changing everywhere on a constant basis and cloning things just means multiplying the amount of work we have to do in order to make changes.

Additionally, the fidelity of the test mixer will always be soso relative to the real thing. Will you be able to trust that the proxy will actually work with the real mixer, given it was tested against this fake mixer? I don't think so.

kyessenov:
How about implementing this test in mixer repo then?

qiwzhang:
No, it should be at istio/proxy repo.  Its goal is to make sure proxy sending correct attributes to mixer.

src/envoy/mixer/integration_test/setup.go
```diff
@@ -0,0 +1,64 @@
...
+package test
+
+import (
+	"log"
+)
+
+const (
+	// These ports should match with used envoy.conf
+	// Default is using one in this folder.
+	ServerProxyPort = 29090
```
wattli: This value is unused, remove?

src/envoy/mixer/integration_test/BUILD
```diff
@@ -0,0 +1,58 @@
...
+        "mixer_test.go",
+    ],
+    data = [
+        "envoy.conf",
+        "//src/envoy/mixer:envoy",
+    ],
+    library = ":go_default_library",
+    # shared memory path /envoy_shared_memory_0 used by Envoy
+    # hot start is not working in sandbox mode.
+    local = True,
```
kyessenov: Can you try with `--sandbox_tmpfs_path=/dev/shm`?

src/envoy/mixer/integration_test/setup.go
kyessenov: It's used in the template. I think the idea is to be able to run config standalone.

src/envoy/mixer/integration_test/BUILD
qiwzhang: Thanks.  Tried with  --sandbox_tmpfs_path=/dev/shm.  Error is:

critical][assert] panic: cannot open shared memory region /envoy_shared_memory_0 check user permissions: external/envoy_git/source/exe/hot_restart.cc:32

Tried with --sandbox_tmpfs_path=/envoy_shared_memory_0,  the error is

src/main/tools/linux-sandbox-pid1.cc:245: "mount(tmpfs, envoy_shared_memory_0, tmpfs, MS_NOSUID | MS_NODEV | MS_NOATIME, NULL)": No such file or directory

PR #179, opened 2017-03-16
Use opaque config to turn on/off forward attribute and mixer filter

kyessenov:
* merged the two filters together
* forward attribute is on by default, requires opaque config to turn it off
* mixer filter is off by default, requires opaque config to turn it on
(client-side mixer filter config is preferred)
* mixer filter is decoder only (drop encoder capability)

No tests, so please help reviewing.


mandarjog:
can you add an example of the opaque config?

src/envoy/mixer/http_filter.cc
```diff
@@ -129,9 +152,22 @@ class Instance : public Http::StreamFilter, public Http::AccessLog::Instance {
   bool initiating_call_;
   int check_status_code_;
 
+  // mixer control switch (on by default)
+  bool mixer_disabled() {
```
qiwzhang: Can we only call this once and save the result to a member variable?

src/envoy/mixer/http_filter.cc
```diff
@@ -149,6 +185,24 @@ class Instance : public Http::StreamFilter, public Http::AccessLog::Instance {
   FilterHeadersStatus decodeHeaders(HeaderMap& headers,
                                     bool end_stream) override {
     Log().debug("Called Mixer::Instance : {}", __func__);
+
+    // forward attributes (off by default)
+    if (!config_->forward_attributes().empty()) {
+      // detect disable switch in the opaque config
+      auto route = decoder_callbacks_->route()->routeEntry();
```
qiwzhang: We may need to optimize this. We don't want to call this for each request.

 Maybe create a map<route, bool>  in the Config to cache the result.
Or two maps, one for mixer_control and one for mixer_forward.

mandarjog:
Just to record here, we should switch the defaults.
forwarder will always forward (and can be switched off)
mixer_attributes are off and should be turned on thru opaque config.
Rationale is that there are *many many* more outbound routes than inbound. So you would need to change this only in 1 place.

src/envoy/mixer/http_filter.cc
mandarjog: the information is available inside the `per route` opaque  config. Not sure if it can be cached on the instance?

src/envoy/mixer/http_filter.cc
kyessenov: is decodeHeaders always called first? if it is, then I can cache. If not, it's safer to recompute it.

qiwzhang: If you cache it in Config, and with route pointer as key,   then you don't care which one is called first.
If cache miss, check the config json,  otherwise, use cache result.

kyessenov: This will work up until filters start changing routes. Just beware.

PR #182, opened 2017-03-17
Fix a bug in opaque config change and test it out

kyessenov:
Note that we need to define `mixer_server` on the client side. We need to do a follow-up to make that config optional.

### DIFF
--- a/src/envoy/mixer/README.md
+++ b/src/envoy/mixer/README.md
@@ -46,14 +46,13 @@ This Proxy will use Envoy and talk to Mixer server.
 * Then issue HTTP request to proxy.
 
 ```
+  # request to server-side proxy
   curl http://localhost:9090/echo -d "hello world"
+  # request to client-side proxy that gets sent to server-side proxy
+  curl http://localhost:7070/echo -d "hello world"
 ```
 
 ## How to configurate HTTP filters
-
-This module has two HTTP filters:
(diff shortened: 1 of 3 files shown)


src/envoy/mixer/README.md
```diff
@@ -65,35 +64,21 @@ This filter will intercept all HTTP requests and call Mixer. Here is its config:
       "name": "mixer",
       "config": {
```
qiwzhang: could you update type?

src/envoy/mixer/README.md
```diff
@@ -65,35 +64,21 @@ This filter will intercept all HTTP requests and call Mixer. Here is its config:
...
- 	    }
+         }
     }
 ```
 
 Notes:
-* attributes: these attributes will be forwarded to the upstream istio/proxy.
+* mixer_server is required
+* mixer_attributes: these attributes will be send to the mixer
+* forward_attributes: these attributes will be forwarded to the upstream istio/proxy.
```
qiwzhang: Could you also document how to set opage_config in route

src/envoy/mixer/README.md
kyessenov: done

src/envoy/mixer/README.md
kyessenov: done

PR #184, opened 2017-03-17
Update mixer client to mixer api with gogoproto.

qiwzhang:


kyessenov:
This is a breaking change to the protocol right? This works with the recent mixer only?

qiwzhang:
Yes, This PR will fix the API incompatible change in mixer api.

PR #186, opened 2017-03-17
Move .bazelrc to tools/bazel.rc

lizan:
.bazelrc in source root will hide ~/.bazelrc customization.


kyessenov:
This is against convention in other istio projects, where we merge project bazelrc with CI's bazelrc

lizan:
Need @sebastienvas comment here. `tools/bazel.rc` gets read by CI too so there is no need to cat them together.

sebastienvas:
Yes @kyessenov is right. In order to make it work properly we need to update the pipeline lib first and then update the repo.

sebastienvas:
I think we have a different one for CI because there are different constraints. I guess we need to move all bazelrc to tools.

sebastienvas:
I need to update https://github.com/istio/istio-testing/blob/master/src/org/istio/testutils/Bazel.groovy#L54 first though

lizan:
Hmm. OK, moving all bazelrcs to tools sounds good.

sebastienvas:
Should I merge this ?

PR #192, opened 2017-03-21
Make quota call

qiwzhang:
It is configurated in the Envoy mixer filter config.  It is not configured, it will not call.

### DIFF
--- a/src/envoy/mixer/README.md
+++ b/src/envoy/mixer/README.md
@@ -67,7 +67,7 @@ This filter will intercept all HTTP requests and call Mixer. Here is its config:
          "mixer_attributes" : {
             "attribute_name1": "attribute_value1",
             "attribute_name2": "attribute_value2",
-            "quota.name": "RequestCount",
+            "quota.name": "RequestCount"
          },
          "forward_attributes" : {
             "attribute_name1": "attribute_value1",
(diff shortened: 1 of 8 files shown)


src/envoy/mixer/README.md
```diff
@@ -66,7 +66,8 @@ This filter will intercept all HTTP requests and call Mixer. Here is its config:
          "mixer_server": "${MIXER_SERVER}",
          "mixer_attributes" : {
             "attribute_name1": "attribute_value1",
-            "attribute_name2": "attribute_value2"
+            "attribute_name2": "attribute_value2",
+            "quota.name": "RequestCount",
```
qiwzhang: No, they will be converted to protobuf field in mixer client. Please checkout this PR
https://github.com/istio/mixerclient/pull/44

src/envoy/mixer/http_control.cc
```diff
@@ -114,6 +114,23 @@ HttpControl::HttpControl(const std::string& mixer_server,
   ::istio::mixer_client::MixerClientOptions options;
   options.mixer_server = mixer_server;
   mixer_client_ = ::istio::mixer_client::CreateMixerClient(options);
+
+  // Extract quota attributes
+  auto it = config_attributes_.find(::istio::mixer_client::kQuotaName);
```
qiwzhang: Don't like to use namespace at mixer_client level.  We should move these two variables into Attributes class, then use Attributes namespace.   

Will do that later.

src/envoy/mixer/envoy.conf.template
```diff
@@ -42,7 +42,8 @@
                   "mixer_server": "${MIXER_SERVER}",
                   "mixer_attributes": {
                       "target.uid": "POD222",
-                      "target.namespace": "XYZ222"
+                      "target.namespace": "XYZ222",
+		      "quota.name": "RequestCount"
```
qiwzhang: Done

PR #195, opened 2017-03-21
Update envoy and update configs

lizan:


wattli:
https://screenshot.googleplex.com/p1wmP0ic4Vi, some build error.

lizan:
Working on that, seems some incompatibility with travis, jenkins passed.

lizan:
Seems the issue with gcc 4.8, which travis uses, with gcc 4.9 there is no compile error. Working on a fix to that.

wattli:
Thanks.

PR #197, opened 2017-03-22
Enable check cache and refactory mixer config loading 

qiwzhang:
1) Add MixerConfig struct to load all configs from Json 
2) Hook up check_cache_keys to enable check cache.

### DIFF
--- a/src/envoy/mixer/README.md
+++ b/src/envoy/mixer/README.md
@@ -83,8 +83,8 @@ This filter will intercept all HTTP requests and call Mixer. Here is its config:
 
 Notes:
 * mixer_server is required
-* mixer_attributes: these attributes will be send to the mixer
-* forward_attributes: these attributes will be forwarded to the upstream istio/proxy.
+* mixer_attributes: these attributes will be send to the mixer in both Check and Report calls.
+* forward_attributes: these attributes will be forwarded to the upstream istio/proxy. It will send them to mixer in Check and Report calls.
 * quota_name, quota_amount are used for making quota call. quota_amount is default to 1 if missing.
 * check_cache_keys is to cache check calls. If missing or empty, check calls are not cached.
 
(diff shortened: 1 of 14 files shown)


qiwzhang:
@geeknoid how about the design of mixer envoy filter config to enable quota and check_cache (in the README.md).   Are you ok with that?

geeknoid:
Yeah, the config looks fine. Ultimately, we'll want a more dynamic way to get the set of cache keys, but this will do for now. And the name of the quota will depend on what the name of the quota descriptor is in the mixer's config.

src/envoy/mixer/config.h
```diff
@@ -0,0 +1,53 @@
...
+namespace Mixer {
+
+// A config for mixer filter
+struct MixerConfig {
+  // the mixer server address
+  std::string mixer_server;
+
+  // These static attributes will be send to mixer in both
+  // Check and Report.
+  std::map<std::string, std::string> mixer_attributes;
```
chowchow316: I may lose some context here. Just wondering what is the relation between `mixer_attributes` and `forward_attributes?` And why we need a standalone quota attributes?

qiwzhang: mixer_attributes will be send to mixer server
forward_attributes will be forwarded to upstream so that next hop proxy can send them to mixer server.  
Above attributes are for Check and Report.
Quota is a separate call.  Here we use some standalone attributes to trigger the calls.

chowchow316: Will `mixer_attributes` and `forward_attributes` have overlap/different entries?

kyessenov:
Is there a description of the protocol somewhere? Is it client or server side making the quota call and why do we need to hard code attributes?

src/envoy/mixer/config.h
qiwzhang: If they have overlapped entries,  mixer_attributes win.

qiwzhang:
I updated the README.  Will that help?

src/envoy/mixer/README.md
```diff
@@ -98,3 +98,16 @@ By default, mixer filter forwards attributes and does not invoke mixer server. Y
...
+
+Quota (rate limiting) is enforced by the mixer. Mixer needs to be configured with Quota in its global config and service config. Its quota config will have
+"quota name", its limit within a window.  If "Quota" is added but param is missing, the default config is: quota name is "RequestCount", the limit is 10 with 1 second window. Essentially, it is imposing 10 qps rate limiting.
+
+Mixer client can be configured to make Quota call for all requests.  If "quota_name" is specified in the mixer filter config, mixer client will call Quota with the specified quota name.  If "quota_amount" is specified, it will call with that amount, otherwise the used amount is 1.
+
+
+## How to pass some attributes from client proxy to mixer.
+
+Usually client proxy is not configured to call mixer (it can be enabled in the route opaque_config). Client proxy can pass some attributes to mixer by using "forward_attributes" field.  Its attributes will be send to the upstream proxy (the server proxy). If the server proxy is calling mixer, these attributes will be send to the mixer.
```
chowchow316: will be sent to

qiwzhang:
Hi @kyessenov. For now, it is the proxy that calling Mixer enforced Quota.  Normally, server proxy is calling Mixer so it is server side.

src/envoy/mixer/README.md
qiwzhang: Thanks. Done

PR #201, opened 2017-03-24
Split the big integration test into separate small tests.

qiwzhang:
1) Envoy config can be customized. 
2) quota and check_cache can be tested separately
3) add test_suite.bzl to generate a go_test for each xxx_test.go file.


src/envoy/mixer/integration_test/attributes.go
```diff
@@ -45,7 +45,21 @@ func verifyStringMap(actual map[string]string, expected map[string]interface{})
 	return nil
 }
 
-// Please see the comment at top of mixer_test.go for verification rules
+// Attributes verification rules:
```
chowchow316: Is this rule applied to real situation or just for go test pipeline?

src/envoy/mixer/integration_test/check_cache_test.go
```diff
@@ -0,0 +1,40 @@
...
+func TestCheckCache(t *testing.T) {
+	s, err := SetUp(t, basicConfig+","+checkCacheConfig)
+	if err != nil {
+		t.Fatalf("Failed to setup test: %v", err)
+	}
+	defer s.TearDown()
+
+	url := fmt.Sprintf("http://localhost:%d/echo", ClientProxyPort)
+
+	// Issues a GET echo request with 0 size body
```
chowchow316: shall we add body here for test?

src/envoy/mixer/integration_test/failed_request_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	rpc "github.com/googleapis/googleapis/google/rpc"
+)
+
+const (
+	mixerAuthFailMessage = "Unauthenticated by mixer."
+)
+
+// Check attributes from a fail GET request from mixer
+const checkAttributesMixerFail = `
```
chowchow316: It seems that this is identical to `checkAttributesOkGet`. Should these two have some differences?

src/envoy/mixer/integration_test/attributes.go
qiwzhang: Just for integration test.

src/envoy/mixer/integration_test/check_cache_test.go
qiwzhang: body is irreverent here, we just test chache for Check calls to mixer.

src/envoy/mixer/integration_test/failed_request_test.go
qiwzhang: They maybe the same.  But It is easy to read if it is in the same file

PR #202, opened 2017-03-24
Allow for HTTP based function from Firebase rules

sarvaniv:
The changes in this PR will allow firebase rules to fetch the required information to do authorization from a REST endpoint. This endpoint could be a Google Datastore or Cloud SQL or spanner endpoint. ESP will invoke the HTTP request on behalf of Firebase rules and provide the response back to the Firebase service as a JSON value. This value can directly used as a Map in firebase rules to check for various attributes or policies to make authorization decisions.

Testing:

Integration testing via Cloud Datastore works.
Added more Unit tests to cover the new implementation.
Note that the old implementation of basic firebase rules should still work and the unit tests for those remain.

### DIFF
--- a/contrib/endpoints/src/api_manager/auth/service_account_token.h
+++ b/contrib/endpoints/src/api_manager/auth/service_account_token.h
@@ -65,6 +65,7 @@ class ServiceAccountToken {
     JWT_TOKEN_FOR_SERVICE_CONTROL = 0,
     JWT_TOKEN_FOR_CLOUD_TRACING,
     JWT_TOKEN_FOR_FIREBASE,
+    JWT_TOKEN_FOR_AUTHORIZATION_SERVICE,
     JWT_TOKEN_TYPE_MAX,
   };
   // Set audience.  Only calcualtes JWT token with specified audience.
--- a/contrib/endpoints/src/api_manager/auth/service_account_token.h
+++ b/contrib/endpoints/src/api_manager/auth/service_account_token.h
@@ -65,6 +65,8 @@ class ServiceAccountToken {
     JWT_TOKEN_FOR_SERVICE_CONTROL = 0,
     JWT_TOKEN_FOR_CLOUD_TRACING,
     JWT_TOKEN_FOR_FIREBASE,
+
+    // JWT token for accessing the http endpoints defined in Firebase Rules.
     JWT_TOKEN_FOR_AUTHORIZATION_SERVICE,
     JWT_TOKEN_TYPE_MAX,
   };
--- a/contrib/endpoints/src/api_manager/check_security_rules.cc
+++ b/contrib/endpoints/src/api_manager/check_security_rules.cc
@@ -17,14 +17,12 @@
 #include <iostream>
 #include <sstream>
 #include "contrib/endpoints/src/api_manager/auth/lib/json_util.h"
-#include "contrib/endpoints/src/api_manager/proto/security_rules.pb.h"
+#include "contrib/endpoints/src/api_manager/firebase_rules/firebase_request.h"
 #include "contrib/endpoints/src/api_manager/utils/marshalling.h"
 
-using ::google::api_manager::auth::GetProperty;
 using ::google::api_manager::auth::GetStringValue;
+using ::google::api_manager::firebase_rules::FirebaseRequest;
 using ::google::api_manager::utils::Status;
-using ::google::protobuf::Map;
-using ::google::protobuf::util::error::Code;
 
 namespace google {
 namespace api_manager {
@@ -33,69 +31,24 @@ namespace {
 const char kFailedFirebaseReleaseFetch[] = "Failed to fetch Firebase Release";
 const char kFailedFirebaseTest[] = "Failed to execute Firebase Test";
 const char kInvalidResponse[] = "Invalid JSON response from Firebase Service";
-const char kTestSuccess[] = "SUCCESS";
-const char kHttpGetMethod[] = "GET";
-const char kHttpPostMethod[] = "POST";
-const char kHttpHeadMethod[] = "HEAD";
-const char kHttpOptionsMethod[] = "OPTIONS";
-const char kHttpDeleteMethod[] = "DELETE";
-const char kFirebaseCreateMethod[] = "create";
-const char kFirebaseGetMethod[] = "get";
-const char kFirebaseDeleteMethod[] = "delete";
-const char kFirebaseUpdateMethod[] = "update";
 const char kV1[] = "/v1";
-const char kTestQuery[] = ":test?alt=json";
+const char kHttpGetMethod[] = "GET";
 const char kProjects[] = "/projects";
 const char kReleases[] = "/releases";
 const char kRulesetName[] = "rulesetName";
 const char kTestResults[] = "testResults";
-const char kState[] = "state";
-const char kToken[] = "token";
-const char kAuth[] = "auth";
-const char kRequest[] = "request";
 const char kContentType[] = "Content-Type";
 const char kApplication[] = "application/json";
-
-void SetProtoValue(const std::string &key,
-                   const ::google::protobuf::Value &value,
-                   ::google::protobuf::Value *head) {
-  ::google::protobuf::Struct *s = head->mutable_struct_value();
-  Map<std::string, google::protobuf::Value> *fields = s->mutable_fields();
-  (*fields)[key] = value;
-}
 
 std::string GetReleaseName(const context::RequestContext &context) {
   return context.service_context()->service_name() + ":" +
          context.service_context()->service().apis(0).version();
 }
 
-std::string GetRulesetTestUri(const context::RequestContext &context,
-                              const std::string &ruleset_id) {
-  return context.service_context()->config()->GetFirebaseServer() + kV1 + "/" +
-         ruleset_id + kTestQuery;
-}
-
 std::string GetReleaseUrl(const context::RequestContext &context) {
   return context.service_context()->config()->GetFirebaseServer() + kV1 +
          kProjects + "/" + context.service_context()->project_id() + kReleases +
          "/" + GetReleaseName(context);
-}
-
-std::string GetOperation(const std::string &httpMethod) {
-  if (httpMethod == kHttpPostMethod) {
-    return kFirebaseCreateMethod;
-  }
-
-  if (httpMethod == kHttpGetMethod || httpMethod == kHttpHeadMethod ||
-      httpMethod == kHttpOptionsMethod) {
-    return kFirebaseGetMethod;
-  }
-
-  if (httpMethod == kHttpDeleteMethod) {
-    return kFirebaseDeleteMethod;
-  }
-
-  return kFirebaseUpdateMethod;
 }
 
 // An AuthzChecker object is created for every incoming request. It does
@@ -111,38 +64,30 @@ class AuthzChecker : public std::enable_shared_from_this<AuthzChecker> {
              std::function<void(Status status)> continuation);
 
  private:
-  // Helper method that invokes the test firebase service api.
-  void CallTest(const std::string &ruleset_id,
-                std::shared_ptr<context::RequestContext> context,
-                std::function<void(Status status)> continuation);
+  // Helper method that invokes Http requests for firebase rules.
+  void CallNextRequest(std::function<void(Status status)> continuation);
 
   // Parse the respose for GET RELEASE API call
   Status ParseReleaseResponse(const std::string &json_str,
                               std::string *ruleset_id);
 
-  // Parses the response for the TEST API call
-  Status ParseTestResponse(context::RequestContext &context,
-                           const std::string &json_str);
-
-  // Builds the request body for the TESP API call.
-  Status BuildTestRequestBody(context::RequestContext &context,
-                              std::string *result_string);
-
   // Invoke the HTTP call
   void HttpFetch(const std::string &url, const std::string &method,
                  const std::string &request_body,
+                 auth::ServiceAccountToken::JWT_TOKEN_TYPE token_type,
                  std::function<void(Status, std::string &&)> continuation);
 
   // Get the auth token for Firebase service
-  const std::string &GetAuthToken() {
-    return sa_token_->GetAuthToken(
-        auth::ServiceAccountToken::JWT_TOKEN_FOR_FIREBASE);
+  const std::string &GetAuthToken(
+      auth::ServiceAccountToken::JWT_TOKEN_TYPE token_type) {
+    return sa_token_->GetAuthToken(token_type);
   }
 
   std::shared_ptr<AuthzChecker> GetPtr() { return shared_from_this(); }
 
   ApiManagerEnvInterface *env_;
   auth::ServiceAccountToken *sa_token_;
+  std::unique_ptr<FirebaseRequest> request_;
 };
 
 AuthzChecker::AuthzChecker(ApiManagerEnvInterface *env,
@@ -165,6 +110,7 @@ void AuthzChecker::Check(
   // Fetch the Release attributes.
   auto checker = GetPtr();
   HttpFetch(GetReleaseUrl(*context), kHttpGetMethod, "",
+            auth::ServiceAccountToken::JWT_TOKEN_FOR_FIREBASE,
             [context, final_continuation, checker](Status status,
                                                    std::string &&body) {
               std::string ruleset_id;
@@ -182,39 +128,38 @@ void AuthzChecker::Check(
               // If the parsing of the release body is successful, then call the
               // Test Api for firebase rules service.
               if (status.ok()) {
-                checker->CallTest(ruleset_id, context, final_continuation);
+                checker->request_ = std::unique_ptr<FirebaseRequest>(
+                    new FirebaseRequest(ruleset_id, checker->env_, context));
+                checker->CallNextRequest(final_continuation);
               } else {
                 final_continuation(status);
               }
             });
 }
 
-void AuthzChecker::CallTest(const std::string &ruleset_id,
-                            std::shared_ptr<context::RequestContext> context,
-                            std::function<void(Status status)> continuation) {
-  std::string body;
-  Status status = BuildTestRequestBody(*context.get(), &body);
-  if (!status.ok()) {
-    continuation(status);
+void AuthzChecker::CallNextRequest(
+    std::function<void(Status status)> continuation) {
+  if (request_->IsDone()) {
+    continuation(request_->RequestStatus());
     return;
   }
 
   auto checker = GetPtr();
-  HttpFetch(GetRulesetTestUri(*context, ruleset_id), kHttpPostMethod, body,
-            [context, continuation, checker, ruleset_id](Status status,
-                                                         std::string &&body) {
+  auto httpRequest = request_->GetHttpRequest();
+  HttpFetch(httpRequest.url, httpRequest.method, httpRequest.body,
+            httpRequest.token_type,
+            [continuation, checker](Status status, std::string &&body) {
 
+              checker->env_->LogError(std::string("Response Body = ") + body);
               if (status.ok()) {
-                checker->env_->LogDebug(
-                    std::string("Test API succeeded with ") + body);
-                status = checker->ParseTestResponse(*context.get(), body);
+                checker->request_->UpdateResponse(body);
+                checker->CallNextRequest(continuation);
               } else {
                 checker->env_->LogError(std::string("Test API failed with ") +
                                         status.ToString());
                 status = Status(Code::INTERNAL, kFailedFirebaseTest);
+                continuation(status);
               }
-
-              continuation(status);
             });
 }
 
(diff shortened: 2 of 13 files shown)


contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  return Status::OK;
+}
+
+bool FirebaseRequest::AllFunctionCallsProcessed() {
+  return func_call_iter_ == response_.test_results(0).function_calls().end();
+}
+
+Status FirebaseRequest::AddFunctionMock(proto::TestRulesetRequest *request,
```
qiwzhang: Is this for testing?  If so, it should be in a separate file for testing.   We don't want to mix testing code with real code.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.h
```diff
@@ -0,0 +1,116 @@
...
+  std::string service_name_;
+  std::string firebase_server_;
+  Status current_status_;
+  bool is_done_;
+  std::vector<std::pair<FunctionCall, std::string>> funcs_with_result_;
+  RepeatedPtrField<FunctionCall>::const_iterator func_call_iter_;
+  TestRulesetResponse response_;
+  HttpRequest *next_request_;
+  HttpRequest firebase_http_request_;
+  HttpRequest external_http_request_;
```
qiwzhang: what is the different between external_http_request and firebase_http_request?  can we share them?  Are they run parallel?

Need more comment.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
sarvaniv: No This is not for testing. These are private methods in the class. The FunctionMock is the name of a message in the TestRulesetRequest protobuf.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.h
sarvaniv: They are not run in parallel. I will add more comments.

contrib/endpoints/src/api_manager/auth/service_account_token.h
```diff
@@ -65,6 +65,7 @@ class ServiceAccountToken {
     JWT_TOKEN_FOR_SERVICE_CONTROL = 0,
     JWT_TOKEN_FOR_CLOUD_TRACING,
     JWT_TOKEN_FOR_FIREBASE,
+    JWT_TOKEN_FOR_AUTHORIZATION_SERVICE,
```
liminw: Let's add a comment for this one, like "JWT token for accessing the http endpoints that user defines in Firebase Rules".

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -182,39 +128,38 @@ void AuthzChecker::Check(
...
 }
 
-void AuthzChecker::CallTest(const std::string &ruleset_id,
-                            std::shared_ptr<context::RequestContext> context,
-                            std::function<void(Status status)> continuation) {
-  std::string body;
-  Status status = BuildTestRequestBody(*context.get(), &body);
-  if (!status.ok()) {
-    continuation(status);
+void AuthzChecker::CallNextRequest(
```
liminw: Can you add a comment for what the function is doing?

contrib/endpoints/src/api_manager/context/service_context.cc
```diff
@@ -76,6 +76,18 @@ ServiceContext::ServiceContext(std::unique_ptr<ApiManagerEnvInterface> env,
...
   service_account_token_.SetAudience(
       auth::ServiceAccountToken::JWT_TOKEN_FOR_FIREBASE, kFirebaseAudience);
+
+  if (config_->server_config() &&
+      !config_->server_config()
+           ->api_check_security_rules_config()
+           .authorization_service_audience()
+           .empty()) {
+    service_account_token_.SetAudience(
+        auth::ServiceAccountToken::JWT_TOKEN_FOR_AUTHORIZATION_SERVICE,
```
liminw: As we discussed, I think it makes sense to ask API producer to enter the accepted audience as a http.invoke() function parameter, instead of having it statically in ESP config. There are two advantages. First, the rules are allowed to call multiple http endpoints, each of which accepts different JWT audience. Second, ESP service config is not bound to rules logic. API producer can change the rules to call a different http endpoint without the need to change ESP config and redeploy.

contrib/endpoints/src/api_manager/firebase_rules/BUILD
```diff
@@ -0,0 +1,41 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
```
liminw: 2017

contrib/endpoints/src/api_manager/check_security_rules.cc
sarvaniv: There is a comment that already exists in the class declaration. look at line 69. Let me know if you need me to add anything more to it.

contrib/endpoints/src/api_manager/auth/service_account_token.h
sarvaniv: Done.

contrib/endpoints/src/api_manager/firebase_rules/BUILD
sarvaniv: Done.

contrib/endpoints/src/api_manager/context/service_context.cc
sarvaniv: While that was what I had initially suggested we do about the audiences, I started having second thoughts once I started writing the firebase rules. For example, the fact that we are using something called "audiences" to generated an auth token is ESP specific and we are only generating these for Google based services. I am not sure how this will translate to custom auth cases.  I think we should think more about this and ask experts on the best way to do this. But this change should not affect this PR. We can address this as a part of another PR .

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -0,0 +1,402 @@
...
+void SetProtoValue(const std::string &key,
+                   const ::google::protobuf::Value &value,
+                   ::google::protobuf::Value *head) {
+  ::google::protobuf::Struct *s = head->mutable_struct_value();
+  Map<std::string, google::protobuf::Value> *fields = s->mutable_fields();
+  (*fields)[key] = value;
+}
+
+// Convert HTTP method to Firebase specific method.
+std::string GetOperation(const std::string &httpMethod) {
```
lizan: nit return `const std::string&`

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -0,0 +1,402 @@
...
+  // request.
+  SetStatus(UpdateRulesetRequestBody(RepeatedPtrField<FunctionCall>()));
+  if (!current_status_.ok()) {
+    return;
+  }
+
+  next_request_ = &firebase_http_request_;
+}
+
+bool FirebaseRequest::IsDone() { return is_done_; }
```
lizan: nit: is_done() for accessors.

contrib/endpoints/src/api_manager/check_security_rules_test.cc
```diff
@@ -295,6 +351,56 @@ class CheckSecurityRulesTest : public ::testing::Test {
...
+    std::string json_str;
+    utils::ProtoToJson(response, &json_str, utils::JsonOptions::DEFAULT);
+    return json_str;
+  }
+
+  void SetProtoValue(const std::string &key,
+                     const ::google::protobuf::Value &value,
+                     ::google::protobuf::Value *head) {
+    ::google::protobuf::Struct *s = head->mutable_struct_value();
+    Map<std::string, google::protobuf::Value> *fields = s->mutable_fields();
```
lizan: nit: `auto fields = head->mutable_struct_value()->mutable_fields()`

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
sarvaniv: As I thought. This gives a bunch of warnings. Looks like the compiler is not liking this even if I am returing on const strings that are already declared ...

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc:61:12: warning: returning reference to temporary [-Wreturn-local-addr]
     return kFirebaseCreateMethod;
            ^
contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc:66:12: warning: returning reference to temporary [-Wreturn-local-addr]
     return kFirebaseGetMethod;
            ^
contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc:70:12: warning: returning reference to temporary [-Wreturn-local-addr]
     return kFirebaseDeleteMethod;
            ^
contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc:73:10: warning: returning reference to temporary [-Wreturn-local-addr]
   return kFirebaseUpdateMethod;

contrib/endpoints/src/api_manager/context/service_context.cc
liminw: "aud" is a standard JWT claim. It's not ESP specific. Any service that accepts JWT defines what audience it accepts. Google services accept JWT audience that matches the service name. Not all services do the same, and the service name can have different formats. That's why we need to ask API producers to specify the accepted audience in the function specification when they write rules. 

I am ok with having this change in a separate PR.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.h
```diff
@@ -0,0 +1,155 @@
...
+  // This variable points to either firebase_http_request_ or
+  // external_http_request_. This will allow the state machine to understand if
+  // the response received is for TestRuleset or user defined HTTP endpoint. If
+  // next_request points to firebase_http_request_, upon receiving a response,
+  // the state machine will convert the response to TestRulesetResponse and
+  // process the response. If next_request_ points to external_http_request_,
+  // then the reponse provided via UpdateResponse is converted into a
+  // protobuf::Value. This value is initialized to nullptr and will be nullptr
+  // once is_done_ is set to true.
+  HttpRequest *next_request_;
```
qiwzhang: this implies state.   Should we define a state variable for that?

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
qiwzhang: I think if you define your const as
const std::string kFirebaseCreateMethod,  then you can return const std::string&.  It will not do data copy.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.h
```diff
@@ -0,0 +1,155 @@
...
+#include <utility>
+#include <vector>
+#include "contrib/endpoints/include/api_manager/utils/status.h"
+#include "contrib/endpoints/src/api_manager/context/request_context.h"
+#include "contrib/endpoints/src/api_manager/proto/security_rules.pb.h"
+
+using ::google::api_manager::utils::Status;
+using TestRulesetResponse = ::google::api_manager::proto::TestRulesetResponse;
+using FunctionCall = TestRulesetResponse::TestResult::FunctionCall;
+using ::google::protobuf::RepeatedPtrField;
```
qiwzhang: per c++ style guide,  name space alias should not be used in the header file. It can only be used in the cc file. Otherwise, it will easily cause confusion.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -0,0 +1,402 @@
...
+      firebase_server_ + kV1 + "/" + ruleset_name + kTestQuery;
+  firebase_http_request_.method = kHttpPostMethod;
+  firebase_http_request_.token_type =
+      auth::ServiceAccountToken::JWT_TOKEN_FOR_FIREBASE;
+  external_http_request_.token_type =
+      auth::ServiceAccountToken::JWT_TOKEN_FOR_AUTHORIZATION_SERVICE;
+
+  // Update the first request to be sent which is the TestRulesetRequest
+  // request.
+  SetStatus(UpdateRulesetRequestBody(RepeatedPtrField<FunctionCall>()));
```
qiwzhang: Here I am confused now.   It will help if you can add comment about the data flow:
What ESP should do first, what Firebase will response,  and how ESP will handle the response.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
sarvaniv: Here is Google's c++ style guide's decision on global or static std::string objects. To my understanding, this seems conflicting advice. Let me know if we have decided not to follow this guideline and I am also curious as to why we have decided this is not the best thing to do ...

https://engdoc.corp.google.com/eng/doc/devguide/cpp/styleguide.shtml?cl=head#Static_and_Global_Variables

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
sarvaniv: I will add the comments. So far as const std::string goes, let me know about why we are not following the google style guide in this case as described here:

https://engdoc.corp.google.com/eng/doc/devguide/cpp/styleguide.shtml?cl=head#Static_and_Global_Variables

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.h
sarvaniv: We don't need the state variable. I have updated the comments so they refer to the exact method and no reference to a State Machine.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
sarvaniv: I have added the full message flow sequence in the header file. Let me know if you think this makes it better to understand the flow.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -111,38 +64,31 @@ class AuthzChecker : public std::enable_shared_from_this<AuthzChecker> {
...
+                 auth::ServiceAccountToken::JWT_TOKEN_TYPE token_type,
                  std::function<void(Status, std::string &&)> continuation);
 
   // Get the auth token for Firebase service
-  const std::string &GetAuthToken() {
-    return sa_token_->GetAuthToken(
-        auth::ServiceAccountToken::JWT_TOKEN_FOR_FIREBASE);
+  const std::string &GetAuthToken(
+      auth::ServiceAccountToken::JWT_TOKEN_TYPE token_type) {
+    return sa_token_->GetAuthToken(token_type);
```
sarvaniv: Removed it.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -17,14 +17,12 @@
 #include <iostream>
 #include <sstream>
 #include "contrib/endpoints/src/api_manager/auth/lib/json_util.h"
-#include "contrib/endpoints/src/api_manager/proto/security_rules.pb.h"
+#include "contrib/endpoints/src/api_manager/firebase_rules/firebase_request.h"
```
sarvaniv: I initially started off with authz but then decided to firebase_rules since this is very explicit as to what is being used. 
1) I don't think auth is the right place for this since authentication and authorization are two different things.
2) Here are few ways things can go: 
a) If we continue using firebase rules for authorization, then this is directory is still good. 
b) If we decide to support another form of authz along with firebase rules, then I think we can create a directory called authz and move the firebase_rules under that directory and add another directory that appropriately names the authorization mechanism we will support.
c) We decide to discontinue using firebase rules, then we delete this directory. 

I think keeping this as firebase_rules makes a lot of sense since that is what we are doing.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -182,39 +129,38 @@ void AuthzChecker::Check(
...
+  if (request_->is_done()) {
+    continuation(request_->RequestStatus());
     return;
   }
 
   auto checker = GetPtr();
-  HttpFetch(GetRulesetTestUri(*context, ruleset_id), kHttpPostMethod, body,
-            [context, continuation, checker, ruleset_id](Status status,
-                                                         std::string &&body) {
+  auto httpRequest = request_->GetHttpRequest();
```
sarvaniv: Done.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.h
sarvaniv: Done.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -111,38 +64,31 @@ class AuthzChecker : public std::enable_shared_from_this<AuthzChecker> {
...
+  const std::string &GetAuthToken(
+      auth::ServiceAccountToken::JWT_TOKEN_TYPE token_type) {
+    return sa_token_->GetAuthToken(token_type);
   }
 
   std::shared_ptr<AuthzChecker> GetPtr() { return shared_from_this(); }
 
   ApiManagerEnvInterface *env_;
   auth::ServiceAccountToken *sa_token_;
+  std::unique_ptr<FirebaseRequest> request_;
```
sarvaniv: Done.

contrib/endpoints/src/api_manager/context/service_context.cc
sarvaniv: Yes. aud is a standard JWT claim. My argument is not that the users don't understand JWTs except that having to provide "aud" here is more confusing. Here is the reason why I think it is more confusing. As a firebase rules author, my thought would be that the user is already authenticated since the user has already provided a valid JWT. Now, the rules file will used the JWT claims to authroize the user. But instead of doing just that, we are also forcing the user to think more about what ESP is going to do which is generate more JWTs. I also don't think it is limiting the user from using multiple authroization sources like cloud datastore AND cloud SQL. Actually, we should encourage our users to not have their firebase rules tied to their datastore schema. This is really painful for the user in the long term for both providing the rules and well as schema changes. Instead, the rules should access some endpoint that is executed as a part of app running in say AppEngine. This allows them to use the provided client libraries that allow for better testing and maintenance. Lets discuss more with Wencheng and Wayne.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -0,0 +1,402 @@
...
+  Status status = utils::JsonToProto(context_->auth_claims(), &claims);
+  if (!status.ok()) {
+    return status;
+  }
+
+  auto *variables = test_case->mutable_request()->mutable_struct_value();
+  auto *fields = variables->mutable_fields();
+
+  path.set_string_value(context_->request()->GetRequestPath());
+  (*fields)[kPath] = path;
```
sarvaniv: No. SetProtoValue will not work here. This is because of the the way mutable_* works. Since I need all the kPath and kAuth in the same map, I cannot use SetProtoValue.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -0,0 +1,402 @@
...
+
+  method.set_string_value(
+      GetOperation(context_->request()->GetRequestHTTPMethod()));
+  (*fields)[kMethod] = method;
+
+  SetProtoValue(kToken, claims, &token);
+  (*fields)[kAuth] = token;
```
sarvaniv: Same reasoning as above.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -182,39 +129,38 @@ void AuthzChecker::Check(
               // If the parsing of the release body is successful, then call the
               // Test Api for firebase rules service.
               if (status.ok()) {
```
sarvaniv: That is another way but here is how that code will look.

if (status.ok()) {                
                checker->env_->LogDebug(            
                    std::string("GetReleasName succeeded with ") + body);
                status = checker->ParseReleaseResponse(body, &ruleset_id);  
                  if (status.ok()) {
                       checker->request_handler_ = std::unique_ptr<FirebaseRequest>(
                    new FirebaseRequest(ruleset_id, checker->env_, context));
                checker->CallNextRequest(final_continuation);     
                    }  else { // I cannot remove this else statement since this will not go into the else part below
                          final_continuation(status)
                     }                                                                                                      
              } else {
                checker->env_->LogError(std::string("GetReleaseName for ") +
                                        GetReleaseUrl(*context.get()) + 
                                        " with status " + status.ToString());
                status = Status(Code::INTERNAL, kFailedFirebaseReleaseFetch);                                                                                                       
              }
         
The code does not look as good with status checks within status checks. Which is why I have written code in the way it is.

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -111,38 +64,31 @@ class AuthzChecker : public std::enable_shared_from_this<AuthzChecker> {
...
-  void CallTest(const std::string &ruleset_id,
-                std::shared_ptr<context::RequestContext> context,
-                std::function<void(Status status)> continuation);
+  // This method invokes the Firebase TestRuleset API endpoint as well as user
+  // defined endpoints provided by the TestRulesetResponse.
+  void CallNextRequest(std::function<void(Status status)> continuation);
 
   // Parse the respose for GET RELEASE API call
   Status ParseReleaseResponse(const std::string &json_str,
                               std::string *ruleset_id);
```
sarvaniv: Here is the Google syle guide link that says output variables should be pointers. The ParseReleaseResponse will set the ruleset_id and making it a pointer is a way of saying that it will be modified. It should also be the last parameter.

https://engdoc.corp.google.com/eng/doc/devguide/cpp/styleguide.shtml?cl=head#Function_Parameter_Ordering

contrib/endpoints/src/api_manager/check_security_rules.cc
```diff
@@ -111,38 +64,31 @@ class AuthzChecker : public std::enable_shared_from_this<AuthzChecker> {
...
  private:
-  // Helper method that invokes the test firebase service api.
-  void CallTest(const std::string &ruleset_id,
-                std::shared_ptr<context::RequestContext> context,
-                std::function<void(Status status)> continuation);
+  // This method invokes the Firebase TestRuleset API endpoint as well as user
+  // defined endpoints provided by the TestRulesetResponse.
+  void CallNextRequest(std::function<void(Status status)> continuation);
 
   // Parse the respose for GET RELEASE API call
```
sarvaniv: Done

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
qiwzhang: I don't like it either.  I think we have code using std::string as static const variables.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -0,0 +1,405 @@
...
+Status FirebaseRequest::RequestStatus() { return current_status_; }
+
+void FirebaseRequest::UpdateResponse(const std::string &body) {
+  if (is_done()) {
+    env_->LogError(
+        "Receive a response body when no HTTP request is outstanding");
+    return;
+  }
+
+  if (next_request_ == nullptr) {
```
qiwzhang: we should use GOOGLE_DCHECK() here then.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -0,0 +1,405 @@
...
+    return HttpRequest();
+  }
+
+  return *next_request_;
+}
+
+Status FirebaseRequest::RequestStatus() { return current_status_; }
+
+void FirebaseRequest::UpdateResponse(const std::string &body) {
+  if (is_done()) {
```
qiwzhang: is it is code bug?  it should be GOOGLE_DCHECK()

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -0,0 +1,402 @@
...
+
+  func_call_iter_ = response_.test_results(0).function_calls().begin();
+  return Status::OK;
+}
+
+std::vector<std::pair<FunctionCall, std::string>>::const_iterator
+FirebaseRequest::Find(const FunctionCall &func_call) {
+  return std::find_if(funcs_with_result_.begin(), funcs_with_result_.end(),
+                      [func_call](std::tuple<FunctionCall, std::string> item) {
+                        return MessageDifferencer::Equals(std::get<0>(item),
```
sarvaniv: As we discussed, creating an string and then comparing is better than just comparing. When ever we get a TestRulesetResponse, creating a serialized String to generate the string key and them trying to do the comparison will be more slower.

PR #207, opened 2017-03-30
Update operation time reported to service control

gurgenh:
Do not use request start time as the operation start time when reporting
the operation to service control. Service control requires the operation
time to be recent (on the order of seconds). Given that a request can
have arbitrary latency, the start time of the request might be arbitrarily
old.


contrib/endpoints/src/api_manager/service_control/proto.cc
```diff
@@ -797,7 +797,7 @@ void SetOperationCommonFields(const OperationInfo& info,
     op->set_consumer_id(std::string(kConsumerIdApiKey) +
                         std::string(info.api_key));
   }
-  *op->mutable_start_time() = CreateTimestamp(info.request_start_time);
+  *op->mutable_start_time() = current_time;
```
qiwzhang: This code is used for sending gRPC streaming metrics too.  For gRPC streaming, we send bytes, and messaeg_count every 30s.   they are accumulated;  All of them have same start_time, but the numbers are accumulated.   For example, last one is 100, this one is 150,  this 150 includes last one's 100, the delta is only 50.   
If we change start time this way, it may not work.

PR #219, opened 2017-04-06
Send delta metrics for intermediate reports.

chowchow316:


contrib/endpoints/src/api_manager/request_handler.h
```diff
@@ -60,6 +62,10 @@ class RequestHandler : public RequestHandlerInterface {
   std::shared_ptr<context::RequestContext> context_;
 
   std::shared_ptr<CheckWorkflow> check_workflow_;
+
+  int64_t last_request_bytes;
```
qiwzhang: It is better to be done in RequestContext.

contrib/endpoints/src/api_manager/context/request_context.cc
```diff
@@ -257,8 +259,12 @@ void RequestContext::FillReportRequestInfo(
...
-    info->response_bytes = request_->GetGrpcResponseBytes();
+    // Make sure we send delta metrics for intermediate reports.
+    info->request_bytes = request_->GetGrpcRequestBytes() - last_request_bytes;
+    info->response_bytes =
+        request_->GetGrpcResponseBytes() - last_response_bytes;
+    last_request_bytes += info->request_bytes;
+    last_response_bytes += info->response_bytes;
   } else {
     info->request_size = response->GetRequestSize();
     info->response_size = response->GetResponseSize();
```
qiwzhang: Have to handle final one too.  delta could be negative for the final one since it is getting from a different call. If negative, make it 0.

contrib/endpoints/src/api_manager/context/request_context.h
```diff
@@ -177,6 +177,10 @@ class RequestContext {
 
   // The time point of last intermediate report
   std::chrono::steady_clock::time_point last_report_time_;
+
+  // The accumulated data sent till last intermediate report
+  int64_t last_request_bytes;
```
lizan: nit: `_` at the end

chowchow316: Done.

PR #244, opened 2017-04-12
Allow HTTP functions in firebase rules to specify audience

sarvaniv:
We need the right audience in order to generate the JWT token to invoke into the authorization endpoint. The code change allows the audience to be specified from the firebase rules function call. 

- Updated tests to reflect this change.
- Updated parameter test to check that audience is always present as the last parameter.

### DIFF
--- a/contrib/endpoints/src/api_manager/auth/service_account_token.cc
+++ b/contrib/endpoints/src/api_manager/auth/service_account_token.cc
@@ -59,8 +59,18 @@ void ServiceAccountToken::SetAudience(JWT_TOKEN_TYPE type,
   jwt_tokens_[type].set_audience(audience);
 }
 
-const std::string& ServiceAccountToken::GetAuthToken(JWT_TOKEN_TYPE type,
-                                                     bool ignore_cache) {
+const std::string& ServiceAccountToken::GetAuthToken(JWT_TOKEN_TYPE type) {
+  return GetAuthToken(type, jwt_tokens_[type].audience());
+}
+
+const std::string& ServiceAccountToken::GetAuthToken(
+    JWT_TOKEN_TYPE type, const std::string& audience) {
+  bool ignore_cache = false;
+  if (jwt_tokens_[type].audience() != audience) {
+    SetAudience(type, audience);
+    ignore_cache = true;
+  }
+
   // Uses authentication secret if available.
   if (!client_auth_secret_.empty()) {
     GOOGLE_CHECK(type >= 0 && type < JWT_TOKEN_TYPE_MAX);
@@ -77,12 +87,6 @@ const std::string& ServiceAccountToken::GetAuthToken(JWT_TOKEN_TYPE type,
     return jwt_tokens_[type].token();
   }
   return access_token_.token();
-}
-
-const std::string& ServiceAccountToken::GetAuthToken(
-    JWT_TOKEN_TYPE type, const std::string& audience, bool ignore_cache) {
-  SetAudience(type, audience);
-  return GetAuthToken(type, ignore_cache);
 }
 
 Status ServiceAccountToken::JwtTokenInfo::GenerateJwtToken(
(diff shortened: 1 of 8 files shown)


contrib/endpoints/src/api_manager/auth/service_account_token.cc
```diff
@@ -78,6 +78,12 @@ const std::string& ServiceAccountToken::GetAuthToken(JWT_TOKEN_TYPE type) {
   return access_token_.token();
 }
 
+const std::string& ServiceAccountToken::GetAuthToken(
+    JWT_TOKEN_TYPE type, const std::string& audience) {
+  SetAudience(type, audience);
```
qiwzhang: Token is cached, you have to clear the cache if audience is different.

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -348,6 +354,13 @@ Status FirebaseRequest::CheckFuncCallArgs(const FunctionCall &func) {
         std::string(func.function() + " Arguments 1 and 2 should be strings"));
   }
 
+  if (func.args(func.args_size()-1).kind_case() != google::protobuf::Value::kStringValue) {
```
qiwzhang: it seems the format is not right.

contrib/endpoints/src/api_manager/auth/service_account_token.cc
sarvaniv: I see. Thank you for pointing this out.

contrib/endpoints/src/api_manager/auth/service_account_token.cc
```diff
@@ -59,11 +59,12 @@ void ServiceAccountToken::SetAudience(JWT_TOKEN_TYPE type,
...
 }
 
-const std::string& ServiceAccountToken::GetAuthToken(JWT_TOKEN_TYPE type) {
+const std::string& ServiceAccountToken::GetAuthToken(JWT_TOKEN_TYPE type,
+                                                     bool ignore_cache) {
   // Uses authentication secret if available.
   if (!client_auth_secret_.empty()) {
     GOOGLE_CHECK(type >= 0 && type < JWT_TOKEN_TYPE_MAX);
-    if (!jwt_tokens_[type].is_valid(0)) {
+    if (ignore_cache || !jwt_tokens_[type].is_valid(0)) {
```
qiwzhang: It is better to compare the new audience with old audience,  if diff, wipe out the cache.

contrib/endpoints/src/api_manager/auth/service_account_token.h
```diff
@@ -77,8 +77,17 @@ class ServiceAccountToken {
   // Gets the auth token to access Google services.
   // If client auth secret is specified, use it to calcualte JWT token.
   // Otherwise, use the access token fetched from metadata server.
+  // If ignore_cache is true then a JWT token is regenerated even if the current
+  // cached JWT token is valid.
   const std::string& GetAuthToken(JWT_TOKEN_TYPE type);
```
liminw: The comments need to be updated. There is no "ignore_cache" parameter.

contrib/endpoints/src/api_manager/auth/service_account_token.h
```diff
@@ -77,8 +77,17 @@ class ServiceAccountToken {
...
 
+  // Gets the auth token to access Google services. This method accepts an
```
liminw: The target service does not have to be Google services. Correct?

contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -305,6 +310,8 @@ Status FirebaseRequest::SetNextRequest() {
       auto call = *func_call_iter_;
       external_http_request_.url = call.args(0).string_value();
       external_http_request_.method = call.args(1).string_value();
+      external_http_request_.audience =
+          call.args(call.args_size() - 1).string_value();
```
liminw: "audience" is an optional field here. For datastore, we provide a default audience if not specified. 

How do you handle the case that "audience" field is provided by in the rules for datastore? Do you overwrite the audience field with the default datastore audience?

How do you handle the case that "audience" is not provided? What is the default audience?

sarvaniv: It is not optional right now since there is no default behavior implemented right now. There is no default audience being specified at all.

contrib/endpoints/src/api_manager/auth/service_account_token.h
sarvaniv: Yes. Will do.

PR #246, opened 2017-04-13
Add HTTP request headers to Firebase TestRuleset request.

sarvaniv:
Currently we send the JWT token payload to the TestRuleset. This change also sends HTTP request headers in TestRulesetRequest. This change also requires a change on ESP. The code changes are done on ESP and I am working on a t-test in ESP to test those changes.

Updated the unit test.

### DIFF
--- a/contrib/endpoints/include/api_manager/request.h
+++ b/contrib/endpoints/include/api_manager/request.h
@@ -72,7 +72,7 @@ class Request {
                                            const std::string &value) = 0;
 
   // Gets all HTTP request headers.
-  virtual void GetHeaders(std::map<std::string, std::string> *headers) { }
+  virtual void GetHeaders(std::map<std::string, std::string> *headers) {}
 };
 
 }  // namespace api_manager
--- a/contrib/endpoints/include/api_manager/request.h
+++ b/contrib/endpoints/include/api_manager/request.h
@@ -70,6 +70,9 @@ class Request {
   // Adds a header to backend. If the header exists, overwrite its value
   virtual utils::Status AddHeaderToBackend(const std::string &key,
                                            const std::string &value) = 0;
+
+  // Gets all HTTP request headers.
+  virtual void GetHeaders(std::map<std::string, std::string> *headers) { }
 };
 
 }  // namespace api_manager
--- a/contrib/endpoints/include/api_manager/request.h
+++ b/contrib/endpoints/include/api_manager/request.h
@@ -71,7 +71,11 @@ class Request {
   virtual utils::Status AddHeaderToBackend(const std::string &key,
                                            const std::string &value) = 0;
 
-  // Gets all HTTP request headers.
+  // Gets all HTTP request headers. This method cannot be made into a pure
+  // virtual function until all the implementations of this class also implement
+  // the GetHeaders method. Changing this to a pure virtual function should be
+  // done with a lot of care and making sure that all implementations of this
+  // class have implemented this method.
   virtual void GetHeaders(std::map<std::string, std::string> *headers) {}
 };
 
--- a/contrib/endpoints/include/api_manager/request.h
+++ b/contrib/endpoints/include/api_manager/request.h
@@ -76,7 +76,7 @@ class Request {
   // the GetHeaders method. Changing this to a pure virtual function should be
   // done with a lot of care and making sure that all implementations of this
   // class have implemented this method.
-  virtual void GetHeaders(std::map<std::string, std::string> *headers) {}
+  virtual void GetHeaders(std::map<std::string, std::string> *headers) = 0;
 };
 
 }  // namespace api_manager
--- a/contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
+++ b/contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
@@ -38,6 +38,7 @@ const std::string kToken = "token";
 const std::string kAuth = "auth";
 const std::string kPath = "path";
 const std::string kMethod = "method";
+const std::string kHeaders = "headers";
 const std::string kHttpGetMethod = "GET";
 const std::string kHttpPostMethod = "POST";
 const std::string kHttpHeadMethod = "HEAD";
@@ -180,6 +181,7 @@ Status FirebaseRequest::UpdateRulesetRequestBody(
   ::google::protobuf::Value claims;
   ::google::protobuf::Value path;
   ::google::protobuf::Value method;
+  ::google::protobuf::Value headers;
 
   Status status = utils::JsonToProto(context_->auth_claims(), &claims);
   if (!status.ok()) {
@@ -188,6 +190,10 @@ Status FirebaseRequest::UpdateRulesetRequestBody(
 
   auto *variables = test_case->mutable_request()->mutable_struct_value();
   auto *fields = variables->mutable_fields();
+  GetHeaders(&headers);
+  if (headers.kind_case() != ::google::protobuf::Value::kNullValue) {
+    (*fields)[kHeaders] = headers;
+  }
 
   path.set_string_value(context_->request()->GetRequestPath());
   (*fields)[kPath] = path;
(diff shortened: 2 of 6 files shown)


contrib/endpoints/src/api_manager/firebase_rules/firebase_request.cc
```diff
@@ -216,6 +222,26 @@ Status FirebaseRequest::UpdateRulesetRequestBody(
...
+
+  std::map<std::string, std::string> req_headers;
+  context_->request()->GetHeaders(&req_headers);
+  if (req_headers.size() == 0) {
+    headers->set_null_value(::google::protobuf::NullValue::NULL_VALUE);
+    return;
+  }
+
+  auto *map = headers->mutable_struct_value()->mutable_fields();
+  for (auto const &kvp : req_headers) {
```
qiwzhang: const auto&

contrib/endpoints/include/api_manager/request.h
```diff
@@ -70,6 +70,9 @@ class Request {
   // Adds a header to backend. If the header exists, overwrite its value
   virtual utils::Status AddHeaderToBackend(const std::string &key,
                                            const std::string &value) = 0;
+
+  // Gets all HTTP request headers.
+  virtual void GetHeaders(std::map<std::string, std::string> *headers) {}
```
liminw: Shouldn't this be "= 0"? I suppose we are going to have a follow up change in ESP repo to provide the implementation of GetHeaders()?

sarvaniv: I have provided all the information in the PR description. This cannot be 0. If this is 0, then ESP that will point to this proxy's commit will have issues creating a request object since it does not implement a pure virtual function. Having this empty body will fix that issue.

sarvaniv: We can convert this into a pure virtual function but we have to wait until we merged back to main whenever that is going to happen. Until then this code cannot be changed.

liminw: Makes sense. Probably you can have a comment here saying that this is a workaround needed until ESP implementation is done, plus proxy and ESP repo are synced.

sarvaniv: On a second thought we have to very careful about making this pure virtual since Envoy also implements this class. Let's not be too eager to change this into a pure virtual function.

qiwzhang: Envoy is not using this class.   I think it is fine to make this pure since ESP is always using SHA.   When you update proxy SHA at ESP to point to this one,  that PR should add its override too.
Only one special Jenkins test will fail.  For each PR in istio/proxy we have a Jenkins test: change proxy SHA at ESP to the new one blindly and run test.  This test will fail.  But it is fine since we know this is a broken change.

lizan: Envoy is no longer implementing this interface, so this method can be `= 0` here. You should submit a PR to ESP to have an implementation (empty if it is not ready) right after merging this PR.

contrib/endpoints/include/api_manager/request.h
```diff
@@ -70,6 +70,13 @@ class Request {
   // Adds a header to backend. If the header exists, overwrite its value
   virtual utils::Status AddHeaderToBackend(const std::string &key,
                                            const std::string &value) = 0;
+
+  // Gets all HTTP request headers. This method cannot be made into a pure
+  // virtual function until all the implementations of this class also implement
+  // the GetHeaders method. Changing this to a pure virtual function should be
+  // done with a lot of care and making sure that all implementations of this
+  // class have implemented this method.
+  virtual void GetHeaders(std::map<std::string, std::string> *headers) {}
```
lizan: Actually having a empty implementation here make it harder to notice some implementation don't implement this, since the build won't break.

contrib/endpoints/include/api_manager/request.h
sarvaniv: But if we decide to merge into master, then the presubmits will fail if I make it pure virtual function. I can't make it pure virtual right now. This merge will have to happen with an empty body.

sarvaniv: Lizan, if Envoy is not implementing this interface, then who is implementing it? Since we need someone to implement this request class to use it in proxy?

lizan: Which presubmit will fail? I don't think it happens.

lizan: Currently Envoy is not depending on any code in `contrib/endpoints` except `transcoding`, this interface is only implemented in ESP.

sarvaniv: As discussed with Wayne, making this pure virtual right now can break other's using the latest proxy and ESP code. This is the safest thing to do. I also want to know the full Envoy story. Someone is implementing the request class that can be used on Proxy when used with Envoy. I don't know who implements this class and we use this Request class everywhere in the proxy code. Let's take the safest route and not break anyone. I think its good karma :-)

contrib/endpoints/include/api_manager/request.h
sarvaniv: The build will break if someone in the team have their ESP point to the latest proxy code as I do a lot during my development/testing. Especially when we merge these changes into master. We cannot be too far off when we merge both proxy and ESP changes into the master and depending on how things go, this is not something we should be too sure about. I must disagree on this. I will make it a pure virtual function once we are in master.

contrib/endpoints/include/api_manager/request.h
qiwzhang: I agreed with Lizan.  If you worry about the breakage of ESP, you should immediately make ESP PR with dummy implementation.

sarvaniv: You mean in master branch? I don't care about this branch since we are just 2 people on this branch. But do you want me to first do this change on master so that we can merge both the changes together?

sarvaniv: Otherwise, when we merge into master, we need to remember an additional step that before we do anything on ESP branch, we provide the dummy implementation in ESP and then merge firebase branch into master.

sarvaniv: I am actually not completely sure what is the big deal since this will become a pure virtual function soon and in a way that will make everyone's lives easy and also without the pressure of having to time the merges for proxy and ESP. Just wondering the reason why this is such a concern?

contrib/endpoints/include/api_manager/request.h
lizan: I didn't notice that this is `firebase` branch. Feel free to merge this but I would block this when you try to merge to master.

This class here is a pure interface, so it shouldn't have any (even if it is dummy) implementations at any time.

Currently the only implementation of this interface is in ESP, and ESP depends on istio/proxy on specific commit. So I don't think making this pure virtual will break anything.

sarvaniv: While ESP depends on a specific commit, the general guideline for merging proxy code is that we point ESP to the local proxy and make sure that ESP tests also pass before we merge changes. But this change is a breaking change if we make it pure virtual right away. Is it OK to block peoples' progress? Instead we will have this temporary fix that will allow people to keep doing their work while we can work on any blocking issues that may come up during the merge. I think the first principle is to never block people and let things go smoothly for everyone. Sometimes perfection is not the best solution :-)

lizan: No, some refactoring cannot be done without breaking ESP, (e.g. #225 and ESP change). It is OK to merge them, then to mitigate the issue, have a dummy (of course if you have real one it is better) implementation ASAP in ESP too.

This is not a big deal, but comparing 2 solutions, (A) adding a dummy implementation first hand v.s. (B) changing this method to pure virtual are almost same effort (in terms of lines of change, number of PRs), when we do (B), it is hard to track whether the fix is made as nothing is breaking anymore.

sarvaniv: If this method does not become a pure virutal function within a few hours of merging ESP into master, please hold me responsible for that. We can convert it into a pure virtual function without breaking builds or trying to keep merges to ESP and proxy very close when there is at least 1-2 hour delay for tests to run. Given that we are merging into main from firebase,  this may take longer if we find any issues. I am hoping to make a strong case that we should be able to merge this code into master.

sarvaniv: The primary difference between A and B, is that the merges are smooth and no one is blocked since he merge cannot be instantaneous. I think making builds non-breaking is the best thing we can do to our fellow team members and should make this a strong requirement for all merges. I would personally dislike being blocked on build breaks.

lizan: the only case you're blocking is someone want to update the istio/proxy commit in ESP, since this is happening in firebase branch, so it doesn't block anyone.

sarvaniv: I am not talking about firebase branch. I am talking about the case when we merge into master. As I said this will be very temporary change and I will make sure that the method becomes pure virtual ASAP once we merge to master. I still don't know what is the issue here...

sarvaniv: Alright, I will make this pure virutal and before merging into master, I will have a dummy implementation that I will check into ESP branch. Thank you all for the input.

PR #247, opened 2017-04-14
build: HttpTemplate as separate target

lizan:
Envoy transcoding will depends on this.

### DIFF
--- a/contrib/endpoints/src/api_manager/BUILD
+++ b/contrib/endpoints/src/api_manager/BUILD
@@ -51,7 +51,6 @@ cc_library(
         "api_manager_impl.h",
         "config.h",
         "gce_metadata.h",
-        "http_template.h",
         "method_impl.h",
         "path_matcher.h",
         "path_matcher_node.h",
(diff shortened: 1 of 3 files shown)


contrib/endpoints/src/api_manager/BUILD
```diff
@@ -164,7 +174,7 @@ cc_test(
     ],
     linkstatic = 1,
     deps = [
-        ":api_manager",
+        ":http_template",
```
qiwzhang: You only need http_template?   How about patch_match files?

lizan: Right, `PathMatcher` and `PathMatcherBuilder` will be in next PR :) I'm trying to keep PRs as small as possible.

qiwzhang: They works together, so they should be in the same PR.

lizan: `HttpTemplate` doesn't have any dependency so it is easier to be separated, `PathMatcher` needs a lot of refactoring.

PR #248, opened 2017-04-14
refactor: remove service_name from PathMatcher

lizan:


contrib/endpoints/src/api_manager/path_matcher.cc
```diff
@@ -236,18 +236,10 @@ void ExtractBindingsFromQueryParameters(
 
 }  // namespace
 
-PathMatcher::PathMatcher(PathMatcherBuilder& builder)
-    : default_root_ptr_(builder.default_root_ptr_->Clone()),
-      strict_service_matching_(builder.strict_service_matching_),
+PathMatcher::PathMatcher(PathMatcherBuilder&& builder)
+    : root_ptr_(std::move(builder.root_ptr_)),
       custom_verbs_(builder.custom_verbs_),
```
qiwzhang: we can move custom_verbs too.

lizan: Done.

PR #250, opened 2017-04-17
Updated commit id of cloudendpoints/service-control-client-cxx

mangchiandjjoe:


contrib/endpoints/src/api_manager/service_control/aggregated_test.cc
```diff
@@ -312,20 +312,6 @@ TEST_F(QuotaAllocationTestWithRealClient, AllocateQuotaTest) {
                  [](Status status) { ASSERT_TRUE(status.ok()); });
 }
 
-TEST_F(QuotaAllocationTestWithRealClient, AllocateQuotaFailedTest) {
```
qiwzhang: It is better if we keep this test case.
First call is passed. I think second call should fail.

mangchiandjjoe: Fixed test case.

contrib/endpoints/src/api_manager/service_control/aggregated_test.cc
```diff
@@ -321,8 +321,11 @@ TEST_F(QuotaAllocationTestWithRealClient, AllocateQuotaFailedTest) {
   info.metric_cost_vector = &metric_cost_vector_;
 
   FillOperationInfo(&info);
-  sc_lib_->Quota(info, nullptr, [](Status status) {
-    ASSERT_TRUE(status.code() == Code::RESOURCE_EXHAUSTED);
+  sc_lib_->Quota(info, nullptr, [info, this](Status status) {
```
qiwzhang: Please add comment.
Quota cache always allows the first call. The negative will take effect
only after remote server call is replied.  In this case, it will be the second call.

PR #251, opened 2017-04-19
Refactor PathMatcher to an independent BUILD target

lizan:
- Templatized PathMatcher


contrib/endpoints/src/api_manager/path_matcher.h
```diff
@@ -129,6 +130,369 @@ class PathMatcherBuilder {
   friend class PathMatcher<Method>;
 };
 
+namespace {
+
+std::vector<std::string>& split(const std::string& s, char delim,
```
qiwzhang: These are not template functions, can they remain in cc file.   I hate to see duplicated code.

contrib/endpoints/src/api_manager/path_matcher.h
```diff
@@ -129,6 +130,369 @@ class PathMatcherBuilder {
...
+      *p++ = part[i];
+      i += 1;
+    }
+  }
+
+  unescaped.resize(p - begin);
+  return unescaped;
+}
+
+template <class VariableBinding>
```
qiwzhang: Are you going to use different VariableBinding type?   try to understand why you need to templatize this class.

contrib/endpoints/src/api_manager/path_matcher.h
```diff
@@ -120,10 +124,10 @@ class PathMatcherBuilder {
   // be multiple templates in different services on a server. Consider moving
   // this to PathMatcherNode.
   std::set<std::string> custom_verbs_;
-  typedef PathMatcher::MethodData MethodData;
+  typedef typename PathMatcher<Method>::MethodData MethodData;
```
qiwzhang: try to understand why this "typename" is needed here.

contrib/endpoints/src/api_manager/path_matcher.h
lizan: These are in unnamed namespace and referred by those templated methods. If you have them in cc file the linker would complain. Having them outside of unnamed namespace would pollute the namespace.

contrib/endpoints/src/api_manager/path_matcher.h
lizan: This is a temporary solution to make PathMatcher independent. I have a [TODO](https://github.com/lizan/proxy/blob/42990b8f82eb29bbf880aa3ca8d5b69aca9f7ba7/contrib/endpoints/src/api_manager/path_matcher.h#L56) to not template this.

contrib/endpoints/src/api_manager/path_matcher.h
lizan: Since compiler would complain without it :)

See http://stackoverflow.com/questions/18385418/c-meaning-of-a-statement-combining-typedef-and-typename

Issue #165, opened 2017-03-14
Attribute value extraction


jasminejaksic:
Extract attributes such as source.name.

ZackButcher:
Istio's canonical attributes are listed [here](https://github.com/istio/istio.github.io/blob/master/docs/reference/attributes.md). There's a few more that the proxy can provide but is not today:

* request.method - the HTTP method of the request, e.g. POST, PUT, GET, etc.
* request.referer - HTTP `Referer` header
* request.scheme - this can be hard coded to HTTP since attributes are produced in an HTTP filter, I think
* request.user-agent - HTTP `User-Agent` header
* request.id - "This should be a guid or a psuedo-guid with a low probability of collision in a temporal window measured in days or weeks."
* source.ip - "The IP address of the client that sent the request."
* source.port 
* origin.ip - "The IP address of the originating application."
* origin.port
* target.ip - "	The IP address the request was sent to."
* target.port

@smawson we need some clarification on the `source`/`origin` attributes: AFAIK `source.{ip, port}` should be extracted from TCP connection and `origin.{ip, port}` should be extracted from the `x-forward-for` headers. Is that correct?

We rely on the manager to populate `target.ip` and `target.port` in config it pushes to proxy, @kyessenov can you confirm?

smawson:
We could ignore origin.ip and origin.port for now if we want to, that's meant to be passed from the ingress proxy. So proxy-specific how it wants to do that, x-forward-for would be reasonable if it is supported by envoy easily.

source.ip/port are just from the connection, yeah.

ZackButcher:
Per a chat with @smawson we're going to punt on the `origin` attributes till we nail down ingress, but we should still try to get in the others for alpha. @qiwzhang is there anything that you know of in Envoy that'd make getting any of these hard? If not I might try my hand at adding them.

qiwzhang:
Hi @ZackButcher It will be great if you can help.  I am kind of short handed.   

it will be straight forward to get these request.* attributes.
source.ip we could get it from XFF, but not sure about source.port

It will be hard to get target.ip and port in Envoy.  @lizan or @kyessenov  may know how

ZackButcher:
So adding all of the request attributes is easy and I'll be sending out a PR implementing it soon. However, source.ip/port are harder because the mixerclient does not yet support IP_ADDRESS attributes. I'm eyeing up what it'd take to implement that in the mixerclient.

qiwzhang:
[...]

PR #256, opened 2017-04-20
Implement support for the remaining canonical request attributes.

ZackButcher:
Populate `request.{method, referer, scheme, user-agent}` attributes for `Check` calls.

We still need to implement `source.ip` and `source.port` but the info isn't in hand as we're assembling attributes to forward for the `Check` call. Today it's only readily available when we go to `Report`, and I'm not proficient enough to plumb it through the correct way.


ZackButcher:
https://github.com/istio/proxy/issues/165

PR #257, opened 2017-04-21
Handle possible white spaces in audiences string in service config.

sarvaniv:
Added service config string in tests that reproduces the issue without the fix.

### DIFF
--- a/contrib/endpoints/src/api_manager/check_auth_test.cc
+++ b/contrib/endpoints/src/api_manager/check_auth_test.cc
@@ -84,6 +84,53 @@ const char kServiceConfig[] =
     "  environment : \"http://127.0.0.1:8081\"\n"
     "}\n";
 
+const char kServiceConfigWithAudiences[] =
+    "name: \"endpoints-test1.cloudendpointsapis.com\"\n"
+    "authentication {\n"
+    "    providers: [\n"
+    "    {\n"
+    "      id: \"issuer1\"\n"
+    "      issuer: \"https://issuer1.com\"\n"
+    "      audiences: \"bad-audience, "
+    "http://endpoints-test.cloudendpointsapis.com\"\n"
+    "    },\n"
+    "    {\n"
+    "      id: \"openid_fail\"\n"
+    "      issuer: \"http://openid_fail\"\n"
+    "      audiences: \"bad-audience, endpoints-test.cloudendpointsapis.com\"\n"
+    "    },\n"
+    "    {\n"
+    "      id: \"issuer2\"\n"
+    "      issuer: \"https://issuer2.com\"\n"
+    "      jwks_uri: \"https://issuer2.com/pubkey\"\n"
+    "      audiences: \"bad-audience,endpoints-test.cloudendpointsapis.com\"\n"
+    "    }\n"
+    "    ],\n"
+    "    rules: {\n"
+    "      selector: \"ListShelves\"\n"
+    "      requirements: [\n"
+    "      {\n"
+    "        provider_id: \"issuer1\"\n"
+    "      },\n"
+    "      {\n"
+    "        provider_id: \"openid_fail\"\n"
+    "      },\n"
+    "      {\n"
+    "        provider_id: \"issuer2\"\n"
+    "      }\n"
+    "      ]\n"
+    "    }\n"
+    "}\n"
+    "http {\n"
+    "  rules {\n"
+    "    selector: \"ListShelves\"\n"
+    "    get: \"/ListShelves\"\n"
+    "  }\n"
+    "}\n"
+    "control {\n"
+    "  environment : \"http://127.0.0.1:8081\"\n"
+    "}\n";
+
 // Auth token generated with the following header and payload.
 //{
 // "alg": "RS256",
@@ -318,7 +365,7 @@ const char kUserInfo_kSub2_kIss[] =
 const char kUserInfo_kSub_kIss2[] =
     "eyJpc3N1ZXIiOiJodHRwczovL2lzc3VlcjIuY29tIiwiaWQiOiJlbmQtdXNlci1pZCJ9";
 
-class CheckAuthTest : public ::testing::Test {
+class CheckAuthTest : public ::testing::TestWithParam<std::string> {
  public:
   void SetUp() {
     std::unique_ptr<MockApiManagerEnvironment> env(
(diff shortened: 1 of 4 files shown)


contrib/endpoints/src/api_manager/method_impl.cc
```diff
@@ -51,17 +50,17 @@ void MethodInfoImpl::addAudiencesForIssuer(const string &issuer,
...
-  // Audience list is comma-delimited.
-  while (getline(ss, audience, ',')) {
-    if (!audience.empty()) {  // Only adds non-empty audience.
-      std::string aud = utils::GetUrlContent(audience);
-      if (!aud.empty()) {
-        audiences.insert(aud);
-      }
+
+  // Audience list is comma-delimited with possible white spaces.
+  char *tokens = strtok(const_cast<char *>(audiences_list.c_str()), " ,");
```
qiwzhang: this const_cast is dangerous.   I suspect strtok will modify the buffer.

contrib/endpoints/src/api_manager/check_auth_test.cc
```diff
@@ -594,23 +640,28 @@ TEST_F(CheckAuthTest, TestBadAudience) {
...
 // Positive test: audience is service name with http prefix and a trailing
 // slash.
-TEST_F(CheckAuthTest, TestHttpSlashAudience) {
+TEST_P(CheckAuthTest, TestHttpSlashAudience) {
   TestValidToken(kTokenHttpSlashAud);
 }
 
+INSTANTIATE_TEST_CASE_P(ServiceConfigsWithAndWithoutAudiencesCheck,
+                        CheckAuthTest,
+                        testing::Values(kServiceConfig,
```
qiwzhang: we only need one test case to test space,  do we need all tests to cover it?

sarvaniv: Actually, there are multiple tests that check for various audience fields in the token. This will just ensure that there are no regressions with this fix.

contrib/endpoints/src/api_manager/method_impl.cc
sarvaniv: Very nice catch. As discussed, I will pass this by value.

contrib/endpoints/src/api_manager/check_auth_test.cc
qiwzhang: then you and new comers have to make sure these two configs are highly related.  if you want to modify one, you have to modify another.   I don't like such tests.  Just testing a simple audience parsing,  you have to make the whole test so complicated.

contrib/endpoints/src/api_manager/method_impl.cc
qiwzhang: this still dangerous.  std::string will not expect anyone modifying its buffer.  Even you have your own copy of string,  internally, it could still use ref_count, since it is not expecting you to modify the buffer.

You can create a
char buf[str.size() + 1);

use that.

contrib/endpoints/src/api_manager/check_auth_test.cc
sarvaniv: sure. I understand your concern. Let me see if I can just add a new test with just an update to existing configuration.

contrib/endpoints/src/api_manager/method_impl.cc
sarvaniv: Maybe I am missing something here. I thought string are mutable in C++. Also, this is a call be value so I already have a copy of the original string that I can change. You mean you want me to pass by reference and then create another char buffer? I am unable to see the difference unless you think somethin like: str[i] = 'x' is dangerous?

sarvaniv: I take my words back. Looks like modifying a const char* returned by c_str() results in undefined behavior. I will fix this. Thank you for pointing this.

sarvaniv: Actually, there is code in fetch_metadata.cc that is already doing this:

 esp_get_service_account_auth_token(
                          const_cast<char *>(body.data()), body.length(), ...)

The code passes this as input to grpc parser which only accepts a char * and NOT a const char *. Do we also need to fix this code?

sarvaniv: The documentation for grpc clearly states that the input is modified.

/* The next two functions are going to parse the input string, and
 * modify it in the process, in order to use its space to store
 * all of the keys and values for the returned object tree.
 *
 * They assume UTF-8 input stream, and will output UTF-8 encoded
 * strings in the tree. The input stream's UTF-8 isn't validated,
 * as in, what you input is what you get as an output.
 *
 * All the keys and values in the grpc_json objects will be strings
 * pointing at your input buffer.
 *
 * Delete the allocated tree afterward using grpc_json_destroy().
 */
grpc_json* grpc_json_parse_string_with_len(char* input, size_t size);

sarvaniv: I think we have to fix the code in fetch_metadata.cc. This is indeed bad behavior. Thank you for pointing out and insisting on the change. Much appreciated :-)

contrib/endpoints/src/api_manager/method_impl.cc
```diff
@@ -51,17 +50,19 @@ void MethodInfoImpl::addAudiencesForIssuer(const string &issuer,
...
-  stringstream ss(audiences_list);
-  string audience;
-  // Audience list is comma-delimited.
-  while (getline(ss, audience, ',')) {
-    if (!audience.empty()) {  // Only adds non-empty audience.
-      std::string aud = utils::GetUrlContent(audience);
-      if (!aud.empty()) {
-        audiences.insert(aud);
-      }
+  char *audiences_copy = new char[audiences_list.length()+1];
```
qiwzhang: memory leak.  
you can allocate it from stack by 

char buf[str.slze()  + 1];

contrib/endpoints/src/api_manager/method_impl.cc
```diff
@@ -51,17 +50,19 @@ void MethodInfoImpl::addAudiencesForIssuer(const string &issuer,
...
+  std::strcpy(audiences_copy, audiences_list.c_str());
```
qiwzhang: prefer strncpy  or memcpy

PR #258, opened 2017-04-24
Merge firebase to master

sarvaniv:


.bazelrc.jenkins
```diff
@@ -0,0 +1,8 @@
+# This is from Bazel's former travis setup, to avoid blowing up the RAM usage.
+startup --host_jvm_args=-Xmx8192m
+startup --host_jvm_args=-Xms8192m
```
qiwzhang: this is not firebase related.  where is it coming from

PR #265, opened 2017-04-25
[DO NOT MERGE] Update envoy

lizan:
bazel clean jenkins to test


kyessenov:
Manager is not going to work with recent envoy, so you can ignore that error. Is Travis still supported?

lizan:
Let's see last change in #260 will work for travis or not.

PR #266, opened 2017-04-26
transcoding: support path binding

lizan:


lizan:
Not yet, I will write one in follow up PRs.

src/envoy/transcoding/config.cc
```diff
@@ -82,49 +84,103 @@ Config::Config(const Json::Object& config, Server::Instance& server) {
...
+          break;
+        default:
+          break;
+      }
+
+      pmb.Register("POST", "/" + service->full_name() + "/" + method->name(),
+                   "", method_info);
+    }
+  }
+
```
qiwzhang: It is better to log a error if no methods added.  It is to detect config error, such as a typo in the services field or missing services field.

qiwzhang: or even throw an exception

lizan: throw an exception for services not in descriptor.

PR #271, opened 2017-04-27
Ensure that attributes are sent with the Quota call

mandarjog:
quota() needs all the attributes that check() needs
Operator may apply conditional quota using attributes in selectors.

Note:
This effectively disables quota caching and prefetching.
Quota caching was "off" by default thru config anyway, so that is not a big problem.

mixerClient::GenerateSignature should be updated to exclude non identifying attributes.

### DIFF
--- a/src/envoy/mixer/http_control.cc
+++ b/src/envoy/mixer/http_control.cc
@@ -210,11 +210,19 @@ void HttpControl::Check(HttpRequestDataPtr request_data, HeaderMap& headers,
   SetStringAttribute(kOriginUser, origin_user, &request_data->attributes);
   log().debug("Send Check: {}", request_data->attributes.DebugString());
 
-  auto check_on_done = [this, on_done](const Status& status) {
+  auto check_on_done = [this, on_done, request_data](const Status& status) {
     if (status.ok()) {
       if (!quota_attributes_.attributes.empty()) {
-        log().debug("Send Quota: {}", quota_attributes_.DebugString());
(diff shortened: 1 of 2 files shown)


mandarjog:
Related to https://github.com/istio/mixer/issues/574

src/envoy/mixer/http_control.cc
```diff
@@ -210,11 +210,19 @@ void HttpControl::Check(HttpRequestDataPtr request_data, HeaderMap& headers,
...
   log().debug("Send Check: {}", request_data->attributes.DebugString());
 
-  auto check_on_done = [this, on_done](const Status& status) {
+  auto check_on_done = [this, on_done, request_data](const Status& status) {
     if (status.ok()) {
       if (!quota_attributes_.attributes.empty()) {
-        log().debug("Send Quota: {}", quota_attributes_.DebugString());
-        mixer_client_->Quota(quota_attributes_, on_done);
+        for (const auto& attribute : quota_attributes_.attributes) {
+          request_data->attributes.attributes[attribute.first] =
```
qiwzhang: this request_data->attributes is send to Report call too.  Do we want to add quota metrics in the report call.  If so,  we are fine.  Otherwise, we may want to clone a new attribute for Quota Call.

mandarjog: I think it is ok to send those in report call too. 

At present we need this to get rate limits working. We need to reopen the caching discussion.

PR #274, opened 2017-04-27
set consumer_id to producer project id for Quota call

mangchiandjjoe:
Another PR  for adding more tests on cloudendpoints/esp  will follow.


qiwzhang:
Please don't put bug id in the PR.  bug id is internal.

mangchiandjjoe:
closing the PR and will create another one

PR #275, opened 2017-04-27
set consumer_id to producer project id for Quota call

mangchiandjjoe:


contrib/endpoints/src/api_manager/service_control/testdata/allocate_quota_request_without_api_key.golden
```diff
@@ -0,0 +1,36 @@
...
+      int64_value: 1
+    }
+  }
+  quota_metrics {
+    metric_name: "metric_second"
+    metric_values {
+      int64_value: 2
+    }
+  }
+  quota_mode: BEST_EFFORT
```
qiwzhang: This can be a separate PR.

Just realized that quota_mode should NOT be fixed as BEST_EFFORT.
Only use BEST_EFFORT with cache and with aggregated tokens.
1) no cache case, it is NORMAL.
2) first call out of new cache item,  or negative,  it is NORMAL.

mangchiandjjoe: I will create another PR for this. BTW, do we have a document for this?

qiwzhang: It is in the proto file.

mangchiandjjoe:
Created a new PR including this

https://github.com/istio/proxy/pull/289

qiwzhang:
Still prefer to have separate PR for separate feature.

PR #277, opened 2017-04-28
add global context

qiwzhang:
This is in preparation for service rollout feature to support multiple services.
==Create a GlobalContext object and move global objects from ServiceContext.
== Server_config is stored in GlobalContext and in Config by using share_ptr.

It seems that incremental change can be done in the master branch without impacting basic features.  So I am trying to develop this feature in the master branch.

### DIFF
--- a/contrib/endpoints/src/api_manager/api_manager_impl.cc
+++ b/contrib/endpoints/src/api_manager/api_manager_impl.cc
@@ -24,23 +24,17 @@ using ::google::api_manager::proto::ServerConfig;
 namespace google {
 namespace api_manager {
 
-namespace {
-
-std::shared_ptr<ApiManager> CreateApiManager(
-    std::unique_ptr<ApiManagerEnvInterface> env,
-    std::unique_ptr<Config> config) {
-  return std::shared_ptr<ApiManager>(
-      new ApiManagerImpl(std::move(env), std::move(config)));
-}
-
-}  // namespace
-
 ApiManagerImpl::ApiManagerImpl(std::unique_ptr<ApiManagerEnvInterface> env,
-                               std::unique_ptr<Config> config)
-    : service_context_(
-          new context::ServiceContext(std::move(env), std::move(config))) {
+                               const std::string& server_config)
+    : global_context_(
+          new context::GlobalContext(std::move(env), server_config)) {
   check_workflow_ = std::unique_ptr<CheckWorkflow>(new CheckWorkflow);
   check_workflow_->RegisterAll();
+}
+
+void ApiManagerImpl::AddConfig(std::unique_ptr<Config> config) {
+  service_context_ = std::make_shared<context::ServiceContext>(
+      global_context_, std::move(config));
 }
 
 std::unique_ptr<RequestHandlerInterface> ApiManagerImpl::CreateRequestHandler(
@@ -52,8 +46,7 @@ std::unique_ptr<RequestHandlerInterface> ApiManagerImpl::CreateRequestHandler(
 std::shared_ptr<ApiManager> ApiManagerFactory::GetOrCreateApiManager(
     std::unique_ptr<ApiManagerEnvInterface> env,
     const std::string& service_config, const std::string& server_config) {
-  std::unique_ptr<Config> config =
-      Config::Create(env.get(), service_config, server_config);
+  std::unique_ptr<Config> config = Config::Create(env.get(), service_config);
   if (config == nullptr) {
     return nullptr;
   }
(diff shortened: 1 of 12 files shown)


contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -24,25 +24,19 @@ using ::google::api_manager::proto::ServerConfig;
...
-      new ApiManagerImpl(std::move(env), std::move(config)));
-}
-
-}  // namespace
-
 ApiManagerImpl::ApiManagerImpl(std::unique_ptr<ApiManagerEnvInterface> env,
-                               std::unique_ptr<Config> config)
-    : service_context_(
-          new context::ServiceContext(std::move(env), std::move(config))) {
+                               const std::string& server_config)
```
sarvaniv: I am not sure if server_config should be a part of global config. A bunch of the configuration seem to be service configuration related [which can override service configuration] and we don't want to override all services with a server configuration override.

qiwzhang: You can have only one server_config.   It is suppose to override all service configs.

contrib/endpoints/src/api_manager/config.h
```diff
@@ -40,9 +40,19 @@ class Config {
   // any of protobuf format json, binary or text. It can be nullptr if there is
   // not server_config.
   static std::unique_ptr<Config> Create(ApiManagerEnvInterface *env,
+                                        const std::string &service_config);
+  // For unit test only
```
sarvaniv: I am wary of code that is marked as "for test only". This usually means the test has a different code path than what is used in production.

qiwzhang: Just the object constructions.   test codes are using a lot of mocking.  this is to make of object creating easier.  it should not impact the main feature testing.

contrib/endpoints/src/api_manager/context/service_context.cc
```diff
@@ -24,57 +23,25 @@ namespace context {
...
-      config_->server_config()->has_service_control_config() &&
-      config_->server_config()
-          ->service_control_config()
-          .intermediate_report_min_interval()) {
-    intermediate_report_interval_ = config_->server_config()
-                                        ->service_control_config()
-                                        .intermediate_report_min_interval();
-  }
+      service_control_(CreateInterface()) {
+  config_->set_server_config(global_context_->server_config());
```
sarvaniv: Why do you need this method? The global context should be available when config is built right? The reason I ask is the number of places we need to maintain the server_config. It is available in global_context, so why not have config_ object built with global_context since that is where the env_ object is also present and we don't have to keep passing the configs around. That way, if server_config is updated, it is updated in the global context and everyone who know the global context will see the new server config.

contrib/endpoints/src/api_manager/context/service_context.cc
```diff
@@ -93,59 +60,18 @@ MethodCallInfo ServiceContext::GetMethodCallInfo(
...
-    if (tracing_config.has_samling_config()) {
-      minimum_qps = tracing_config.samling_config().minimum_qps();
-    }
-  }
-
-  return std::unique_ptr<cloud_trace::Aggregator>(new cloud_trace::Aggregator(
-      &service_account_token_, url, aggregate_time_millisec, cache_max_size,
-      minimum_qps, env_.get()));
+      service_control::Aggregated::Create(
+          config_->service(), global_context_->server_config().get(), env(),
```
sarvaniv: Just pass in the global_context here instead of parts of it.

qiwzhang: Same circular dependency issue.  service_control/*  can depends on api_manager/*  and context/*,  but not the other way around.  I could not pass GlobalContext to service_control objects.

qiwzhang: To be exactly,  service_control/* can depend on api_manager/*   but not api_manager/context/*

contrib/endpoints/src/api_manager/context/global_context.h
```diff
@@ -0,0 +1,117 @@
...
+  ApiManagerEnvInterface *env() { return env_.get(); }
+
+  // the service account token store
+  auth::ServiceAccountToken *service_account_token() {
+    return &service_account_token_;
+  }
+
+  // certs and jwt_cache
+  auth::Certs &certs() { return certs_; }
+  auth::JwtCache &jwt_cache() { return jwt_cache_; }
```
lizan: `Certs` and `JwtCache` shouldn't live here, different service have different trusted public keys, thus they are bounded to service configs.

qiwzhang: Done.  thanks

contrib/endpoints/src/api_manager/context/service_context.cc
qiwzhang: There are places we don't have GlobalContext when creating config object (in api_manager_impl.cc).  It can be refactored, but I don't want this PR to have other refractory code.

PR #280, opened 2017-05-01
Pass shared_ptr to check callback.

qiwzhang:
Under load, the filter object could be destroyed before Check callback is called if Mixer check is taking too long. This fix is to pass a shared_ptr to the callback function to make sure the object is adding ref_count and object is not destroyed.

### DIFF
--- a/src/envoy/mixer/http_filter.cc
+++ b/src/envoy/mixer/http_filter.cc
@@ -122,7 +122,8 @@ class Config : public Logger::Loggable<Logger::Id::http> {
 typedef std::shared_ptr<Config> ConfigPtr;
 
 class Instance : public Http::StreamDecoderFilter,
-                 public Http::AccessLog::Instance {
+                 public Http::AccessLog::Instance,
+                 public std::enable_shared_from_this<Instance> {
  private:
   std::shared_ptr<HttpControl> http_control_;
   ConfigPtr config_;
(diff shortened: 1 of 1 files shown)


mandarjog:
Can you add description about what was the root cause and how this is fixing it ?
why it happened under load ?

kyessenov:
Is the issue here that the check call outlives the filter? Why would proxy destroy a filter in progress?

qiwzhang:
Maybe due to downstream connection reset.  It seems that decoder_callback object is gone.  So I think the Envoy connectionManager is destroying everything.

kyessenov:
would you mind asking Matt on gitter if using shared pointers right here? I'm sure this will be a common problem for various filters that folks would be contributing to Envoy.

qiwzhang:
I may have found the root cause.   We did not handle stream reset correctly.  If stream reset, we have to abort the callback.

kyessenov:
we gotta try fault injection to explore the error space in mixer-proxy protocol :D

mandarjog:
@qiwzhang do you still need to use `public std::enable_shared_from_this<Instance>` now that you are checking for `responded stream`.

qiwzhang:
Yes, to make sure my Instance object not get deleted.  The callback function is checking its state_ variable.  If it is deleted, it will be checking the random data.

PR #282, opened 2017-05-03
Use a struct to create ApiManager.

qiwzhang:
* Define a struct ApiManagerInput to pass in all data to create ApiManager.
* Remove the map in the ApiManagerFactory.


qiwzhang:
Please don't review this

PR #290, opened 2017-05-04
Added /producer/by_consumer metrics.

mangchiandjjoe:


mandarjog:
@louiscryan  @qiwzhang Does `contrib/endpoints` code need to be in the proxy repo? 
Looks like it could be a separate repo which can depend on proxy.

qiwzhang:
current plan is: @lizan will move grpc transcoding code to a separate grpc repo so Istio/proxy (envoy) can use it.  After that,  we will move api_manager code (for ESP) back to the private ESP repo.

kyessenov:
Before we start moving large amount of code around again, can we decide if we ever need API management features in the proxy. I understand that the two core piece apart from transcoding are monitoring/logging per method and authentication/authorization. Are we confident we can do both purely on the mixer side?

qiwzhang:
@kyessenov you are right,  beside grpc transcoding,  there are auth, and calling Chemist directly.  we need to answer:  where will jwt auth checking be for istio/proxy?   will istio/proxy ever call Chemist directly?

PR #292, opened 2017-05-04
Add stress test for envoy.

qiwzhang:


src/envoy/mixer/integration_test/mixer_server.go
```diff
@@ -32,14 +32,16 @@ import (
...
 	count    int
 	r_status rpc.Status
 }
 
-func newHandler() *Handler {
+func newHandler(stress bool) *Handler {
 	return &Handler{
+		stress:   stress,
 		bag:      nil,
 		ch:       make(chan int, 10), // Allow maximum 10 requests
```
mandarjog: probably don't need to create a channel during `stress` ?

src/envoy/mixer/integration_test/stress_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	defer s.TearDown()
+
+	url := fmt.Sprintf("http://localhost:%d/echo", ClientProxyPort)
+
+	var count uint64 = 0
+	for k := 0; k < concurrent; k++ {
+		go func() {
+			for true {
+				if _, err := http.Get(url); err != nil {
```
mandarjog: is this creating 1 connection per thread?

Can you use `wrk` to perform the actual stress test? 
"https://storage.googleapis.com/istio-build-deps/wrk-linux"

kyessenov: You need to set http context deadline or request timeout. Go default http request has no deadline.

qiwzhang: I am not sure how go http package implement http client.  Since I did not re-use http.Client object, I believe each request uses its own connection.   This is even better since some failure condition is caused by the down-stream connection reset.  So this "stress" test is not maximum stress.  It is more stress on error conditions.

For now, go client can issue 3K qps.   That is good enough stress.  not need to switch to use wrk.

kyessenov: just suggesting settings _some_ deadline, since `http.Get` will just hang forever if server is unresponsive.

src/envoy/mixer/integration_test/mixer_server.go
qiwzhang: Done

src/envoy/mixer/integration_test/stress_test.go
qiwzhang: Done added timeout

PR #293, opened 2017-05-08
add weighted selector.

qiwzhang:


contrib/endpoints/src/api_manager/weighted_selector.cc
```diff
@@ -0,0 +1,53 @@
...
+  counts_.resize(list_.size());
+  std::fill(counts_.begin(), counts_.end(), 0);
+}
+
+float WeightedSelector::score(int index) {
+  return float(counts_[index]) / list_[index].second;
+}
+
+const std::string& WeightedSelector::Select() {
+  if (list_.size() == 1) {
```
sarvaniv: What if the list were empty? There is no checks to ensure that an empty list is no a valid input.

contrib/endpoints/src/api_manager/weighted_selector.cc
```diff
@@ -0,0 +1,53 @@
...
+    return list_[0].first;
+  }
+
+  auto min_score = score(0);
```
sarvaniv: This assumes list cannot be empty.

contrib/endpoints/src/api_manager/weighted_selector.cc
```diff
@@ -0,0 +1,53 @@
...
+  std::fill(counts_.begin(), counts_.end(), 0);
+}
+
+float WeightedSelector::score(int index) {
+  return float(counts_[index]) / list_[index].second;
+}
+
+const std::string& WeightedSelector::Select() {
+  if (list_.size() == 1) {
+    return list_[0].first;
```
sarvaniv: Do you not have to increment counts in this case?

contrib/endpoints/src/api_manager/weighted_selector.cc
sarvaniv: I don't think you need the == 1 check. The code should work even if you don't have the check. Rather you should check for an empty list instead here.

qiwzhang: Added check with size=0 case,  special handling of size=1 is for optimization.

sarvaniv: But then you are not incrementing counts when size is 1. Is this intentional?

qiwzhang: Added check with size=0 case,  special handling of size=1 is for optimization

contrib/endpoints/src/api_manager/weighted_selector.cc
qiwzhang: size=0 case should be handled before here.

qiwzhang:
Yes, we don't need to update the count for size=1. it is not used.

PR #302, opened 2017-05-10
[DO NOT MERGE] temporary PR for initial review

mangchiandjjoe:


contrib/endpoints/src/api_manager/context/service_context.h
```diff
@@ -108,6 +131,12 @@ class ServiceContext {
 
   // The service control object.
   std::unique_ptr<service_control::Interface> service_control_;
+
```
qiwzhang: not need to track using instances for a service config.  My current thinking is:  we can just leave the service in the memory even nobody is using it,   it will not take too much memory since client cache will be empty after idle for a white, and all items should be removed on expiration.

contrib/endpoints/src/api_manager/api_manager_impl.h
```diff
@@ -73,6 +79,9 @@ class ApiManagerImpl : public ApiManager {
 
   // A weighted service selector.
   std::unique_ptr<WeightedSelector> service_selector_;
+
+  // Access control to service_context_map_
+  std::mutex service_context_mutex_;
```
qiwzhang: Most of api-manager code are not supposed to run in multiple thread environment.  Nginx is single threaded.  We need to do a lot of code hardening if it is moved to multiple threaded environment. We need a overall design to make api-manager thread safe.

mangchiandjjoe:
I got it. Let me cancel the PR which is not necessary for now.

PR #304, opened 2017-05-11
Basic tests for transcoding

lizan:


src/envoy/transcoding/config.cc
```diff
@@ -135,8 +135,9 @@ Config::Config(const Json::Object& config, Server::Instance& server) {
           break;
       }
 
-      pmb.Register("POST", "/" + service->full_name() + "/" + method->name(),
-                   "", method_info);
+      //      pmb.Register("POST", "/" + service->full_name() + "/" +
```
qiwzhang: why do we comment this out?

lizan: This was a debug registration before supporting path binding. Removed.

src/envoy/transcoding/config.cc
```diff
@@ -134,10 +134,6 @@ Config::Config(const Json::Object& config) {
         default:
           break;
       }
-
-      //      pmb.Register("POST", "/" + service->full_name() + "/" +
```
qiwzhang: So method->options().GetExentsion() is always there even you did not specify the "options" in the proto file.  Is that right?

lizan: if you don't have options in the proto, GetExtension() will just return a default instance of HttpRule, which is fine.

PR #305, opened 2017-05-12
transcoding: Handle RequestStatus error

lizan:


lizan:
@qiwzhang now this is ready. PTAL

lizan:
Add some unit tests. More integration tests (based on ESP t-tests) will be in follow up PR.

PR #314, opened 2017-05-15
transcoding: Basic integration test

lizan:


src/envoy/transcoding/integration_test.cc
```diff
@@ -0,0 +1,101 @@
...
+ public:
+  /**
+   * Global initializer for all integration tests.
+   */
+  void SetUp() override {
+    fake_upstreams_.emplace_back(
+        new FakeUpstream(0, FakeHttpConnection::Type::HTTP2));
+    registerPort("upstream_0",
+                 fake_upstreams_.back()->localAddress()->ip()->port());
+    createTestServer("src/envoy/transcoding/test/integration.json", {"http"});
```
qiwzhang: This is Envoy way of doing integration test.  It is not a true integration test.   Ideally,  we should run Envoy binary, backend and http client as separate processes.  Similar to //src/envoy/mixer/integration_test/ which is using go to write client and server and testing code.

src/envoy/transcoding/integration_test.cc
```diff
@@ -0,0 +1,101 @@
...
+
+  auto response_data = Grpc::Common::serializeBody(response_pb);
+  request_stream->encodeData(*response_data, false);
+
+  request_stream->encodeTrailers(
+      Http::TestHeaderMapImpl{{"grpc-status", "0"}, {"grpc-message", ""}});
+
+  response->waitForEndStream();
+  EXPECT_TRUE(request_stream->complete());
+  EXPECT_TRUE(response->complete());
```
qiwzhang: did you verify the response as JSON format after transcode?

src/envoy/transcoding/integration_test.cc
lizan: This is intended, since all the code in src/envoy/transcoding will be moved to lyft/envoy repo. Taking dependency on our go integration framework doesn't make sense.

src/envoy/transcoding/integration_test.cc
lizan: Done

PR #315, opened 2017-05-15
Update gRPC to v1.3.2

lizan:


repositories.bzl
```diff
@@ -174,3 +174,266 @@ def transcoding_repositories(bind=True):
...
+            name = "grpc++",
+            actual = "@grpc_git//:grpc++",
+        )
+
+        native.bind(
+            name = "grpc_lib",
+            actual = "@grpc_git//:grpc++_codegen_proto",
+        )
+
+def googleapis_repositories(protobuf_repo="@protobuf_git//", bind=True):
```
qiwzhang: I don't see places using this function.

lizan: in contrib/endpoints/repositoreis.bzl for service_control_client, but transcoding filter also use this repo, so moved to top level.

qiwzhang: I don't see the transcoding using it.  Did you forget to include a file?

lizan: https://github.com/istio/proxy/blob/master/src/envoy/transcoding/BUILD#L54

transcoding is using the repo, not function.

To avoid confusion, moved the loading part to top level WORKSPACE.

costinm:
Please hold - my build is crashing under load. I'll try another clean build.

( to repro: run with "wrk -c 1000 -t 100 -d 30s ", with mixer enabled)

kyessenov:
there's breaking filter API change coming from envoy, wait for matt to finish it before rebasing?

lizan:
@kyessenov I'm aware of the filter API change, which is not related to this PR though. I will make follow up changes to comply with that.

costinm:
It seems the crash happens without this as well, will file a separate issue.

PR #317, opened 2017-05-16
Fixed proxy crash on callQuota.

qiwzhang:
Need to check state to make sure it is not reset yet.


mandarjog:
Can we add a wrk test or similar to put it under some load ?

I know it won't be a real load test,  but Better than nothing .

Using "iPhone 9";

qiwzhang:
Will do, created https://github.com/istio/proxy/issues/318 to track it.

PR #328, opened 2017-05-18
Fix copyright

geeknoid:


LICENSE
```diff
@@ -188,7 +188,7 @@ Copyright 2016 Google Inc. All rights reserved.
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2016 Google Inc
+   Copyright 2017 Istio Authors
```
lizan: - No need to update the year
- Change line 1 too

PR #339, opened 2017-05-23
transcoding: Buffer unary response and add status to header

lizan:


src/envoy/transcoding/filter.cc
```diff
@@ -140,13 +142,17 @@ void Instance::setDecoderFilterCallbacks(
...
-  log().debug("Transcoding::Instance::encodeHeaders");
+  log().debug("Transcoding::Instance::encodeHeaders {}", end_stream);
   if (error_) {
     return Http::FilterHeadersStatus::Continue;
   }
 
   if (transcoder_) {
+    response_headers_ = &headers;
     headers.insertContentType().value(kJsonContentType);
+    if (!method_->server_streaming() && !end_stream) {
```
qiwzhang: Is it possible method_ is still nullptr?

src/envoy/transcoding/filter.cc
```diff
@@ -184,6 +193,29 @@ Http::FilterTrailersStatus Instance::encodeTrailers(Http::HeaderMap& trailers) {
     if (data.length()) {
       encoder_callbacks_->addEncodedData(data);
     }
+
+    if (!method_->server_streaming()) {
+      const Http::HeaderEntry* grpc_status_header = trailers.GrpcStatus();
+      if (grpc_status_header) {
+        uint64_t grpc_status_code;
+        if (!StringUtil::atoul(grpc_status_header->value().c_str(),
```
qiwzhang: what is this line for?   just check status_code is integer?   after atoul, grpc_status_code is not used.

src/envoy/transcoding/filter.cc
lizan: not when transcoder_ is not nullptr.

src/envoy/transcoding/filter.cc
lizan: yes

PR #343, opened 2017-05-30
ApiManager uses ConfigManager to download the service config

mangchiandjjoe:


contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -25,43 +25,53 @@ ApiManagerImpl::ApiManagerImpl(std::unique_ptr<ApiManagerEnvInterface> env,
...
   }
 
   check_workflow_ = std::unique_ptr<CheckWorkflow>(new CheckWorkflow);
   check_workflow_->RegisterAll();
 }
 
-void ApiManagerImpl::AddConfig(const std::string &service_config,
-                               bool deploy_it) {
+bool ApiManagerImpl::AddConfig(const std::string &service_config,
+                               std::string &config_id) {
```
qiwzhang: per style guide,  passing data out should use *, not use &

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -75,11 +85,41 @@ utils::Status ApiManagerImpl::Init() {
     global_context_->cloud_trace_aggregator()->Init();
   }
 
-  for (auto it : service_context_map_) {
-    if (it.second->service_control()) {
-      it.second->service_control()->Init();
-    }
+  if (service_context_map_.size() > 0) {
+    // Already initialized
+    return utils::Status::OK;
```
qiwzhang: this is wrong.   We have to call service_context->service_control()->Init() for each service_context. 

These sevice_context was created at master process.  (at line 61)

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -75,11 +85,41 @@ utils::Status ApiManagerImpl::Init() {
...
+            if (AddConfig(item.first, config_id)) {
+              rollouts.push_back({config_id, item.second});
+            } else {
+              initialization_status_ =
+                  utils::Status(Code::ABORTED, "Invalid service config");
+              return;
+            }
+          }
+
+          for (auto it : service_context_map_) {
```
qiwzhang: not to call all service_contexts.  only ones created by this call.   It is better to call it at line 103

contrib/endpoints/src/api_manager/api_manager_impl.h
```diff
@@ -73,6 +81,15 @@ class ApiManagerImpl : public ApiManager {
...
   std::unique_ptr<WeightedSelector> service_selector_;
+
+  // A config manager
+  std::unique_ptr<ConfigManager> config_manager_;
+
+  // Initialization status.
+  //  - Code::UNKNOWN   Not initialized yet. The default value.
+  //  - Code::OK        Successfully initialized
+  //  - Code::ABORTED   Initialization was failed
+  utils::Status initialization_status_;
```
qiwzhang: let us call it 

init_service_config_loading_status

contrib/endpoints/src/api_manager/api_manager_test.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Google Inc. All Rights Reserved.
+// Copyright 2017 Google Inc. All Rights Reserved.
```
qiwzhang: again, only new files use 2017

contrib/endpoints/include/api_manager/api_manager.h
```diff
@@ -87,6 +87,12 @@ class ApiManager {
   virtual utils::Status GetStatistics(
       ApiManagerStatistics *statistics) const = 0;
 
+  // Returns the ApiManager initialization status.
+  //  - Code::UNKNOWN   Not initialized yet. The default value.
+  //  - Code::OK        Successfully initialized
+  //  - Code::ABORTED   Initialization was failed
+  virtual utils::Status InitializationStatus() = 0;
```
mangchiandjjoe: nginx module is checking the status.
While the ApiManager is waiting for the initialization of the ConfigManager, the status is UNKNOWN and nginx posts the event to handle it later.
If the status is ABORTED, that means ApiManager was failed to load service configs. module returns NGX_DECLINED.
If it is OK, then  ctx->request_handler->Check will be called.

Before the initialization, service_context_map_ is empty.

qiwzhang: The interface between Nginx and api_manager should be as simple as possible.  This should be done by request_handler::Check().

BTW,  if initial_status is for the status of api_manager constructor,   we don't need it.   We only care about the one called by Init().

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Changed

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Moved the initialization into Init() from Constructor

contrib/endpoints/src/api_manager/api_manager_impl.h
mangchiandjjoe: Changed

contrib/endpoints/src/api_manager/api_manager_test.cc
mangchiandjjoe: Reverted

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Google Inc. All Rights Reserved.
+// Copyright 2017 Google Inc. All Rights Reserved.
```
mangchiandjjoe: Reverted

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Moved initialization into AddConfig()

mangchiandjjoe:
I am still working on the PendingRequestHandler class.

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -25,43 +25,45 @@ ApiManagerImpl::ApiManagerImpl(std::unique_ptr<ApiManagerEnvInterface> env,
                                const std::string &service_config,
                                const std::string &server_config)
     : global_context_(
-          new context::GlobalContext(std::move(env), server_config)) {
-  if (!service_config.empty()) {
-    AddConfig(service_config, true);
-  }
-
+          new context::GlobalContext(std::move(env), server_config)),
+      config_manager_(nullptr),
```
qiwzhang: no need to init smart pointer.

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -75,11 +77,43 @@ utils::Status ApiManagerImpl::Init() {
...
+    std::string config_id;
+    if (AddConfig(service_config_, &config_id)) {
+      DeployConfigs({{config_id, 100}});
+
+      config_loading_status_ = utils::Status::OK;
+    } else {
+      config_loading_status_ =
+          utils::Status(Code::ABORTED, "Invalid service config");
     }
+    return utils::Status::OK;
```
qiwzhang: return config_loading_status_?

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -25,43 +25,45 @@ ApiManagerImpl::ApiManagerImpl(std::unique_ptr<ApiManagerEnvInterface> env,
...
+  } else {
+    if (service_name != global_context_->service_name()) {
+      auto err_msg = std::string("Mismatched service name; existing: ") +
+                     global_context_->service_name() + ", new: " + service_name;
+      global_context_->env()->LogError(err_msg);
+      return false;
     }
   }
+
+  config_id->assign(config->service().id());
```
qiwzhang: just

*config_id = config_service().id();

is easier to read

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -25,43 +25,45 @@ ApiManagerImpl::ApiManagerImpl(std::unique_ptr<ApiManagerEnvInterface> env,
...
+
+  service_context_map_[*config_id] = std::make_shared<context::ServiceContext>(
```
qiwzhang: to avoid a map look up.   use

auto context_service = make_shared
context_service->Init()

then add it to the map.

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -75,11 +77,43 @@ utils::Status ApiManagerImpl::Init() {
...
+      global_context_,
+      [this](const utils::Status &status,
+             const std::vector<std::pair<std::string, int>> &configs) {
+        if (status.ok()) {
+          std::vector<std::pair<std::string, int>> rollouts;
+          std::string config_id;
+
+          for (auto item : configs) {
+            if (AddConfig(item.first, &config_id)) {
+              rollouts.push_back({config_id, item.second});
```
qiwzhang: move line 99 to above 102

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -75,11 +77,43 @@ utils::Status ApiManagerImpl::Init() {
...
+            } else {
+              config_loading_status_ =
+                  utils::Status(Code::ABORTED, "Invalid service config");
```
qiwzhang: if one fails, status maybe ok,  if all fails,  then mark it fail

qiwzhang:
Try to see if it is possible to use a separate PR for request_hander change and the callback handing from config_manager.

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Removed else then check rollouts.size() == 0

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Moved

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Modified

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Modified

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Removed

mangchiandjjoe:
This version of ApiManager is not compatible with  esp at the moment.
I need to create a branch for the feature or need to wait for the compatible version to merge.

qiwzhang:
which part broke the compatibility?   we did not change api_manager interface.  all changes are inside istio/proxy api_manager.

mangchiandjjoe:
I moved the service context initialization from the constructor to Init
So service_name of the global context was not set at the point of ApiManager instance creation.

esp module is checking service is empty. https://github.com/cloudendpoints/esp/blob/master/src/nginx/module.cc#L452

And lc->esp->Enabled() will return false until the initialization. https://github.com/cloudendpoints/esp/blob/master/src/nginx/module.cc#L542

These two conditions should be handled to support service config download.

I can modify the ApiManager to avoid first condition for backward compatibility.

mangchiandjjoe:
I fixed backward compatibility issue and tested.

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -76,16 +88,13 @@ utils::Status ApiManagerImpl::Init() {
...
-      config_loading_status_ =
-          utils::Status(Code::ABORTED, "Invalid service config");
+  if (!service_context_map_.empty()) {
+    for (auto it : service_context_map_) {
+      if (it.second->service_control()) {
+        it.second->service_control()->Init();
+      }
     }
+
     return config_loading_status_;
```
qiwzhang: why there is a "return" here?  don't we want to construct config_manager?

mangchiandjjoe: If the constructor got the valid service_config text, then ApiManager doesn't need to create an instance of the ConfigManager.

qiwzhang: this is ok for now.  But if rollout_strategy=managed.  you will still need config_mgr to update configs even service_config is passed in from constructor.   So just write it for the final version.

mangchiandjjoe: Added rollout_strategy condition check to return here.

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -95,7 +101,10 @@ utils::Status ApiManagerImpl::Init() {
       }
     }
 
-    return config_loading_status_;
+    if (global_context_->rollout_strategy().compare(
+            std::string(kConfigRolloutStrategy)) != 0) {
+      return config_loading_status_;
```
qiwzhang: We don't need check here.  My idea is: you have to make constructing config_manager safe even service_config_ is not empty.   

BTW,  std::string has operator==,  it will be easy to read if using ==

contrib/endpoints/src/api_manager/api_manager_impl.h
```diff
@@ -91,6 +91,9 @@ class ApiManagerImpl : public ApiManager {
   //  - Code::OK        Successfully initialized
   //  - Code::ABORTED   Initialization was failed
   utils::Status config_loading_status_;
+
+  // service_config provided by esp module
+  std::string service_config_;
```
qiwzhang: how about set service_config_ to global_context?

qiwzhang:
Sorry,  it seems it is too complicated to support service_config with using config_mgr.  So I changed my mind.  We don't need to support it.   Let us go with your original idea;  if service_config is valid,  not config_mgr.

contrib/endpoints/src/api_manager/api_manager_impl.cc
qiwzhang: You can define kConfigRolloutStrategy as std::string, so no need to cast for each request.

BTW,  it is better called kConfigRolloutManaged  or kConfigRolloutFixed.

contrib/endpoints/include/api_manager/api_manager.h
```diff
@@ -87,6 +87,12 @@ class ApiManager {
   virtual utils::Status GetStatistics(
       ApiManagerStatistics *statistics) const = 0;
 
+  // Returns the ApiManager initialization status.
+  //  - Code::UNKNOWN   Not initialized yet. The default value.
+  //  - Code::OK        Successfully initialized
+  //  - Code::ABORTED   Initialization was failed
+  virtual utils::Status ConfigLoadingStatus() = 0;
```
qiwzhang: this should be removed,  why it is still here?

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -21,47 +21,66 @@
...
+}  // namespace unknown
+
 ApiManagerImpl::ApiManagerImpl(std::unique_ptr<ApiManagerEnvInterface> env,
                                const std::string &service_config,
                                const std::string &server_config)
     : global_context_(
-          new context::GlobalContext(std::move(env), server_config)) {
+          new context::GlobalContext(std::move(env), server_config)),
+      config_loading_status_(
+          utils::Status(Code::UNKNOWN, "Not initialized yet")) {
```
qiwzhang: use UNAVAILABLE, better than UNKNOWN

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -21,47 +21,66 @@
...
+    }
   }
 
   check_workflow_ = std::unique_ptr<CheckWorkflow>(new CheckWorkflow);
   check_workflow_->RegisterAll();
 }
 
-void ApiManagerImpl::AddConfig(const std::string &service_config,
-                               bool deploy_it) {
+bool ApiManagerImpl::AddConfig(const std::string &service_config,
```
qiwzhang: return Status here instead of bool so error message can be passed.

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -75,11 +94,45 @@ utils::Status ApiManagerImpl::Init() {
...
+            config_loading_status_ =
+                utils::Status(Code::ABORTED, "Invalid service config");
+            return;
+          }
+
+          DeployConfigs(std::move(rollouts));
+        }
+        config_loading_status_ = status;
+      }));
+  config_manager_->Init();
```
qiwzhang: it is better to pass in callback function at Init() call.  Most likely,  onTimer() call needs different callback function.

contrib/endpoints/src/api_manager/api_manager_impl.h
```diff
@@ -52,11 +53,19 @@ class ApiManagerImpl : public ApiManager {
   utils::Status GetStatistics(ApiManagerStatistics *statistics) const override;
 
   // Add a new service config.
-  void AddConfig(const std::string &service_config, bool deploy_it);
+  // Return true if service_config is valid, otherwise return false.
+  // config_id will be updated when the deployment was successful
+  bool AddConfig(const std::string &service_config, std::string *config_id,
+                 bool initialize);
```
qiwzhang: per style guide,  return parameters should be at last.

contrib/endpoints/src/api_manager/api_manager_test.cc
```diff
@@ -83,6 +146,99 @@ TEST_F(ApiManagerTest, CorrectStatistics) {
...
+      std::move(env), kServiceConfig1, kServerConfigWithServiceNameConfigId));
+
+  // Already initialized
+  EXPECT_TRUE(api_manager);
+  EXPECT_EQ("OK", api_manager->ConfigLoadingStatus().ToString());
+
+  api_manager->Init();
+
+  // No change is expected
+  EXPECT_EQ("OK", api_manager->ConfigLoadingStatus().ToString());
```
qiwzhang: You can create api_manager_impl class and call its ConfigLoadingStatus() function.  The key is not to expose this function at api_manager interface.

contrib/endpoints/src/api_manager/api_manager_test.cc
```diff
@@ -83,6 +146,99 @@ TEST_F(ApiManagerTest, CorrectStatistics) {
...
 
+// ApiManager instance should be initialized on ApiManager instance creation
+TEST_F(ApiManagerTest, InitializedOnApiManagerInstanceCreation) {
+  std::unique_ptr<MockApiManagerEnvironment> env(
+      new ::testing::NiceMock<MockApiManagerEnvironment>());
+
+  // ServiceManagement service should not be called
+  EXPECT_CALL(*(env.get()), DoRunHTTPRequest(_))
+      .Times(0)
+      .WillOnce(Invoke([this](HTTPRequest *req) {
```
qiwzhang: Times(1) is good enough.  Not need Invoke

qiwzhang: sorry,  Times(0)

contrib/endpoints/include/api_manager/api_manager.h
mangchiandjjoe: Removed

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Changed

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Changed

contrib/endpoints/src/api_manager/api_manager_impl.h
mangchiandjjoe: Changed

contrib/endpoints/src/api_manager/api_manager_test.cc
mangchiandjjoe: Removed WillOnce

contrib/endpoints/src/api_manager/api_manager_test.cc
mangchiandjjoe: Changed

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Moved.

PR #346, opened 2017-06-01
Merge unary mixer api

qiwzhang:


src/envoy/mixer/grpc_transport.cc
```diff
@@ -0,0 +1,90 @@
...
+using ::google::protobuf::util::Status;
+using StatusCode = ::google::protobuf::util::error::Code;
+
+namespace Envoy {
+namespace Http {
+namespace Mixer {
+namespace {
+
+// gRPC request timeout
+const std::chrono::milliseconds kGrpcRequestTimeoutMs(5000);
```
kyessenov: how does this interact with cluster timeouts?

qiwzhang: I did not read Envoy code,  I guess it will be min(this_timeout, and cluster_time_out) for each request

PR #352, opened 2017-06-05
Add empty secret string check.

sarvaniv:


contrib/endpoints/src/api_manager/auth/service_account_token.cc
```diff
@@ -37,16 +37,19 @@ const int kClientSecretAuthTokenExpiration(3600 - 100);
...
 Status ServiceAccountToken::SetClientAuthSecret(const std::string& secret) {
-  client_auth_secret_ = secret;
 
-  for (unsigned int i = 0; i < JWT_TOKEN_TYPE_MAX; i++) {
-    if (!jwt_tokens_[i].audience().empty()) {
-      Status status = jwt_tokens_[i].GenerateJwtToken(client_auth_secret_);
-      if (!status.ok()) {
-        if (env_) {
-          env_->LogError("Failed to generate auth token.");
+  if (!secret.empty()) {
```
qiwzhang: for easy reading.   it is better bail out earlier, such as

if (empty) {
   return
}
...

sarvaniv: Done.

PR #360, opened 2017-06-08
Creates RequestContext if ApiManager was initialized

mangchiandjjoe:


contrib/endpoints/src/api_manager/request_handler.cc
```diff
@@ -29,6 +29,17 @@ using google::devtools::cloudtrace::v1::Traces;
 namespace google {
 namespace api_manager {
 
+RequestHandler::RequestHandler(ApiManagerImpl* api_manager,
+                               std::shared_ptr<CheckWorkflow> check_workflow,
+                               std::unique_ptr<Request> request_data)
+    : api_manager_(api_manager),
+      check_workflow_(check_workflow),
+      request_data_(std::move(request_data)) {
+  if (api_manager->IsConfigLoadingSucceeded()) {
```
lizan: What is the reason we want to create RequestContext during constructor?

If we do so, do we need other places that creating RequestContext?

contrib/endpoints/src/api_manager/request_handler.cc
```diff
@@ -29,6 +29,17 @@ using google::devtools::cloudtrace::v1::Traces;
...
+    CreateRequestContext();
```
qiwzhang: why do we need this?  what cases will fail?

contrib/endpoints/src/api_manager/request_handler.cc
mangchiandjjoe: If ApiManager was initialized in the constructor with service_config, then 
 request_handler->GetServiceConfigId() should return config_id.

This affects transcoder_factory.
https://github.com/cloudendpoints/esp/blob/master/src/nginx/module.cc#L75

contrib/endpoints/src/api_manager/request_handler.cc
mangchiandjjoe: Same as above.

contrib/endpoints/src/api_manager/request_handler.cc
qiwzhang: This is a real issue.

While config is loading,  if a grpc transcoding request comes,  it needs config_id to find service_config to build trancoder_factory. 

Check() can be delayed by using its callback.   Not sure how easy to delay transcoding factory.
@lizan ?

qiwzhang: Hi @mangchiandjjoe 

Here:  https://github.com/cloudendpoints/esp/blob/master/src/nginx/grpc.cc#L138

we may have to check api_manager loading status,   If it is in progress,  delay it.

Then, create trancode_factory here

qiwzhang: Hi @mangchiandjjoe 

With this case, I like to go back to your old way:   expose load_config_status in api_manager interface.    Let Nginx check it,  if still in progress,   go back to event loop to wait for it.  We will do this for both Check/Report and GrpcTranscoding.

I will like to keep the current pending_requests_callback code.   But in principal,  dead code should be removed.  If it is not used, it should be removed.

mangchiandjjoe: I got it. I will remove deprecated codes after new implement verified.
Let me close this PR.

Thanks.

PR #361, opened 2017-06-08
Removed pending requests management

mangchiandjjoe:
- Reverted request_handler.cc and request_handler.h
- Added IsConfigLoadingInProgress  and IsConfigLoadingSucceeded to ApiManager interface
- Removed pending_request_callbacks_ related codes


contrib/endpoints/include/api_manager/api_manager.h
```diff
@@ -87,6 +87,12 @@ class ApiManager {
   virtual utils::Status GetStatistics(
       ApiManagerStatistics *statistics) const = 0;
 
+  // Return true if config loading is still in progress
+  virtual bool IsConfigLoadingInProgress() = 0;
+
+  // Return true if config loading was succeeded
+  virtual bool IsConfigLoadingSucceeded() = 0;
```
qiwzhang: can we only expose one function for checking load_config_status?

mangchiandjjoe: Replaced.

PR #364, opened 2017-06-10
create one client cache per worker thread

qiwzhang:
Each cache has a mutex to protect it.  It will be faster for each thread to have its own cache so they will not be blocked for the mutex.


qiwzhang:
which mutex?    the one inside the mixer client library?

I still want the library to be thread safe so it will not impose callers to protect it.  beside,  not contented mutex is fast.

PR #365, opened 2017-06-12
Added service_configs map to server config

mangchiandjjoe:


contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -44,6 +44,12 @@ message ServerConfig {
   // The service name.
   string service_name = 8;
 
+  // Maps service configuration files to their corresponding traffic percentage.
+  // Key is the service configuration file, Value is the traffic percentage
+  // which must be greater than 0.0 and the sum must equal to 100.0.
+  map<string, double> service_configs = 12;
```
qiwzhang: I prefer to define a message and use repeated field.  call it 

init_service_configs

qiwzhang: also describe the relationship between this and config_id

contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -196,6 +194,15 @@ message ServiceManagementConfig {
   int32 refresh_interval_ms = 2;
 }
 
+// Maps service configuration files to their corresponding traffic percentage.
+message ServiceConfigTrafficPercentage {
+  // The service configuration file name
+  string service_config = 1;
```
qiwzhang: call it service_config_file_full_path

contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -44,15 +44,13 @@ message ServerConfig {
...
-  map<string, double> service_configs = 12;
-
-  // TODO(jaebong): This will be deprecated once service_configs is applied
-  // The service config id to use at the start of ESP.
+  // The service config id to use at the start of ESP. It has higher precedence
+  // than init_service_configs
   string config_id = 9;
 
+  // Init Service Configs.
+  repeated ServiceConfigTrafficPercentage init_service_configs = 12;
```
qiwzhang: add
// only use when service_id is not specified and not in the metadata either.

contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -196,6 +194,15 @@ message ServiceManagementConfig {
   int32 refresh_interval_ms = 2;
 }
 
+// Maps service configuration files to their corresponding traffic percentage.
+message ServiceConfigTrafficPercentage {
+  // The service configuration file name
+  string service_config = 1;
+  // The traffic percentage which must be greater than 0.0 and the sum must
+  // equal to 100.0.
```
qiwzhang: no, we don't need to add such restriction.  let is call it 

traffic_weight.   

the traffic is spitted according to the weights. sum of them not need to be 100%

qiwzhang: internally was implemented as weight.

contrib/endpoints/src/api_manager/proto/server_config.proto
qiwzhang: Actually, it should be:
init_service_configs first, 

config_id is not supported,  it require init_loading,   we don't support it,  it should be removed.

contrib/endpoints/src/api_manager/proto/server_config.proto
mangchiandjjoe: Can we ignore this comment?

    // Maps service configuration IDs to their corresponding traffic percentage.
    // Key is the service configuration ID, Value is the traffic percentage
    // which must be greater than 0.0 and the sum must equal to 100.0.

https://github.com/googleapis/googleapis/blob/master/google/api/servicemanagement/v1/resources.proto#L216

contrib/endpoints/src/api_manager/proto/server_config.proto
mangchiandjjoe: Renamed

contrib/endpoints/src/api_manager/proto/server_config.proto
mangchiandjjoe: Added

qiwzhang: Now, init_config is from service_content passed to construct of esp,  or this one.  That is it.

Ideally, we should change passing service_content to the ESP constructor to server_config too.

contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -44,11 +44,8 @@ message ServerConfig {
   // The service name.
   string service_name = 8;
 
-  // The service config id to use at the start of ESP. It has higher precedence
-  // than init_service_configs
-  string config_id = 9;
-
   // Init Service Configs.
+  // only use when service_id is not specified and not in the metadata either.
```
qiwzhang: remove this comment too

contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -44,8 +44,8 @@ message ServerConfig {
   // The service name.
   string service_name = 8;
```
qiwzhang: I think we can remove "service_name" too.

mangchiandjjoe: Removed

PR #373, opened 2017-06-20
Changed the type of init_service_configs and added rollout_id

mangchiandjjoe:


contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -195,6 +195,16 @@ message ServiceConfigTrafficPercentage {
   double traffic_percentage = 2;
 }
 
+message InitServiceConfigs {
```
qiwzhang: call it
 ServiceConfigRollout

contrib/endpoints/src/api_manager/proto/server_config.proto
```diff
@@ -42,7 +42,7 @@ message ServerConfig {
   ServiceManagementConfig service_management_config = 11;
 
   // Init Service Configs.
-  repeated ServiceConfigTrafficPercentage init_service_configs = 12;
+  InitServiceConfigs init_service_configs = 12;
```
qiwzhang: call the field name as

ServiceConfigRollout service_config_rollout = 12;

contrib/endpoints/src/api_manager/config_manager.cc
```diff
@@ -47,6 +47,11 @@ ConfigManager::ConfigManager(
     }
   }
 
+  if (global_context_->server_config()->has_init_service_configs()) {
```
qiwzhang: not copy it in constructor.   only set the current_rollout_id_ in line 81 of api_manager_impl.cpp after DeployConfigs() call.

mangchiandjjoe: Do you want to add set_current_rollout_id() to ConfigManager and call it in ApiManager?

I didn't include using ConfigManager in ApiManager this PR to reduce the size of the code review.

contrib/endpoints/src/api_manager/proto/server_config.proto
mangchiandjjoe: Renamed

contrib/endpoints/src/api_manager/proto/server_config.proto
mangchiandjjoe: Renamed

contrib/endpoints/src/api_manager/config_manager.cc
qiwzhang: Yes,  you may have to set it from api_manager.

So you are saying ConfigManager is not created yet in api_manager?  If so,  we can use separate PR to create it and use it.

mangchiandjjoe: ApiManager creates the instance of ConfigManager only after the deployment was successful.
It would be safe to copy the value here.

qiwzhang: I see.  Then it is good.

PR #375, opened 2017-06-20
ApiManager handles managed rollout_strategy

mangchiandjjoe:


contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -95,6 +84,33 @@ ApiManagerImpl::ApiManagerImpl(std::unique_ptr<ApiManagerEnvInterface> env,
...
+    const std::vector<std::pair<std::string, int>> &&configs, bool initialize) {
+  std::vector<std::pair<std::string, int>> list;
+  for (auto item : configs) {
+    std::string config_id;
+    if (AddConfig(item.first, initialize, &config_id).ok()) {
+      list.push_back({config_id, round(item.second)});
+    } else {
+      std::string msg = "Invalid service config";
+      global_context_->env()->LogError(msg);
+      return utils::Status(Code::ABORTED, msg);
```
qiwzhang: For multiple versions,  not bail out if one version fails.

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -153,6 +169,25 @@ utils::Status ApiManagerImpl::Init() {
     }
   }
 
+  if (global_context_->rollout_strategy() == kConfigRolloutManaged) {
+    config_manager_.reset(new ConfigManager(
+        global_context_,
+        [this](const utils::Status &status,
+               const std::vector<std::pair<std::string, int>> &configs) {
```
qiwzhang: you need to use && if you want to use std::move.   also remove "const"

contrib/endpoints/src/api_manager/api_manager_test.cc
```diff
@@ -249,6 +350,47 @@ TEST_F(ApiManagerTest, InitializedByConfigManager) {
...
+  EXPECT_TRUE(api_manager->Enabled());
+  EXPECT_EQ("bookstore.test.appspot.com", api_manager->service_name());
+  EXPECT_EQ("2017-05-01r0", api_manager->service("2017-05-01r0").id());
+
+  api_manager->Init();
+
+  EXPECT_TRUE(api_manager->Enabled());
+  EXPECT_EQ("2017-05-01r0", api_manager->service("2017-05-01r0").id());
+
+  auto service = api_manager->SelectService();
```
qiwzhang: I don't understand why calling SelectService() will cause version to be updated.   It should be OnTimer() call.   But I don't see OnTimer() call.

mangchiandjjoe: In this test, ApiManager was initialized in the constructor and Init() function. 

When ApiManager creates an instance of ConfigManager in Init(), the constructor of ConfigManager invokes the callback.

https://github.com/istio/proxy/pull/375/files/3529d1ef9d24ececcb583f27ebc52acbf9a70bc7#diff-43539418cb3efc25680906e605c5c19eR189

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: As discussed, I will keep current implementation.

contrib/endpoints/src/api_manager/api_manager_impl.cc
mangchiandjjoe: Fixed

contrib/endpoints/src/api_manager/api_manager_impl.cc
qiwzhang: ok

mangchiandjjoe:
I will revert the latest commit and create another PR.

PR #377, opened 2017-06-21
Added ConfigLoadingStatus to ApiManager interface

mangchiandjjoe:


contrib/endpoints/include/api_manager/api_manager.h
```diff
@@ -87,6 +87,12 @@ class ApiManager {
   virtual utils::Status GetStatistics(
       ApiManagerStatistics *statistics) const = 0;
 
+  // Returns the config loading status.
+  //  - Code::UNAVAILABLE Not initialized yet. The default value.
+  //  - Code::OK          Successfully initialized
+  //  - Code::ABORTED     Initialization was failed
+  virtual utils::Status ConfigLoadingStatus() = 0;
```
qiwzhang: Can we do add another function 

status LoadServiceRollouts();

It will be much easier to read.

We can add comment say, this can be called only once,
the data is from server_config.

mangchiandjjoe: The purpose of the ConfigLoadingStatus was to get access the config_loading_status_ of  the ApiManager. If we add LoadServiceRollouts, we can remove the ConfigLoadingStatus.

LoadServiceRollouts should be called only once right after the ApiManager instance was created.

contrib/endpoints/src/api_manager/api_manager_impl.cc
```diff
@@ -80,8 +86,7 @@ ApiManagerImpl::ApiManagerImpl(std::unique_ptr<ApiManagerEnvInterface> env,
     config_loading_status_ = utils::Status(Code::ABORTED, err_msg);
   }
 
-  check_workflow_ = std::unique_ptr<CheckWorkflow>(new CheckWorkflow);
-  check_workflow_->RegisterAll();
+  return config_loading_status_;
```
qiwzhang: We don't need a class variable to store this status.   The main reason to make it  in the Load() function so it can return the status right away,  not need to store it.

mangchiandjjoe: I agree. Removed.

PR #378, opened 2017-06-22
Update Envoy to latest

lizan:


ldemailly:
who is making the pilot changes to pick this up ?

lizan:
@ldemailly pilot-regression has been failing for a while, not particularly related to this change. https://testing.istio.io/job/proxy/job/pilot-regression/375/ is the last successful one and https://testing.istio.io/job/proxy/job/pilot-regression/376/ is where it starts to fail.

ldemailly:
@lizan I know pilot has broken the e2e tests but I think the internal tests are still working and if this PR changed anything that pilot needs to know, there need to be a PR for that change ? or minimally a PR to pick this new version

cc @kyessenov

lizan:
@ldemailly I don't think there is any changes in Envoy side would break. https://testing.istio.io/job/proxy/job/pilot-regression/376/ is from #363 and I suspect that breaks the test. Since it changes mixer protocol. @qiwzhang ?

PR #383, opened 2017-07-08
Added mixer filter for TCP connection.

qiwzhang:


src/envoy/mixer/tcp_filter.cc
```diff
@@ -0,0 +1,206 @@
...
+      return nullptr;
+    }
+
+    Http::Mixer::TcpConfigPtr tcp_config(
+        new Http::Mixer::TcpConfig(config, server));
+    return [tcp_config](Network::FilterManager& filter_manager) -> void {
+      std::shared_ptr<Http::Mixer::TcpInstance> instance =
+          std::make_shared<Http::Mixer::TcpInstance>(tcp_config);
+      filter_manager.addReadFilter(Network::ReadFilterSharedPtr(instance));
+      filter_manager.addWriteFilter(Network::WriteFilterSharedPtr(instance));
```
douglas-reid: curious: do we want to filter both sides (read & write)?

qiwzhang: It seems that it is the only way to get write_bytes.   Other approaches may require Envoy core code change.

src/envoy/mixer/tcp_filter.cc
```diff
@@ -0,0 +1,206 @@
...
+  }
+};
+
+}  // namespace Mixer
+}  // namespace Http
+
+namespace Server {
+namespace Configuration {
+
+class TcpMixerFilter : public NetworkFilterConfigFactory {
```
lizan: use `NamedNetworkFilterConfigFactory` instead, `NetworkFilterConfigFactory` is deprecated.

qiwzhang: Done. Thanks

PR #386, opened 2017-07-12
Copy HTTP trace headers to gRPC metadata

qiwzhang:


src/envoy/mixer/grpc_transport.cc
```diff
@@ -31,10 +31,20 @@ const std::chrono::milliseconds kGrpcRequestTimeoutMs(5000);
 // The name for the mixer server cluster.
 const char* kMixerServerClusterName = "mixer_server";
 
+// HTTP trace headers that should pass to gRPC metadata from origin request.
+const LowerCaseString kHttpTraceHeders[]{
```
lizan: nit: `kHttpTraceHeaders`

src/envoy/mixer/grpc_transport.cc
```diff
@@ -31,10 +31,20 @@ const std::chrono::milliseconds kGrpcRequestTimeoutMs(5000);
...
+    LowerCaseString("x-request-id"),      LowerCaseString("x-b3-traceid"),
```
qiwzhang: x-request-id is useful for debugging, I will keep it.

src/envoy/mixer/grpc_transport.h
```diff
@@ -28,13 +28,31 @@ namespace Envoy {
 namespace Http {
 namespace Mixer {
 
+// A helper class to pass Upstream::ClusterManager& in the lambda capture
+// If passing Upstream::ClusterManager& directly in the lambda capture,
```
lizan: I don't think this is happening. can you elaborate?

qiwzhang: You are right, have to use & in the lambda capture to pass a reference type. Thanks.
The whole class is removed.

src/envoy/mixer/grpc_transport.cc
```diff
@@ -72,6 +82,18 @@ bool GrpcTransport::IsMixerServerConfigured(Upstream::ClusterManager& cm) {
   return cm.get(kMixerServerClusterName) != nullptr;
 }
 
+void GrpcTransport::onPreRequestCustomizeHeaders(Http::HeaderMap& headers) {
+  if (!headers_) return;
+
+  for (const auto& key : kHttpTraceHeaders) {
+    const HeaderEntry* entry = headers_->get(key);
```
lizan: AFAIK get doesn't work for inline headers (e.g. every header other than "x-ot-span-context"), this might be an issue for envoy itself.
https://github.com/lyft/envoy/blob/master/source/common/http/header_map_impl.cc#L338

Do you have a test to confirm the behavior of this PR?

qiwzhang: It works.  I have tested. 

Apparently,  headers_ has everything.   
inline_headers_ is just an optimization to access the headers_

qiwzhang: I manually tested it, it works.  I could not integration-test it since I have not figured out how to extract gRPC metadata in go,  It may require change mixer go code to extract gRPC metadata.

lizan: It's in the context
https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md

But at least we don't get the optimization for most of the inline headers. Is that OK? `get` is a linear search.

qiwzhang: I did not realized get() is NOT map lookup.  I am changing code to use inline_headers directly.

qiwzhang: Done optimize the header copy

PR #389, opened 2017-07-25
Use dockerfiles for proxy

kyessenov:
Use standard dockerfiles to push base image for proxy and debug proxy


docker/Dockerfile.debug
```diff
@@ -0,0 +1,20 @@
...
+    netcat \
+    tcpdump \
+    net-tools \
+    libc6-dbg gdb valgrind \
+    vim \
+    nano \
+    emacs \
+    wrk
+ADD proxy-redirection-clear /usr/local/bin/proxy-redirection-clear
+ADD proxy-redirection-restore /usr/local/bin/proxy-redirection-restore
```
ayj: These are an artifact of the traffic redirection scheme via iptables. Do they need to be included in the base proxy image?

Jenkinsfile
```diff
@@ -51,13 +51,11 @@ def presubmit(gitUtils, bazel) {
...
     bazel.updateBazelRc()
-    stage('Push Binary') {
+    stage('Binary push') {
       sh 'script/release-binary'
     }
     stage('Docker Push') {
-      def images = 'envoy,envoy_debug'
-      def tags = "${gitUtils.GIT_SHORT_SHA},\$(date +%Y-%m-%d-%H.%M.%S),latest"
-      utils.publishDockerImages(images, tags, 'release')
+      sh 'script/release-docker'
```
ayj: Do we need to use `gitUtils.GIT_SHORT_SHA` or is `git show -q HEAD --pretty=format:%h` always equivalent? 

cc @sebastienvas

docker/Dockerfile.debug
kyessenov: They are tools to simplify iptables manipulation and not likely to change much. It's best to cache tools and put them into base image to speed up docker builds.

Jenkinsfile
kyessenov: I'd like to avoid jenkins tools since I want to be able to run this script locally.

docker/Dockerfile.debug
ayj: I would have expected these to go into the base sidecar image in istio/pilot along with `prepare_proxy.sh`.

kyessenov: Yes, but bazel violates layering of docker images. it doesn't do good job caching layers on top of pulled images.

Jenkinsfile
sebastienvas: Well have to redo that with Prow anyway.

PR #392, opened 2017-07-31
jwt-auth: JWT verification class

mtakigiku:


src/envoy/auth/BUILD
```diff
@@ -18,11 +18,19 @@ envoy_cc_binary(
 
 envoy_cc_library(
     name = "http_filter_lib",
-    srcs = ["http_filter.cc"],
-    hdrs = ["http_filter.h"],
+    srcs = [
+        "http_filter.cc",
+        "jwt.cc",
```
lizan: can you add another target for `jwt.{cc,h}`, and add it to deps in this target?

src/envoy/auth/jwt_test.cc
```diff
@@ -0,0 +1,90 @@
...
+  auto payload_str = Base64url::decode(jwt_payload_encoded);
+  EXPECT_STREQ(payload_str.c_str(), payload.c_str());
+}
+
+TEST_F(JwtTest, Jwt_decode) {
+  Jwt ob = Jwt();
+  auto result = ob.decode(jwt, pubkey);
+
+  bool valid = result.first;
+  rapidjson::Document* payload = result.second;
```
lizan: use `std::unique_ptr`

src/envoy/auth/jwt.h
```diff
@@ -0,0 +1,48 @@
...
+
+}  // Util
+
+class Jwt {
+ public:
+  std::pair<bool, rapidjson::Document*> decode(const std::string& jwt,
+                                               const std::string& pkey_pem);
+
+ private:
+  EVP_PKEY* evp_pkey_from_str(const std::string& pkey_pem);
```
lizan: use camelCase for function names

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,176 @@
...
+
+}  // namespace Base64url
+
+namespace Util {
+
+uint8_t* unsigned_c_str(const std::string& str) {
+  return reinterpret_cast<uint8_t*>(const_cast<char*>(str.c_str()));
+}
+
+std::vector<std::string> split(std::string str, char delimiter) {
```
lizan: https://github.com/lyft/envoy/blob/master/source/common/common/utility.h#L148

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,176 @@
...
+  }
+  return Base64::decode(input);
+}
+
+}  // namespace Base64url
+
+namespace Util {
+
+uint8_t* unsigned_c_str(const std::string& str) {
+  return reinterpret_cast<uint8_t*>(const_cast<char*>(str.c_str()));
```
lizan: Why do you need to do a `const_cast` here? It is unsafe and I think the BoringSSL functions takes `const uint8_t*`s.

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,176 @@
...
+}  // namespace Util
+
+EVP_PKEY* Jwt::evp_pkey_from_str(const std::string& pkey_pem) {
+  std::string pkey_der = Base64::decode(pkey_pem);
+  RSA* rsa = RSA_public_key_from_bytes(Util::unsigned_c_str(pkey_der),
+                                       pkey_der.length());
+  if (rsa == NULL) {
+    return NULL;
+  }
+  EVP_PKEY* key = EVP_PKEY_new();
```
lizan: use `bssl::UniquePtr<EVP_PKEY>` (for all owned BoringSSL pointers in this file)

src/envoy/auth/jwt.h
```diff
@@ -0,0 +1,48 @@
...
+namespace Util {
+
+uint8_t* unsigned_c_str(const std::string& str);
+std::vector<std::string> split(std::string str, char c);
+
+}  // Util
+
+class Jwt {
+ public:
+  std::pair<bool, rapidjson::Document*> decode(const std::string& jwt,
```
lizan: If you the 1st of the pair is `false` iff the document is `nullptr` then you don't need the `bool`.

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,176 @@
...
+const EVP_MD* Jwt::evp_md_from_alg(const std::string& alg) {
+  /*
+   * may use
+   * EVP_sha384() if alg == "RS384" and
+   * EVP_sha512() if alg == "RS512"
+   */
+  if (alg == "RS256") {
+    return EVP_sha256();
+  } else {
+    return NULL;
```
lizan: `nullptr` instead of `NULL`. (throughout)

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,176 @@
...
+
+end:
+  if (md_ctx != NULL) {
+    EVP_MD_CTX_destroy(md_ctx);
+  }
+  return result;
+}
+
+bool Jwt::verify_signature(const std::string& pkey_pem, const std::string& alg,
+                           const std::string signature,
```
lizan: `const std::string&` ?

src/envoy/auth/jwt.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace Envoy {
+namespace Http {
+namespace Auth {
+
+namespace Base64url {
```
lizan: the two namespace can be moved to `.cc` file if you don't expect other files to use it.

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,176 @@
...
+  if (key != NULL) EVP_PKEY_free(key);
+
+  return valid;
+}
+
+std::pair<bool, rapidjson::Document*> Jwt::decode(const std::string& jwt,
+                                                  const std::string& pkey_pem) {
+  // return value in failure cases
+  std::pair<bool, rapidjson::Document*> failed =
+      std::make_pair(false, (rapidjson::Document*)NULL);
```
qiwzhang: use unique_ptr?

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,154 @@
...
+                         Util::unsignedCStr(signed_data), signed_data.length());
+}
+
+std::unique_ptr<rapidjson::Document> Jwt::decode(const std::string& jwt,
+                                                 const std::string& pkey_pem) {
+  std::vector<std::string> jwt_split = StringUtil::split(jwt, '.');
+  if (jwt_split.size() != 3) {
+    return nullptr;
+  }
+  std::string header_base64url_encoded = jwt_split[0];
```
lizan: nti: const std::string& for these 3 lines.

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,154 @@
...
+  }
+}
+
+bool Jwt::verifySignature(bssl::UniquePtr<EVP_PKEY> key, const std::string& alg,
+                          const uint8_t* signature, size_t signature_len,
+                          const uint8_t* signed_data, size_t signed_data_len) {
+  bssl::UniquePtr<EVP_MD_CTX> md_ctx(EVP_MD_CTX_create());
+  const EVP_MD* md = evpMdFromAlg(alg);
+
+  assert(md != nullptr);
```
lizan: This shouldn't be a crash either.

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,154 @@
...
+      break;
+    case 2:
+      input += "==";
+      break;
+    case 3:
+      input += "=";
+      break;
+    default:
+      /* invalid base64url input*/
+      assert(0);
```
lizan: This should just be a verification failure, not crash.

src/envoy/auth/jwt_test.cc
```diff
@@ -0,0 +1,67 @@
...
+
+  EXPECT_TRUE((*payload)["sub"].IsString());
+  std::string sub = (*payload)["sub"].GetString();
+  EXPECT_STREQ("test@example.com", sub.c_str());
+
+  EXPECT_TRUE((*payload)["exp"].IsInt64());
+  int64_t exp = (*payload)["exp"].GetInt64();
+  EXPECT_EQ(1501281058LL, exp);
+}
+
```
lizan: More test for verification failures.

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,154 @@
...
+  } else {
+    return nullptr;
+  }
+}
+
+bool Jwt::verifySignature(bssl::UniquePtr<EVP_PKEY> key, const std::string& alg,
+                          const uint8_t* signature, size_t signature_len,
+                          const uint8_t* signed_data, size_t signed_data_len) {
+  bssl::UniquePtr<EVP_MD_CTX> md_ctx(EVP_MD_CTX_create());
+  const EVP_MD* md = evpMdFromAlg(alg);
```
mtakigiku: I tried to do like `bssl::UniquePtr<const EVP_MD> md(...)`, but failed with an error
`
external/boringssl/src/include/openssl/base.h:379:25: error: 'Free' is not a member of 'bssl::internal::DeleterImpl<const env_md_st>'
     DeleterImpl<T>::Free(ptr);
`

How to handle this?

lizan: I don't think you need `UniquePtr` for EVP_MD, the instance is owned by the library.

src/envoy/auth/jwt_test.cc
```diff
@@ -0,0 +1,167 @@
...
+  const std::string pubkey =
+      "MIIBCgKCAQEAtw7MNxUTxmzWROCD5BqJxmzT7xqc9KsnAjbXCoqEEHDx4WBlfcwk"
+      "XHt9e/2+Uwi3Arz3FOMNKwGGlbr7clBY3utsjUs8BTF0kO/poAmSTdSuGeh2mSbc"
+      "VHvmQ7X/kichWwx5Qj0Xj4REU3Gixu1gQIr3GATPAIULo5lj/ebOGAa+l0wIG80N"
+      "zz1pBtTIUx68xs5ZGe7cIJ7E8n4pMX10eeuh36h+aossePeuHulYmjr4N0/1jG7a"
+      "+hHYL6nqwOR3ej0VqCTLS0OloC0LuCpLV7CnSpwbp2Qg/c+MDzQ0TH8g8drIzR5h"
+      "Fe9a3NlNRMXgUU5RqbLnR9zfXr7b9oEszQIDAQAB";
+};
+
+TEST_F(JwtTest, Jwt_decode) {
```
lizan: nit: CamelCase for test cases

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,168 @@
...
+                          const std::string& signature,
+                          const std::string& signed_data) {
+  return verifySignature(evpPkeyFromStr(pkey_pem), alg,
+                         Util::unsignedCStr(signature), signature.length(),
+                         Util::unsignedCStr(signed_data), signed_data.length());
+}
+
+std::unique_ptr<rapidjson::Document> Jwt::decode(const std::string& jwt,
+                                                 const std::string& pkey_pem) {
+  if (std::count(jwt.begin(), jwt.end(), '.') != 2) {
```
lizan: we might need something similar to [`grpc_jwt_verifier_status`](https://github.com/grpc/grpc/blob/master/src/core/lib/security/credentials/jwt/jwt_verifier.h#L38) to return the failure reason. This can be added as a TODO here and in follow up PRs.

src/envoy/auth/jwt.h
```diff
@@ -0,0 +1,35 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace Envoy {
+namespace Http {
+namespace Auth {
+
+class Jwt {
+ public:
```
lizan: Those function looks all static, no?

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,168 @@
+#include "jwt.h"
```
qiwzhang: please add standard istio copy right header to all files

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,168 @@
...
+#include <vector>
+
+namespace Envoy {
+namespace Http {
+namespace Auth {
+
+namespace Base64url {
+
+std::string decode(std::string input) {
+  std::replace(input.begin(), input.end(), '-', '+');
```
qiwzhang: please add comment on why doing such replacement.

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,168 @@
...
+      return "";
+  }
+  return Base64::decode(input);
+}
+
+}  // namespace Base64url
+
+namespace Util {
+
+const uint8_t* unsignedCStr(const std::string& str) {
```
qiwzhang: rename to CastToUChar?

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,168 @@
...
+  return reinterpret_cast<const uint8_t*>(str.c_str());
+}
+
+}  // namespace Util
+
+bssl::UniquePtr<EVP_PKEY> Jwt::evpPkeyFromStr(const std::string& pkey_pem) {
+  std::string pkey_der = Base64::decode(pkey_pem);
+  bssl::UniquePtr<RSA> rsa(RSA_public_key_from_bytes(
+      Util::unsignedCStr(pkey_der), pkey_der.length()));
+  if (rsa == nullptr) {
```
qiwzhang: can we use std::unique_ptr?  what is bssl::UniquePtr?

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,168 @@
...
+  const std::string& header_base64url_encoded = jwt_split[0];
+  const std::string& payload_base64url_encoded = jwt_split[1];
+  const std::string& signature_base64url_encoded = jwt_split[2];
+  std::string signed_data = jwt_split[0] + '.' + jwt_split[1];
+
+  /*
+   * verification
+   */
+  rapidjson::Document header_json;
+  if (header_json.Parse(Base64url::decode(header_base64url_encoded).c_str())
```
qiwzhang: I think "decode" function may modify the string,  but its type is "config string &", compiler should complain.

lizan: the input of decode is `std::string`, so it is a copy.

src/envoy/auth/jwt.cc
lizan: wondering why presubmit doesn't complain this...

src/envoy/auth/jwt.h
mtakigiku: Yes they are.
Just because at first I passed the public key to Jwt class object as an argument of its constructor. I'll make it static.

src/envoy/auth/jwt.cc
mtakigiku: [`bssl::UniquePtr<T>`](https://github.com/google/boringssl/blob/master/include/openssl/base.h#L439)
is `std::unique_ptr` with specified deleter (e.g. `RSA_free()` in this case)

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,197 @@
...
+    case 3:
+      input += "=";
+      break;
+    default:
+      /*
+       * for invalid base64url input, return empty string
+       */
+      return "";
+  }
+  return Base64::decode(input);
```
mtakigiku: The comment in the implementation of `Base64::decode()` says it skips any invalid character (i.e. `[^-A-Za-z0-9+/=]`), not raising an error.
It is not suitable, is it?

Another problem is, its implementation actually does not correctly skip invalid characters (in other words, I think it contains a bug)

mtakigiku: I don't think it's a bug, it's just assuming the validity of input, as 4n-th or (4n+1)-th character never be a padding in a correct base64 string.
So what we need is to check the validity of input string of base64url.

src/envoy/auth/jwt_test.cc
```diff
@@ -0,0 +1,175 @@
...
+  invalid_jwt[jwt.length() - 1] = jwt[jwt.length() - 1] != 'a' ? 'a' : 'b';
+
+  auto payload = Jwt::decode(invalid_jwt, pubkey);
+
+  EXPECT_TRUE(payload == nullptr);
+}
+
+TEST_F(JwtTest, InvalidPublickey) {
+  auto invalid_pubkey = pubkey;
+  invalid_pubkey[pubkey.length() - 1] =
```
lizan: why do you need the ternary statement?

mtakigiku: not needed. Fixed it.

mtakigiku:
- to check base64url input validity
- to return verification failure reason
- to support jwk

Should these remaining TODOs be done in this PR, or following PRs?
I think I will do them in this PR.

src/envoy/auth/jwt.h
```diff
@@ -0,0 +1,52 @@
...
+  static std::unique_ptr<rapidjson::Document> decode(
```
qiwzhang: if all members are static,  can we use namespace?

src/envoy/auth/jwt.h
```diff
@@ -0,0 +1,52 @@
...
+      const std::string& jwt, const std::string& pkey_pem);
+
+ private:
+  static bssl::UniquePtr<EVP_PKEY> evpPkeyFromStr(const std::string& pkey_pem);
+  static const EVP_MD* evpMdFromAlg(const std::string& alg);
+  static bool verifySignature(bssl::UniquePtr<EVP_PKEY> key,
```
qiwzhang: can we move static private functions into .cc file in anonymous namespace?

src/envoy/auth/jwt.cc
```diff
@@ -31,6 +31,7 @@
 namespace Envoy {
 namespace Http {
 namespace Auth {
+namespace {
 
 namespace Base64url {
```
qiwzhang: It seems that Base64url namespace is inside anonymous namespace.   Does it work?

I thought the anonymous namespace should the last one.

src/envoy/auth/jwt.h
lizan: We might need to eventually convert this to a class, public key as member. That should be in follow up PRs.

src/envoy/auth/jwt.cc
mtakigiku: I just didn't understand about unnamed namespaces. Did you mean
```
namespace Base64url{
namespace{
...
}
}
```
or
```
namespace{ // Base64url
...
}
```
?

lizan: neither, since it is in `.cc` file, you don't need to namespace it, just remove the Base64url and Util namespace is find. have all of them in a unnamed namespace

mtakigiku: So we need either rename `decode()` (of base64url) to `base64UrlDecode()`, or modify all its call to `Auth::decode()` to avoid collision. The former is better?

src/envoy/auth/jwt.cc
```diff
@@ -95,20 +93,14 @@ std::string decode(std::string input) {
...
-
-namespace Util {
-
 const uint8_t *castToUChar(const std::string &str) {
   return reinterpret_cast<const uint8_t *>(str.c_str());
 }
 
-}  // namespace Util
-
 bssl::UniquePtr<EVP_PKEY> evpPkeyFromStr(const std::string &pkey_pem) {
```
qiwzhang: should we following Google code style to have function name started with Upper case if it is camel case?

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,232 @@
...
+namespace Envoy {
+namespace Http {
+namespace Auth {
+namespace {
+
+// Conversion table is taken from
+// https://opensource.apple.com/source/QuickTimeStreamingServer/QuickTimeStreamingServer-452/CommonUtilitiesLib/base64.c
+//
+// and modified the position of 62 ('+' to '-') and 63 ('/' to '_')
+const uint8_t REVERSE_LOOKUP_TABLE_BASE64URL[256] = {
```
qiwzhang: const variable usually names as kCamelCase

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,232 @@
...
+    64, 64, 64, 64, 64, 64, 64, 64, 64};
+
+bool IsNotBase64UrlChar(int8_t c) {
+  return REVERSE_LOOKUP_TABLE_BASE64URL[static_cast<int32_t>(c)] & 64;
+}
+
+std::string Base64UrlDecode(std::string input) {
+  /*
+   * if input contains non-base64url character, return empty string
+   */
```
qiwzhang: use // for short comment

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,232 @@
...
+
+const uint8_t *CastToUChar(const std::string &str) {
+  return reinterpret_cast<const uint8_t *>(str.c_str());
+}
+
+bssl::UniquePtr<EVP_PKEY> EvpPkeyFromStr(const std::string &pkey_pem) {
+  std::string pkey_der = Base64::decode(pkey_pem);
+  bssl::UniquePtr<RSA> rsa(
+      RSA_public_key_from_bytes(CastToUChar(pkey_der), pkey_der.length()));
+  if (rsa == nullptr) {
```
qiwzhang: since rsa is unique_ptr,  you can do

if (!rsa)

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,232 @@
...
+  const std::string &header_base64url_encoded = jwt_split[0];
+  const std::string &payload_base64url_encoded = jwt_split[1];
+  const std::string &signature_base64url_encoded = jwt_split[2];
+  std::string signed_data = jwt_split[0] + '.' + jwt_split[1];
+
+  /*
+   * verification
+   */
+  rapidjson::Document header_json;
+  if (header_json.Parse(Base64UrlDecode(header_base64url_encoded).c_str())
```
qiwzhang: header_base64url_encoded is "const string&" type,  but Base64UrlDecode() expecting "string" since the function is modifying it.   Compiler should fail it

src/envoy/auth/jwt.cc
```diff
@@ -0,0 +1,232 @@
...
+
+  // if signature is invalid, it will not decode the payload
+  if (!valid) {
+    return nullptr;
+  }
+
+  /*
+   * decode payload
+   */
+  std::unique_ptr<rapidjson::Document> payload_json_ptr(
```
qiwzhang: just payload_json,   not need to add "_ptr"

src/envoy/auth/jwt.h
```diff
@@ -0,0 +1,39 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace Envoy {
+namespace Http {
+namespace Auth {
+namespace Jwt {
+
+std::unique_ptr<rapidjson::Document> Decode(const std::string& jwt,
```
qiwzhang: please add function comment for exposed public functions

src/envoy/auth/jwt_test.cc
```diff
@@ -0,0 +1,282 @@
...
+      "CjZiNWIzZjgzYTk0Zjg1MjFkMDhiNmYyNzY1MTM1N2MyYWI0MzBkM2FlYjg5MTFmNjM5ZGNj"
+      "MGM2NTcxNThmOWUKOWQ1ZDM2NWFkNGVjOTgwYmNkY2RiMDUzM2MzYjY2MjJmYWJiMDViNjNk"
+      "NjIxMDJiZDkyZDE3ZjZkZDhiMTBkOQo1YjBlMDRiZWU2MDBjNjRhNzM0ZGE1ZGY2YjljMGY5"
+      "ZDM1Mzk3MjcyNDcyN2RjMTViYjk1MjMwZjdmYmU5MzYx";
+
+  /*
+   * jwt with header replaced by
+   * "{"alg":256,"typ":"JWT"}"
+   */
+  const std::string jwt_with_alg_is_not_string =
```
qiwzhang: const variable name as kCamelCase

src/envoy/auth/jwt.cc
mtakigiku: I think Base64UrlDecode() is receiving a copy and not modifying the original

lizan: Compiler wouldn't fail, Base64UrlDecode receives a copy.

src/envoy/auth/jwt_test.cc
```diff
@@ -0,0 +1,282 @@
...
+}
+
+TEST_F(JwtTest, JsonBadInputHeader) {
+  auto payload = Jwt::Decode(kJwtWithBadJsonHeader, kPublicKey);
+  EXPECT_TRUE(payload == nullptr);
+}
+
+TEST_F(JwtTest, JsonBadInputPayload) {
+  auto payload = Jwt::Decode(kJwtWithBadJsonPayload, kPublicKey);
+  EXPECT_TRUE(payload == nullptr);
```
qiwzhang: EXPECT_FALSE(payload);

Issue #391, opened 2017-07-27
Use source.user instead of origin.user as the authentication result


myidpt:
Envoy can only get the immediate caller's identity through the SSL certificate, so the identity to Mixer should be source.user instead of origin.user. We need to change this to source.user:
https://github.com/istio/proxy/blob/71b204349729e368213fb5ad6fb3834d8acfdd76/src/envoy/mixer/mixer_control.cc#L42

Reference:
https://istio.io/docs/reference/config/mixer/attribute-vocabulary.html

PR #393, opened 2017-08-01
Change origin.user to source.user.

myidpt:
The mTLS authentication result only represents the identity of the immediate user,
so this result should be put into source.user. This resolves issue #391.

### DIFF
--- a/src/envoy/mixer/mixer_control.cc
+++ b/src/envoy/mixer/mixer_control.cc
@@ -39,7 +39,7 @@ namespace Mixer {
 namespace {
 
 // Define attribute names
-const std::string kOriginUser = "origin.user";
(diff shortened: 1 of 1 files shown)


wattli:
Thanks for working on this.

yutongz:
build proxy/pilot-regression

yutongz:
@myidpt  I fixed it. So the problem is pilot changed [WORKSPACE](https://github.com/istio/pilot/commit/e502d9b8cad19d13b254f9816c491941d21b1688#diff-09498dbadf45966909850dc8a47ebb13R338). The key word is modified from "PROXY" to "ISTIO_PROXY_BUCKET" but Jenkins has no idea of it.

PR #395, opened 2017-08-03
jwt-auth: support JWK public keys

mtakigiku:


src/envoy/auth/jwt.cc
```diff
@@ -114,9 +156,10 @@ const EVP_MD *EvpMdFromAlg(const std::string &alg) {
   }
 }
 
-bool VerifySignature(bssl::UniquePtr<EVP_PKEY> key, const std::string &alg,
-                     const uint8_t *signature, size_t signature_len,
-                     const uint8_t *signed_data, size_t signed_data_len) {
+bool VerifySignature(const bssl::UniquePtr<EVP_PKEY> &key,
```
lizan: same above, use raw pointer instead of const unique_ptr&

src/envoy/auth/jwt.cc
```diff
@@ -139,75 +182,155 @@ bool VerifySignature(bssl::UniquePtr<EVP_PKEY> key, const std::string &alg,
   return true;
 }
 
-bool VerifySignature(const std::string &pkey_pem, const std::string &alg,
-                     const std::string &signature,
+bool VerifySignature(const bssl::UniquePtr<EVP_PKEY> &key,
```
lizan: ditto

src/envoy/auth/jwt_test.cc
```diff
@@ -172,6 +172,61 @@ class JwtTest : public testing::Test {
...
+  // }
+  // Payload:
+  // {
+  //   "iss": "628645741881-"
+  //     "noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com",
+  //   "sub": "628645741881-"
+  //     "noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com",
+  //   "aud": "http://myservice.com/myapi"
+  // }
+  const std::string kTokenNoKid =
```
lizan: format is weird...

src/envoy/auth/jwt.cc
```diff
@@ -139,75 +182,155 @@ bool VerifySignature(bssl::UniquePtr<EVP_PKEY> key, const std::string &alg,
...
+  return VerifySignature(key, alg, CastToUChar(signature), signature.length(),
+                         CastToUChar(signed_data), signed_data.length());
 }
 
 }  // namespace
 
 namespace Jwt {
+namespace {
+
+class Internal {
```
qiwzhang: Please have a meaningful class name.
Also add class comment to describe its purpose

src/envoy/auth/jwt.cc
```diff
@@ -90,19 +102,49 @@ const uint8_t *CastToUChar(const std::string &str) {
...
+  return EvpPkeyFromRsa(bssl::UniquePtr<RSA>(
+      RSA_public_key_from_bytes(CastToUChar(pkey_der), pkey_der.length())));
+}
+
+BIGNUM *BigNumFromBase64UrlString(const std::string &s) {
+  std::string s_decoded = Base64UrlDecode(s);
+  if (s_decoded == "") {
+    return nullptr;
+  }
+  return BN_bin2bn(CastToUChar(s_decoded), s_decoded.length(), NULL);
```
mtakigiku: the pointer to the allocated BIGNUM will be returned. Is it a problem?

src/envoy/auth/jwt.cc
```diff
@@ -90,19 +102,49 @@ const uint8_t *CastToUChar(const std::string &str) {
...
   return key;
 }
 
+bssl::UniquePtr<EVP_PKEY> EvpPkeyFromStr(const std::string &pkey_pem) {
+  std::string pkey_der = Base64::decode(pkey_pem);
+  return EvpPkeyFromRsa(bssl::UniquePtr<RSA>(
+      RSA_public_key_from_bytes(CastToUChar(pkey_der), pkey_der.length())));
+}
+
+BIGNUM *BigNumFromBase64UrlString(const std::string &s) {
```
mtakigiku: Using UniquePtr<BIGNUM> doesn't work because it causes that rsa->n and rsa->e are deleted before signature validation. Any idea to resolve this?

src/envoy/auth/jwt.cc
lizan: so when it gets freed?

mtakigiku: In `RSA_free(rsa)`, `BN_clear_free(rsa->n)` etc. are called

mtakigiku: I guess, this `BigNumFromBase64UrlString()` should be defined in `RsaFromJwk()` rather than under namespace

lizan: it is fine either, you can also

return bssl::UniquePtr and release it when assigns to rsa->n

src/envoy/auth/jwt.cc
```diff
@@ -90,17 +102,48 @@ const uint8_t *CastToUChar(const std::string &str) {
...
+
+bssl::UniquePtr<RSA> RsaFromJwk(const std::string &n, const std::string &e) {
+  bssl::UniquePtr<RSA> rsa(RSA_new());
   if (!rsa) {
+    // Couldn't create RSA key.
     return nullptr;
   }
-  bssl::UniquePtr<EVP_PKEY> key(EVP_PKEY_new());
-  EVP_PKEY_set1_RSA(key.get(), rsa.get());
+  auto BigNumFromBase64UrlString = [](const std::string &s) -> BIGNUM * {
```
lizan: Ugh, sorry I meant lambda doesn't make it clearer, and this is inefficient since you're making an function object every time.

src/envoy/auth/jwt.cc
mtakigiku: I don't understand how to use UniquePtr correctly. 
If we release BIGNUMs just after assigning it to rsa->n,e, it does not work.

mtakigiku: I'm confused.
It seems that these BIGNUMs must not be released until verification completes, but which implies RSA_free() didn't release them.

mtakigiku: maybe one way to resolve this is to make a class and put all stuffs as members

lizan:
The unique_ptr release is to release the ownership, not freeing the object

src/envoy/auth/jwt_test.cc
```diff
@@ -174,6 +174,139 @@ class JwtTest : public testing::Test {
...
+      "\"b3319a147514df7ee5e4bcdee51350cc890cc89e\",\"n\": "
+      "\"qDi7Tx4DhNvPQsl1ofxxc2ePQFcs-L0mXYo6TGS64CY_"
+      "2WmOtvYlcLNZjhuddZVV2X88m0MfwaSA16wE-"
+      "RiKM9hqo5EY8BPXj57CMiYAyiHuQPp1yayjMgoE1P2jvp4eqF-"
+      "BTillGJt5W5RuXti9uqfMtCQdagB8EC3MNRuU_KdeLgBy3lS3oo4LOYd-"
+      "74kRBVZbk2wnmmb7IhP9OoLc1-7-9qU1uhpDxmE6JwBau0mDSwMnYDS4G_ML17dC-"
+      "ZDtLd1i24STUw39KH0pcSdfFbL2NtEZdNeam1DDdk0iUtJSPZliUHJBI_pj8M-2Mn_"
+      "oA8jBuI8YKwBqYkZCN1I95Q\",\"e\": \"AQAB\"}]}";
+
+  const std::string kPrivateKey =
```
lizan: where is this used?

mtakigiku: This is just a note

lizan: Yes, let's keep this as a comment

lizan:
@mtakigiku Can you try squash this to one commit? Not sure why CLA is still waiting.

mtakigiku:
squashed

lizan:
The filter is still WIP, this PR is just the part for JWT validation, not built into the sidecar yet.

PR #396, opened 2017-08-03
Refactor build to allow creation of the .deb file and for consistency

costinm:
Pilot and mixer are using the 'new' style, with all deps declared in the workspace, I've also
moved the custom BUILD to separate files, to make it more readable, and removed the old
repositories.

The main addition are boilerplates for .deb, pulling the pilot-agent from the pilot repo.


qiwzhang:
Can we split this into multiple smaller PR?   For example, one for just building pilot-agent.   another one for removing sub-folder repositories.bzl.

kyessenov:
This is introducing a circular dependency. Is that OK?

costinm:
The idea is to automate the generation of the WORKSPACE - use repo or godep or other tools
to manage the SHAs, and simply copy all custom BUILD files and deps you need. Having a clear BUILD file that can be copied is simpler than cut&pasting py variables.

lizan:
Please document that intention and how to do WORKSPACE generation.

lizan:
Also FYI, per https://groups.google.com/d/msgid/istio-dev/CAAAxnCQD%3DQaQCfSkBn4JbofB9jcXF4bUbMYRBJ5rMWtvErKeGA%40mail.gmail.com?utm_medium=email&utm_source=footer
I will simplify the dependency loading from envoy by directly rely on their repository_rules soon.

ldemailly:
we may have to do something like this if the full solution takes too long but I really think we should try to push for the build flattening to happen - in istio/istio or istio/build rather than hack the proxy build
https://github.com/istio/istio/issues/392
cc @sebastienvas

PR #397, opened 2017-08-03
Make binary target visible, so it can be called from other workspaces

costinm:


lizan:
Also FYI, per https://groups.google.com/d/msgid/istio-dev/CAAAxnCQD%3DQaQCfSkBn4JbofB9jcXF4bUbMYRBJ5rMWtvErKeGA%40mail.gmail.com?utm_medium=email&utm_source=footer 
I will simplify the dependency loading from envoy by directly rely on their `repository_rule`s soon.

PR #401, opened 2017-08-10
Clean up tcp filter attributes.

qiwzhang:
1) Added context.protocol for both http and tcp
2) following for tcp
context.time
connection related attributes.

### DIFF
--- a/src/envoy/mixer/mixer_control.cc
+++ b/src/envoy/mixer/mixer_control.cc
@@ -24,6 +24,7 @@
 #include "src/envoy/mixer/thread_dispatcher.h"
 #include "src/envoy/mixer/utils.h"
 
+using namespace std::chrono;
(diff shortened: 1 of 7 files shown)


src/envoy/mixer/mixer_control.cc
```diff
@@ -24,6 +24,7 @@
 #include "src/envoy/mixer/thread_dispatcher.h"
 #include "src/envoy/mixer/utils.h"
 
+using namespace std::chrono;
```
lizan: per style guide, no `using namespace`

qiwzhang: I remember it is OK in cc file
not in header file

qiwzhang: Done.  removed "using namespace"

PR #402, opened 2017-08-11
Add per-route mixer attributes.

qiwzhang:
Get mixer_attributes and mixer_forward_attributes from route opaque config.

### DIFF
--- a/src/envoy/mixer/README.md
+++ b/src/envoy/mixer/README.md
@@ -131,7 +131,23 @@ By default, the mixer filter only forwards attributes and does not call mixer se
    }
 ```
 
-This route opaque config reverts the behavior by sending requests to mixer server but not forwarding any attributes.
+Above route opaque config reverts the behavior by sending requests to mixer server but not forwarding any attributes.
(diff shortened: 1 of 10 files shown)


src/envoy/mixer/README.md
```diff
@@ -131,7 +131,23 @@ By default, the mixer filter only forwards attributes and does not call mixer se
...
+Mixer attributes and forward attributes can be set per-route in the route opaque config.
+
+```
+ "routes": [
+   {
+     "timeout_ms": 0,
+     "prefix": "/",
+     "cluster": "service1",
+     "opaque_config": {
+      "mixer_attributes.key1": "value1",
```
lizan: does `opaque_config` need to be flatten? not "opaque_config": { "mixer": "attributes"... } ?

qiwzhang: Envoy only supports flag name value pairs in opaque config

https://github.com/lyft/envoy/blob/master/include/envoy/router/router.h#L250

PR #403, opened 2017-08-11
Add proper user and sudo setup and lsof command

ldemailly:
~Dropped emacs and nano (anyone using those and not knowing basic vi
commands ?)~
~Dropped wrk (curl (and fortio) should be plenty)~

by popular request, only additions: lsof and sudo setup

### DIFF
--- a/docker/Dockerfile.debug
+++ b/docker/Dockerfile.debug
@@ -11,10 +11,11 @@ RUN apt-get update && \
     net-tools \
     libc6-dbg gdb valgrind \
     vim \
-    nano \
-    emacs \
-    wrk
+    lsof \
+    sudo
 ADD proxy-redirection-clear /usr/local/bin/proxy-redirection-clear
 ADD proxy-redirection-restore /usr/local/bin/proxy-redirection-restore
 ADD envoy /usr/local/bin/envoy
+RUN useradd -m --uid 1337 istio-proxy
+RUN echo "istio-proxy ALL=NOPASSWD: ALL" >> /etc/sudoers
 ENTRYPOINT /usr/local/bin/envoy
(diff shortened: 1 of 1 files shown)


ldemailly:
Instead of "i have no name" you now get:
```
$  kubectl exec raw-svc-deployment-3298127744-0g269 -c istio-proxy -ti bash
istio-proxy@raw-svc-deployment-3298127744-0g269:/$ sudo bash
root@raw-svc-deployment-3298127744-0g269:/#
```

docker/Dockerfile.debug
```diff
@@ -11,10 +11,11 @@ RUN apt-get update && \
     net-tools \
     libc6-dbg gdb valgrind \
     vim \
-    nano \
-    emacs \
-    wrk
```
kyessenov: let's not remove wrk

ldemailly: ok, but what do you use it for ? in proxy container?

rkpagadala: shriram wanted nano and emacs I think. Lets not start a vi vs emacs.

ldemailly: we should have a separate "dev/testing env" image than one for proxy_debug
it seems quite wasteful to have 3 editors and a load tester but for the sake of getting the PR in I guess I'll "add" only

rshriram: you can drop emacs.. but please do keep nano. I am vi handicapped. Second, do we need another tier of debug images? May be sometime in future, but I was under the impression that our debug is kinda kitchen sink, and not a production friendly image. For production, this would be super stripped down for both perf and security. We would not be running valgrind, nor would we have vim/emacs/nano/fortio or anything, except envoy, agent and iptables script

ldemailly:
updated, ptal

ldemailly:
@rshriram , @kyessenov , @rkpagadala can one of you accept/review  ?

PR #404, opened 2017-08-12
Add support for building a standalone deb package

costinm:
Includes a more narrow iptables capture, and systemd control.

The startup script will detect if pilot-agent is installed, and use it - otherwise will start envoy with 
the included config file, without the cert monitoring and other agent features.


tools/deb/BUILD
```diff
@@ -0,0 +1,51 @@
...
+
+pkg_tar(
+    name = "envoy-bin",
+    srcs = [
+        "istio-iptables.sh",
+        "istio-start.sh",
+        "//src/envoy/mixer:envoy",
+    ],
+    mode = "0755",
+    package_dir = "/usr/local/bin",
```
lizan: A deb package installed binaries and scripts should go to `/usr/bin` (managed by package system), not to `/usr/local/bin` (for locally compiled installation).

costinm: AFAIK that's for distribution-provided binaries ( i.e. if debian or ubuntu would build the package). 
Google packages typically go to /opt/ (chrome), that would be my first choice.

However - for now I think it's important to minimize disruption, the docker images use /usr/local/bin and it's hardcoded into pilot agent - no need to confuse users with 2 locations. After we decide where it should go - we can update agent, docker, etc and keep things consistent.

ldemailly:
it doesn't pass the build 
ERROR: error loading package 'tools/deb': Package 'tools/deb' contains errors.

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
ldemailly: this is modified from pilot's right ? can you add a link to the original one in the header (and in a pilot PR so people are aware they need to modify both some times) ?
with a TODO: unify/have 1

tools/deb/istio-start.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
ldemailly: I think you can assume /bin/bash
(or write /bin/sh compatible code)

tools/deb/istio-start.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Script to run Envoy for Istio.
+
```
ldemailly: add set -e as for instance if failing to read the sidecar.env it would be dangerous to run the iptables script

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,111 @@
...
+      ;;
+  esac
+done
+
+if [[ -z "${ENVOY_PORT-}" ]]; then
+  ENVOY_PORT=15001
+fi
+
+if [[ -z "${ENVOY_UID-}" ]]; then
+  ENVOY_UID=`id -u ${ENVOY_USER-istio}`
```
ldemailly: nit : ` is discouraged nowadays, use $(id ...)

more important: so this will fail if the user is missing ? (which is good)
maybe explicitly check for ENVOY_UID not found with an explicit error message ?

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,111 @@
...
+      ;;
+    \?)
+      echo "Invalid option: -$OPTARG" >&2
+      usage
+      exit 1
+      ;;
+  esac
+done
+
+if [[ -z "${ENVOY_PORT-}" ]]; then
```
ldemailly: what's the - doing when not giving a default value ?

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,111 @@
...
+  ENVOY_PORT=15001
+fi
+
+if [[ -z "${ENVOY_UID-}" ]]; then
+  ENVOY_UID=`id -u ${ENVOY_USER-istio}`
+  ENVOY_UID=${ENVOY_UID},0
```
ldemailly: so we should document the svc should not be running as root

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,111 @@
...
+      exit 1
+      ;;
+  esac
+done
+
+if [[ -z "${ENVOY_PORT-}" ]]; then
+  ENVOY_PORT=15001
+fi
+
+if [[ -z "${ENVOY_UID-}" ]]; then
```
ldemailly: same

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,111 @@
...
+iptables -t nat -X ISTIO_REDIRECT
+iptables -t nat -N ISTIO_REDIRECT                                             -m comment --comment "istio/redirect-common-chain"
+iptables -t nat -A ISTIO_REDIRECT -p tcp -j REDIRECT --to-port ${ENVOY_PORT}  -m comment --comment "istio/redirect-to-envoy-port"
+
+IFS=,
+if [ "${ISTIO_LOCAL_PORTS}" == "*" ]; then
+    # Redirect all inbound traffic to Envoy.
+    iptables -t nat -A PREROUTING -j ISTIO_REDIRECT                               -m comment --comment "istio/install-istio-prerouting"
+else
+    for port in ${SERVICE_PORTS}; do
```
ldemailly: bug ? meant to be ISTIO_LOCAL_PORTS ?

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,111 @@
...
+iptables -t nat -N ISTIO_OUTPUT                                               -m comment --comment "istio/common-output-chain"
+
+# Jump to the ISTIO_OUTPUT chain from OUTPUT chain for all tcp
+# traffic. '-j RETURN' bypasses Envoy and '-j ISTIO_REDIRECT'
+# redirects to Envoy.
+iptables -t nat -A OUTPUT -p tcp -j ISTIO_OUTPUT                              -m comment --comment "istio/install-istio-output"
+
+# Redirect app calls to back itself via Envoy when using the service VIP or endpoint
+# address, e.g. appN => Envoy (client) => Envoy (server) => appN.
+iptables -t nat -A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -j ISTIO_REDIRECT     -m comment --comment "istio/redirect-implicit-loopback"
```
ldemailly: I'm a bit confused by this one - it's on lo (loopback) but excludes localhost ?
cc @ayj

tools/deb/istio-iptables.sh
costinm: I started with the one in pilot - but I don't think it's safe (or good idea) to attempt to unify, quite the opposite.

tools/deb/istio-iptables.sh
costinm: Fixed.

tools/deb/istio-iptables.sh
costinm: Fixed.

tools/deb/istio-iptables.sh
costinm: Done

tools/deb/istio-iptables.sh
costinm: Yes - will go to the docs ( unless we decide to allow it ), for now I want to minimize risks

tools/deb/istio-iptables.sh
costinm: Done

tools/deb/istio-iptables.sh
costinm: That's what we do in k8s. My understanding is that if destination address is a local address (i.e. send to the IP address of local machine), it'll be routed to loopback interface, but the DST IP will be the actual IP of the machine, not 127.0.0.1.

tools/deb/istio-start.sh
costinm: Done

tools/deb/BUILD
lizan: See https://wiki.debian.org/FilesystemHierarchyStandard
`/opt` is for "Pre-compiled, non ".deb" binary distribution (tar'ed..) goes here."

https://wiki.debian.org/HowToPackageForDebian clearly says "DO NOT INSTALL ANY FILES" under `/usr/local`

For now we can do `/usr/local` if it involve many components, but eventually this should go to `/usr`.

tools/deb/istio-iptables.sh
ldemailly: yes there are risks but I do think cross reference is useful - including highlighting the differences

tools/deb/istio-iptables.sh
```diff
@@ -1,78 +1,46 @@
...
+if [ ! -r $ISTIO_SIDECAR_CONFIG ]; then
+  echo "Missing sidecar config: $ISTIO_SIDECAR_CONFIG"
+  exit 1
 fi
 
-if [[ -z "${ENVOY_UID-}" ]]; then
-  ENVOY_UID=`id -u ${ENVOY_USER-istio}`
+. $ISTIO_SIDECAR_CONFIG
+
+if [ -z "${ENVOY_UID:-}" ]; then
```
ldemailly: you still have `:-}` here (does that change anything vs just the variable ? (I don't know))

tools/deb/istio-iptables.sh
```diff
@@ -1,78 +1,46 @@
...
+  # Default to the UID of ENVOY_USER
+  ENVOY_UID=$(id -u ${ENVOY_USER:-istio})
+  if [ ! $? == 0 ]; then
```
ldemailly: should be `$? != 0` or `$? -ne 0`
the above is non standard (wouldn't work with bourne shell / test command) (and also longer/more confusing than != )

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,111 @@
...
+fi
+
+
+# Create a new chain for redirecting inbound traffic to the common Envoy port.
+iptables -t nat -X ISTIO_REDIRECT
+iptables -t nat -N ISTIO_REDIRECT                                             -m comment --comment "istio/redirect-common-chain"
+iptables -t nat -A ISTIO_REDIRECT -p tcp -j REDIRECT --to-port ${ENVOY_PORT}  -m comment --comment "istio/redirect-to-envoy-port"
+
+IFS=,
+if [ "${ISTIO_LOCAL_PORTS}" == "*" ]; then
```
ayj: This assumes a single local address/interface. Is that okay for 0.3, or do we need to support multiple address/port tuples for inbound redirection?

tools/deb/istio-iptables.sh
```diff
@@ -1,78 +1,46 @@
...
+  ENVOY_UID=$(id -u ${ENVOY_USER:-istio})
+  if [ ! $? == 0 ]; then
+     echo "Invalid istio user $ENVOY_UID $ENVOY_USER"
+     exit 1
+  fi
   ENVOY_UID=${ENVOY_UID},0
 fi
 
-if [ "$1" = "clean" ]; then
+if [ "${1:-}" = "clean" ]; then
```
ldemailly: I'm confused why you add the :- (maybe I wasn't clear in my earlier comment to remove them)

ldemailly:
some of my comments don't show up maybe because of subsequent commits, but if you open show outdated they would

tools/deb/istio-iptables.sh
ayj: Costin's explanation is correct. The four combinations of inbound/outbound traffic redirection are possible (see below). This rule makes (1) work with local-to-local traffic. Without it we'd end up with (3) which breaks mTLS.
```
1) appN => Envoy (client) => Envoy (server) => appN.    # Istio for local invocation of service
2) appN => appN.                                        # bypass Istio completely 
3) appN => Envoy (client) > appN.                       # not allowed
4) appN => Envoy (server) => appN.                      # not allowed
```

ayj: Yes, @lizan added this in https://github.com/istio/pilot/pull/494 as a fix for  https://github.com/istio/pilot/pull/494

1) inject and apply https://github.com/kubernetes/contrib/blob/master/keepalived-vip/examples/echoheaders.yaml
2) exec into echoheader pod
3) `curl localhost:8080` is variant 2
4) `curl echoheaders:8080` is variant 1

ldemailly:
```
ERROR: /home/jenkins/workspace/proxy/presubmit/tools/deb/BUILD:3:1: //tools/deb:envoy-bin: no such attribute 'srcs' in 'pkg_tar' rule.
ERROR: /home/jenkins/workspace/proxy/presubmit/tools/deb/BUILD:14:1: //tools/deb:istio-conf: no such attribute 'srcs' in 'pkg_tar' rule.
ERROR: /home/jenkins/workspace/proxy/presubmit/tools/deb/BUILD:24:1: //tools/deb:istio-systemd: no such attribute 'srcs' in 'pkg_tar' rule.
```

costinm:
PTAL

tools/deb/BUILD
costinm: Agreed it shouldn't be /usr/local - but separate PRs needed across pilot.

On /opt - I know it's controversial, but seems to be ok for google-chrome. I'll check what k8s does. 

We can fine tune this - but we need to get testing started.

lizan: Sure let's tune this later. Can you add a TODO here or file an issue?

ldemailly:
```
ERROR: missing input file '//tools/deb:envoy.json'.
ERROR: /home/jenkins/workspace/proxy/presubmit/tools/deb/BUILD:24:1: error executing shell command: 'bazel-out/host/bin/external/bazel_tools/tools/build_defs/pkg/build_tar --flagfile=bazel-out/local-fastbuild/bin/tools/deb/istio-systemd.args' failed: bash failed: error executing command 
  (exec env - \
    PATH=/usr/lib/google-cloud-sdk/bin:/usr/local/go/bin:/usr/local/bin:/usr/lib/google-cloud-sdk/bin:/usr/local/go/bin:/usr/local/bin:/usr/lib/google-cloud-sdk/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
  /bin/bash -c 'bazel-out/host/bin/external/bazel_tools/tools/build_defs/pkg/build_tar --flagfile=bazel-out/local-fastbuild/bin/tools/deb/istio-systemd.args')
```

tools/deb/istio-iptables.sh
```diff
@@ -37,8 +38,15 @@ iptables -t nat -X ISTIO_OUTPUT
 iptables -t nat -N ISTIO_REDIRECT                                             -m comment --comment "istio/redirect-common-chain"
 iptables -t nat -A ISTIO_REDIRECT -p tcp -j REDIRECT --to-port ${ENVOY_PORT:-15001}  -m comment --comment "istio/redirect-to-envoy-port"
 
-IFS=,
-if [ ! -z "${ISTIO_LOCAL_PORTS:-}" ]; then
+# Makes sure SSH is not redirectred
+iptables -t nat -A PREROUTING -p tcp --dport 22 -j RETURN
```
ldemailly: 👍

tools/deb/sidecar.env
```diff
@@ -0,0 +1,27 @@
...
+# ENVOY_PORT=15001
+
+# User running Envoy. For testing you can use a regular user ID - however running iptables requires
+# root or netadmin capability.
+# ENVOY_USER=istio
+
+# Uncomment to enable debugging
+# ISTIO_DEBUG="-l debug"
+
+# SVC_IP=
```
ldemailly: add missing newline (use vi or set your editors to add newlines)

tools/deb/postinst.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
ldemailly: /bin/bash (please apply to all files)

tools/deb/postinst.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -e
+
+action="$1"
```
ldemailly: not used? is that safe/ok ?

tools/deb/postinst.sh
```diff
@@ -0,0 +1,25 @@
...
+umask 022
+
+if ! getent passwd istio >/dev/null; then
+    addgroup --system istio
+    adduser --system --group --home /var/lib/istio istio
+fi
+
+if [ ! -e /etc/istio ]; then
+   # Backward compat.
+   ln -s /var/lib/istio /etc/istio
```
ldemailly: which code hardcodes /etc/istio ? should we fix that (before 0.2 if not now)

tools/deb/postinst.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+mkdir -p /var/lib/istio/envoy
```
ldemailly: deb don't have the notion of an install prefix that we're supposed to use ? or variables for var/lib etc ?

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,87 @@
...
+. $ISTIO_SIDECAR_CONFIG
+
+if [ -z "${ENVOY_UID:-}" ]; then
+  # Default to the UID of ENVOY_USER
+  ENVOY_UID=$(id -u ${ENVOY_USER:-istio})
+  if [ ! $? == 0 ]; then
+     echo "Invalid istio user $ENVOY_UID $ENVOY_USER"
+     exit 1
+  fi
+  ENVOY_UID=${ENVOY_UID},0
```
ldemailly: rename to PROTECTED_UIDS or UNINTERCEPTED_UIDS here and where used maybe ? (would be clearer)

tools/deb/BUILD
ldemailly: we shouldn't have hardcoded paths (expect /usr/local/bin/...) anywhere - do we ? but agreed we can start somewhere and TODO the rest (same question as below for /etc/istio vs /var/lib/...)

tools/deb/istio-start.sh
```diff
@@ -0,0 +1,30 @@
...
+. $ISTIO_SIDECAR_CONFIG
+
+/usr/local/bin/istio-iptables.sh
+
+
+if [ -f /usr/local/bin/pilot-agent ]; then
+  exec su - istio /usr/local/bin/pilot-agent proxy > /var/log/istio/istio.log 2>&1
+else
+  if [ "" = "$SVC_IP" ]; then
```
ldemailly: -z "$SVC_IP"

ldemailly:
this is pretty cool this is taking shape btw, thanks for all the work to get there !

tools/deb/istio-start.sh
costinm: done ( a bit differently, but same idea)

tools/deb/BUILD
costinm: Added some customization, more later.

rshriram:
@costinm where are we going to host the debian package?

tools/deb/istio-start.sh
```diff
@@ -0,0 +1,30 @@
...
+    SVC_IP=$(hostname --ip-address)
+  fi
+  # Run envoy directly - agent not installed. This should be used only for debugging/testing standalone envoy
+  exec su - istio -c "/usr/local/bin/envoy -c /var/lib/istio/envoy/envoy.json --restart-epoch 0 --drain-time-s 2 --parent-shutdown-time-s 3 --service-cluster istio-proxy --service-node '$SVC_IP|mysvc.default|cluster.local' $ISTIO_DEBUG"
+fi
```
rshriram: I am conflicted about having this packaging reside in proxy, given that it depends on pilot agent. We already depend on istio/proxy in istio/pilot. Perhaps in istio/istio, where things are more stable?

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
rshriram: Missing copyright. Please add copyrights wherever possible, esp since this is being supplied as a package.

tools/deb/istio-iptables.sh
rshriram: +1. We really grokked on this script back in Feb. I would really like to understand what is different here, and make sure that we don't end up having to deal with two iptable variants without understanding what the heck is going on. In fact, I suggest that if possible, we have a single script with if platform==k8s else.. (where platform can be set in the beginning) and keep that script in pilot.

tools/deb/istio-start.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
rshriram: missing copyrights.

tools/deb/istio-10.conf
```diff
@@ -0,0 +1,4 @@
+[Service]
+Environment="ISTIO_EXTRA_ARGS="
+ExecStart=
+ExecStart=/opt/istio/bin/pilot $ISTIO_EXTRA_ARGS
```
rshriram: should be called pilot agent.. and where is the agent?

costinm:
Where to host and how to sign are TBD - I'm trying to build and test it
first :-)

costinm:
If we had a unified build - I agree, that's what I tried first.

The current build model is to use separate repos, each producing its own
artifacts - with pilot picking binaries from
web to build some combined packages. Yes, we already do this - but it
doesn't seem using docker as intermediary
is best build practice or optimal.

A side benefit of doing it in proxy is that we can also test/debug it in
isolation or with custom config - since LDS the generated
config is very static, the agent only checks for cert updates ( but there
are other ways to handle that)

costinm:
On Wed, Aug 16, 2017 at 1:28 PM, Shriram Rajagopalan <
notifications@github.com> wrote:

> *@rshriram* requested changes on this pull request.
>
> missing copyrights
>
Ack, will update all new files (original didn't have them, but for deb we
should follow the deb rules).


> ------------------------------
>
> In tools/deb/istio-iptables.sh
> <https://github.com/istio/proxy/pull/404#discussion_r133555477>:
>
> > @@ -0,0 +1,87 @@
> +#!/bin/bash
>
> Missing copyright. Please add copyrights wherever possible, esp since this
> is being supplied as a package.
> ------------------------------
>
> In tools/deb/istio-iptables.sh
> <https://github.com/istio/proxy/pull/404#discussion_r133556491>:
>
> > @@ -0,0 +1,111 @@
> +#!/bin/bash
>
> +1. We really grokked on this script back in Feb. I would really like to
> understand what is different here, and make sure that we don't end up
> having to deal with two iptable variants without understanding what the
> heck is going on. In fact, I suggest that if possible, we have a single
> script with if platform==k8s else.. (where platform can be set in the
> beginning) and keep that script in pilot.
>
It is possible - just more risky. The k8s container has hardcoded UIDs and
captures all - for raw VM we need more control and different defaults,
also needs to be idempotent, cleanup, etc.

In 0.3 we may simply include the istio-proxy.deb in the sidecar docker
(like we do with iptables, etc), and pick the new script, but I would like
to
have once cycle to test/get feedback.




> ------------------------------
>
> In tools/deb/istio-start.sh
> <https://github.com/istio/proxy/pull/404#discussion_r133556579>:
>
> > @@ -0,0 +1,30 @@
> +#!/bin/bash
>
> missing copyrights.
> ------------------------------
>
> In tools/deb/istio-10.conf
> <https://github.com/istio/proxy/pull/404#discussion_r133557723>:
>
> > @@ -0,0 +1,4 @@
> +[Service]
> +Environment="ISTIO_EXTRA_ARGS="
> +ExecStart=
> +ExecStart=/opt/istio/bin/pilot $ISTIO_EXTRA_ARGS
>
> should be called pilot agent.. and where is the agent?
>

Forgot about this, merged it from the initial attempt.

tools/deb/istio-iptables.sh
ldemailly: @rshriram we should unify eventually we agree but for now we want to make sure we don't leave a VM in unreachable state, and we need to iterate on the VM iptables faster/make more changes than the regular ones. So the unification will be post 0.2

rshriram: Sure. But atleast provide the diff or comment container lines and add vm lines so that others can understand what is different here vs vm. I would like to review these rules but it's going to take a while if I have to run through the entire thing from scratch. Having the what changed portion would be very helpful.

tools/deb/istio-iptables.sh
rshriram: Once again folks, this script is the crux of our entire system unfortunately. This is how we do the magical transparent proxying. we should have one shared across repos before final 0.2 release (can tolerate separate ones for couple of initial test releases). 
 However, I would like to keep the divergence between two versions to the absolute minimum possible. If variable names are not clear, let's fix in the pilot repo as well. If you want, I can take a crack at this script. 

cc @ayj @louiscryan @kyessenov
@tgraf FYI since you were asking about our pain points for transparent proxying.

rshriram:
I think we should have the single deb package from 0.2 itself. Would rather have this subsystem stabilized now than later. If this means shipping from proxy repo, that's fine. Btw, keep in mind that proxy is going to go away eventually and get merged with Envoy upstream. We just haven't gotten around to it yet.

costinm:
I want to avoid destabilizing the k8s script, or making it too complex. The
needs of k8s are much simpler - we hardcode
the UID of istio user, capture everything, etc. No need to increase the
risks - just to 'reuse' few lines of code.

I can make this script backward compatible ( i.e. by default behave like
the one in k8s) - and in 0.3 remove the old
script.

Regarding of location - it's driven by the build system. I don't think we
should do hacks like build the binary in one workspace
and upload/download/copy the binary to another workspace - if envoy is
build in proxy repo, that's the natural place for
the envoy .deb file and all artifacts. The docker image with envoy is
created in this workspace as well, and that's how
typical .deb packages are built ( from the source code repo ).

Having the iptable script along envoy is useful because it is possible to
test and use envoy with manual config - there
is little specific to pilot in the iptables script.

For 0.3 I'm planning to also change how the proxy docker is created - use
the .deb file directly - but I don't like
useless risks and breakages. After the .deb gets more testing we can start
using it more.

costinm:
If you really want to combine the 2 scripts - I can make this script fully
backward compatible with the other (i.e. if no
env file exists, behave exactly like the old one, including CLI.

I think we can have it included in the proxy docker image as well - and
probably use that as both 'initializer' and sidecar
( no need to have 2 different docker images, one with iptables+script and
one with envoy).

ldemailly:
let's keep the scripts separate for now - just put a reference to each other in the comment header

tools/deb/istio-iptables.sh
ldemailly: @rshriram the original script has a single uid which is the envoy process'es UID, this script supports a list which is why the name should change.

not to beat a dead horse but the other scripts has 4 rules and this one has to be more sophisticated to not kill the vm. we will merge LATER

docker/Dockerfile.debug
```diff
@@ -16,9 +16,11 @@ RUN apt-get update && \
     wrk \
     lsof \
     sudo
+# Deprecated: use "/usr/local/bin/istio-iptables.sh clear"
```
ldemailly: do you need to add the script to the images then ?

kyessenov:
We definitely don't want to merge the init and proxy containers. They have different security privileges - proxy does not have network admin capability.

kyessenov:
Besides, we also need to allow non-transparent mode, and that no init/iptables magic at all.

tools/deb/description
```diff
@@ -0,0 +1 @@
+Istio proxy provides the envoy sidecar used by the http://istio.io project.
```
ldemailly: (same nit again) vi this file (add a newline)

ps: please go over all my previous comments - tia

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,167 @@
...
+# Create a new chain for redirecting inbound traffic to the common Envoy port.
+# In the ISTIO_INBOUND and ISTIO_OUTBOUND chains, '-j RETURN' bypasses Envoy
+# and '-j ISTIO_REDIRECT' redirects to Envoy.
+iptables -t nat -N ISTIO_REDIRECT
+iptables -t nat -A ISTIO_REDIRECT -p tcp -j REDIRECT --to-port ${ENVOY_PORT:-15001}
+
+# Handling of inbound ports. Traffic will be redirected to Envoy, which will process and forward
+# to the local service.
+if [ -n "${ISTIO_INBOUND_PORTS:-}" ]; then
+  iptables -t nat -N ISTIO_INBOUND
```
rshriram: what if inbound_ports is missing? What is this env for?

tools/deb/istio-start.sh
```diff
@@ -0,0 +1,54 @@
...
+if [ -z "${ISTIO_SVC_IP:-}" ]; then
+  ISTIO_SVC_IP=$(hostname --ip-address)
+fi
+
+if [ -z "${POD_NAME:-}" ]; then
+  POD_NAME=$(hostname -s)
+fi
+
+if [ -f ${ISTIO_BIN_BASE}/pilot-agent ]; then
+  exec su -s /bin/bash -c "INSTANCE_IP=${ISTIO_SVC_IP} POD_NAME=${POD_NAME} POD_NAMESPACE=${NS} exec ${ISTIO_BIN_BASE}/pilot-agent proxy > ${ISTIO_LOG_DIR}/istio.log 2>&1" istio-proxy
```
ldemailly: why su ? what does this do vs just running the agent (
exec ... directly)

tools/deb/istio-start.sh
```diff
@@ -0,0 +1,54 @@
...
+else
+  ENVOY_CFG=${ENVOY_CFG:-${ISTIO_CFG}/envoy/envoy.json}
+  # Run envoy directly - agent not installed. This should be used only for debugging/testing standalone envoy
+  exec su -s /bin/bash -c "exec ${ISTIO_BIN_BASE}/envoy -c $ENVOY_CFG --restart-epoch 0 --drain-time-s 2 --parent-shutdown-time-s 3 --service-cluster istio-proxy --service-node 'sidecar~${ISTIO_SVC_IP}~mysvc.${NS}~${NS}svc.cluster.local' $ISTIO_DEBUG" istio-proxy
```
ldemailly: same question

tools/deb/istio-iptables.sh
ayj: The k8s iptable rules have a corresponding test (see https://github.com/istio/pilot/blob/master/docker/prepare_proxy_test.sh). It would be nice to see something similar for the VM rules. This will give us more confidence when we go to merge the VM and k8s rules.

To address @rshriram's comment, one way to make this reviewer easier is to commit a copy of the prepare_proxy.sh first and rebase these changes on top of that. The PR would then include the diff vs. the version we're already familiar with and highlight how far the VM script has diverged from k8s script.

docker/Dockerfile.debug
```diff
@@ -15,6 +15,7 @@ RUN apt-get update && \
     emacs \
     wrk \
     lsof \
+    busybox \
```
ldemailly: what's in that package that isn't in ubuntu:xenial and the other already ?

docker/Dockerfile.debug
```diff
@@ -23,8 +24,10 @@ ADD istio-iptables.sh /usr/local/bin/istio-iptables.sh
 ADD istio-start.sh /usr/local/bin/istio-start.sh
 ADD envoy.json /var/lib/istio/envoy/envoy.json
 ADD envoy /usr/local/bin/envoy
-RUN useradd -m --uid 1337 istio-proxy
```
ldemailly: useradd (and chown) part probably need to be in non debug too (but not sudo)

tools/deb/istio-start.sh
```diff
@@ -44,8 +40,20 @@ if [ -z "${POD_NAME:-}" ]; then
   POD_NAME=$(hostname -s)
 fi
 
+
+# Init option will only initialize iptables. Can be used
+if [[ ${1-} == "init" || ${1-} == "-p" ]] ; then
+  # Update iptables, based on current config. This is for backward compatibility with the init image mode.
+  # The sidecar image can replace the k8s init image, to avoid downloading 2 different images.
+  ${ISTIO_BIN_BASE}/istio-iptables.sh $*
```
ldemailly: should be "${@}"
but I'm concerned we're turning this into a multi command error prone shell script

tools/deb/test/envoy_local.json
```diff
@@ -0,0 +1,314 @@
+{
```
ldemailly: is this the shortest local config we can get, seems pretty long... and how will get updated?

tools/deb/test/run_iptables_test.sh
```diff
@@ -6,11 +6,12 @@
 # Script requires a working docker on the test machine
 # It is run in the proxy dir, will create a docker image with proxy deb installed
```
ldemailly: should this also run the iptable test found along the original script (in the pilot repo)

tools/deb/istio-start.sh
```diff
@@ -0,0 +1,62 @@
...
+  exit 0
+fi
+
+# Update iptables, based on config file
+${ISTIO_BIN_BASE}/istio-iptables.sh
+
+if [ -f ${ISTIO_BIN_BASE}/pilot-agent ]; then
+  exec su -s /bin/bash -c "INSTANCE_IP=${ISTIO_SVC_IP} POD_NAME=${POD_NAME} POD_NAMESPACE=${POD_NAMESPACE} exec ${ISTIO_BIN_BASE}/pilot-agent proxy > ${ISTIO_LOG_DIR}/istio.log 2>&1" istio-proxy
```
ldemailly: it'd be super awesome if you answer previous comments - can you do a full pass over every previous comment - including saying "disagree" or something rather than having comments up in the air where I don't know if you didn't see it or what

tools/deb/istio-iptables.sh
costinm: Undefined variables - for some reason most of our scripts are set to report error on undef (which is not bad)

tools/deb/istio-iptables.sh
costinm: Whitelist of ports we capture on 'inbound' - i.e. ports services run on. 
 The .env config file has more detailed comments on all settings.

I also have the * option ( which means everything except 22 - bad things happen if we try to capture 22 ), with
an exclude list.

costinm: If missing - we don't capture anything. That's for the case the VM is only using services, without providing any.

tools/deb/istio-start.sh
costinm: Done

tools/deb/istio-start.sh
costinm: The intent is for iptables script to be safe if sidecar.env is missing - and it should be possible to run it.
I'm still testing and tweaking the special cases - probably this can be improved in separate PR.

tools/deb/istio-start.sh
costinm: The script runs as root. 

We don't want to run the agent as root - in k8s that's handled by setting the uid on the pod.

tools/deb/istio-start.sh
costinm: Same answer :-)

Bad karma running stuff as root ( and besides the iptables are set for istio-proxy user )

tools/deb/istio-start.sh
costinm: I kind of tried all 3 options - if/when we switch to 'repo' or other 'unified' build it won't matter. But in the current bazel world - with manual SHA copy and total mess, I think proxy is the best place - that's where most of the sidecar is built.
It's also what would be expected with bazel - you build envoy in this repo, it should be packaged here too, same as we for docker.

One thing I'm conflicted about is the pilot-agent - right now I'm creating a separate .deb ( for same reasons - should be packaged where it is built, as would be expected ), k8s also has one deb per component. However it would be 
easier to make an exception and pull the agent into proxy.  

Can we iterate - I don't want to bike shed the location decision in the PR review :-)

tools/deb/postinst.sh
costinm: Done. Not sure why actually - env bash is supposed to be more portable, I've seen systems where bash is in /usr/bin without a link in bin. But long ago.

tools/deb/istio-start.sh
costinm: Done - sorry it took so long, I've been trying to add some basic tests and make sure everything works.

tools/deb/istio-start.sh
costinm: I'm adding tests - some of the commands are intended for reusing the script in k8s. 
We need to have some flexibility here - in future we can replace the error prone shell script 
with an error prone golang binary.

kube-proxy does all the iptables manipulation in go, and seems to work with CAP_NET_ADMIN only - but it's too complex and something like that is better in separate PR.

tools/deb/test/envoy_local.json
costinm: It's a test config - will get updated as we update tests. It is longer than the previous test template (in src/mixer/...) because it adds http_proxy tests and is closer to the pilot generated config.
We do have integration tests - but it's useful to also have smaller tests.

tools/deb/test/run_iptables_test.sh
costinm: AFAIK that script is not run automatically, doesn't seem to pass - and is a different type of test.
This is small test - using golden files. We should have both kind of tests - I'm working on an
integration test that would include iptables (using a GCE machine).

docker/Dockerfile.debug
costinm: If we have emacs - I don't think we need to worry too much about the few utilities that busybox is duplicating or adds :-) Added it for 'httpd' util, helps in some tests, and traceroute. 

We should probably have a separate thread about cleaning up the list - IMHO emacs is a bit too much, and wrk and few others could be added in separate images to load aside, don't need to be in the proxy.

tools/deb/test/run_iptables_test.sh
ayj: Thanks for the heads-up. I created https://github.com/istio/pilot/issues/1095 to track the issue.

tools/deb/BUILD
```diff
@@ -0,0 +1,52 @@
...
+    extension = "tar.gz",
+    deps = [
+        ":envoy-bin",
+        ":istio-conf",
+        ":istio-systemd",
+    ],
+)
+
+pkg_deb(
+    name = "istio-proxy",
```
rshriram: can we call this envoy-istio ? That is more truthful and would be in line with the terminology we use everywhere in docs. We try to avoid istio proxy in docs.

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,167 @@
...
+IFS=,
+if [ -n "${IP_RANGES_INCLUDE:-}" ]; then
+    for cidr in ${IP_RANGES_INCLUDE}; do
+        iptables -t nat -A ISTIO_OUTPUT -d ${cidr} -j ISTIO_REDIRECT
+    done
+    iptables -t nat -A ISTIO_OUTPUT -j RETURN
+else
+    iptables -t nat -A ISTIO_OUTPUT -j ISTIO_REDIRECT
+fi
+
```
rshriram: Thanks for the comments..

tools/deb/istio-iptables.sh
```diff
@@ -0,0 +1,167 @@
...
+# Handling of inbound ports. Traffic will be redirected to Envoy, which will process and forward
+# to the local service.
+if [ -n "${ISTIO_INBOUND_PORTS:-}" ]; then
+  iptables -t nat -N ISTIO_INBOUND
+  iptables -t nat -A PREROUTING -p tcp -j ISTIO_INBOUND
+
+  # Makes sure SSH is not redirectred
+  iptables -t nat -A ISTIO_INBOUND -p tcp --dport 22 -j RETURN
+
+  if [ "${ISTIO_INBOUND_PORTS:-}" == "*" ]; then
```
rshriram: what is * here?

tools/deb/postinst.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
rshriram: copyright!

docker/Dockerfile.release
```diff
@@ -1,4 +1,11 @@
 FROM ubuntu:xenial
+RUN apt-get update && \
```
rshriram: copyright

docker/Dockerfile.debug
```diff
@@ -15,10 +15,19 @@ RUN apt-get update && \
     emacs \
     wrk \
     lsof \
+    busybox \
```
rshriram: copyright

tools/deb/sidecar.env
```diff
@@ -0,0 +1,50 @@
+# Environment variables used to configure istio startup
```
rshriram: copyright

tools/deb/test/build_docker.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
```
rshriram: copyright. :(

tools/deb/test/iptables_tests.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
```
rshriram: copyright

tools/deb/test/run_iptables_test.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# Test 'istio-iptables' script, verify the rules generate the expected iptables.
```
rshriram: copyright

tools/deb/test/run_pilot_test.sh
```diff
@@ -0,0 +1,149 @@
+#!/bin/bash
+
+# Run a local pilot, local envoy, [local mixer], using a hand-crafted config.
```
rshriram: copyright

tools/deb/BUILD
costinm: Do you mind istio-envoy ? It would be nice for all istio- related debs to be close to each other in 
"dpkg -l". 

Technically we should have a meta-package istio-proxy, with possible selections - but don't know how to do that, and right now we only have envoy. My only concern is that a user may not know
what 'envoy' is - istio-proxy-envoy may at least let user guess it's the proxy/sidecar.

I'm fine either way - but since the name is already used in scripts, let's decide on one color :-)

tools/deb/istio-iptables.sh
costinm: Wildcard - do you have other preference ? It means to capture all ports, like we do in the docker image ( well, except 22)

tools/deb/postinst.sh
costinm: done

docker/Dockerfile.release
costinm: None of the dockerfiles in our repos (proxy or pilot) have one - is it required ?

tools/deb/sidecar.env
costinm: Same - equivalent config files ( yaml, etc) don't seem to have it. 
Not sure what are the requirements for non-source code.

costinm:
Shriram: if Docker and config files require copyright, we should open a bug to update all other files across repos, checked pilot. Otherwise - I don't think only this PR should add them. 

I'm getting close - just the install integration test still in progress, I would appreciate settling the 
names / locations for 0.2 - we can revisit it later. 

Also please consider what the default should be on iptables - on inbound I think it's clear "off by default",
to allow incremental migration. The question is for outbound - if CIDR is not set, right now I'm
defaulting to 'all outbound captured', like in docker. However I have doubts about it :-), in particular
for users with ufw or other iptables. Added some more comments and TODOs as well.

tools/deb/BUILD
ldemailly: https://istio.io/search/?q=%22istio%20proxy%22 has more hits than "envoy istio" or "istio envoy"
the repo is called proxy, may eventually have more than 1 proxy implementation, I don't think it needs to be named envoy here  (maybe sidecar is better than proxy)

docker/Dockerfile.release
ldemailly: curious about that too, I think it's covered at the repo level anyway - and for short files it's a lot of overhead ?

tools/deb/description
ldemailly: ping :-)

tools/deb/istio-start.sh
ldemailly: please fix $* it will bug out with spaces in params

tools/deb/postinst.sh
ldemailly: add TODO

tools/deb/test/run_gce_test.sh
```diff
@@ -0,0 +1,351 @@
...
+# - IP of the rawvm (will get it from gcloud if not set)
+#
+function istioConfigHttpService() {
+  local NAME=${1:-}
+  local PORT=${2:-}
+  local IP=${3:-}
+
+  # The 'name: http' is critical - without it the service is exposed as TCP
+
+  cat << EOF | kubectl apply -f -
```
ldemailly: to be replaced by istioctl register (https://github.com/istio/pilot/pull/1100) ideally (though that means building or downloading istioctl)

ldemailly:
I have only a few nits/leftover unaddressed comments (see inline comments) after which I think this is good to go - thx !
(though I didn't read carefully the long tests)

tools/deb/description
costinm: Done. Just curious, why is CRLF at end required ?

tools/deb/istio-start.sh
costinm: Done, thanks for the info - didn't know about this.

tools/deb/postinst.sh
costinm: added in the BUILD (which needs to be updated first)

tools/deb/test/run_gce_test.sh
costinm: Ack, we should add it in a separate PR (both kubeclt and istioctl need to be tested). 

Note that for istioctl we also want to test that it can be run from the VM itself, as we discussed for the DHCP/non fixed address. And in context of the new consul updates, we probably need to extend it to support consul as well ( in 0.3)

tools/deb/istio-start.sh
```diff
@@ -46,7 +46,7 @@ fi
 if [[ ${1-} == "init" || ${1-} == "-p" ]] ; then
   # Update iptables, based on current config. This is for backward compatibility with the init image mode.
   # The sidecar image can replace the k8s init image, to avoid downloading 2 different images.
-  ${ISTIO_BIN_BASE}/istio-iptables.sh $*
+  ${ISTIO_BIN_BASE}/istio-iptables.sh ${@}
```
ldemailly: this must be `"${@}"` (or `"$@"`)
and is not cosmetic (to preserve the number of arguments in presence of spaces etc)

tools/deb/test/run_pilot_test.sh
```diff
@@ -37,6 +37,13 @@ function build_all() {
     (cd $GOPATH/src/istio.io; git clone https://github.com/costinm/pilot -b deb)
   fi
 
+  if [[ -d $GOPATH/src/istio.io/istio ]]; then
+    (cd $GOPATH/src/istio.io/istio; git pull upstream master)
+  else
+    #(cd $GOPATH/src/istio.io; git clone https://github.com/istio/pilot)
+    (cd $GOPATH/src/istio.io; git clone https://github.com/costinm/istio)
```
ldemailly: probably don't want to merge that ?

tools/deb/test/run_gce_test.sh
ldemailly: if kubectl can be run, istioctl can be run

consul has it's own registration but yes in 0.3 we should look into unification, not in 0.2

ok to postpone using istioctl here for right now but any suggestion on how we'd get istioctl binary here ?

ldemailly:
approved but fix the 2 small issues

script/release-docker
```diff
@@ -16,35 +16,68 @@
 #
 ################################################################################
 #
+
+# Build the docker images, and upload them to gcloud.
+#
+# For development, you can pass "debug" parameter to only build debug image and skip the upload.
+# You can override DEBUG_IMAGE_NAME, RELEASE_IMAGE_NAME to specify a different destination for upload.
+# You can also set TAG, PROJECT, REPO - the upload will go to $REPO/$PROJECT/envoy[-debug]:$TAG
```
rshriram: can you copy the code from pilot/bin/e2e.sh (-hub -tag etc. ?) So that we have consistency across repos. And its more flexible here (no env vars

script/release-docker
```diff
@@ -16,35 +16,68 @@
...
+
+BAZEL_TARGET_DIR="bazel-bin/src/envoy/mixer"
+
+BAZEL_TARGET="${BAZEL_TARGET_DIR}/envoy"
+RELEASE_IMAGE_NAME=${RELEASE_IMAGE_NAME:-${REPO}/${PROJECT}/envoy:${TAG}}
+DEBUG_IMAGE_NAME=${DEBUG_IMAGE_NAME:-${REPO}/${PROJECT}/envoy-debug:${TAG}}
+
+if [[ ${1-} != "debug" ]] ; then
+   gcloud docker --authorize-only
+fi
```
rshriram: also sync with istio/pilot/bin/push-docker script if possible

tools/deb/BUILD
rshriram: istio-sidecar-envoy. For precisely the same reasons as @ldemailly stated above, we are packaging Envoy (not nginx or linkerd). So, it needs to be qualified. Don't care if its istio-prox-envoy or istio-envoy or whatever.. But please do use envoy in the name. 

And FYI, from the style guide
```Envoy

We prefer to use “Envoy” as it’s a more concrete term than “proxy” and will resonate if used consistently throughout the docs.

Synonyms:

“Envoy sidecar” - ok
“Envoy proxy” - ok
“The Istio proxy” – best to avoid unless you’re talking about advanced scenarios where another proxy might be used.
“Sidecar” – mostly restricted to conceptual docs
“Proxy – only if context is obvious
```

tools/deb/BUILD
```diff
@@ -0,0 +1,55 @@
...
+    architecture = "amd64",
+    built_using = "bazel",
+    data = ":debian-data",
+    description_file = "description",
+    homepage = "http://istio.io",
+    maintainer = "The Istio Authors <istio-dev@googlegroups.com>",
+    package = "istio-proxy",
```
rshriram: some *-envoy

tools/deb/postinst.sh
```diff
@@ -0,0 +1,43 @@
...
+################################################################################
+set -e
+
+action="$1"
+oldversion="$2"
+
+umask 022
+
+if ! getent passwd istio-proxy >/dev/null; then
+    addgroup --system istio-proxy
```
rshriram: call it istio please. (like Docker, which just has a docker user.. no docker-daemon/docker-x.)

tools/deb/description
ldemailly: only LF - because that's how most source code file are afaik, they are line based and contains N lines not N-1 ?

tools/deb/postinst.sh
ldemailly: that quite different. we may run proxy and mixer on same machine as different users

ldemailly:
@linsun is interested in potentially using this to make outbound capture only (for ingress like services) - so your idea of making it backward compatible may come in very handy ! we would need to make a docker image though etc...

linsun:
@ldemailly yes, we are interested in having the front side envoy capture turned off and just have the backside capture when istio is integrated in bluemix.   thank you for pointing to me this ISTIO_INBOUND_PORTS param, i'll try this next week.

PR #410, opened 2017-08-15
jwt-auth: failure status of JWT verification

mtakigiku:
Moved JWT verification functions into member functions of a class.
After verification, status ("OK" or the failure reason) is stored as a member of the class.


src/envoy/auth/jwt.cc
```diff
@@ -102,102 +129,78 @@ const uint8_t *CastToUChar(const std::string &str) {
...
 
-bssl::UniquePtr<EVP_PKEY> EvpPkeyFromRsa(RSA *rsa) {
-  if (!rsa) {
-    return nullptr;
-  }
-  bssl::UniquePtr<EVP_PKEY> key(EVP_PKEY_new());
-  EVP_PKEY_set1_RSA(key.get(), rsa);
-  return key;
-}
+class EvpPkeyGetter {
```
qiwzhang: please add class comment.

src/envoy/auth/jwt.cc
```diff
@@ -102,102 +129,78 @@ const uint8_t *CastToUChar(const std::string &str) {
...
+ public:
+  // It holds failure reason.
+  Status status_;
```
qiwzhang: member variables should be private

src/envoy/auth/jwt.cc
```diff
@@ -207,30 +210,38 @@ class Verifier {
  public:
   rapidjson::Document header;
   std::string alg;
```
qiwzhang: member variables should be private,  should tailing with _.

src/envoy/auth/jwt.h
```diff
@@ -26,18 +26,127 @@
...
+  // This function should be called before Decode().
+  virtual JwtVerifier& SetPublicKey(const std::string& pkey) = 0;
+
+  // This function verifies JWT signature and returns the decoded payload as a
+  // JSON if the signature is valid.
+  // If verification failed, it returns nullptr, and status_ holds the failture
+  // reason.
+  virtual std::unique_ptr<rapidjson::Document> Decode(
+      const std::string& jwt) = 0;
+  Status status_;
```
mtakigiku: Fixed

src/envoy/auth/jwt.cc
mtakigiku: Fixed

src/envoy/auth/jwt.cc
mtakigiku: Fixed

src/envoy/auth/jwt.cc
mtakigiku: Fixed

src/envoy/auth/jwt.cc
```diff
@@ -257,93 +290,195 @@ class Verifier {
...
+                       const uint8_t *signature, size_t signature_len,
+                       const uint8_t *signed_data, size_t signed_data_len) {
+    bssl::UniquePtr<EVP_MD_CTX> md_ctx(EVP_MD_CTX_create());
+    const EVP_MD *md = EvpMdFromAlg(alg);
+
+    if (!md) {
+      UpdateStatus(Status::ALG_NOT_IMPLEMENTED);
+      return false;
+    }
+    if (!md_ctx) {
```
mtakigiku: so should we call abort() if they are false?

lizan: No, trying to dereference a nullptr will automatically cause crash.

mtakigiku: fixed

src/envoy/auth/jwt.cc
```diff
@@ -102,138 +126,139 @@ const uint8_t *CastToUChar(const std::string &str) {
...
-    return false;
-  }
-  if (EVP_DigestVerifyUpdate(md_ctx.get(), signed_data, signed_data_len) != 1) {
-    return false;
-  }
-  if (EVP_DigestVerifyFinal(md_ctx.get(), signature, signature_len) != 1) {
-    return false;
+  bssl::UniquePtr<EVP_PKEY> EvpPkeyFromRsa(RSA *rsa) {
+    if (!rsa) {
+      UpdateStatus(Status::PUBKEY_RSA_OBJECT_NULL);
```
lizan: when will this happen? Parse error?

mtakigiku: Almost right. Now, there are 3 cases that it will happen:
- when `RSA_public_key_from_bytes()` returned `nullptr`, which is a parse error of PEM public key
- when `RSA_new()` returned `nullptr`, which I guess should be treated as crash
- when `BigNumFromBase64UrlString()` returned `nullptr`, which is a pares error of "n" or "e" field of a JWK

So, I guess this status should be renamed like `PUBKEY_PARSE_ERROR`.

mtakigiku:
I thought that we could keep `JwtVerifier` object after calling `SetPublicKey()` method, but your suggestion seems to be better. I'll fix it.

mtakigiku:
And, any reason to take `EvpPkeyGetter` as param? I think just passing `EVP_PKEY*` is fine.

src/envoy/auth/jwt.cc
mtakigiku: did

mtakigiku:
Now there are a lot of error status but we can classify them into three different kinds:
- error of JWT
- error of public key
- unmatch between JWT and public key
Any idea for implementation design to make this clear?

lizan:
I think it is fine, we can do that in follow up PRs.

mtakigiku:
And I would like to change the interface `JwtVerifier` in later PRs, so as to it has members like:
- `void Setup(string jwt)`
- `bool Verify(Pubkeys key)`
- `json Header()`
- `json Payload()`

sebastienvas:
build proxy/presubmit

sebastienvas:
build proxy/presubmit

sebastienvas:
build proxy/presubmit

sebastienvas:
build proxy/presubmit

src/envoy/auth/jwt_test.cc
```diff
@@ -329,90 +331,129 @@ TEST_F(JwtTest, InvalidSignature) {
   auto invalid_jwt = kJwt;
   invalid_jwt[kJwt.length() - 1] = kJwt[kJwt.length() - 1] != 'a' ? 'a' : 'b';
```
mtakigiku: Modifying the last character may make the signature part an invalid base64url input

mtakigiku: fixed

Issue #411, opened 2017-08-15
Allow simultaneous check and forward attributes


kyessenov:
Needed for ingress.

qiwzhang:
With #417 , #412 is reverted,  need to re-open

kyessenov:
It appears #412 is not the cause of seg faults. @rshriram is investigating.

PR #412, opened 2017-08-15
Fix mixer forward attribute bug.

qiwzhang:
Split CheckHTTP() into two calls:  BuildHttpCheck and SendCheck so that after BuildHttpCheck additional headers can be added.
Do the similar splitting for ReportHTTP, CheckTCP and ReportTCP


kyessenov:
thanks

qiwzhang:
https://github.com/istio/proxy/issues/411

PR #415, opened 2017-08-16
Fail open on 5xx from mixer or missing mixer clusters

rshriram:


qiwzhang:
Please use clang-format 3.8.0 to format the c++ code.   It is too much change.

Checking mixer failure is done here

https://github.com/istio/mixerclient/blob/master/include/options.h#L46

I think this change should be in mixerclient too.

rshriram:
I used clang-format-3.8 -style=Google and this is what it gave me.

src/envoy/mixer/utils.cc
```diff
@@ -13,6 +13,7 @@
  * limitations under the License.
  */
 
+#include "common/http/headers.h"
```
lizan: This should go after `#include "src/envoy/mixer/utils.h"` (also add an empty line), and breaks style check.

qiwzhang:
Here is the mixer client change:  https://github.com/istio/mixerclient/pull/96

rshriram:
@qiwzhang can I close this, assuming you have fixed the client?

PR #416, opened 2017-08-17
Use mock mixer

qiwzhang:


src/envoy/mixer/integration_test/check_report_test.go
```diff
@@ -87,7 +87,7 @@ const reportAttributesOkGet = `
      "date": "*",
      "content-type": "text/plain; charset=utf-8",
      "content-length": "0",
-     ":status": "200",
+     ":status": "'200'",
```
lizan: why there is extra `'`s?

src/envoy/mixer/integration_test/repositories.bzl
```diff
@@ -76,16 +73,19 @@ DESCRIPTOR_FILE_GROUP = [
     "mixer/v1/config/descriptor/value_type.proto",
 ]
 
-gogoslick_proto_library(
-    name = "mixer/v1/config",
+# gogoslick_proto_compile cannot be used here. it generates Equal, Size, and
```
lizan: Do you mean `gogoslick_proto_library`? It isn't clear to me why this causes compilation issues though. Does mixer do this too?

Issue #394, opened 2017-08-02
Mac build


ldemailly:
envoy is now buildable on macOS but when I build I get
```
$ bazel build -c opt src/envoy/mixer:envoy
[...]
WARNING: /private/var/tmp/_bazel_dl/eb119874c7b30dcafcbb687c93e2d621/external/envoy/bazel/repositories.bzl:31:5: External dep build exited with return code: 2.
WARNING: /private/var/tmp/_bazel_dl/eb119874c7b30dcafcbb687c93e2d621/external/envoy/bazel/repositories.bzl:32:5: grep: /proc/cpuinfo: No such file or directory
.
WARNING: /private/var/tmp/_bazel_dl/eb119874c7b30dcafcbb687c93e2d621/external/envoy/bazel/repositories.bzl:33:5: /private/var/tmp/_bazel_dl/eb119874c7b30dcafcbb687c93e2d621/external/envoy_deps/./repositories.sh: line 9: md5sum: command not found

External dependency cache directory /private/var/tmp/_bazel_dl/eb119874c7b30dcafcbb687c93e2d621/external/envoy_deps_cache_
.
ERROR: /private/var/tmp/_bazel_dl/eb119874c7b30dcafcbb687c93e2d621/external/envoy/bazel/repositories.bzl:178:13: no such package '@envoy_deps//': External dep build failed and referenced by '//external:spdlog'.
ERROR: Analysis of target '//src/envoy/mixer:envoy' failed; build aborted.
```

lizan:
we haven't rebased to latest envoy contains the build fix.

ldemailly:
it's close but still not there (for me, am I missing something ?)
```
ERROR: /private/var/tmp/_bazel_ldemailly/0811656893d1e79d77d5501da4894649/external/envoy/source/common/network/BUILD:11:1: C++ compilation of rule '@envoy//source/common/network:address_lib' failed (Exit 1).
external/envoy/source/common/network/address_impl.cc:81:15: error: use of undeclared identifier 'SOCK_NONBLOCK'
  int flags = SOCK_NONBLOCK;
              ^
1 error generated.
Target //src/envoy/mixer:envoy failed to build
```

ldemailly:
the current (master) error seems not mac specific:

ERROR: /Users/ldemailly/go/src/istio.io/proxy/src/envoy/mixer/BUILD:44:1: C++ compilation of rule '//src/envoy/mixer:filter_lib' failed (Exit 1).
[...]

PR #419, opened 2017-08-17
Fixing compile error

ldemailly:
See for before error
https://github.com/istio/proxy/issues/394#issuecomment-323203852


ldemailly:
@qiwzhang please confirm overriding is what we want here

PR #420, opened 2017-08-18
jwt-auth: Prototype of JWT verification filter

mtakigiku:
implementation detail is in [`ImplDesign.md`](https://github.com/mtakigiku/proxy/blob/jwt-auth-filter/src/envoy/auth/ImplDesign.md).


mtakigiku:
Updated config loading part according to [this](https://github.com/mtakigiku/proxy/tree/jwt-auth-readme/src/envoy/auth#config-format)

src/envoy/auth/config.cc
```diff
@@ -0,0 +1,138 @@
...
+#include <vector>
+
+namespace Envoy {
+namespace Http {
+namespace Auth {
+
+void AsyncClientCallbacks::onSuccess(MessagePtr &&response) {
+  auto len = response->body()->length();
+  std::string body(static_cast<char *>(response->body()->linearize(len)), len);
+  auto status = std::string(response->headers().Status()->value().c_str());
```
lizan: `std::string status = response->headers().Status()->value().c_str()`

src/envoy/auth/config.cc
```diff
@@ -0,0 +1,138 @@
...
+#include <string>
+#include <vector>
+
+namespace Envoy {
+namespace Http {
+namespace Auth {
+
+void AsyncClientCallbacks::onSuccess(MessagePtr &&response) {
+  auto len = response->body()->length();
+  std::string body(static_cast<char *>(response->body()->linearize(len)), len);
```
lizan: can body() be null? extract body inside `if`?

src/envoy/auth/config.cc
```diff
@@ -0,0 +1,138 @@
...
+  pos = pos == std::string::npos ? 0 : pos + 3;  // Start position of host
+  auto pos1 = uri.find("/", pos);
+  if (pos1 == std::string::npos) pos1 = uri.length();
+  std::string host = uri.substr(pos, pos1 - pos);
+  std::string path = "/" + uri.substr(pos1 + 1);
+
+  MessagePtr message(new RequestMessageImpl());
+  message->headers().insertMethod().value().setReference(
+      Http::Headers::get().MethodValues.Get);
+  message->headers().insertPath().value().append(path.c_str(), path.size());
```
lizan: .value(path)?

src/envoy/auth/config.cc
mtakigiku: fixed

src/envoy/auth/config.cc
mtakigiku: fixed

src/envoy/auth/config.cc
mtakigiku: fixed

mtakigiku:
Removed `ImplDesign.md` since [this](https://github.com/mtakigiku/proxy/tree/jwt-auth-readme/src/envoy/auth#config-format) is newer

mtakigiku:
I'm really regretting making this big PR and would like to make some comments.
This PR actually consists of some parts:

- Loading Envoy config
  `JwtAuthConfig::Load()` in `config.cc` is loading the filter config in Envoy config. It calls `IssuerInfo::Preload()` for each issuer. Filter object has a `JwtAuthConfig` object keeping the config.
  `IssuerInfo` object keeps the config of an issuer. `IssuerInfo::Preload()` (in `config.cc`) parses the config JSON for an issuer. In the case URI is provided, `IssuerInfo` keeps the URI and cluster name and the public key will be fetched later, namely in `JwtVerificationFilter::decodeHeaders()`.

- Fetching public key
  In `JwtVerificationFilter::decodeHeaders()`, requests for public keys to issuers if needed, using `AsyncClientCallbacks` in `config.cc`.
  `JwtVerificationFilter::ReceivePubkey()` is passed to AsyncClient as a callback, which will call `JwtVerificationFilter::CompleteVerification()` after all responses are received.
  
- Verifying JWT
  In `JwtVerificationFilter::CompleteVerification()` it verifies the JWT in HTTP header.

src/envoy/auth/config.cc
```diff
@@ -0,0 +1,138 @@
...
+        loaded_ = true;
+        return true;
+      } else if (json_pubkey->hasObject("value")) {
+        // Public key is written in this JSON.
+        pkey_ = json_pubkey->getString("value");
+        loaded_ = true;
+        return true;
+      }
+    }
+  }
```
qiwzhang: if required fields are missing, instead of just drop it,  it is better to log some error messages.

src/envoy/auth/config.cc
```diff
@@ -0,0 +1,138 @@
...
+      json.getInteger("pubkey_cache_expiration_sec", 600);
+
+  // Empty array if key "audience" does not exist
+  audiences_ = json.getStringArray("audience", true);
+
+  issuers_.clear();
+  if (json.hasObject("issuers")) {
+    for (auto issuer_json : json.getObjectArray("issuers")) {
+      auto issuer =
+          std::shared_ptr<IssuerInfo>(new IssuerInfo(issuer_json.get()));
```
qiwzhang: use std::make_shared

src/envoy/auth/config.cc
```diff
@@ -0,0 +1,138 @@
...
+namespace Auth {
+
+void AsyncClientCallbacks::onSuccess(MessagePtr &&response) {
+  auto len = response->body()->length();
+  std::string body(static_cast<char *>(response->body()->linearize(len)), len);
+  auto status = std::string(response->headers().Status()->value().c_str());
+  if (status == "200") {
+    cb_(true, body);
+  } else {
+    cb_(false, "");
```
qiwzhang: can we log status code if not 200. for easy debugging

src/envoy/auth/config.cc
```diff
@@ -0,0 +1,138 @@
...
+  }
+}
+void AsyncClientCallbacks::onFailure(AsyncClient::FailureReason) {
+  cb_(false, "");
```
qiwzhang: any error message we can log?

src/envoy/auth/config.h
```diff
@@ -0,0 +1,108 @@
...
+  bool loaded_;          // If the public key is loaded or not
+  std::string uri_;      // URI for public key
+  std::string cluster_;  // Envoy cluster name for public key
+
+  bool Preload(Json::Object *json);
+  std::string name_;       // e.g. "https://accounts.example.com"
+  std::string pkey_type_;  // Format of public key. "jwks" or "pem"
+  std::string pkey_;       // Public key
+
+  std::unique_ptr<AsyncClientCallbacks> async_client_cb_;
```
qiwzhang: I feel async_client_cb_ should not be part of this class.  It is just temporarily used.

src/envoy/auth/config.h
```diff
@@ -0,0 +1,108 @@
...
+
+// Struct to hold an issuer's information.
+struct IssuerInfo {
+  IssuerInfo(Json::Object *json) : loaded_(false) { failed_ = !Preload(json); }
+  bool failed_;          // True if Preload() or fetching public key failed
+  bool loaded_;          // If the public key is loaded or not
+  std::string uri_;      // URI for public key
+  std::string cluster_;  // Envoy cluster name for public key
+
+  bool Preload(Json::Object *json);
```
qiwzhang: can we move all functions together?

src/envoy/auth/http_filter.cc
```diff
@@ -13,36 +13,100 @@
...
-  return FilterHeadersStatus::Continue;
+
+  // list up issuers whose public key should be fetched
+  for (const auto& iss : config_->issuers_) {
+    if (!iss->failed_ && !iss->loaded_) {
+      calling_issuers_[iss->name_] = iss;
+    }
+  }
+  // send HTTP requests to fetch public keys
+  if (!calling_issuers_.empty()) {
```
qiwzhang: If the first client request starts an async call to fetch public key,  before it is responded,  what do we do for the second client request?

src/envoy/auth/http_filter.cc
```diff
@@ -51,5 +115,90 @@ void JwtVerificationFilter::setDecoderFilterCallbacks(
   decoder_callbacks_ = &callbacks;
 }
 
+void JwtVerificationFilter::ReceivePubkey(HeaderMap& headers,
+                                          std::string issuer_name, bool succeed,
+                                          const std::string& pubkey) {
+  auto iss_it = calling_issuers_.find(issuer_name);
```
qiwzhang: do we handle case where async call has been timeout,  and client request object has been destroyed.

src/envoy/auth/http_filter.cc
```diff
@@ -51,5 +115,90 @@ void JwtVerificationFilter::setDecoderFilterCallbacks(
...
+             */
+
+            /*
+             * TODO: replace appropriately
+             */
+            headers.addReferenceKey(AuthorizedHeaderKey(), payload_str);
+
+            // Remove JWT from headers.
+            headers.remove(kAuthorizationHeaderKey);
+            goto end;
```
qiwzhang: please don't use "goto"

src/envoy/auth/http_filter.cc
```diff
@@ -51,5 +115,90 @@ void JwtVerificationFilter::setDecoderFilterCallbacks(
...
+             * TODO: check exp claim
+             */
+
+            /*
+             * TODO: replace appropriately
+             */
+            headers.addReferenceKey(AuthorizedHeaderKey(), payload_str);
+
+            // Remove JWT from headers.
+            headers.remove(kAuthorizationHeaderKey);
```
qiwzhang: why add the header and delete it right away?

src/envoy/auth/config.cc
mtakigiku: we should use ENVOY_LOG to log?

src/envoy/auth/config.cc

src/envoy/auth/config.h
mtakigiku: done

src/envoy/auth/http_filter.cc
mtakigiku: First one is adding a new header("Istio-Auth-UserInfo"), and second one is deleting Authorization header
(I guess you thought `AuthorizedHeaderKey()` and `kAuthorizationHeaderKey` are the same)

src/envoy/auth/http_filter.cc
mtakigiku: fixed

src/envoy/auth/config.cc

src/envoy/auth/config.cc

src/envoy/auth/config.cc
```diff
@@ -0,0 +1,162 @@
...
+  message->headers().insertHost().value(host);
+
+  cm_.httpAsyncClientForCluster(cluster_->name())
+      .send(std::move(message), *this, timeout_);
+}
+
+IssuerInfo::IssuerInfo(Json::Object *json) {
+  ENVOY_LOG(debug, "IssuerInfo: {}", __func__);
+  if (json->hasObject("name") && json->hasObject("pubkey")) {
+    name_ = json->getString("name");
```
mtakigiku: `getString(string)` may throw exception when the object was not a string, so use `getString(string, string)` instead.

mtakigiku: fixed

src/envoy/auth/config.cc
```diff
@@ -0,0 +1,178 @@
...
+    user_info_type_ = UserInfoType::kHeaderPayloadBase64Url;
+  } else {
+    user_info_type_ = UserInfoType::kPayloadBase64Url;
+  }
+
+  pubkey_cache_expiration_sec_ =
+      json.getInteger("pubkey_cache_expiration_sec", 600);
+
+  // Empty array if key "audience" does not exist
+  audiences_ = json.getStringArray("audience", true);
```
mtakigiku: `getStringArray(string, bool)" may throw exception when given non-array

mtakigiku: fixed

src/envoy/auth/config.cc

src/envoy/auth/http_filter.cc
mtakigiku: fixed

src/envoy/auth/config.h
```diff
@@ -38,21 +38,25 @@ class AsyncClientCallbacks : public AsyncClient::Callbacks,
  public:
   AsyncClientCallbacks(Upstream::ClusterManager &cm, const std::string &cluster,
                        std::function<void(bool, const std::string &)> cb)
-      : cm_(cm),
+      : cancel_([this]() { cancelled_ = true; }),
```
qiwzhang: I think you may have to hold "request" object and call its cancel method.  As 

https://github.com/istio/proxy/blob/master/src/envoy/mixer/grpc_transport.cc#L115

mtakigiku: fixed

src/envoy/auth/config.h
mtakigiku: fixed

src/envoy/auth/http_filter.cc
mtakigiku: It would be done in following PRs for regular updates of public key

PR #425, opened 2017-08-23
jwt-auth: changed the interface of the class for JWT verification

mtakigiku:
as commented in the [former PR](https://github.com/istio/proxy/pull/410#issuecomment-323219386)


src/envoy/auth/jwt.cc
```diff
@@ -208,7 +196,38 @@ class EvpPkeyGetter {
   }
 };
 
-// Class to decode and verify JWT. Setup() must be called before
+std::pair<bool, std::string> GetStringFromJson(const rapidjson::Document &d,
```
lizan: How about reusing envoy's JsonLoader and JsonObject?
https://github.com/lyft/envoy/blob/master/include/envoy/json/json_object.h

src/envoy/auth/jwt_test.cc
```diff
@@ -307,11 +307,23 @@ class JwtTestWithJwk : public testing::Test {
       "37BD2KXH3jN9_EBocxwcxhkPwSLtP3dgkfls_f5GoWCgmp-c5ycIskCDcIjxRnPjg";
 };
 
+#define TEST_SETUP(public_key, jwt, verified) \
```
lizan: Use "Value Parameterized Tests".

See https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md

src/envoy/auth/jwt.cc
mtakigiku: you mean, replacing all `rapidjson::Document` in this lib with envoy's JsonObject? Or just for this function?

lizan: yes replacing rapidjson::Document if possible.

src/envoy/auth/jwt_test.cc
mtakigiku: Do you mean something like this?:
```
vector<tuple<string, string, bool> > testdata = {
  make_tuple(jwt1, pubkey1, verify_result1),
  make_tuple(jwt2, pubkey2, verify_result2),
  ...
};
INSTANTIATE_TEST_CASE_P(foobar, JwtTest, testing::ValuesIn(testdata));
...
```

src/envoy/auth/jwt.cc
mtakigiku: done

src/envoy/auth/jwt_test.cc
mtakigiku: done

src/envoy/auth/jwt_test.cc
```diff
@@ -307,242 +316,206 @@ class JwtTestWithJwk : public testing::Test {
...
 }
 
-TEST_F(JwtTestWithJwk, JwkBadPublicKey) {
-  std::string invalid_pubkey = R"EOF({"keys":[]})EOF";
-  JwtVerifier v;
-  auto payload = v.Decode(*Pubkeys::ParseFromJwks(invalid_pubkey), kJwtNoKid);
-  EXPECT_FALSE(payload);
-  EXPECT_EQ(v.GetStatus(), Status::JWK_NO_VALID_PUBKEY);
-}
+INSTANTIATE_TEST_CASE_P(TestCases, JwtTest, testing::ValuesIn(TestData()));
```
qiwzhang: I am worrying about if one of tests fail,  how easy to identify which one file,  does each test has a name to associate with it?

src/envoy/auth/jwt.h
```diff
@@ -134,12 +183,12 @@ class JwtVerifier : public WithStatus {
 class Pubkeys : public WithStatus {
  public:
   Pubkeys(){};
-  static std::unique_ptr<Pubkeys> ParseFromPem(const std::string &pkey_pem);
-  static std::unique_ptr<Pubkeys> ParseFromJwks(const std::string &pkey_jwks);
+  static std::unique_ptr<Pubkeys> ParseFromPem(const std::string& pkey_pem);
```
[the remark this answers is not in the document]
mtakigiku: done

src/envoy/auth/jwt_test.cc
mtakigiku: No, each test has only a number (corresponding to what-th test it is), and I agree that it is not so easy to see if it fails...

src/envoy/auth/jwt.cc
```diff
@@ -246,22 +228,48 @@ class Verifier {
...
+    // Parse header json
+    header_str_base64url_ = jwt_split[0];
+    header_str_ = Base64UrlDecode(jwt_split[0]);
+    try {
+      header_ = Json::Factory::loadFromString(header_str_);
+    } catch (...) {
       UpdateStatus(Status::JWT_HEADER_PARSE_ERROR);
       return false;
-    }
+    };
```
lizan: ?

src/envoy/auth/jwt_test.cc
lizan: IMO it is fine with numbers.

mtakigiku: I fixed it anyway

src/envoy/auth/jwt.cc
```diff
@@ -344,7 +355,81 @@ class Verifier {
   }
 };
 
-}  // namespace
+JwtVerifier::JwtVerifier(const std::string &jwt) {
+  impl_ = new Impl();
```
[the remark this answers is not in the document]
mtakigiku: Got a compile error if I use unique_ptr:
```
/usr/include/c++/5/bits/unique_ptr.h:74:22: error: invalid application of 'sizeof' to incomplete type 'Envoy::Http::Auth::JwtVerifier::Impl'
```
How can we fix it?

PR #426, opened 2017-08-23
Not to use therad local dispatch to delete grpc transport object

qiwzhang:
My suspicious is:  grpc callback function onSuccess() may be called in a different thread than filter->set_callback thread in some cases during race conditions.  This causes thread local dispatcher was not set for that thread. This caused assertion failure.

Just store the pending free objects in a thread local list.  For each new grpc call,  free the objects in the pending list.  So dispatcher is not needed.

thread local dispatcher is still used by timer_callback for report_batch.  It should be fine; it is called during Report() call which should be at the same thread as fitler->set_callback.   Beside, it only needs to call once for the whole process.

### DIFF
--- a/src/envoy/mixer/grpc_transport.cc
+++ b/src/envoy/mixer/grpc_transport.cc
@@ -52,16 +52,16 @@ inline void CopyHeaderEntry(const HeaderEntry* entry,
 }
 
 thread_local std::vector<GrpcTransport*> pending_freelist;
-  
+
 }  // namespace
 
 void GrpcTransport::FreePendingGrpcObject() {
-    for (GrpcTransport* obj : pending_freelist) {
-      delete obj;
-    }
-    pending_freelist.clear();
+  for (GrpcTransport* obj : pending_freelist) {
+    delete obj;
   }
-  
+  pending_freelist.clear();
+}
+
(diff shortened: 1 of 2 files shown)


lizan:
#427 would solve this

src/envoy/mixer/grpc_transport.cc
```diff
@@ -51,19 +51,32 @@ inline void CopyHeaderEntry(const HeaderEntry* entry,
...
+  pending_freelist.clear();
+}
+
 }  // namespace
 
 GrpcTransport::GrpcTransport(Upstream::ClusterManager& cm,
                              const HeaderMap* headers)
-    : channel_(NewChannel(cm)), stub_(channel_.get()), headers_(headers) {}
+    : channel_(NewChannel(cm)), stub_(channel_.get()), headers_(headers) {
+  FreePendingGrpcObject();
```
rshriram: This is not so clean. Is it possible to launch a timer to periodically cleanup the free list? This is just leaking memory, because if I do 100 GRPC calls followed by 1hr of silence, I am going to have 100 GrpcTransport objects lying around. I need to do 100 more calls to delete the previous 100 objects

PR #428, opened 2017-08-23
Not check cluster existence

lizan:
Grpc::AsyncClient holds a reference to ClusterManager and cluster_name, the mixer cluster can be configured after MixerControl construction without problem.


src/envoy/mixer/mixer_control.h
```diff
@@ -86,15 +86,16 @@ class MixerControl final : public Logger::Loggable<Logger::Id::http> {
 
   // Envoy cluster manager for making gRPC calls.
   Upstream::ClusterManager& cm_;
+
+  CheckTransport::AsyncClientPtr check_client_;
```
qiwzhang: it is more clear in the destructor to clear mixer_client first.

rshriram:
Do we still need this?

lizan:
No

PR #432, opened 2017-08-24
jwt-auth: add README.md

mtakigiku:
Added README.md and some scripts (backend, issuer) to run proxy.
Some features are not implemented yet, so this also serves a design of the configuration specification.


mtakigiku:
I noticed this later, but I don't think we should merge it now since it's WIP and merging this README now is confusing.

lizan:
Then can you add "WIP" or "DO NOT MERGE" in the subject of this PR?

mtakigiku:
I added some notes for not implemented parts, so please merge it if it seems to be good

PR #433, opened 2017-08-24
Use per request AsyncClient

lizan:
@qiwzhang can you test this to see if the segfault is still there?


src/envoy/mixer/grpc_transport.cc
```diff
@@ -36,6 +36,9 @@ const LowerCaseString kB3Sampled("x-b3-sampled");
 const LowerCaseString kB3Flags("x-b3-flags");
 const LowerCaseString kOtSpanContext("x-ot-span-context");
 
+// The name for the mixer server cluster.
+const char* kMixerServerClusterName = "mixer_server";
```
rshriram: Do all our configs have to have this cluster name hardcoded ? Also istio-mixer is better.

PR #436, opened 2017-08-24
Update Envoy to latest

lizan:


src/envoy/mixer/grpc_transport.cc
```diff
@@ -92,9 +92,10 @@ void GrpcTransport<RequestType, ResponseType>::onSuccess(
 
 template <class RequestType, class ResponseType>
 void GrpcTransport<RequestType, ResponseType>::onFailure(
-    Grpc::Status::GrpcStatus status) {
-  ENVOY_LOG(debug, "{} failed with code: {}", descriptor().name(), status);
```
qiwzhang: could you add debug to print request data when sending?

lizan: let's do that in separate PR, `const RequestType&` has different lifecycle so it might cause issue.

qiwzhang: can we add ENVOY_LOG(debug, )  here

https://github.com/lizan/proxy/blob/f8ec3d61e37f0a0da4683abf87ff1cad4f759001/src/envoy/mixer/grpc_transport.cc#L61

PR #437, opened 2017-08-24
fix attribute forwarding bug again.

qiwzhang:
It was reverted as suspected to cause crash.  Now crash is fixed,  put the change back.

### DIFF
--- a/src/envoy/mixer/http_filter.cc
+++ b/src/envoy/mixer/http_filter.cc
@@ -212,7 +212,7 @@ class Instance : public Http::StreamDecoderFilter,
     mixer_disabled_ = mixer_disabled();
     if (mixer_disabled_) {
       if (!forward_disabled()) {
-        mixer_control_->ForwardAttributes(
+        mixer_control_.ForwardAttributes(
             headers, GetRouteStringMap(kPrefixForwardAttributes));
       }
       return FilterHeadersStatus::Continue;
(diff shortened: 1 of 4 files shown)


src/envoy/mixer/mixer_control.h
```diff
@@ -55,38 +56,37 @@ class MixerControl final : public ThreadLocal::ThreadLocalObject,
...
-                  int check_status_code);
-
-  // Make mixer check call for Tcp requests.
-  void CheckTcp(HttpRequestDataPtr request_data,
-                Network::Connection& connection, std::string origin_user,
-                ::istio::mixer_client::DoneFunc on_done);
+  // Build check request attributes for HTTP.
+  void BuildHttpCheck(HttpRequestDataPtr request_data, HeaderMap& headers,
+                      const ::istio::proxy::mixer::StringMap& map_pb,
+                      std::string origin_user,
```
lizan: const reference?

src/envoy/mixer/http_filter.cc
```diff
@@ -230,11 +229,28 @@ class Instance : public Http::StreamDecoderFilter,
       origin_user = ssl->uriSanPeerCertificate();
     }
 
+    // Extract attributes from x-istio-attributes header
+    ::istio::proxy::mixer::StringMap forwarded_attributes;
+    const HeaderEntry* entry = headers.get(Utils::kIstioAttributeHeader);
+    if (entry) {
+      std::string str(entry->value().c_str(), entry->value().size());
```
lizan: const

qiwzhang: Done

src/envoy/mixer/mixer_control.h
qiwzhang: Done

PR #439, opened 2017-08-25
Cancel check call on destroy

### LINKED ISSUE #438: proxy crash at race condition with LDS update
The back trace log is here:  https://pastebin.com/PUm4G5sC

The theory is  that Mixer Config is deleted by Mixer filter instance since the latter holds a ref count to the former (https://github.com/istio/proxy/blob/master/src/envoy/mixer/http_filter.cc#L132).  A filter instance deletion will be delayed if there is a gRPC call in the progress (https://github.com/istio/proxy/blob/master/src/envoy/mixer/http_filter.cc#L254).  When the grpc call is completed, the instance will be deleted at that thread.   Then mixer config will be deleted at that thread too.  

But since mixer config holds envoy thread local slot (https://github.com/istio/proxy/blob/master/src/envoy/mixer/http_filter.cc#L105).  It has to be deleted at main thread.  it causes assertion (https://github.com/lyft/envoy/blob/master/source/common/thread_local/thread_local_impl.cc#L58).

lizan:
depends on https://github.com/istio/mixerclient/pull/102

Fixes #438

### DIFF
--- a/src/envoy/mixer/tcp_filter.cc
+++ b/src/envoy/mixer/tcp_filter.cc
@@ -58,12 +58,11 @@ typedef std::shared_ptr<TcpConfig> TcpConfigPtr;
 
 class TcpInstance : public Network::Filter,
                     public Network::ConnectionCallbacks,
-                    public Logger::Loggable<Logger::Id::filter>,
-                    public std::enable_shared_from_this<TcpInstance> {
+                    public Logger::Loggable<Logger::Id::filter> {
  private:
   enum class State { NotStarted, Calling, Completed, Closed };
 
-  TcpConfigPtr config_;
+  istio::mixer_client::CancelFunc cancel_check_;
   MixerControl& mixer_control_;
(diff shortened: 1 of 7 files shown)


src/envoy/mixer/tcp_filter.cc
```diff
@@ -125,14 +131,12 @@ class TcpInstance : public Network::Filter,
...
       filter_callbacks_->connection().readDisable(true);
-      auto instance = GetPtr();
       calling_check_ = true;
       mixer_control_.BuildTcpCheck(
           request_data_, filter_callbacks_->connection(), origin_user);
-      mixer_control_.SendCheck(request_data_, nullptr,
-                               [instance](const Status& status) {
-                                 instance->completeCheck(status);
-                               });
+      mixer_control_.SendCheck(
```
qiwzhang: cancel_check callback need to be saved.

PR #442, opened 2017-08-26
use envoy v1.4.0

ldemailly:
v1.4.0 (a8507f67225cdd912712971bf72d41f219eb74ed) slightly newer than 3f3fc94f85766b70ca6239dfbba7067e95ee0bdb

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -39,7 +39,7 @@ bind(
 git_repository(
     name = "envoy",
     remote = "https://github.com/lyft/envoy.git",
-    commit = "3f3fc94f85766b70ca6239dfbba7067e95ee0bdb",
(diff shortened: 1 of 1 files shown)


ldemailly:
wouldn't it be wise to get some stability and use 1.4.0 for istio 0.2 - at least until a bug fix requires update?

master of envoy is already full speed ahead on 1.5.x and may not be stable just when we need stability?

lizan:
There is almost no difference in term of stability v1.4.0 vs master. Master of envoy is considered release candidate quality and the versioning is more about config/API deprecation and backward compatibility.

See:
https://github.com/lyft/envoy/blob/master/CONTRIBUTING.md#release-cadence

PR #444, opened 2017-08-29
remove private repo/branches

ldemailly:
follow up on #404 missing (ah!) code review updates

### DIFF
--- a/tools/deb/test/run_pilot_test.sh
+++ b/tools/deb/test/run_pilot_test.sh
@@ -31,21 +31,19 @@ function build_all() {
 
 
   if [[ -d $GOPATH/src/istio.io/pilot ]]; then
-    (cd $GOPATH/src/istio.io/pilot; git pull upstream master)
+    (cd $GOPATH/src/istio.io/pilot; git fetch && git pull upstream master)
   else
-    #(cd $GOPATH/src/istio.io; git clone https://github.com/istio/pilot)
-    (cd $GOPATH/src/istio.io; git clone https://github.com/costinm/pilot -b deb)
+    (cd $GOPATH/src/istio.io; git clone https://github.com/istio/pilot)
   fi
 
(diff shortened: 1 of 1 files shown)


tools/deb/test/run_pilot_test.sh
```diff
@@ -31,21 +31,19 @@ function build_all() {
 
 
   if [[ -d $GOPATH/src/istio.io/pilot ]]; then
-    (cd $GOPATH/src/istio.io/pilot; git pull upstream master)
+    (cd $GOPATH/src/istio.io/pilot; git fetch && git pull upstream master)
```
lizan: why do you need to change pull to fetch & pull?
pull is just a short form of fetch + checkout. https://git-scm.com/docs/git-pull

ldemailly: when I'm on master and/or my own branches I think I need to fetch for pull to do anything so I've had the habit of always doing "git fetch && git pull" but maybe the "upstream master" part does the fetch somehow/or doesn't make it apply

it can't hurt can it though ? (though if it's 100% sure it's never any different yes I shouldn't add a noop )

tools/deb/test/run_pilot_test.sh
```diff
@@ -31,21 +31,19 @@ function build_all() {
...
   else
-    #(cd $GOPATH/src/istio.io; git clone https://github.com/istio/pilot)
-    (cd $GOPATH/src/istio.io; git clone https://github.com/costinm/pilot -b deb)
```
ldemailly: this is the main point of this PR, that and line 44's deletion/change

tools/deb/test/run_pilot_test.sh
ldemailly: looks like I must have been dreaming though so will revert those parts

PR #446, opened 2017-08-29
Adding bookinfo to rawvm and rename package

costinm:
Continuing the work on testing rawvm and preparing for demo.


tools/deb/test/build_all.sh
```diff
@@ -0,0 +1,56 @@
...
+# Will checkout mixer, pilot and proxy in the expected locations/
+function build_all() {
+  mkdir -p $GOPATH/src/istio.io
+
+
+  if [[ -d $GOPATH/src/istio.io/pilot ]]; then
+    (cd $GOPATH/src/istio.io/pilot; git pull origin master)
+  else
+    #(cd $GOPATH/src/istio.io; git clone https://github.com/istio/pilot)
+    (cd $GOPATH/src/istio.io; git clone https://github.com/costinm/pilot -b deb)
```
rshriram: ? Your private repos ?

tools/deb/test/build_all.sh
```diff
@@ -0,0 +1,56 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# For e2e tests, build pilot, mixer, istio from head.
```
rshriram: Why should this be in proxy?

tools/deb/test/machine_setup.sh
```diff
@@ -59,6 +73,23 @@ server {
...
+
+# Start bookinfo components
+ruby details/details.rb 9080 &
+echo $! > details.pid
+
+# Note that we run productpage on a different port - 9080 is taken
+(cd productpage ; python productpage.py 9081 &)
+echo $! > productpage.pid
+
+(cd ratings ; node ratings.js 9082 &)
```
rshriram: Where is reviews?

tools/deb/test/run_gce_test.sh
```diff
@@ -304,8 +313,16 @@ function setUp() {
...
+ istioConfigHttpService $TESTVM 80 $LOCAL_IP
+
+ # Bookinfo components on the VM
+ istioConfigService ${TESTVM}-details 9080 $LOCAL_IP
+ istioConfigService ${TESTVM}-productpage 9081 $LOCAL_IP
+ istioConfigService ${TESTVM}-ratings 9082 $LOCAL_IP
+ istioConfigService ${TESTVM}-mysql 3306 $LOCAL_IP
+
+ # Prepare simple test to make sure istio-ingress can reach a http server on the VM
+ istioRoute "/${TESTVM}/" ${TESTVM} 80
```
rshriram: Why do we need this elaborate setup for this simple test? Isn't it enough to run a single nginx service with echo modules to emulate multiple services ? If there is no label based routing test, you just need to test http and tcp.

tools/deb/test/build_all.sh
costinm: That's all we're running in the VM so far - the proxy. 
Eventually we can move them to istio/istio - but I would rather do it after we sanitize
the build system, editing SHAs and uploading/downloading from gcr.io or docker is not the 
cleanest way.

tools/deb/test/build_all.sh
costinm: Ops, sorry.

tools/deb/test/machine_setup.sh
costinm: I would rather leave it in K8S - java and all the other deps.

AFAIK we only want to run one of the components on the VM for the demo, I was not 
sure which, will remove the others.

tools/deb/test/run_gce_test.sh
costinm: Already have the nginx service and echo, for the simple test. 

The script is intended to setup an environment for end-to-end tests, including testing
the bookinfo works on VM. Eventually will run load tests and maybe long-running
stability tests. We basically need to have similar coverage for VM as we have 
for K8S - the setup will be amortized. 

( also it's easier to have a script to setup everything rather than remembering all commands)

tools/deb/test/machine_setup.sh
rshriram: just the mysql. nothing else.

costinm:
Yes, mysql is probably best for a VM and my main goal, but eventually (post
0.2) we want to
test and setup everything on a VM ( with standalone etcd+kubeapiserver for
example).
So any test that we run on K8S is IMHO a target for running on VM.

tools/deb/test/machine_setup.sh
rshriram: Sorry, just realized. We don't want to run the database in a VM, as it doesn't make for a convincing auth/mixer story [DBs have dedicated hardware, access policies embedded in the database engine, and so on]. As a user, I would immediately understand a use case where we run one of the services in the VM (use ratings itself, or perhaps as Kelsey suggested, add a simple Go based binary to the VM - indicating a legacy service). Its relatively easy and straightforward to do this [just ask people to run ratings-v1 in VM and remove the k8s version]. This eliminates all the mongo/mysql trouble as well.

tools/deb/test/run_gce_test.sh
```diff
@@ -235,7 +237,7 @@ spec:
   ports:
```
ldemailly: let's make the switch to istioctl register here
(see the end of https://github.com/istio/istio/blob/rawvm-register-3/samples/rawvm/README.md for an example)

tools/deb/test/run_gce_test.sh
```diff
@@ -29,10 +29,11 @@ export ISTIO_ZONE=${ISTIO_ZONE:-us-west1-c}
 # Name of the k8s cluster running istio control plane. Used to find the service CIDR
```
ldemailly: delete ISTIO_REGION it's not used (line 26)

costinm:
I think that was the use case - DB on dedicated on-prem hardware, user
wants to use it from the k8s cluster as some services are migrated.
I would assume the mysql DB will be the last to move inside k8s (and more
likely will be moved to a managed SQL, not to a pod).

I'm ok with any scenario you want for the demo - my point in the comment
recommending keeping mysql was to also be able to
showcase what I think will be common use case, and for us to test and
evaluate this scenario. In particular the access policies and
TLS side of mysql/postgresql and mongodb are quite interesting, in
particular if the native DB is using mTLS. We can explore this
with a separate app - but bookinfo seems designed to show variety of
languages and technologies.

rshriram:
I too thought using a DB would be nice in a VM, but then the more I think about it, the more strange it sounds. and sorry, I should have caught this earlier and could have avoided all the mongo/mysql drama. Most DBs support TLS for their clients.. And almost all of them have a full fledged fine grained access control at per table/per operation, that their DB admins manage.

Given the fact that these are coveted hardware costing millions of dollars, introducing sidecar agents in the DB cluster adds another point of failure - something that DBAs would not like. [Although, from a pure fault tolerance perspective, it makes no difference to the client/replica, we have to clearly explain this concept to people - which is a pain].

My point is that it would be safer if we said we are emulating a legacy service using a simple binary, instead of showcasing how to install/use istio sidecars in a DB machine. As you said, these are the last pieces to move and typically move to hosted services. People are happy to do just TLS and keep these pieces out of the mesh, while they migrate it to the cloud. 

We would still have support for both mysql & mongo in the bookinfo that show cases integration with a database (whether hosted or on prem). Just that we won't show case running our envoy sidecar on the DB machine, for the VM sample. Make sense?

ldemailly:
as I mentioned I branched off this branch and made changes already to the scripts. I'll deal with merging and PR after the demo

tools/deb/test/istio_common.sh
```diff
@@ -20,6 +20,89 @@
...
+function istioConfigService() {
+  local NAME=${1:-}
+  local PORT=${2:-}
+  local IP=${3:-}
+  local TYPE=${4:-http}
+
+  # The 'name: http' is critical - without it the service is exposed as TCP
+  local ISTIOCTL=$GOPATH/src/istio.io/pilot/bazel-bin/cmd/istioctl/istioctl
+
+  $ISTIOCTL register $NAME $IP $PORT:$TYPE
```
ldemailly: 🎉

PR #450, opened 2017-08-31
Decouple check and report for TCP

### LINKED ISSUE #448: Need to separate check/quota and Report
The mixer filter currently enables or disables both Check and Report via single switch. This is not flexible because in many instances, one would just want to enable mixer for gathering metrics, but cannot tolerate performance overhead of a mixer check call for every service in the chain (or perhaps, they don't want to deal with the fail open/close scenarios).

We need to add a separate mixer_report config in the filter (default on), that would enable reports to be sent. If mixer_report is off, no reports will be sent. mixer_control will be used only for check and quota.

cc @qiwzhang @lizan

rshriram:
Adds a new global option (`disable_tcp_check_calls`) that when set to true will not make check calls to the mixer. By default this option is set to false, i.e., check call will always be made.

Even if check calls are disabled, report calls will still be made as long as the mixer filter is configured. This will enable us to use mixer purely for telemetry alone and then slowly turn on the tcp filter for mixer in services where policy checks are required.

fixes #448


rshriram:
cc @qiwzhang and @lizan .. PTAL

rshriram:
cc @kyessenov . Once this is in, we need to add a global mesh option for this.

qiwzhang:
I have PR https://github.com/istio/proxy/pull/451
It is better than this one.  It allow you to config Check/Report independently at route level

rshriram:
YEs, sorry, I just saw that. I can close this one.

rshriram:
@qiwzhang PTAL. trimmed PR and scoped to TCP only.

lizan:
Shouldn't the config be consistent with #451 ? Have 3 flags like:
```
"mixer_control": "off",
"mixer_check": "on",
"mixer_report": "off",
```

rshriram:
It can, but these are global configs. and so, it would impact http as well. In other words, we would have two mixer_controls, two mixer_reports, etc.

PR #451, opened 2017-08-31
add route flags to disable check and report independently.

qiwzhang:


rshriram:
Oh shoot. I did the same stuff in another pr. Sorry for the duplication

rshriram:
@qiwzhang this is nice.. Can you also add the ability to disable check alone from tcp filter as well ?
I think that would require a global config at mixer filter level.

qiwzhang:
@rshriram OK,  I will use a separate PR for tcp

rshriram:
@qiwzhang I am repurposing my PR to disable only TCP check. And we can use this for HTTP check disabling..

rshriram:
partly fixes #448

PR #452, opened 2017-08-31
Add tests for disable check and report 

qiwzhang:
* Also updated README.md
* Added a flag to disable report batch.

### DIFF
--- a/src/envoy/mixer/config.cc
+++ b/src/envoy/mixer/config.cc
@@ -48,14 +48,6 @@ void ReadString(const Json::Object& json, const std::string& name,
   }
 }
 
-bool ReadBoolean(const Json::Object& json, const std::string& name,
-                 bool defaultValue) {
-  if (json.hasObject(name)) {
-    return json.getBoolean(name, defaultValue);
-  }
-  return defaultValue;
-}
-
 void ReadStringMap(const Json::Object& json, const std::string& name,
                    std::map<std::string, std::string>* map) {
   if (json.hasObject(name)) {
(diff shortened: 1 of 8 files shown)


src/envoy/mixer/config.cc
```diff
@@ -69,9 +78,11 @@ void MixerConfig::Load(const Json::Object& json) {
...
   ReadString(json, kNetworkFailPolicy, &network_fail_policy);
 
-  ReadString(json, kDisableCheckCache, &disable_check_cache);
-  ReadString(json, kDisableQuotaCache, &disable_quota_cache);
-  disable_tcp_check_calls = json.getBoolean(kDisableTcpCheckCalls, false);
+  disable_check_cache = ReadBoolean(json, kDisableCheckCache, false);
+  disable_quota_cache = ReadBoolean(json, kDisableQuotaCache, false);
+  disable_report_batch = ReadBoolean(json, kDisableReportBatch, false);
+
+  disable_tcp_check_calls = ReadBoolean(json, kDisableTcpCheckCalls, false);
```
rshriram: nit: we don't need the ReadBoolean function because the json.getBoolean(name, default) automatically returns default if name is not present.

src/envoy/mixer/integration_test/check_report_disable_test.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	// wait for 2 second to wait for envoy to come up
+	time.Sleep(2 * time.Second)
+	tag = "Check Only"
+	if _, _, err := HTTPGet(url); err != nil {
+		t.Errorf("Failed in request %s: %v", tag, err)
+	}
+	time.Sleep(1 * time.Second)
+	// Only check send, Not report.
```
rshriram: typo: only send check, not report.

src/envoy/mixer/config.cc
qiwzhang: Done

src/envoy/mixer/integration_test/check_report_disable_test.go
qiwzhang: Done

PR #454, opened 2017-09-06
jwt-auth: add an interface to create a public key object 

mtakigiku:
added `Pubkeys::CreateFrom()`

### DIFF
--- a/src/envoy/auth/jwt.cc
+++ b/src/envoy/auth/jwt.cc
@@ -15,6 +15,7 @@
 
 #include "jwt.h"
 
+#include "common/common/assert.h"
 #include "common/common/base64.h"
 #include "common/common/utility.h"
 #include "common/json/json_loader.h"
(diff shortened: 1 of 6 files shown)


src/envoy/auth/jwt.cc
```diff
@@ -433,6 +433,18 @@ std::unique_ptr<Pubkeys> Pubkeys::CreateFromJwks(const std::string &pkey_jwks) {
...
 
+std::unique_ptr<Pubkeys> Pubkeys::CreateFrom(const std::string &pkey,
+                                             Type type) {
+  switch (type) {
+    case Type::JWKS:
+      return CreateFromJwks(pkey);
+    case Type::PEM:
+      return CreateFromPem(pkey);
+    default:
+      abort();
```
lizan: use PANIC macro and indicate the reason.

src/envoy/auth/jwt.h
```diff
@@ -205,9 +205,14 @@ class JwtVerifier : public WithStatus {
 //   if(keys->GetStatus() == Status::OK) { ... }
 class Pubkeys : public WithStatus {
  public:
+  // Format of public key.
+  enum Type { PEM, JWKS };
+
   Pubkeys(){};
   static std::unique_ptr<Pubkeys> CreateFromPem(const std::string& pkey_pem);
```
lizan: make these two private?

mtakigiku: done. Removed these two rather than making them private.

src/envoy/auth/jwt.cc
mtakigiku: done

PR #455, opened 2017-09-06
jwt-auth: JWT lib interface: split verification function into another class

mtakigiku:
- split `class JwtVerifier` into `class Jwt` and `class Verifier`.
- moved "creating `EVP_MD` object" part from the verification part to JWT parsing part.


src/envoy/auth/jwt.h
```diff
@@ -200,6 +211,8 @@ class JwtVerifier : public WithStatus {
   std::string iss_;
   std::string aud_;
   int64_t exp_;
+
+  friend bool Verifier::Verify(const Jwt& jwt, const Pubkeys& pubkeys);
```
qiwzhang: try not to use friend function.

mtakigiku: added TODO comments for this

PR #456, opened 2017-09-06
jwt-auth: make audiences configurable for each issuer

mtakigiku:


src/envoy/auth/config.cc
```diff
@@ -158,6 +167,18 @@ IssuerInfo::IssuerInfo(Json::Object *json) {
   failed_ = true;
 }
 
+bool IssuerInfo::IsValidAudience(const std::string &aud) {
+  if (audiences_.empty()) {
+    return true;
+  }
+  for (auto &a : audiences_) {
```
lizan: variable name `a` is not good, name it `audience`?

src/envoy/auth/config.cc
```diff
@@ -158,6 +167,18 @@ IssuerInfo::IssuerInfo(Json::Object *json) {
   failed_ = true;
 }
 
+bool IssuerInfo::IsValidAudience(const std::string &aud) {
```
lizan: `IsAudienceAllowed`

src/envoy/auth/config.cc
mtakigiku: resolved by using std::find()

PR #458, opened 2017-09-06
jwt-auth: changed additional header name

mtakigiku:


src/envoy/auth/http_filter.cc
```diff
@@ -28,7 +28,7 @@ namespace Envoy {
 namespace Http {
 
 const LowerCaseString& JwtVerificationFilter::AuthorizedHeaderKey() {
-  static LowerCaseString* key = new LowerCaseString("Istio-Auth-UserInfo");
+  static LowerCaseString* key = new LowerCaseString("sec-Istio-Auth-UserInfo");
```
qiwzhang: can we do all lower cases?

PR #459, opened 2017-09-06
Change docker image name to enable auto dependency update

chxchx:
This change names future docker images using the full sha of the head commit, which is consistent with artifacts from mixer, pilot, etc. It is also required to implement the auto dependency update on this repo.


chxchx:
@costinm @kyessenov PTAL

PR #461, opened 2017-09-07
jwt-auth: add feature to update public key regularly with mutex

mtakigiku:


src/envoy/auth/config.cc
```diff
@@ -122,25 +164,27 @@ IssuerInfo::IssuerInfo(Json::Object *json) {
   // Check "value"
   std::string value = json_pubkey->getString("value", "");
   if (value != "") {
+    pkey_ = std::unique_ptr<Pubkey>(new Pubkey());
```
lizan: `.reset`

src/envoy/auth/config.cc
```diff
@@ -122,25 +164,27 @@ IssuerInfo::IssuerInfo(Json::Object *json) {
...
     // Public key is written in this JSON.
     if (pkey_type_ == "pem") {
-      pkey_ = Pubkeys::CreateFromPem(value);
+      pkey_->Update(Pubkeys::CreateFromPem(value).release());
```
lizan: You should be able to pass `unique_ptr` directly and use `std::move` inside.

src/envoy/auth/http_filter.cc
```diff
@@ -53,34 +53,29 @@ FilterHeadersStatus JwtVerificationFilter::decodeHeaders(HeaderMap& headers,
...
-      }
-      calling_issuers_[iss->name_].second =
-          std::unique_ptr<Auth::AsyncClientCallbacks>(
-              new Auth::AsyncClientCallbacks(
-                  config_->cm_, iss->cluster_,
-                  [&](bool succeed, const std::string& pubkey) -> void {
-                    this->ReceivePubkey(headers, iss->name_, succeed, pubkey);
-                  }));
-      calling_issuers_[iss->name_].second->Call(iss->uri_);
+    if (!iss->pkey_->IsNotExpired()) {
```
lizan: This will be easily deadlocked when request A comes and the filter issued requests to public key server, then another request with same JWT comes and reached here, the thread is locked and never being unblocked since you won't be able to process the response from public key server.

src/envoy/auth/config.cc

src/envoy/auth/config.cc

src/envoy/auth/config.cc
```diff
@@ -90,7 +91,48 @@ void AsyncClientCallbacks::Call(const std::string &uri) {
...
 
-IssuerInfo::IssuerInfo(Json::Object *json) {
+IssuerInfo::Pubkey::Pubkey() : update_needed_(false) {}
+
+IssuerInfo::Pubkey::Pubkey(std::chrono::duration<int> valid_period)
+    : valid_period_(valid_period), update_needed_(true) {}
+
+bool IssuerInfo::Pubkey::IsNotExpired() {
+  if (update_needed_) {
+    mutex_pkey_.lock();
```
qiwzhang: this is bad.  Second call to IsNotExpired() will be blocked.

mtakigiku:
filed as an issue for now: https://github.com/istio/proxy/issues/468

lizan:
I'm taking over this.

PR #462, opened 2017-09-07
jwt-auth: check exp claim

mtakigiku:
- added code to check "exp" of JWT
- added integration tests for expired/nonexpired JWTs

### DIFF
--- a/src/envoy/auth/http_filter.cc
+++ b/src/envoy/auth/http_filter.cc
@@ -23,7 +23,6 @@
 #include "server/config/network/http_connection_manager.h"
 
 #include <chrono>
-#include <ctime>
 #include <string>
 
 namespace Envoy {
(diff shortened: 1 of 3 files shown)


src/envoy/auth/http_filter.cc
```diff
@@ -154,9 +156,12 @@ std::string JwtVerificationFilter::Verify(HeaderMap& headers) {
     // Invalid JWT
     return Auth::StatusToString(jwt.GetStatus());
   }
-  /*
-   * TODO: check exp claim
-   */
+
+  // Check "exp" claim.
+  auto unix_timestamp = std::chrono::seconds(std::time(NULL)).count();
```
lizan: Use `system_clock` and http://en.cppreference.com/w/cpp/chrono/time_point/time_since_epoch

PR #463, opened 2017-09-07
Remove envoy binary before bazel build.

qiwzhang:


kyessenov:
@costinm what's the point of iptables script in the docker image? This is a base layer for the binary alone.

PR #464, opened 2017-09-08
rename attribute name target to destination

qiwzhang:
In the integration test (go test),  some target attributes remain since they are used in cache keys.  It is easier to use global dictionary in ReferencedAttributes for cache keys.  This is just for integration test only.


mandarjog:
I will be ready tomorrow.

PR #467, opened 2017-09-08
jwt-auth: return failure reason as message body

mtakigiku:


src/envoy/auth/http_filter.cc
```diff
@@ -201,11 +201,9 @@ void JwtVerificationFilter::CompleteVerification(HeaderMap& headers) {
...
   if (status != "OK") {
     // verification failed
-    /*
-     * TODO: detailed information on message body
-     */
     Code code = Code(401);  // Unauthorized
-    std::string message_body = "Verification Failed";
+    // return failure reason as message body
+    std::string message_body = status;
     Utility::sendLocalReply(*decoder_callbacks_, false, code, message_body);
```
qiwzhang: just pass "status" to the function.

PR #470, opened 2017-09-08
Shrink and better document the debug image

### LINKED ISSUE #423: cleanup proxy_debug image
debug is good, kitchen sink is bad
now that we have `sudo apt-get install` working we don't need to install everyone's random editor and other huge footprint packages

ldemailly:
before: 722MB
after: 236MB
(Not counting envoy itself)

Fixes #423

Needs https://github.com/istio/pilot/pull/1202 to fully work

```release-note
Reduce proxy debug image by 2/3rd (486Mb saved) yet as functional
```

### DIFF
--- a/docker/Dockerfile.debug
+++ b/docker/Dockerfile.debug
@@ -1,22 +1,26 @@
 FROM ubuntu:xenial
+# Do not add more stuff to this list that isn't small or critically useful.
+# If you occasionally need something on the container do
+# sudo /usr/local/bin/proxy-redirection-clear
+# sudo apt-get update && apt-get whichever
+# sudo /usr/local/bin/proxy-redirection-restore
 RUN apt-get update && \
-    apt-get install -y \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.debug
```diff
@@ -1,22 +1,26 @@
...
+      curl \
+      iptables \
+      iproute2 \
+      iputils-ping \
+      dnsutils \
+      netcat \
+      tcpdump \
+      net-tools \
+      libc6-dbg gdb \
+      elvis-tiny \
```
ldemailly: this is like vim but 94k instead of megabytes

and anyone can sudo apt-get their favorite editor if they really have to edit things in the container and don't know `ESC:wq` for save and quit and `i` to insert text

ldemailly:
tcpdump is there
I left busybox which is a weird server executable but worse than the base image just in case someone wants the server side, eventually I'll put fortio there but not now

PR #471, opened 2017-09-08
Remove entrypoint

costinm:
May create confusion in pilot docker images.

(the long term plan is to generate docker images and deb files in istio-build, and migrate to the 
new iptable script and consistent images)


kyessenov:
ENTRYPOINY id not required for docker files. We should sort out the layering after the release...

costinm:
Laurent is right, we had entrypoint before (envoy) - so not clear what broke but it's unlikely this change is causing the problem.

ldemailly:
@kyessenov can you file an issue if you found a bug with istio-start.sh
it's very possible there is one, but I'd like to see the details and then decide how to fix it
(I got the images updated yesterday so the tests would have a chance to pass or find issues, as far as I know there are no issues but again if you have information, please share)

PR #472, opened 2017-09-11
Rename the debian package, fix the iptables defaults based on env

costinm:
Also updates the machie setup for test with mysql.

This is a subset of the previous PR, with only the critical parts to get the rename into the builds.

### DIFF
--- a/tools/deb/istio-iptables.sh
+++ b/tools/deb/istio-iptables.sh
@@ -31,14 +31,16 @@
 # the sidecar image using the .deb file created by proxy.
 
 function usage() {
-  echo "${0} -p PORT -u UID [-h]"
+  echo "${0} [-p PORT -u UID -i] [clean]"
   echo ''
+  echo ' K8s proxy init mode: if -p or -u options are set, the script will be backward compatible with the pod initializer mode.'
   echo '  -p: Specify the envoy port to which redirect all TCP traffic (default $ENVOY_PORT = 150001)'
   echo '  -u: Specify the UID of the user for which the redirection is not'
   echo '      applied. Typically, this is the UID of the proxy container (default to uid of $ENVOY_USER, uid of istio_proxy, or 1337)'
   echo '  -i: Comma separated list of IP ranges in CIDR form to redirect to envoy (optional)'
   echo ''
-  echo 'Using environment variables in $ISTIO_SIDECAR_CONFIG (default: /var/lib/istio/envoy/sidecar.env)'
+  echo ' If called without arguments, will apply settings from config file - in $ISTIO_SIDECAR_CONFIG and $ISTIO_CLUSTER_CONFIG'
+  echo ' Only defined inbound ports and defined outbound ranges will be intercepted.'
 }
 
 set -o nounset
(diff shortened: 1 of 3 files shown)


tools/deb/test/machine_setup.sh
```diff
@@ -17,48 +17,163 @@
...
 EOF
 
-# Start istio
-systemctl start istio
+
+  systemctl restart nginx
+}
+
+function istioMysql() {
+  # The ratings script is using root/password by default.
```
rshriram: Is this for testing? If so, won't it be easier to just install the docker container and port forward? You won't need any of this. You could probably even use mongo container if you want to and take advantage of continuously tested demo, and simplify future integration tests

costinm: It may be easier - but not right :-), the goal is to test a real VM. We do have mongodb in docker in 
the cluster. I kept mysql because eventually I want to enable TLS on both mongo and myslq and test it works with the 'bypass auth' or 'custom certs' when they're ready.


I agree it would be good to test the interaction of iptables and envoy running in the raw VM with a docker with forwarded ports, but I think it's lower priority.

ldemailly:
move this pr directly to istio/istio ?

costinm:
Not so easy without the flat build. Bazel still need the files built
locally, can't cross projects.

ldemailly:
I mean machine_setup.sh

tools/deb/istio-iptables.sh
```diff
@@ -143,6 +150,13 @@ if [ -n "${ISTIO_INBOUND_PORTS:-}" ]; then
           iptables -t nat -A ISTIO_INBOUND -p tcp --dport ${port} -j ISTIO_REDIRECT
       done
   fi
+else
+  # ISTIO_INBOUND_PORTS not set, but script called in k8s backward compatibility mode via args
+  if [ "${K8S_MODE}" == "1" ]; then
+    iptables -t nat -N ISTIO_INBOUND
+    iptables -t nat -A PREROUTING -p tcp -j ISTIO_INBOUND
+    iptables -t nat -A ISTIO_INBOUND -p tcp -j ISTIO_REDIRECT
+  fi
```
ldemailly: else echo warning something is off?

tools/deb/istio-iptables.sh
```diff
@@ -176,6 +191,9 @@ if [ -n "${IP_RANGES_INCLUDE:-}" ]; then
     done
     iptables -t nat -A ISTIO_OUTPUT -j RETURN
 else
-    iptables -t nat -A ISTIO_OUTPUT -j ISTIO_REDIRECT
+    if [ "${K8S_MODE}" == "1" ]; then
+       iptables -t nat -A ISTIO_OUTPUT -j ISTIO_REDIRECT
+    fi
+    # Else: don't intercept anything, to avoid breaking the VM in case Istio is not configured properly
```
ldemailly: echo ...

tools/deb/istio-iptables.sh
costinm: Nothing off - if running in K8S_MODE (i.e. in 0.3, when we use the new script in both
k8s and raw vm), the default will remain unchanged in k8s, for backward compat.

In rawVM we want the default to be to intercept only what is explicitly configured since VMs are more likely to run other services. But there is nothing to warn or wrong.

tools/deb/istio-iptables.sh
costinm: See above - the purpose of this change is to be more conservative on the raw vm, while
keeping the behavior in k8s.

costinm:
Closing it for 0.2 - can wait for next dot release.

PR #473, opened 2017-09-12
Update mixerclient with latest

sebastienvas:


sebastienvas:
jenkins test proxy/pilot-regression

PR #485, opened 2017-09-23
Proper install for cloud builder

sebastienvas:
* Separate release from cloud builder setup to ease reproduction and testing
* Increase to more robust machine
* Use specific bazel.rc for cloudbuilder and use recommanded batch mode for container


ldemailly:
script/push-debian.sh: line 38: ISTIO_VERSION: unbound variable

Issue #483, opened 2017-09-21
Allow Pilot to pass any type of mesh attributes to Mixer


qiwzhang:
Pilot generates Mesh attributes in Envoy Mixer filter config and Route Opaque data.  Now it only supports string type.   Other attribute types should be supported, such as BYTES for IP,  int for port.

Please see https://github.com/istio/mixerclient/issues/112

douglas-reid:
@qiwzhang is there any way we can fix just the `destination.ip` attribute to use BYTES for 0.2.x?

cc: @geeknoid 

having `source.ip` sent as BYTES and `destination.ip` sent as STRING while both being IP_ADDRESS is not good. It makes writing config that refers to IP_ADDRESS attributes a near-impossibility and/or incredibly ugly (example: `match: destination.ip == "10.1.1.9" && source.ip == ip("10.10.10.1")`. It could also lead to some surprising results in metrics, depending on serialization etc.

qiwzhang:
For 0.2, it has to be a hack:  any attribute names from Pilot with suffix as ".ip",  proxy needs to convert to bytes.   Then, we need similar hack for "port" which should be integer, but now it is string.

I think we like to do it properly.  To define a protobuf for Pilot to pass any type of attributes to Proxy and to Mixer

douglas-reid:
Without the hack, we are going to end up either:
- saying no Mixer policy can be written using IP_ADDRESS (and `port`s)
- or, having policies written in an inconsistent fashion -- which will break as soon as we fix the issue.

mandarjog:
I think the hack is necessary for 0.2, without the hack `match: destination.ip == "10.1.1.9" && source.ip == ip("10.10.10.1")` this inconsistent expression is used in our configuration.
It is a low risk hack with just a few lines of code, we should go for it.

The alternative is for Proxy to consume the `attribute-manifest` proto that is already defined in Mixer. That is the authoritative mapping between an attribute name and its type.

geeknoid:
We definitely should do the hack for 0.2.

In the long run, the proxy should rely on the manifest data for attribute
typing contraints.

kyessenov:
This isn't just a proxy issue. Pilot sets the value. The problem is that we can't communicate the right type from pilot to proxy.

qiwzhang:
For 0.2,  Proxy will only convert suffix ".ip" to bytes as a hack.  Hi @kyessenov , could you check if there are other types Proxy should convert, such as  ".port"?
[...]

PR #496, opened 2017-09-25
Convert Pilot mesh ip attributes to BYTES

qiwzhang:


qiwzhang:
https://github.com/istio/proxy/issues/483

src/envoy/mixer/mixer_control.cc
```diff
@@ -382,6 +388,32 @@ void MixerControl::BuildTcpReport(
...
 }
 
+// Mesh attributes from Pilot are all string type.
+// The attributes with ".ip" suffix will be treated
+// as ipv4 and use BYTES attribute type.
+void MixerControl::SetMeshAttribute(const std::string& name,
+                                    const std::string& value,
+                                    Attributes* attr) const {
+  // Check with ".ip" suffix,
+  if (name.length() < kIPSuffix.length() ||
```
mangchiandjjoe: "name" can be ".ip"? otherwise '<=' seems to be better,
Does name.compare need to be case insensitive?

qiwzhang: Thanks.  ".ip" is not allowed,  code is changed.
All istio attribute names are lower cases.  
https://istio.io/docs/reference/config/mixer/attribute-vocabulary.html

costinm:
Wait - some context ? This seems like an API change - is this based on a bug, and was it approved (we're locking down the release).

qiwzhang:
@costinm This hack is needed for 0.2. Here is the detail. https://github.com/istio/proxy/issues/483

src/envoy/mixer/mixer_control.cc
```diff
@@ -382,6 +388,32 @@ void MixerControl::BuildTcpReport(
...
+                                    Attributes* attr) const {
+  // Check with ".ip" suffix,
+  if (name.length() <= kIPSuffix.length() ||
+      name.compare(name.length() - kIPSuffix.length(), kIPSuffix.length(),
+                   kIPSuffix) != 0) {
+    attr->attributes[name] = Attributes::StringValue(value);
+    return;
+  }
+
+  in_addr ipv4_bytes;
```
costinm: Please add a block for ipv6.

The sidecar will run on raw VMs, some on prem using ipv6 - we can't add new code that doesn't support ipv6.

qiwzhang: https://github.com/istio/proxy/pull/503

PR #518, opened 2017-09-29
Remove mixer deps from proxy for automated deps update

sebastienvas:
This is a testing dependency. It currently generates an update to pilot for each update in mixer which should not happen.


qiwzhang:
Then,  each time if istio/api changes global dictionary,  proxy integration test will break.  Someone has to manually update Mixer SHA.

kyessenov:
@qiwzhang @geeknoid isn't this an actual problem that mixer and proxy must agree on precise list of global words and not negotiate the minimum?

kyessenov:
I LGTM, since we need to address the problem behind this, and not rely on test-infra.

PR #595, opened 2017-10-19
Add quota config parser

qiwzhang:


src/envoy/mixer/quota_config.cc
```diff
@@ -0,0 +1,92 @@
...
+    const std::string& value = it->second.str_v;
+
+    switch (match.match_type_case()) {
+      case ::istio::proxy::v1::config::StringMatch::kExact:
+        if (value != match.exact()) {
+          return false;
+        }
+        break;
+      case ::istio::proxy::v1::config::StringMatch::kPrefix:
+        if (value.length() < match.prefix().length() ||
```
lizan: nit: use `StringUtil::startsWith`?

src/envoy/mixer/quota_config.h
```diff
@@ -0,0 +1,54 @@
...
+// attributes to generate required quotas.
+class QuotaConfig {
+ public:
+  QuotaConfig(const ::istio::mixer::v1::config::client::QuotaSpec& spec_pb);
+
+  struct Quota {
+    std::string quota;
+    int64_t charge;
+
+    // For unit-test.
```
lizan: If this is for unit test only, just add a matcher for gtest.

src/envoy/mixer/quota_config.cc
qiwzhang: STL doesn't have StringUtil::startsWith

src/envoy/mixer/quota_config.h
qiwzhang: It is used in 
ASSERT_EQ for std::vector<Quota>
gTest matcher will not help.

qiwzhang: std::vector of Quota

src/envoy/mixer/quota_config_test.cc
```diff
@@ -0,0 +1,123 @@
...
+)";
+
+using QuotaVector = std::vector<QuotaConfig::Quota>;
+
+class QuotaConfigTest : public ::testing::Test {
+ public:
+  void SetUp() {}
+};
+
+TEST_F(QuotaConfigTest, TestNotMatch) {
```
ayj: nit: `TestNotMatch` suggests this spec doesn't match anything. Maybe rename `NotMatch` to `EmptyMatch` here and line 30 to make the intent clearer?

src/envoy/mixer/quota_config.cc
lizan: Envoy has...

qiwzhang: Try not to use Envoy code in this module.  May move it inside MixerClient library.

src/envoy/mixer/quota_config_test.cc
qiwzhang: Done

src/envoy/mixer/quota_config.cc
```diff
@@ -0,0 +1,92 @@
...
+            value.compare(0, match.prefix().length(), match.prefix()) != 0) {
+          return false;
+        }
+        break;
+      case ::istio::proxy::v1::config::StringMatch::kRegex:
+        // TODO: support regex
+        return false;
+        break;
+      default:
+        break;
```
lizan: Let's not allow this fall through and return `true`?

src/envoy/mixer/quota_config.cc
lizan: ok fine.

src/envoy/mixer/quota_config.cc
ayj: Should an `unknown` match type result in a valid match? Seems like the default behavior should be to ignore that case.

qiwzhang: @ayj I agreed with.  We should ignore "Unknown" type

lizan: No, there shouldn't be a unknown clause (it should be removed) and it shouldn't go through. You treated "unimplemented" regex as "no match" above, so the `unknown` should be same. Also leave a comment here.

qiwzhang: OK, I changed "unimplemented" as ignored.  so regex and not_set case are all ignored.

ayj: + 1 for comment explanation rationale. 

Really, this type of error should be caught earlier by validation before proxy receives the config. 

Shouldn't adding unoptimized regex support be fairly straightforward? e.g.

```c++
if (!std::regex_match(value, match.regex())) {
    return false
}
```

PR #597, opened 2017-10-20
Cherry-pick Piotr's critical fix for TCP proxy

costinm:
Needed for 0.2.10, without it TCP proxy will be broken for mysql and gRPC.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -37,12 +37,14 @@ bind(
     actual = "//external:ssl",
 )
 
-ENVOY_SHA = "b02d8a9b5033a61a7b13d98999f5f00c4b85b6a4"  # Sep 26, 2017 (use github to download tclap instead of sourceforge)
 
-http_archive(
+ENVOY_SHA = "577410987eddf1d8f47439475c8569dcd68eeb04"  # Cherry-pick Oct 20
+# Branched from "b02d8a9b5033a61a7b13d98999f5f00c4b85b6a4"  # Sep 26, 2017 (use github to download tclap instead of sourceforge)
+
+git_repository(
     name = "envoy",
-    strip_prefix = "envoy-" + ENVOY_SHA,
-    url = "https://github.com/envoyproxy/envoy/archive/" + ENVOY_SHA + ".zip",
+    commit = ENVOY_SHA,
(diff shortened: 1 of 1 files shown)


WORKSPACE
```diff
@@ -37,12 +37,14 @@ bind(
     actual = "//external:ssl",
 )
 
-ENVOY_SHA = "b02d8a9b5033a61a7b13d98999f5f00c4b85b6a4"  # Sep 26, 2017 (use github to download tclap instead of sourceforge)
 
-http_archive(
+ENVOY_SHA = "577410987eddf1d8f47439475c8569dcd68eeb04"  # Cherry-pick Oct 20
+# Branched from "b02d8a9b5033a61a7b13d98999f5f00c4b85b6a4"  # Sep 26, 2017 (use github to download tclap instead of sourceforge)
+
+git_repository(
```
lizan: Use `http_archive`, this is an intended workaround for Cloud Container Builder.

costinm: This is using istio/envoy fork - is the archive auto-generated ? And what is the issue with GCB ?

lizan: Yes same URL format works.

https://github.com/istio/envoy/archive/577410987eddf1d8f47439475c8569dcd68eeb04.zip

The root cause: https://github.com/bazelbuild/bazel/issues/3786

costinm:
Good news is that we have a plan to fix it, by having repo pool the deps,
so bazel bug won't affect us in future.

costinm:
Fixed the workspace to use http_archive, Seb is trying to find something for the jenkins image that is out of date ( or we won't be able to make more releases from the branch..)

lizan:
It's a JGit bug actually, but anyway http_archive gives you better performance on CI.

presubmit fails because it runs bazel 0.6.1 and this branch has to run with bazel 0.5.x.

rshriram:
Why are we cherrypicking envoy commits? We might as well fork envoy for Istio and totally customize it for our needs.

sebastienvas:
This is a the release-0.2 branch.

costinm:
Jenkins image was updated to build master, causing presubmit failures. The cherry-picked PR has passed the tests on master.

PR #607, opened 2017-10-26
Use mixer client of using protobuf Attributes.

qiwzhang:
Also removed string_map.proto. replace it with mixer::v1::Attributes_StingMap.  It is used to serialize a string map into a string to be passed in the http header.


src/envoy/mixer/repositories.bzl
```diff
@@ -18,10 +18,9 @@
 MIXER_CLIENT = "847a67bd905605f87b3e9a559fca905434fb50c2"
 
 def mixer_client_repositories(bind=True):
```
lizan: ?

qiwzhang: This will be changed after Mixerclient change is submitted.

PR #618, opened 2017-11-08
Integrate with new mixer control library.

qiwzhang:
Code is not ready for submit (Pending on mixerclient PR https://github.com/istio/mixerclient/pull/248  to be submitted).  
But the code is ready for review.


qiwzhang:
@ayj @lizan This PR is ready.  PTAL

src/envoy/mixer/mixer_control.cc
```diff
@@ -220,199 +35,50 @@ class EnvoyTimer : public ::istio::mixer_client::Timer {
   Event::TimerPtr timer_;
 };
 
-}  // namespace
+// Create all environment functions.
```
ayj: It doesn't have to be addressed in this PR, but can we reconcile use of `platform` and `environment` between istio/proxy and istio/mixerclient?

src/envoy/mixer/config.cc
```diff
@@ -64,31 +59,54 @@ void ReadStringMap(const Json::Object& json, const std::string& name,
...
+      json.getBoolean(kDisableTcpCheckCalls, false));
 }
 
-void MixerConfig::ExtractQuotaAttributes(Attributes* attr) const {
-  if (!quota_name.empty()) {
-    int64_t amount = 1;  // default amount to 1.
-    if (!quota_amount.empty()) {
-      amount = std::stoi(quota_amount);
-    }
+void MixerConfig::CreateLegacyConfig(
```
ayj: s/CreateLegacyConfig/CreateLegacyRouteConfig? Otherwise I was expecting this to refer to the legach top-level config.

src/envoy/mixer/config.cc
```diff
@@ -64,31 +59,54 @@ void ReadStringMap(const Json::Object& json, const std::string& name,
 }  // namespace
 
 void MixerConfig::Load(const Json::Object& json) {
-  ReadStringMap(json, kMixerAttributes, &mixer_attributes);
-  ReadStringMap(json, kForwardAttributes, &forward_attributes);
+  ReadStringMap(json, kMixerAttributes, http_config.mutable_mixer_attributes());
```
ayj: I'm assuming the goal is for this PR to preserve existing behavior and v2 config will be added in a follow-up PR?

src/envoy/mixer/mixer_control.cc
qiwzhang: Mixerclient is using Environment,  I will try to replace all "platform" with environment

src/envoy/mixer/config.cc
qiwzhang: Will do

src/envoy/mixer/config.cc
qiwzhang: Yes.  that is the plan.

src/envoy/mixer/mixer_control.cc
qiwzhang: Hmm,  "platform" is not used.

ayj: See https://github.com/istio/mixerclient/pull/248/files#diff-52908b68b42c30e44f87d00288171c3fR50

qiwzhang: I missed that one.  Will change it in MixerClient

PR #620, opened 2017-11-15
Use new istio/api with cc_proto build.

qiwzhang:


sebastienvas:
build proxy/presubmit

PR #623, opened 2017-11-16
Add conffiles to avoid overriding on upgrade

costinm:
On upgrade sidecar.env will be preserved.

### DIFF
--- a/tools/deb/conffiles
+++ b/tools/deb/conffiles
@@ -1,0 +1,3 @@
+/var/lib/istio/envoy/xsidecar.env
+/var/lib/istio/envoy/cluster.env
+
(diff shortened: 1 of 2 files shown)


tools/deb/conffiles
```diff
@@ -0,0 +1,3 @@
+/var/lib/istio/envoy/xsidecar.env
```
ldemailly: extra x ?

costinm: Sorry, used it for debugging (to not override my previous config). 

No, it doesn't offer to merge - asks user if he wants to override. 
An option would be to provide a sidecar.env.sample or none at all - the provisioning 
tool will create it. 

But one step at a time - right now the critical problem is that sidecar.env is replaced on
upgrade losing any config, and that's the standard .deb solution.

PR #660, opened 2017-11-22
Allow Cloud Builder to be used instead of Jenkins

mattdelco:
Jenkins was apparently killed off without a replacement.  This change allows Cloud Builder to be used instead.  It also looks like the version was changed from 0.2 to 0.3 without updating the scripts.


script/release-binary
```diff
@@ -48,7 +48,7 @@ gsutil cp "${BINARY_NAME}" "${SHA256_NAME}" "${DST}/"
 BINARY_NAME="istio-proxy-${SHA}.deb"
 SHA256_NAME="istio-proxy-${SHA}.sha256"
 bazel build --config=release //tools/deb:istio-proxy
-BAZEL_TARGET="bazel-bin/tools/deb/istio-proxy_0.2-dev_amd64.deb"
+BAZEL_TARGET="bazel-bin/tools/deb/istio-proxy_0.3-dev_amd64.deb"
```
hklai: So these versions need to change every month?

mattdelco: Possibility, though I'm not clear on the options to not include the version in the filename.  In proxy's tools/deb/BUILD the debug one apparently gets its version from a hard-coded:
 "version = "0.3-dev"
the release one apparently uses:
version_file = "version"
thought I can't figure out where the file comes from (there's a place in push-debian.sh that tries to create the file based on what's provided via the -v parameter), but I'm not sure this script is involved in this particular build (and regardless I didn't track down where the version originates from -- I'm certainly not specifying a version anywhere (and it it's coming from the istio repo's istio.RELEASE file then this use of the file should probably be discontinued).

mattdelco: The BAZEL_TARGET only seems to be used once with "cp" so another option is to replace the "0.2" or "0.3" with a "*", though that'll break if multiple files of different versions are present (which could only happen in a dev tree who did an incremental re-build after a version bump).

Issue #687, opened 2017-11-30
Update integration_test to use moved mixer mock test


qiwzhang:
It is moved from mixer/test/server.go → mixer/pkg/mockapi/server.go

By this PR

https://github.com/istio/istio/pull/1843

PR #690, opened 2017-12-01
Use latest mixer mockapi package.

qiwzhang:
```release-note
NONE
```


qiwzhang:
https://github.com/istio/proxy/issues/687

Issue #693, opened 2017-12-01
Envoy crashes 


qiwzhang:
Run into Envoy assertion here  at

https://github.com/envoyproxy/envoy/blob/master/source/common/http/conn_manager_impl.cc#L1030

The log is

https://storage.googleapis.com/istio-prow/pull/istio_proxy/691/proxy-presubmit/34/build-log.txt

qiwzhang:
The bug was introduced by https://github.com/istio/proxy/pull/680  at 76ed00adea006e25878f20daa58c456848243999

And fixed by
Fix Envoy crash with v2 config and post. #697

PR #697, opened 2017-12-01
Fix Envoy crash with v2 config and post.

qiwzhang:
HTTP mixer filter::decodeData return CONTINUE,  while Check is still in progress.  This bug was introduced a couple days ago.   It will happen with v2 config and POST requests.
Envoy hits assertion on the "stopped" flag.

```release-note
NONE
```

### DIFF
--- a/src/envoy/mixer/http_filter.cc
+++ b/src/envoy/mixer/http_filter.cc
@@ -265,14 +265,11 @@ class Instance : public Http::StreamDecoderFilter,
 
   bool initiating_call_;
 
-  bool mixer_check_disabled_;
-  bool mixer_report_disabled_;
-
   // check mixer on/off flags in route opaque data
-  void check_mixer_route_flags() {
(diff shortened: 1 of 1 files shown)


qiwzhang:
https://github.com/istio/proxy/issues/693

src/envoy/mixer/http_filter.cc
```diff
@@ -265,33 +265,30 @@ class Instance : public Http::StreamDecoderFilter,
 
   bool initiating_call_;
 
-  bool mixer_check_disabled_;
-  bool mixer_report_disabled_;
-
   // check mixer on/off flags in route opaque data
-  void check_mixer_route_flags() {
+  void check_mixer_route_flags(bool* check_disabled, bool* report_disabled) {
```
kyessenov: is this a good style to pass bool pointers? Can this be refactored to "set_mixer_route_flags(flags)"?

PR #702, opened 2017-12-05
Add circleci automation for proxy

costinm:
Initial integration - will need to be extended to upload artifacts, currently only builds the
deb file.

Change also includes a drone config file, using the same image as circleci. 
Drone is easy to debug and run locally ( it's an open source project, very similar with circleci ),
we don't plan to use it for official builds but I've been using it for some time.


costinm:
Example build:
https://circleci.com/gh/costinm/proxy/36#tests/containers/0

costinm:
Full build + test (on std 2CPU/4G): 15 min
Next build + test, with bazel cache saved: 2 min !

.circleci/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM    circleci/openjdk:latest
+
+# Based on https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/bazel/Dockerfile
+RUN \
+    sudo sh -c 'echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" > /etc/apt/sources.list.d/bazel.list '
+
+RUN \
+    curl https://storage.googleapis.com/bazel-apt/doc/apt-key.pub.gpg | sudo apt-key add -
+
+RUN sudo apt-get update && \
```
sebastienvas: I think you want to stick to a version as we don't always want to use the latest.

.circleci/Makefile
```diff
@@ -0,0 +1,21 @@
+#PROJECT ?= istio-testing
+VERSION ?= 0.4.4
+#HUB ?= gcr.io/
+
+PROJECT ?= costinm
```
sebastienvas: That can't be the right default

.cloudbuild/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM   gcr.io/cloud-builders/bazel
```
sebastienvas: Same here, you cannot just use the last version. At least not until bazel is GA.

Makefile
```diff
@@ -24,17 +26,28 @@ TAG ?=
...
+
 check:
 	@script/check-license-headers
 	@script/check-style
 
 artifacts: build
 	@script/push-debian.sh -c opt -p $(ARTIFACTS_DIR)
 
+deb:
+	bazel build tools/deb:istio-proxy  ${BAZEL_BUILD_ARGS}
```
sebastienvas: I think you want ${BAZEL_BUILD_ARGS} before your targets

.circleci/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    sudo apt-get -y install bazel \
+    wget software-properties-common make cmake python python-pip \
+    bc libtool automake zip time g++-6 gcc-6 \
+    rsync golang
+
```
rshriram: Can you reuse the dockerfile in istio/istio circleci/ ? We can have one big fat docker image that is pushed to docker hub.

.circleci/Makefile
```diff
@@ -0,0 +1,21 @@
...
+HUB ?=
+
+# Note: The build directory is the root of the istio/test-infra repository, not ./
+image:
+	docker build -t ${HUB}$(PROJECT)/proxy-builder:$(VERSION) -f Dockerfile .
+
+
+push:
+	gcloud docker -- push "${HUB}$(PROJECT)/proxy-builder:$(VERSION)"
```
rshriram: What does this do? Why do we need to push to gcloud ?

.circleci/config.yml
```diff
@@ -0,0 +1,39 @@
+version: 2
+
+jobs:
+  build:
+    #working_directory: src/proxy
+    environment:
+    docker:
+      - image: costinm/proxy-builder:0.4.4
```
rshriram: You can copy from istio/istio. We have access to xlarge resources now.

.circleci/Dockerfile
costinm: I'm not sure - in general the latest is supposed to be better :-) I would rather find problems with new bazel earlier. We do need to tag the docker images used for releases, so we can repro, and of course bazel must take backward compatibility seriously.

.circleci/Makefile
costinm: I've used this for testing - not sure if we should use gcr.io or dockerhub (istio/istio is using the later), 
and what name to use ( 'ci' is used by istio/istio)

.circleci/Dockerfile
costinm: I don't think it's the best option - fat is also slow and complex. Modular 'micro-services' like images seem better. For example the istio/istio is based on golang - doesn't need jdk and bazel or gcc. I'm working  on another image to build on ARM - which needs arm and android cross compilers and cmake, but don't need java or bazel or go.

If you add the cross compilers - it's getting too fat. 

Even for istio/istio - I think one image to build with bazel and generate artifacts, and one with only go for the normal builds is cleaner.

costinm: About gcloud - it's a leftover, I agree it should be under dockerhub.

But I was not sure about naming - istio/ci:bazel0.8-... ?  Or should we use 
istio/ci-bazel, etc ?

.circleci/Makefile
costinm: Not needed, will remove. I was trying to find which is faster...

.circleci/config.yml
costinm: Not on my fork, where I've been testing. 
It's an interesting problem - we want people to be able to fork and use circleci on the fork. The xlrage will be ignored - but if I increase the parallelism (-j 16), the build will fail on the small
machine. Not sure how to customize the build, maybe some env settings ? 

Even with the small image - the build seems pretty fast, in particular for incremental builds.

rshriram: I don’t think we can optimize for forks here. It won’t work until they sign up for circle ci. Let’s optimjze for our build time. 15mins is a long time.

.circleci/Dockerfile
rshriram: Envoy’s build image is about 1.3GB. Circleci caches the images. And since these images don’t change often, it’s often very helpful for people to download a single build container and build all they want from there, just like the repo tool concept. Internal build systems can replicate the setup from a single dockerfile and go about work. 

We use the same build for api and c++ in Envoy as an example. 

While I am up for the smaller images, I would like us to first get quickly to a streamlined build with minimal overhead. Once we have a well oiled pipeline, we can split up images if we find things to be slow.

costinm:
On Tue, Dec 5, 2017 at 7:21 PM, Shriram Rajagopalan <
notifications@github.com> wrote:

> *@rshriram* commented on this pull request.
> ------------------------------
>
> In .circleci/Dockerfile
> <https://github.com/istio/proxy/pull/702#discussion_r155138210>:
>
> > @@ -0,0 +1,17 @@
> +FROM    circleci/openjdk:latest
> +
> +# Based on https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/bazel/Dockerfile
> +RUN \
> +    sudo sh -c 'echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" > /etc/apt/sources.list.d/bazel.list '
> +
> +RUN \
> +    curl https://storage.googleapis.com/bazel-apt/doc/apt-key.pub.gpg | sudo apt-key add -
> +
> +RUN sudo apt-get update && \
> +    sudo apt-get -y install bazel \
> +    wget software-properties-common make cmake python python-pip \
> +    bc libtool automake zip time g++-6 gcc-6 \
> +    rsync golang
> +
>
> Envoy’s build image is about 1.3GB. Circleci caches the images. And since
> these images don’t change often, it’s often very helpful for people to
> download a single build container and build all they want from there, just
> like the repo tool concept. Internal build systems can replicate the setup
> from a single dockerfile and go about work.
>

That was the problem - at least in my tests I didn't see the cache, maybe
it's for the paid version.

I'm trying to automate the build for non-x86 architecture as well - and for
alpine, which needs as separate image anyways ( musl ) -
and didn't want to bloat the image.

I don't mind a single image with all the tools - I can send a PR to add
what's missing in istio/istio image, but we should baze
it on jdk. I was planning to bake kubectl/etcd and the other tools into the
base image as well, they only change when
k8s is updated and can stay in the image.



> We use the same build for api and c++ in Envoy as an example.
>
> While I am up for the smaller images, I would like us to first get quickly
> to a streamlined build with minimal overhead. Once we have a well oiled
> pipeline, we can split up images if we find things to be slow.
>
Since I finally got it working with the smaller image - and changing the
istio/istio image has its own set of risks and testing,
can we start  with it and incrementally merge the images ? Technically this
image can be used to build istio/istio with bazel
as well, so the only 'small' image is the go-only one, and changing that is
a separate PR anyways..

rshriram:
On Tue, Dec 5, 2017 at 9:31 PM Costin Manolache <notifications@github.com>
wrote:

> On Tue, Dec 5, 2017 at 7:21 PM, Shriram Rajagopalan <
> notifications@github.com> wrote:
>
> > *@rshriram* commented on this pull request.
> > ------------------------------
> >
> > In .circleci/Dockerfile
> > <https://github.com/istio/proxy/pull/702#discussion_r155138210>:
> >
> > > @@ -0,0 +1,17 @@
> > +FROM circleci/openjdk:latest
> > +
> > +# Based on
> https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/bazel/Dockerfile
> > +RUN \
> > + sudo sh -c 'echo "deb [arch=amd64]
> http://storage.googleapis.com/bazel-apt stable jdk1.8" >
> /etc/apt/sources.list.d/bazel.list '
> > +
> > +RUN \
> > + curl https://storage.googleapis.com/bazel-apt/doc/apt-key.pub.gpg |
> sudo apt-key add -
> > +
> > +RUN sudo apt-get update && \
> > + sudo apt-get -y install bazel \
> > + wget software-properties-common make cmake python python-pip \
> > + bc libtool automake zip time g++-6 gcc-6 \
> > + rsync golang
> > +
> >
> > Envoy’s build image is about 1.3GB. Circleci caches the images. And since
> > these images don’t change often, it’s often very helpful for people to
> > download a single build container and build all they want from there,
> just
> > like the repo tool concept. Internal build systems can replicate the
> setup
> > from a single dockerfile and go about work.
> >
>
> That was the problem - at least in my tests I didn't see the cache, maybe
> it's for the paid version.


Take a look at this: https://github.com/istio/istio/pull/1952

I think to build Envoy, we just need a Docker image per platform. Circleci
does not offer arm builds though

costinm:
The arm builds are cross-compiled - the problem is installing the cross
compilers and the android ndk take lots
of space and doesn't seem right.

I'm not sure about installing bazel (or minikube, etc) on each build ...

.cloudbuild/Dockerfile
lizan: Can we just use envoyproxy/envoy-build as base image, maybe with SHA tagged. That ensure we meet envoy build requirements when we update it.

costinm: Good point - however Shriram recommended to use the same image with istio/istio. But we may base that one on envoyproxy. 

Only problem is that envoy-build has even more deps (clang, etc).

For now - I'm removing cloudbuild, it didn't work, will let others handle it...

Makefile
costinm: Done.

PR #711, opened 2017-12-08
Implements function GetJWTPayload in mixer filter.

JimmyCYJ:
**What this PR does / why we need it**: Overrides function CheckData::GetJWTPayload() in /proxy/src/envoy/mixer/http_filter.cc, so that mixer filter is able to generate authentication attributes.

**Release note**:

```release-note
```


src/envoy/mixer/http_filter.cc
```diff
@@ -257,6 +258,37 @@ class CheckData : public HttpCheckData,
     }
     return false;
   }
+
+  bool GetJWTPayload(
+      std::map<std::string, std::string>* payload) const override {
+    if (payload == nullptr) {
+      return false;
+    }
+    const std::string kSecIstioAuthUserinfo("sec-istio-auth-userinfo");
```
qiwzhang: call this function https://github.com/istio/proxy/blob/master/src/envoy/auth/http_filter.cc#L31
to get the key.  Don't define another one here

src/envoy/mixer/http_filter.cc
```diff
@@ -257,6 +258,37 @@ class CheckData : public HttpCheckData,
...
+    std::string sec_istio_auth_userinfo;
+    // Currently the value in "sec-istio-auth-userinfo" header is a string of
+    // base64url-encoded payload JSON. Decode payload to populate auth
+    // attributes.
+    // TODO: Support extracting auth attributes from all types of contents in
+    // "sec-istio-auth-userinfo" header.
+    if (FindHeaderByName(kSecIstioAuthUserinfo, &sec_istio_auth_userinfo)) {
```
qiwzhang: call headers_.get() directly so you can use LowerCastString class

src/envoy/mixer/http_filter.cc
```diff
@@ -257,6 +258,37 @@ class CheckData : public HttpCheckData,
...
+      Json::ObjectSharedPtr json_object =
+          Auth::Jwt::LoadFromBase64UrlEncodedPayload(sec_istio_auth_userinfo);
+      if (json_object != nullptr) {
```
qiwzhang: Use JsonObje::iterate() function to get all map elements,  users can define any map elements as they want.  we have to extract all of them.

PR #718, opened 2017-12-12
Add (experimental) support for building on ARM platforms and using cmake

costinm:

Proof of concept for Istio running on Raspberry Pi and other ARM platforms, 
and for using cmake for better IDE integration.

This doesn't impact the code - it simply adds few jobs to the circleci integration.

This also uses repo tool for dep management - which would allow us to catch dependency 
breakages earlier, and may allow a flat build, where we create the full proxy docker image, 
including agent, from this git tree.

```release-note
NONE
```


kyessenov:
It goes without saying that the experimental stuff should not be made as "required" in circle ci jobs. Other than that, thanks for investing into ARM/pi :)

costinm:
Of course, it should not be required.

I'm planning to also add one target for alpine/musl ( it works, but I
didn't integrate it yet in circleci ), and hopefully community will
contribute other CPUs and platforms, and we may include custom/unofficial
builds with other external plugins ( squash ? )

kyessenov:
squash wants to upstream its filter, so let's just wait for it become part
of the main build.

PR #719, opened 2017-12-12
Strip out "spiffe://" in the identity

JimmyCYJ:
**What this PR does / why we need it**:Strips out the prefix "spiffe://" from the URI in the SAN field of the peer certificate.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/envoy/mixer/utils.cc
+++ b/src/envoy/mixer/utils.cc
@@ -20,6 +20,12 @@ namespace Envoy {
 namespace Http {
 namespace Utils {
 
+namespace {
+
+const std::string kSPIFFEPrefix("spiffe://");
+
+}  // namespace
+
 std::map<std::string, std::string> ExtractHeaders(const HeaderMap& header_map) {
   std::map<std::string, std::string> headers;
   header_map.iterate(
@@ -55,11 +61,10 @@ bool GetSourceUser(const Network::Connection* connection, std::string* user) {
     Ssl::Connection* ssl = const_cast<Ssl::Connection*>(connection->ssl());
     if (ssl != nullptr) {
       std::string result = ssl->uriSanPeerCertificate();
-      std::string prefix = "spiffe://";
-      // Strip out the prefix "spiffe://" in the identity.
-      std::size_t found = result.find(prefix);
-      if (found == 0) {
-        *user = result.substr(prefix.size());
+      if (result.length() >= kSPIFFEPrefix.length() &&
+          result.compare(0, kSPIFFEPrefix.length(), kSPIFFEPrefix) == 0) {
+        // Strip out the prefix "spiffe://" in the identity.
+        *user = result.substr(kSPIFFEPrefix.size());
         return true;
       }
     }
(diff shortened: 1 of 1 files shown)


kyessenov:
What's the motivation?

src/envoy/mixer/utils.cc
```diff
@@ -54,8 +54,14 @@ bool GetSourceUser(const Network::Connection* connection, std::string* user) {
   if (connection) {
     Ssl::Connection* ssl = const_cast<Ssl::Connection*>(connection->ssl());
     if (ssl != nullptr) {
-      *user = ssl->uriSanPeerCertificate();
-      return true;
+      std::string result = ssl->uriSanPeerCertificate();
```
wattli: Is there any test we need to update as well?

JimmyCYJ: I have run all existing tests in /proxy via bazel test ..., and all of them pass.

src/envoy/mixer/utils.cc
```diff
@@ -54,8 +54,14 @@ bool GetSourceUser(const Network::Connection* connection, std::string* user) {
...
+      std::string prefix = "spiffe://";
```
qiwzhang: define it as global const  in annoymous namespace

src/envoy/mixer/utils.cc
```diff
@@ -54,8 +54,14 @@ bool GetSourceUser(const Network::Connection* connection, std::string* user) {
...
+      // Strip out the prefix "spiffe://" in the identity.
+      std::size_t found = result.find(prefix);
```
qiwzhang: use compare like this

https://github.com/istio/mixerclient/blob/master/src/attributes_builder.cc#L34

find() is very expensive

src/envoy/mixer/utils.cc
```diff
@@ -54,8 +60,13 @@ bool GetSourceUser(const Network::Connection* connection, std::string* user) {
...
     if (ssl != nullptr) {
-      *user = ssl->uriSanPeerCertificate();
-      return true;
+      std::string result = ssl->uriSanPeerCertificate();
+      if (result.length() >= kSPIFFEPrefix.length() &&
+          result.compare(0, kSPIFFEPrefix.length(), kSPIFFEPrefix) == 0) {
+        // Strip out the prefix "spiffe://" in the identity.
+        *user = result.substr(kSPIFFEPrefix.size());
+        return true;
+      }
```
qiwzhang: what about else? the one doesn't match the prefix?

wattli: if not, we don't need to do anything. So far, all id will have this prefix.

qiwzhang: I think we still want to get source.user even it doesn't match spiff prefix.

qiwzhang: what format will it look like?   will it have similar format of xyz://  If so,  we want to strip it too.

wattli: Good point, as mentioned, so far, all Ids have prefix "spiffe://", i don't think we have plan to support others in the near future.

JimmyCYJ: From offline discussion with Wayne, if there is no "spiffe://" prefix, assign the uri to user and return true. This helps to keep consistent behavior for cases where uri does not begin with "spiffe://".

PR #724, opened 2017-12-13
Use shorter name for proxy debian files

mattdelco:
Last week I implemented a request to get rid of the behavior that was causing "-release" to get added to the basename of debian files.  I did this by unifying the release and non-release rules to use a version_file = "//:deb_version" in the pkg_deb() rules, whereby the top-level BUILD generates the version file using:

"echo $${ISTIO_VERSION:-\"0.3.0-dev\"} > \"$@\""

This seems to be causing the postsubmit to fail in proxy.  The issue is that script/release-binary expects a name like:

istio-proxy_0.3-dev_amd64.deb

but after my change the version part is not included in the filename:

istio-proxy__amd64.deb

I'm not sure if this is a bazel bug or intentional feature when the version goes from a static value to a file reference.  I've confirmed before (and reconfigured again) that the version # in the deb is appropriate.  I've also checked that the bazel-genfiles/deb_version.txt does contain valid contents.  I tried running "bazel build //:deb_version.txt" as a separate step in case there was a race in bazel, but this didn't change the behavior.

I was going to make a change to change the name references to "istio-proxy__amd64.deb", but the shorter "istio-proxy.deb" is just a symlink to that (and the shorter name seems to be what everyone else in istio is using) so that seems like the safer option for me to use.  I'm not sure why the amd64 version was used in the first place unless there is/was a cross-compilation case.

```release-note
NONE
```

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -25,6 +25,14 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+genrule(
+    name = "deb_version",
+    srcs = [],
+    outs = ["deb_version.txt"],
+    cmd = "echo $${ISTIO_VERSION:-\"0.3.0-dev\"} > \"$@\"",
+    visibility = ["//visibility:public"],
+)
+
 load("@io_bazel_rules_go//go:def.bzl", "go_prefix")
 
 go_prefix("istio.io/proxy")
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,14 @@
+reviewers:
+  - qiwzhang
+  - lizan
+  - sebastienvas
+  - geeknoid
+  - rshriram
+  - linsun
+approvers:
+  - qiwzhang
+  - lizan
+  - sebastienvas
+  - geeknoid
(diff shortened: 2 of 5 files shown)


rkpagadala:
We reverted your previous commit, to see if that helps. (It did)
https://github.com/istio/proxy/commit/13669ce4244e248f6360675068d2905edaed8b81

Can you please add your commit back in again. Sorry about that.

mattdelco:
Original change reapplied.  I noticed the following too, which should be alright as-is but also has potential for simplification:

tools/deb/test/build_docker.sh:cp -f $BAZEL_TARGET/istio-proxy_*_amd64.deb tools/deb/test/istio-proxy_amd64.deb

I'm guessing this will be handled eventually given the current outlook on the build systems.

mattdelco:
The proxy repo doesn't have any OWNERS files so I'm not sure it's possible at the present time for anyone to satisfy the requirements of the merge bot

PR #726, opened 2017-12-14
make control plane mTLS for vms by default

rkpagadala:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/deb/istio-start.sh
+++ b/tools/deb/istio-start.sh
@@ -26,12 +26,20 @@ if [[ -r ${ISTIO_SIDECAR_CONFIG} ]]; then
   . $ISTIO_SIDECAR_CONFIG
 fi
 
+# Load config variables ISTIO_SYSTEM_NAMESPACE, CONTROL_PLANE_AUTH_POLICY
+ISTIO_CLUSTER_CONFIG=${ISTIO_CLUSTER_CONFIG:-/var/lib/istio/envoy/cluster.env}
+if [[ -r ${ISTIO_CLUSTER_CONFIG} ]]; then
+  . $ISTIO_CLUSTER_CONFIG
+fi
+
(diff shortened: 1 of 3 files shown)


tools/deb/istio-start.sh
```diff
@@ -54,7 +62,7 @@ fi
 ${ISTIO_BIN_BASE}/istio-iptables.sh
 
 if [ -f ${ISTIO_BIN_BASE}/pilot-agent ]; then
-  exec su -s /bin/bash -c "INSTANCE_IP=${ISTIO_SVC_IP} POD_NAME=${POD_NAME} POD_NAMESPACE=${NS} exec ${ISTIO_BIN_BASE}/pilot-agent proxy --serviceCluster $SVC > ${ISTIO_LOG_DIR}/istio.log" istio-proxy
+  exec su -s /bin/bash -c "INSTANCE_IP=${ISTIO_SVC_IP} POD_NAME=${POD_NAME} POD_NAMESPACE=${NS} exec ${ISTIO_BIN_BASE}/pilot-agent proxy --serviceCluster $SVC --discoveryAddress istio-pilot.${ISTIO_SYSTEM_NAMESPACE} --controlPlaneAuthPolicy $CONTROL_PLANE_AUTH_POLICY > ${ISTIO_LOG_DIR}/istio.log" istio-proxy
```
ldemailly: what would happen if --discoveryAddress is not passed?

rkpagadala: The namespace part of the discovery address is used by pilot-agent to construct spiffe uri to authenticate. So it needs to be correct otherwise the client will reject the secure connection. The default value works for plain connections.

ldemailly:
cross linking: https://github.com/istio/istio/pull/2154

Issue #886, opened 2018-01-02
make prow presubmit to run asan and tsan test paralleled 


qiwzhang:
Now, "make test" is running regular test, asan, and tsan sequentially.     prow/presubmit test failed as timeout since each test is about 30 minutes. (https://storage.googleapis.com/istio-prow/pull/istio_proxy/771/proxy-presubmit/287/build-log.txt)

It is better to run these tests paralleled.

PR #888, opened 2018-01-02
Remove 5 seconds sleep for integration tests.

qiwzhang:

To speed up integration tests.  It is taking too long for presubmit tests.
* sleep(5) is removed when detecting Envoy starting up.  Instead using net.dial to check the http ports.
* also removed "envoy" binary in src/envoy/mixer/BUILD which is only used by start_envoy script.

**Release note**:

```release-note
None
```

### DIFF
--- a/src/envoy/mixer/integration_test/envoy.go
+++ b/src/envoy/mixer/integration_test/envoy.go
@@ -87,8 +87,8 @@ func (s *Envoy) Start() error {
 	if err == nil {
 		url := fmt.Sprintf("http://localhost:%v/server_info", AdminPort)
 		WaitForHttpServer(url)
-		WaitForAPort(ClientProxyPort)
-		WaitForAPort(ServerProxyPort)
+		WaitForPort(ClientProxyPort)
+		WaitForPort(ServerProxyPort)
 	}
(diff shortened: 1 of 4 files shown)


qiwzhang:
https://github.com/istio/proxy/issues/886 is not needed if this works.

src/envoy/mixer/integration_test/envoy.go
```diff
@@ -88,11 +87,8 @@ func (s *Envoy) Start() error {
 	if err == nil {
 		url := fmt.Sprintf("http://localhost:%v/server_info", AdminPort)
 		WaitForHttpServer(url)
-		// TODO: need a better way to detect listen ports are up.
-		// In tsan case, even admin port is up, other ports may not.
-		// Test program will crash with connection is refused error.
-		// Wait 2 seconds more for tsan
-		time.Sleep(time.Second * 5)
+		WaitForAPort(ClientProxyPort)
```
kyessenov: nit: No need to have the article "a" here

qiwzhang: Done.  Thanks

Issue #728, opened 2017-12-15
Request with invalid Sec-Istio-Auth-Userinfo header cause Envoy crash


kryzthov:
When sending a request with a dummy `Sec-Istio-Auth-Userinfo:Foo` header, Envoy crashes:
```
terminate called after throwing an instance of 'Envoy::Json::Exception'
  what():  JSON supplied is not valid. Error(offset 0, line 1): Invalid value.

[2017-12-15 23:54:38.645][25][critical][backtrace] bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:101] Caught Aborted, suspect faulting address 0x53900000017
[2017-12-15 23:54:38.646][25][critical][backtrace] bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:85] Backtrace obj</lib/x86_64-linux-gnu/libc.so.6> thr<25> (use tools/stack_decode.py):
[2017-12-15 23:54:38.646][25][critical][backtrace] bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:95] thr<25> #0 0x7f96c53c7428
[...]
ERROR: logging before flag.Parse: W1215 23:54:38.656636       1 agent.go:204] Epoch 0 terminated with an error: signal: aborted (core dumped)
...
```

kryzthov:
Also, I would imagine the `Sec-Istio-Auth-Userinfo` header should probably not be forwarded.

PR #889, opened 2018-01-02
Fix crash with invalid sec-istio-auth-userinfo header

qiwzhang:

Fix proxy crash with invalid sec-istio-auth-userinfo header

**Release note**:

```release-note
None
```

### DIFF
--- a/src/envoy/mixer/http_filter.cc
+++ b/src/envoy/mixer/http_filter.cc
@@ -287,18 +287,32 @@ class CheckData : public HttpCheckData,
     if (!entry) {
       return false;
     }
-    std::string payload_str = Auth::Base64UrlDecode(
-        std::string(entry->value().c_str(), entry->value().size()));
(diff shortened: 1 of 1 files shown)


qiwzhang:
https://github.com/istio/proxy/issues/728

src/envoy/mixer/http_filter.cc
```diff
@@ -287,18 +287,32 @@ class CheckData : public HttpCheckData,
...
-          // will throw execption if value type is not string.
-          try {
-            (*payload)[key] = obj.asString();
-          } catch (...) {
-          }
-          return true;
-        });
+    std::string value(entry->value().c_str(), entry->value().size());
+    std::string payload_str = Auth::Base64UrlDecode(value);
+    // Return is an empty string if Base64 decode fails.
```
JimmyCYJ: I think "is" is redundant, remove "is" after Return.

qiwzhang: Done

PR #907, opened 2018-01-09
Fix a memory leak when a request is canceled.

qiwzhang:

When a request is canceled when it is waiting for Mixer Check call,  there is a memory leak.
This PR fixes this leak.

**Release note**:

```release-note
None
```

### DIFF
--- a/src/envoy/mixer/grpc_transport.cc
+++ b/src/envoy/mixer/grpc_transport.cc
@@ -102,6 +102,12 @@ void GrpcTransport<RequestType, ResponseType>::onFailure(
 }
 
 template <class RequestType, class ResponseType>
+void GrpcTransport<RequestType, ResponseType>::Cancel() {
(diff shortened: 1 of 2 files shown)


src/envoy/mixer/grpc_transport.cc
```diff
@@ -113,7 +119,7 @@ GrpcTransport<RequestType, ResponseType>::GetFunc(Upstream::ClusterManager& cm,
             new Grpc::AsyncClientImpl<RequestType, ResponseType>(
                 cm, kMixerServerClusterName)),
         request, headers, response, on_done);
-    return [transport]() { transport->request_->cancel(); };
+    return [transport]() { transport->Cancel(); };
```
rshriram: This seems fine, but any reason to not use boxed pointers (shared/unique etc.) so that we don't have to deal with these?

lizan: `transport` have longer lifetime than the lambda, otherwise proxy will simply crash.

PR #917, opened 2018-01-12
Remove go integration tests.

qiwzhang:

Remove all go integration tests.  They have been moved to istio repo

**Release note**:

```release-note
None
```


WORKSPACE
```diff
@@ -63,13 +63,8 @@ git_repository(
 load("@mixerapi_git//:api_dependencies.bzl", "mixer_api_for_proxy_dependencies")
 mixer_api_for_proxy_dependencies()
 
-ISTIO_SHA = "cdbdb153fb673cf444649c520ba7ed1b0dc99972"
```
kyessenov: Can you nuke mixer_api dependencies, too?
We want to nuke all bazel to maximum extent per @rshriram

qiwzhang: Yes,  I need to wait for api repo to remove BUILD files.   Otherwise, I could not put my BUILD file on its top.  
@douglas-reid @rshriram not sure if api repos is ready to remove all its BUILD files.

Anyway, it will be a separate PR.

douglas-reid: Ok. Let's kill the BUILD files from istio/api then. I think we were waiting on proxy, so if we don't need them, they can go.

PR #921, opened 2018-01-15
Update envoy sha

rshriram:


kyessenov:
skylark rule changed. I think @lizan had a working bazel update build.

rshriram:
Its been 3 months since we updated Envoy. So much for the bots

mandarjog:
really? it was that old :-(  
@kyessenov this also means that the proxy crash we see in e2e tests is unrelated to base envoy.

qiwzhang:
@lizan could you submit your PR to update proxy to latest Envoy?

rkpagadala:
@rshriram Envoy has always been updated manually, not sure why. Istio eng prod team has been asked to update envoy using bots only last week. We have it on our list to get done.

lizan:
Envoy's C++ interface is never stable so it isn't always simply updating SHA. Bot can perform SHA only changes, but if the build fails we should get notified.

lizan:
I'll have the updating PR by EoD

lizan:
Closing in favor of #922

PR #932, opened 2018-01-23
add envoyproxy as dependency

rkpagadala:
Add envoy proxy as dependency in istio.deps, after this PR is merged envoy proxy will be be updated once every week.

**Release note**:

```release-note
```


qiwzhang:
@lizan do we want to pick up every Envoy commit automatically?   Is it safe?

lizan:
@qiwzhang as long as presubmit passes it should be good.

rkpagadala:
@qiwzhang bot only tries to pick envoy up once a week. Presubmit should pass before the code goes in, and when it fails you will have a week to resolve (or the next week could fail again)

rkpagadala:
Please add /lgtm and PR will get auto merged

PR #736, opened 2017-12-21
Add tsan and asan

qiwzhang:
**What this PR does / why we need it**:

Add tsan and asan test.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
None
```

PR #947, opened 2018-01-25
Bring back accidentally deleted build options in #736

lizan:

#736 accidentally remove `-c opt` from release build and a couple of compile options, bring them back.

cc @costinm @ldemailly 

**Release note**:
```release-note
None
```


ldemailly:
thank explains why debug and non debug perf looked the same ! thanks !
/lgtm

ldemailly:
I wonder if (beside perf tests) we can add some test for this

kyessenov:
We should just aggregate perf charts, so we can detect deviations. I don't
think this needs to be automated.

PR #949, opened 2018-01-25
Use v4only for CircleCI

lizan:
Fix CircleCI presubmit failure

**Which issue this PR fixes*

**Release note**:
```release-note
None
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,8 @@ jobs:
     environment:
     docker:
       - image: istio/ci:go1.9-bazel0.7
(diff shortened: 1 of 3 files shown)


qiwzhang:
@lizan Hmm,   config=clang-tsan failed.  Any ideas?

lizan:
clang-tsan doens't work since the ci image doesn't have clang, we're building clang-tsan with gcc and it fails.

qiwzhang:
@lizan could you remove clang-tsan in circleci?

PR #977, opened 2018-02-01
Add an option in Envoy filter to output JWT sub for authentication

### LINKED ISSUE #976: Add an option in Envoy filter to output JWT sub for authentication
Add an option in the Envoy filter config to output the JWT sub attribute for the authentication purpose.

lei-tang:
Add an option in Envoy filter config to output the JWT sub attribute for the authentication purpose.
When the option "authn_output_jwt_sub" is true, the JWT sub attribute is outputted.
To test, run: "bazel test //src/envoy/auth:jwt_test" or
"bazel test //src/envoy/auth:http_filter_integration_test"

**What this PR does / why we need it**: add an option in Envoy filter config to output the JWT sub attribute for the authentication purpose.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #976

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/auth/config.cc
+++ b/src/envoy/auth/config.cc
@@ -173,6 +173,7 @@ JwtAuthConfig::JwtAuthConfig(
   // TODO: public key expiration should be per issuer.
   pubkey_cache_expiration_sec_ = 600;
   user_info_type_ = UserInfoType::kPayloadBase64Url;
+  authn_output_jwt_sub_ = false;
 }
 
 /*
@@ -191,6 +192,8 @@ JwtAuthConfig::JwtAuthConfig(const Json::Object &config,
   } else {
     user_info_type_ = UserInfoType::kPayloadBase64Url;
   }
+  // By default, the subject field of a JWT is not outputted
+  authn_output_jwt_sub_ = config.getBoolean("authn_output_jwt_sub", false);
 
   pubkey_cache_expiration_sec_ =
       config.getInteger("pubkey_cache_expiration_sec", 600);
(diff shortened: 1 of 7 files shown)


lizan:
What is the purpose of this PR? The sub claim is already included in the sec-istio-auth-userinfo.

lei-tang:
@lizan The purpose of this code is: To implement Istio authentication policy,  the identity related attributes need to be extracted based on the Istio authentication policy configuration.

lizan:
I still don't think you need this, istio attributes are extracted in mixer filter, so you can just extract from userinfo there.

qiwzhang:
Agreed with @lizan.  This PR is not the right way to do this.
Auth attributes are extracted here: https://github.com/istio/mixerclient/blob/master/control/src/http/attributes_builder.cc#L60

lei-tang:
@lizan @qiwzhang Added a few paragraphs in [the Istio authentication design doc ](https://docs.google.com/document/d/1ezP4UuOn3JXEs_cXW4GyPGq-Ppq_XhS9-M-lN6ocOA4/edit?ts=5a736cda#bookmark=id.h49m4lr09aq) to discuss the options for interacting with the existing JWT-auth module when implementing the Istio authentication policy related to JWT.

diemtvu:
I don't think we need this neither. For prototyping, the authN filter can just consume what the existing JWT output, and if we end up in that direction (have separate filters for each authN mechanism, then the authN filter to consolidate attributes) then it's better that this filter output all claims as it (and "output" could be as request-context shared object rather than put back to header)

qiwzhang:
@lei-tang should we close this PR?

lei-tang:
@qiwzhang Yes, we can close this PR now; it has served its purpose of inspiring the discussion on the architecture of the Istio authentication policy implementation interacting with the existing JWT-auth module.

PR #983, opened 2018-02-02
Fix auth filter not to crash with wrong jwks cluster

qiwzhang:

If jwks cluster is wrong,  auth filter will cause Envoy to crash.
This fix will check cluster manager first, and reject the requests.

**Release note**:

```release-note
None
```

### DIFF
--- a/src/envoy/auth/jwt_authenticator.cc
+++ b/src/envoy/auth/jwt_authenticator.cc
@@ -128,7 +128,7 @@ void JwtAuthenticator::FetchPubkey(PubkeyCacheItem* issuer) {
 
   const auto& cluster = issuer->jwt_config().jwks_uri_envoy_cluster();
   if (cm_.get(cluster) == nullptr) {
-    DoneWithStatus(Status::WRONG_JWKS_URI_ENVOY_CLUSTER);
+    DoneWithStatus(Status::FAILED_FETCH_PUBKEY);
     return;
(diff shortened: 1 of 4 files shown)


lizan:
I don't think the cluster not matchc should be exposed to the client querying, it is implementation detail and should not be exposed.

Crash with proper message is fine.

src/envoy/auth/jwt_authenticator.cc
```diff
@@ -126,11 +126,15 @@ void JwtAuthenticator::FetchPubkey(PubkeyCacheItem* issuer) {
   message->headers().insertPath().value(path);
   message->headers().insertHost().value(host);
 
-  request_ = cm_.httpAsyncClientForCluster(
-                    issuer->jwt_config().jwks_uri_envoy_cluster())
-                 .send(std::move(message), *this,
-                       Optional<std::chrono::milliseconds>());
+  const auto& cluster = issuer->jwt_config().jwks_uri_envoy_cluster();
+  if (cm_.get(cluster) == nullptr) {
+    DoneWithStatus(Status::WRONG_JWKS_URI_ENVOY_CLUSTER);
```
lizan: Just use `FAILED_FETCH_PUBKEY`.

qiwzhang: Done

PR #993, opened 2018-02-05
make mixer cluster name configurable

rshriram:
Please see https://github.com/istio/api/pull/358 for context.
Essentially remove hardcoded "mixer_server" and allow users to override the name, in addition to allowing separate clusters for check vs reports so that it makes it easier to load balance/manage the system.

### DIFF
--- a/src/envoy/mixer/config.cc
+++ b/src/envoy/mixer/config.cc
@@ -51,6 +51,12 @@ const std::string kDisableTcpCheckCalls("disable_tcp_check_calls");
 
 const std::string kV2Config("v2");
 
+// The name for the mixer server cluster.
+const std::string kDefaultMixerClusterName("mixer_server");
+// The Json object name for check_cluster and report_cluster
+const std::string kCheckCluster("check_cluster");
+const std::string kReportCluster("report_cluster");
+
 void ReadStringMap(const Json::Object& json, const std::string& name,
                    Attributes* attributes) {
   if (json.hasObject(name)) {
@@ -73,6 +79,9 @@ void ReadTransportConfig(const Json::Object& json, TransportConfig* config) {
   config->set_disable_check_cache(json.getBoolean(kDisableCheckCache, false));
   config->set_disable_quota_cache(json.getBoolean(kDisableQuotaCache, false));
   config->set_disable_report_batch(json.getBoolean(kDisableReportBatch, false));
+
+  config->set_check_cluster(json.getString(kCheckCluster, kDefaultMixerClusterName));
+  config->set_report_cluster(json.getString(kReportCluster, kDefaultMixerClusterName));
 }
 
 bool ReadV2Config(const Json::Object& json, Message* message) {
--- a/src/envoy/mixer/config.cc
+++ b/src/envoy/mixer/config.cc
@@ -123,10 +123,12 @@ void HttpMixerConfig::Load(const Json::Object& json) {
     legacy_quotas.clear();
   }
 
-  check_cluster = transport_config->check_cluster().empty() ? kDefaultMixerClusterName :
-                  transport_config->check_cluster();
-  report_cluster = transport_config->report_cluster().empty() ? kDefaultMixerClusterName :
-                   transport_config->report_cluster();
+  check_cluster = transport_config->check_cluster().empty()
+                      ? kDefaultMixerClusterName
+                      : transport_config->check_cluster();
+  report_cluster = transport_config->report_cluster().empty()
+                       ? kDefaultMixerClusterName
+                       : transport_config->report_cluster();
 }
 
 void HttpMixerConfig::CreateLegacyRouteConfig(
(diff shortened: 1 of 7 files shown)


src/envoy/mixer/grpc_transport.cc
```diff
@@ -110,13 +107,14 @@ void GrpcTransport<RequestType, ResponseType>::Cancel() {
 template <class RequestType, class ResponseType>
 typename GrpcTransport<RequestType, ResponseType>::Func
 GrpcTransport<RequestType, ResponseType>::GetFunc(Upstream::ClusterManager& cm,
+                                                  const std::string& cluster_name,
                                                   const HeaderMap* headers) {
-  return [&cm, headers](const RequestType& request, ResponseType* response,
+  return [&cm, &cluster_name, headers](const RequestType& request, ResponseType* response,
```
lizan: does `cluster_name` outlive the returned function? If not sure, capture by value.

src/envoy/mixer/config.cc
```diff
@@ -73,6 +79,9 @@ void ReadTransportConfig(const Json::Object& json, TransportConfig* config) {
   config->set_disable_check_cache(json.getBoolean(kDisableCheckCache, false));
   config->set_disable_quota_cache(json.getBoolean(kDisableQuotaCache, false));
   config->set_disable_report_batch(json.getBoolean(kDisableReportBatch, false));
+
+  config->set_check_cluster(json.getString(kCheckCluster, kDefaultMixerClusterName));
+  config->set_report_cluster(json.getString(kReportCluster, kDefaultMixerClusterName));
```
qiwzhang: This is the wrong place to set it.    Sorry for the misleading,  hhis function should be called "ReadLegacyTransportConfig"

In the mixer_filter config,  if there is a "v2" field,  
these data will be override by the data from "v2".  

https://github.com/istio/proxy/blob/master/src/envoy/mixer/config.cc#L115

You have to set these two fields after ReadV2Config() call

and check if they are not set by Pilot,  set their default value as "mixer_server"

src/envoy/mixer/grpc_transport.cc
rshriram: I think it does? Because we used to pass kMixerServerClusterName variable to GrpcTransport.

lizan: `kMixerServerClusterName` is a global constant so it definitely outlives. But does the cluster_name outlive?

src/envoy/mixer/config.cc
```diff
@@ -110,13 +117,19 @@ void HttpMixerConfig::Load(const Json::Object& json) {
...
+  TransportConfig* transport_config = http_config.mutable_transport();
+  ReadLegacyTransportConfig(json, transport_config);
 
   has_v2_config = ReadV2Config(json, &http_config);
   if (has_v2_config) {
     // If v2 config is valid, clear v1 legacy_quotas.
     legacy_quotas.clear();
   }
+
+  transport_config->set_check_cluster(
```
qiwzhang: We should check if check_cluster is non-empty.  If it is non-empty, it is set by Pilot, we should not override it

src/envoy/mixer/config.cc
```diff
@@ -25,6 +25,9 @@ using ::istio::mixer_client::AttributesBuilder;
 using ::istio::mixer::v1::config::client::ServiceConfig;
 using ::istio::mixer::v1::config::client::TransportConfig;
 
+#define PROTOBUF_GET_STRING_OR_DEFAULT(message, field_name, default_value)                        \
+  ((message).has_##field_name() ? (message).field_name() : (default_value))
```
qiwzhang: protobuf primary type, such as string,  does not generate "has_"  function.  Only the message type does.
it has default value,  string default is empty.

src/envoy/mixer/mixer_control.cc
```diff
@@ -70,10 +72,13 @@ HttpMixerControl::HttpMixerControl(const HttpMixerConfig& mixer_config,
   ::istio::mixer_control::http::Controller::Options options(
       mixer_config.http_config, mixer_config.legacy_quotas);
 
-  CreateEnvironment(cm, dispatcher, random, &options.env);
+  check_cluster_ = mixer_config.check_cluster;
+  report_cluster_ = mixer_config.report_cluster;
 
-  controller_ = ::istio::mixer_control::http::Controller::Create(options);
+  CreateEnvironment(cm, dispatcher, random, check_cluster_, report_cluster_,
```
qiwzhang: I think you will not need check_cluster_ and report_cluster_ in MixerControl object if you don't pass reference to CreateEnvironment().    Just pass std::string and make a copy

lizan: passing const reference is fine as long as the lamda in GetFunc captures cluster_name as copy, so just remove check_cluster_ and report_cluster_ is good.

rshriram: Oh I needed to persist these variables in mixer control object because there is a GetFunc call in http_filter.cc that does `mixer_control_->...` ..
Ideally, i would have liked to stuff them into the options object, but that happens to be defined in mixerclient (spreading to too many repos).

Does that your point @qiwzhang ?

qiwzhang: I see. That one is per-route Check transport function.  My suggestion is:

in both HTTPMixerControl and TcpMixerControl store a const reference to MixerConfig

const MixerConfig& config_;

Expose it as

const MxerConfig& config() const { return config_; }

use the check_cluster_ as

 config()->check_cluster()

It is better than store anothery copy in MixerControl

qiwzhang: Well, since the PR is merged.  I will clean it up later.

qiwzhang: Here is my PR https://github.com/istio/proxy/pull/1001
to clean this up

PR #1019, opened 2018-02-09
refactor: Lazy initialize mixer filter by route config

lizan:

- Add a filter wrapper to lazy initialize filters until it knows route
- Modify mixer filter to use it

Tested with latest istio/istio integration test manually.

**Release note**:
```release-note
None
```

### DIFF
--- a/src/envoy/common/per_route_http_filter.h
+++ b/src/envoy/common/per_route_http_filter.h
@@ -1,0 +1,170 @@
+/* Copyright 2017 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


lizan:
No, the filters can be still used standalone, like auth. This just allows
you to lazy initialize it by route entries. In mixer filter factory now you
can pass per route config to initialize JwtAuthFilter, though
JwtAuthFilterFactory can still initialize the filter without this.

lizan:
The point of this PR is, you can even write a generic FilterFactory that construct existing filters (e.g. fault injection, grpc transcoding) from the JSON in route entry. All you need is customize your Filter factory, not the stream filter itself.

src/envoy/common/per_route_http_filter.h
```diff
@@ -0,0 +1,170 @@
...
+      inner_filter_ = constructor(request_info.routeEntry());
+    }
+    auto filter = dynamic_cast<AccessLog::Instance *>(inner_filter_.get());
+    if (filter) {
+      filter->log(request_headers, response_headers, request_info);
+    }
+  }
+
+  void onDestroy() override {
+    auto filter = dynamic_cast<Http::StreamFilterBase *>(inner_filter_.get());
```
JimmyCYJ: Is it necessary to use dynamic_cast in every function? Can we use inner_filter_->onDestroy() instead? I think it is recommended to avoid use too many dynamic_cast in production code [[link](https://google.github.io/styleguide/cppguide.html#Run-Time_Type_Information__RTTI_)].

lizan: Not really, a inner_filter_ might not be a instance of StreamFilter when it is only a AccessLog::instance.

Ideally I can avoid all dynamic_cast in this class with a heavy template programming, but I didn't want to go that far at this moment.

lizan:
We got upstream support for per-route config, closing this.

PR #1111, opened 2018-02-13
Update envoy to e2117cb

lizan:
Update Envoy to https://github.com/envoyproxy/envoy/commit/e2117cbf1f7a84e4cfd7a1e00bcc053b1afb0ea7

Align with https://github.com/envoyproxy/envoy/pull/2386

**Release note**:
```release-note
None
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -35,7 +35,7 @@ bind(
 )
 
 # When updating envoy sha manually please update the sha in istio.deps file also
-ENVOY_SHA = "9b42fc27a2d1a1d5fdf7dd7d005660952872d1b6"  # Feb 8, 2018
+ENVOY_SHA = "e2117cbf1f7a84e4cfd7a1e00bcc053b1afb0ea7"  # Feb 12, 2018
 
 http_archive(
     name = "envoy",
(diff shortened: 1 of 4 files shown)


WORKSPACE
```diff
@@ -35,7 +35,7 @@ bind(
 )
 
 # When updating envoy sha manually please update the sha in istio.deps file also
```
rkpagadala: Lizan,
   Please update istio.deps also at the same time.

lizan: Done

lizan:
TSAN failure is likely due to #1124, temporary disable Google gRPC to pass presubmit here, and make another fix later.

prow/proxy-presubmit.sh
```diff
@@ -45,6 +45,8 @@ fi
 
 cd $ROOT
 
+export BAZEL_TEST_ARGS="--test_output=errors --define=google_grpc=disabled"
```
qiwzhang: My concerns is:  test is using different code than release.  This is not testing anything.

qiwzhang: Just realized this is for "auth",  auth is not using grpc so it is fine then.

lizan:
in mixer we are not using Google gRPC either.

PR #1116, opened 2018-02-13
Ignore unknown field when parsing proxy config.

yusuoh:
Ignore unknown protobuf field

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/istio/istio/issues/3428

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/mixer/config.cc
+++ b/src/envoy/mixer/config.cc
@@ -15,7 +15,6 @@
 
 #include "google/protobuf/stubs/status.h"
 #include "google/protobuf/util/json_util.h"
-#include "include/attributes_builder.h"
 #include "src/envoy/mixer/config.h"
 #include "src/envoy/mixer/utils.h"
 
--- a/src/envoy/mixer/config.cc
+++ b/src/envoy/mixer/config.cc
@@ -13,9 +13,11 @@
  * limitations under the License.
  */
 
-#include "src/envoy/mixer/config.h"
 #include "google/protobuf/stubs/status.h"
 #include "google/protobuf/util/json_util.h"
+#include "include/attributes_builder.h"
+#include "src/envoy/mixer/config.h"
+#include "src/envoy/mixer/utils.h"
 
 using ::google::protobuf::Message;
 using ::google::protobuf::util::Status;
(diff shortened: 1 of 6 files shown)


src/envoy/mixer/utils_test.cc
```diff
@@ -0,0 +1,47 @@
...
+#include "src/envoy/mixer/utils.h"
+#include "mixer/v1/config/client/client_config.pb.h"
+#include "test/test_common/utility.h"
+
+using Envoy::Http::Utils::ParseJsonMessage;
+
+namespace {
+
+TEST(UtilsTest, ParseNormalMessage) {
+  std::string config_str =
```
lizan: Use raw string literal (i.e. `R"({...})"` so it is easier to read. JSON keys need `"`s as well.

src/envoy/mixer/config.cc
```diff
@@ -93,8 +94,7 @@ bool ReadV2Config(const Json::Object& json, Message* message) {
     return false;
   }
   std::string v2_str = json.getObject(kV2Config)->asJsonString();
-  Status status =
-      ::google::protobuf::util::JsonStringToMessage(v2_str, message);
+  Status status = Utils::ParseJsonMessage(v2_str, message);
```
lizan: I think we should just use `MessageUtil::loadFromJson` from envoy, the behavior should be consistent with it.

yusuoh: That one is relying on a third-party lib to parse the json. I would rather ensure that the option is set correctly by using the protobuf lib directly, as the logic is simple here and shouldn't involve much overhead.

src/envoy/mixer/utils_test.cc
yusuoh: Done

src/envoy/mixer/config.cc
lizan: No it doesn't use third-party lib to parse:
https://github.com/envoyproxy/envoy/blob/910058d7113e6439d077e2467e26abb7b803000e/source/common/protobuf/utility.cc#L38

Do we see any real problem with unknown fields? If you don't fix upstream, the issue could be still there.

yusuoh: Then it could also be an issue for envoy. I will check the upstream.

yusuoh:
@lizan Could you please LGTM this PR for now to unblock the failed upgrade test? I will investigate the upstream issue separately.

yusuoh:
A little more context, this issue happens during upgrade phase when the envoy proxy receives a JSON config from Pilot which has a different version with the proxy.

PR #1119, opened 2018-02-14
Add back mixerclient stats

JimmyCYJ:
**What this PR does / why we need it**:Revert Mixerclient stats back.
Move Envoy stats into Mixer filter config object.
Keep MixerStatsObject in HttpMixerControl and TcpMixerControl, and pass Envoy stats into MixerStatsObject for periodical stats update.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/envoy/mixer/stats.h
+++ b/src/envoy/mixer/stats.h
@@ -52,10 +52,9 @@ typedef std::function<bool(::istio::mixer_client::Statistics* s)> GetStatsFunc;
 // calls issued by a mixer filter.
 class MixerStatsObject {
  public:
-  MixerStatsObject(Event::Dispatcher& dispatcher, const std::string& name,
-                   Stats::Scope& scope,
+  MixerStatsObject(Event::Dispatcher& dispatcher,
                    ::google::protobuf::Duration update_interval,
-                   GetStatsFunc func);
+                   GetStatsFunc func, MixerFilterStats& stats);
 
  private:
   // This function is invoked when timer event fires.
--- a/src/envoy/mixer/stats.h
+++ b/src/envoy/mixer/stats.h
@@ -52,9 +52,9 @@ typedef std::function<bool(::istio::mixer_client::Statistics* s)> GetStatsFunc;
 // calls issued by a mixer filter.
 class MixerStatsObject {
  public:
-  MixerStatsObject(Event::Dispatcher& dispatcher,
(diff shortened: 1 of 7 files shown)


src/envoy/mixer/stats.h
```diff
@@ -0,0 +1,87 @@
...
+
+typedef std::function<bool(::istio::mixer_client::Statistics* s)> GetStatsFunc;
+
+// MixerStatsObject maintains statistics for number of check, quota and report
+// calls issued by a mixer filter.
+class MixerStatsObject {
+ public:
+  MixerStatsObject(Event::Dispatcher& dispatcher,
+                   ::google::protobuf::Duration update_interval,
+                   GetStatsFunc func, MixerFilterStats& stats);
```
qiwzhang: Put all these "T&" arguments first.   move "stats" right after dispatcher

qiwzhang:
Could you do a manual stress test of Envoy with --concurrent=100 to see if it crash?

src/envoy/mixer/stats.h
JimmyCYJ: Done.

JimmyCYJ:
Please take a look. Thanks!

src/envoy/mixer/stats.h
```diff
@@ -0,0 +1,87 @@
...
+
+ private:
+  // This function is invoked when timer event fires.
+  void OnTimer();
+
+  // Compares old stats with new stats and updates envoy stats.
+  void CheckAndUpdateStats(const ::istio::mixer_client::Statistics& new_stats);
+
+  // A set of Envoy stats for the number of check, quota and report calls.
+  MixerFilterStats stats_;
```
qiwzhang: It is better to use &.   Here you are making a copy.   I am not sure  of a copy of references, or a copy of the stat slots?

JimmyCYJ: It should be a reference, not a copy. Thanks!

ldemailly:
hey, thanks a lot for this - quick question:

what is different between this version and the one that was crashing before ?
why was it crashing before ?

qiwzhang:
The place allocating "stat" struct is different.  "stat" is thread local data.  Before it was allocated on top of another thread local data.  This is the case Envoy doesn't support. Now the change is to allocate it on the Config, which is not thread local data.

PR #1125, opened 2018-02-14
Move mixerclient codes to proxy repo

qiwzhang:

This step is just move the code and make it pass the compiling and can build other components properly.

**Release note**:

```release-note
None
```


lizan:
While I am fine with this change, is this repo re-organize discussed widely?

qiwzhang:
Yes, I think during istio repo re-org.  It was decided that all c++ move into one repo.

qiwzhang:
Here is the Istio repo reorg doc.
https://docs.google.com/document/d/1i1PEwx0vHCD9oR5vg1QJqo_8JfQFhwzOdIYarYdH1NI/edit#heading=h.x9snb54sjlu9

qiwzhang:
It was decided all c++ code moved into one repo

lizan:
/lgtm 

> We should probably also get rid of Bazel in here too at some point.
This will be hard and it is less benefit than that in istio/istio. Unless Envoy switches to another build toolchain. We don't even have a better alternative.

ldemailly:
did you guys coordinate with test and release ?
(the sync bots and the release/build process are getting updated in this PR) ?

cc @hklai @rkpagadala @sebastienvas @mattdelco 

also next time please give a heads up to the oncall too

ldemailly:
also looks like this isn't an import so you lost all history - please consider asking @mandarjog or anyone that helped with mono repo and redoing this migration properly

PR #1131, opened 2018-02-15
Remove unused Dockerfiles and scripts

### LINKED ISSUE #1129: delete docker, deb, iptables related files and scripts which have moved to istio/istio
it's quite confusing that it's been copied and not moved

ditto for iptables stuff, deb stuff etc

nmittler:
The istio/istio Dockerfiles no longer use the Dockerfiles from
this project. The iptables scripts have been moved to istio/istio
as well.
    
Fixes #1129

The istio/istio dependency on these Dockerfiles was removed in https://github.com/istio/istio/pull/3784


ldemailly:
have you diffed this version and the one in istio/istio ? (can you paste the diff here - hopefully no feature was lost in the move)

also can you remove the dockerfile entirely (or is it used?)

nmittler:
> have you diffed this version and the one in istio/istio ? (can you paste the diff here - hopefully no feature was lost in the move)

The new iptables config is done by the `proxy-init` container (not the `istio-proxy` container). The [new script](https://github.com/istio/istio/blob/master/pilot/docker/prepare_proxy.sh) therefore only exists in the `proxy-init` container.  This means that we won't have any of the previous utility scripts on `istio-proxy`. We will, however, have access to the iptables functions since they are installed on the `istio-proxy` image (assuming we've enabled root access on the `istio-proxy` container).  So I'm not sure we've lost much.

> can you remove the dockerfile entirely (or is it used?)

We reference an istio envoy base image from our proxy Dockerfile here:

https://github.com/istio/istio/blob/master/pilot/docker/Dockerfile.proxy#L2

I assume the envoy docker image is produced by these Dockerfiles.

You should note that I'm new to all of this, so take everything I'm saying with an appropriate grain of salt :)

kyessenov:
Last time, Costin forked iptables scripts for VMs for some reason. Apart
from that, I don't have any concerns about moving files around.

ldemailly:
those iptables scripts have 2 usage context we want to unify:

the on VM (mesh expansion) script

the on k8s proxy init container script

The script created for VM mesh expansion is newer and in theory a super set, and it was at some point copied around so there are probably at least 3 copies/variants

before we delete one of the copies I'd like to see a diff

re the docker file, ic, as it is used as a base, you must make sure `proxy-redirection-clear` isn't used or referenced in docs (I'm pretty sure it is referenced for instance
https://github.com/istio/proxy/pull/1131/files#diff-4dcdef193d4ac0269f3d1ba7ea113466R4
)

like any cleanup, the devil is in the long tail of details

docker/Dockerfile.debug
```diff
@@ -1,9 +1,5 @@
 FROM ubuntu:xenial
 # Do not add more stuff to this list that isn't small or critically useful.
-# If you occasionally need something on the container do
```
ldemailly: this is quite useful information, the idea here is not to remove the comment but update it or make sure it still works with the new iptable script or replace with a pointer to new instructions (or maybe move it to the pilot dockerfile or a doc there)

docker/Dockerfile.release
```diff
@@ -3,9 +3,4 @@ RUN apt-get update && \
     apt-get install -y \
     iptables
 ADD envoy /usr/local/bin/envoy
-ADD istio-iptables.sh /usr/local/bin/istio-iptables.sh
-ADD istio-start.sh /usr/local/bin/istio-start.sh
 ADD envoy.json /var/lib/istio/envoy/envoy.json
-
-ENTRYPOINT /usr/local/bin/istio-start.sh
```
ldemailly: wonder what is entry point when unset

docker/Dockerfile.debug
nmittler: @ldemailly since we've moved the iptables work to istio/istio, it would seem that any documentation belongs in those Dockerfiles, not here.  That way, these just become dockerfiles for our own custom envoy deployment and very little more.  WDYT?

nmittler:
@ldemailly any idea what's going on with `mixer_client_lib`? prow doesn't seem happy.

qiwzhang:
Please hold.  I am in the middle of merging mixer client code from other repo

docker/Dockerfile.debug
ldemailly: sounds good, but what's the PR adding them on the other side (sorry if I missed it). I just want to avoid losing the setup/info.

ldemailly:
@costinm I am all for deleting (which is what I asked nate to do and he kindly volunteered to) as long as we first have the PR transferring the information open in istio/istio

nmittler:
@costinm can you verify that my understanding of how all of this current works is correct?

- "proxy" builds the envoy/envoy-debug images which contain proxy's version of `istio-iptables.sh` and `istio-start.sh`. These images have `ENTRYPOINT /usr/local/bin/istio-start.sh`, which runs `istio-iptables.sh` and then starts Envoy.
- The `istio-init` container has its own iptables config script `prepare_proxy.sh`. The init container, which is the first to start up at runtime, will run this script and then exit.
- The `istio-proxy` container, extends the `envoy` container and defines `ENTRYPOINT ["/usr/local/bin/pilot-agent"]`.

So IIUC, the iptables are configured twice: first by `istio-init`, and then again by `istio-proxy`.  This seems  volatile, since the configuration must be identical or strange things could happen.

Does any/all of this sound correct? If so, I think this PR is the solution.

nmittler:
@qiwzhang any update? Can I merge?

nmittler:
@ldemailly 

> @costinm I am all for deleting (which is what I asked nate to do and he kindly volunteered to) as long as we first have the PR transferring the information open in istio/istio

IIUC, what @costinm is saying is that an Istio PR isn't needed since these scripts aren't being used. The scripts already exist in istio/istio and should already be working correctly.  The one thing I'm concerned of is described in [my comment above](https://github.com/istio/proxy/pull/1131#issuecomment-367077453) .. I think we're configuring the iptables twice from 2 different containers.

nmittler:
@geeknoid do the CLA checks work differently on this repo? It seems to be waiting for status to be reported.

nmittler:
@costinm @ldemailly AFAIK, submitting this change should have no impact on tests in istio/istio.  Can you confirm?  And if so, am I free to pull the trigger on this once the CLA bot is happy?

docker/Dockerfile.debug
nmittler: @ldemailly I think this is answered by https://github.com/istio/proxy/pull/1131#pullrequestreview-97355637.  Short answer is: I don't think we need another PR.

nmittler:
@costinm @ldemailly I'm trying out the new proxy build here: https://github.com/istio/istio/pull/3690

qiwzhang:
@costinm @ldemailly could you approve this?  If I have approval from you, I will merge it manually since I don't know how these files are used.

qiwzhang:
@ldemailly circleci test failed is a known issue. not related to this change.

nmittler:
@ldemailly 

> also please link the istio/istio PRs with the matching changes in the description

I'm not sure what you mean.  This PR doesn't have a matching change in istio/istio ... it's pretty independent.

ldemailly:
we used to use all/some of those scripts and images; they now live in istio/istio somewhere and we stopped using the docker image - I just want that when someone steps on deletion they can find the new location easily and track that we are making an ordered change (add/unify in new location first, delete second)

I also want to be able to click to check that the info isn't lost

nmittler:
@ldemailly I've added a link in the description to the istio/istoi PR that removes the dependency

ldemailly:
sgtm

can you also add a link or 2 to the dockerfile and the new location of the unified iptables script

nmittler:
AFAICT, the tsan error is unrelated.  @qiwzhang can you confirm?

PR #1132, opened 2018-02-15
Initialize response code in ReportInfo to 500

### LINKED ISSUE #1126: response.code attribute may use un-initialized value
response.code sometime seen as very big value.

It is possible that it is not initialized.

The code is here

https://github.com/istio/mixerclient/blob/master/control/src/http/attributes_builder.cc#L140

JimmyCYJ:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1126 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/envoy/mixer/http_filter.cc
+++ b/src/envoy/mixer/http_filter.cc
@@ -104,9 +104,10 @@ class Config : public Logger::Loggable<Logger::Id::http> {
     Runtime::RandomGenerator& random = context.random();
     tls_->set([this, &random](Event::Dispatcher& dispatcher)
                   -> ThreadLocal::ThreadLocalObjectSharedPtr {
-      return ThreadLocal::ThreadLocalObjectSharedPtr(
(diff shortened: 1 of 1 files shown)


src/envoy/mixer/http_filter.cc
```diff
@@ -329,6 +329,7 @@ class ReportData : public HttpReportData {
     data->send_bytes = info_.bytesSent();
     data->duration =
         std::chrono::duration_cast<std::chrono::nanoseconds>(info_.duration());
+    data->response_code = 500;
```
qiwzhang: please add comment
// responseCode is for the backend response. If it is not valid, the request is rejected by Envoy
// set the response code for such requests as 500.

JimmyCYJ: Done. Thanks!

PR #1152, opened 2018-02-22
Move TCP filter Check() call under onData().

JimmyCYJ:
**What this PR does / why we need it**:Currently, TCP filter makes a Check() call too early so that SSL handshake is not done at that moment. Delay the Check() call to onData() so that we can collect information about SSL connection.

This PR also adds a stats generator that keeps the stats creation consistent with other stats in Envoy.

**Release note**:

```release-note
NONE
```


JimmyCYJ:
code re-org [PR](https://github.com/istio/proxy/pull/1153) is merged.

PR #1166, opened 2018-02-22
Send unique connection

JimmyCYJ:
**What this PR does / why we need it**: Let TCP filter send connection.id attribute to Mixer in Check() and Report() calls. This connection.id is a downstream connection id, and its format is <uuid>-<connection id>. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/envoy/tcp/mixer/tcp_filter.cc
+++ b/src/envoy/tcp/mixer/tcp_filter.cc
@@ -51,6 +51,7 @@ class TcpConfig : public Logger::Loggable<Logger::Id::filter> {
   TcpMixerConfig mixer_config_;
   ThreadLocal::SlotPtr tls_;
   Utils::MixerFilterStats stats_;
+  // UUID of the Envoy TCP mixer filter.
   const std::string uuid_;
 
  public:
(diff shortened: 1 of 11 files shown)


src/envoy/tcp/mixer/tcp_filter.cc
```diff
@@ -253,6 +260,16 @@ class TcpInstance : public Network::Filter,
     data->duration = std::chrono::duration_cast<std::chrono::nanoseconds>(
         std::chrono::system_clock::now() - start_time_);
   }
+
+  std::string GetConnectionId() const override {
+    char connection_id_str[32];
+    memset(connection_id_str, 0, sizeof(connection_id_str));
+    StringUtil::itoa(connection_id_str, 32,
```
qiwzhang: if we can assume itoa append 0 at the end of string.  we don't need "memset".  could you double check?

src/envoy/tcp/mixer/tcp_filter.cc
```diff
@@ -89,6 +94,7 @@ class TcpInstance : public Network::Filter,
 
   istio::mixerclient::CancelFunc cancel_check_;
   TcpMixerControl& mixer_control_;
+  const std::string uuid_;
```
qiwzhang: not to copy uuid for each connection,  maybe let mixer_control keep mixer_config reference, and directly use the one in config.

src/envoy/tcp/mixer/tcp_filter.cc
JimmyCYJ: Yes, 0 is appended. https://github.com/envoyproxy/envoy/blob/master/source/common/common/utility.cc#L183
memset is removed.

include/control/tcp/report_data.h
```diff
@@ -39,6 +39,9 @@ class ReportData {
     std::chrono::nanoseconds duration;
   };
   virtual void GetReportInfo(ReportInfo* info) const = 0;
+
+  // Get downstream tcp connection id.
+  virtual std::string GetConnectionId() const = 0;
```
qiwzhang: We should set it in Check call

PR #1168, opened 2018-02-22
Add a filter for prototyping authentication policy

lei-tang:
Add a filter for prototyping the implementation of the Istio authentication policy. It currently does:
-- Validate the authentication policy json config with the authentication policy proto.
-- Parse the peer mTLS config and output in log whether requiring mTLS for the connection to peer.

**Release note**:

```release-note
```


src/envoy/http/authn/authn_store.h
```diff
@@ -0,0 +1,70 @@
+/* Copyright 2018 Istio Authors. All Rights Reserved.
```
diemtvu: Can we start with smt simpler, i.e without this config store. I think the caching should be later, per (authn) method. Also, I try to see if we can use the same proto in the API repo as is.  You can take a look at my attempt [here](https://github.com/diemtvu/proxy/tree/policy/src/envoy/http/authn).

src/envoy/http/authn/http_filter.h
```diff
@@ -0,0 +1,55 @@
...
+ public:
+  AuthnFilter(Upstream::ClusterManager& cm, Auth::AuthnStore& store);
+  ~AuthnFilter();
+
+  // Http::StreamFilterBase
+  void onDestroy() override;
+
+  // Http::StreamDecoderFilter
+  FilterHeadersStatus decodeHeaders(HeaderMap& headers, bool) override;
+  FilterDataStatus decodeData(Buffer::Instance&, bool) override;
```
diemtvu: Just curious, do you know if all these decodeXXX needed for the filter, or we just need to overwrite decodeHeaders?

src/envoy/http/authn/http_filter.h
```diff
@@ -0,0 +1,55 @@
...
+#include "src/envoy/http/authn/authn_store.h"
+
+#include "common/common/logger.h"
+#include "server/config/network/http_connection_manager.h"
+
+namespace Envoy {
+namespace Http {
+
+// The authentication filter.
+class AuthnFilter : public StreamDecoderFilter,
```
diemtvu: Let's try not to use abbreviation. Name this AuthenticationFilter

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,85 @@
...
+                          public Logger::Loggable<Logger::Id::http> {
+ public:
+  HttpFilterFactoryCb createFilterFactory(const Json::Object& config,
+                                          const std::string&,
+                                          FactoryContext& context) override {
+    istio::authentication::v1alpha1::Policy proto_config;
+
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+
+    MessageUtil::loadFromJson(config.asJsonString(), proto_config);
```
diemtvu: Use Utils::ParseJsonMessage then you can simply link to the proto defined in api repo instead of copying them. I believe this function simply has some extra validation, which require to envoy validation.proto. We can ignore all that for now.

src/envoy/http/authn/authn_store.h
```diff
@@ -0,0 +1,70 @@
...
+#include "envoy/thread_local/thread_local.h"
+#include "src/envoy/http/authn/policy.pb.validate.h"
+
+namespace Envoy {
+namespace Http {
+namespace Auth {
+
+// The authentication store stores config and caches.
+// It is per-thread and stored in thread local.
+class AuthnStore : public ThreadLocal::ThreadLocalObject {
```
qiwzhang: Do you have cache?  If you don't have cache, you don't need this per-thread object.  just directly use config.

src/envoy/http/authn/http_filter.cc
```diff
@@ -0,0 +1,76 @@
...
+
+FilterHeadersStatus AuthnFilter::decodeHeaders(HeaderMap&, bool) {
+  ENVOY_LOG(debug, "Called AuthnFilter : {}", __func__);
+  state_ = HandleHeaders;
+
+  const ::istio::authentication::v1alpha1::Policy& config = store_.config();
+  int peer_size = config.peers_size();
+  ENVOY_LOG(debug, "AuthnFilter: {} config.peers_size()={}", __func__,
+            peer_size);
+  for (int i = 0; i < peer_size; i++) {
```
qiwzhang: Do we need to go through each peer?

We may need to pre-process your config to optimize the authn check per request.  We are in data path,  we should be as fast as possible.

src/envoy/http/authn/http_filter.h
lei-tang: decodeXXX are pure virtual functions from the base class so they are implemented, which simply logs the events for diagnosis. We can probably place all authentication logic in decodeHeaders().

src/envoy/http/authn/http_filter.h
lei-tang: Done.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done.

src/envoy/http/authn/authn_store.h
lei-tang: Done: cache and per thread storage have been removed. And the code now directly uses config.

src/envoy/http/authn/http_filter.cc
lei-tang: Done. In most cases, only one peer exists. For faster speed, the code has been revised to only go through the first peer.

repositories.bzl
```diff
@@ -185,6 +185,53 @@ filegroup(
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+#
+load("@com_google_protobuf//:protobuf.bzl", "cc_proto_library")
+
+cc_proto_library(
```
qiwzhang: just add this into line 164, use one git_reposity for api repo

src/envoy/http/authn/BUILD
```diff
@@ -0,0 +1,56 @@
...
+################################################################################
+#
+
+package(default_visibility = ["//visibility:public"])
+
+load(
+    "@envoy//bazel:envoy_build_system.bzl",
+    "envoy_cc_binary",
+    "envoy_cc_library",
+    "envoy_cc_test",
```
qiwzhang: you don't need test, binary

src/envoy/http/authn/BUILD
```diff
@@ -0,0 +1,56 @@
...
+)
+
+load(
```
qiwzhang: you don't need this api_proto_library

src/envoy/http/authn/sample/authn.conf
```diff
@@ -0,0 +1,122 @@
...
+            },
+            "access_log": [
+              {
+                "path": "/dev/stdout"
+              }
+            ],
+            "filters": [
+              {
+                "type": "decoder",
+                "name": "authN",
```
qiwzhang: how about call it "istio_authn"

src/envoy/http/authn/sample/authn.conf
```diff
@@ -0,0 +1,122 @@
...
+                "config": {
+                   "destinations": [
```
qiwzhang: I don't see desinations field is used

src/envoy/http/authn/sample/authn.conf
```diff
@@ -0,0 +1,122 @@
...
+                            "issuer": "628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com",
+                            "jwks_uri": "http://localhost:8081/"
+                         }
+                      }
+                   ]
+                }
+              },
+              {
+                "type": "decoder",
+                "name": "jwt-auth",
```
qiwzhang: you don't need to add this filter.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,94 @@
...
+class AuthnFilterConfig : public NamedHttpFilterConfigFactory,
+                          public Logger::Loggable<Logger::Id::http> {
+ public:
+  HttpFilterFactoryCb createFilterFactory(const Json::Object& config,
+                                          const std::string&,
+                                          FactoryContext& context) override {
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+
+    istio::authentication::v1alpha1::Policy* ptr =
+        new istio::authentication::v1alpha1::Policy();
```
qiwzhang: memory leak.
please use unique_ptr

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,94 @@
...
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+    return ProtobufTypes::MessagePtr{
+        new istio::authentication::v1alpha1::Policy};
+  }
+
+  std::string name() override { return "authN"; }
+
+ private:
+  HttpFilterFactoryCb createFilter(
+      const istio::authentication::v1alpha1::Policy* ptr,
```
qiwzhang: per style guide,  use const T&

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,94 @@
...
+      FactoryContext& context) {
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+
+    Upstream::ClusterManager& cm = context.clusterManager();
+
+    return [&cm, ptr](Http::FilterChainFactoryCallbacks& callbacks) -> void {
+      callbacks.addStreamDecoderFilter(
+          std::make_shared<Http::AuthenticationFilter>(
```
qiwzhang: share object should be created outside of lambda, otherwise it is NOT shared.

src/envoy/http/authn/sample/authn.conf
diemtvu: +1 Or istio.authn

src/envoy/http/authn/BUILD
```diff
@@ -0,0 +1,56 @@
...
+    "@envoy_api//bazel:api_build_system.bzl",
+    "api_proto_library",
+)
+
+envoy_cc_library(
```
diemtvu: You can bundle both libs together. I imagine they are unlikely need to be used separately.

src/envoy/http/authn/http_filter.cc
```diff
@@ -0,0 +1,81 @@
...
+  ENVOY_LOG(
+      debug,
+      "Called AuthenticationFilter : {}, return FilterHeadersStatus::Continue;",
+      __func__);
+  return FilterHeadersStatus::Continue;
+}
+
+FilterDataStatus AuthenticationFilter::decodeData(Buffer::Instance&, bool) {
+  ENVOY_LOG(debug, "Called AuthenticationFilter : {}", __func__);
+  state_ = HandleData;
```
diemtvu: Do we need to keep track of state_ ? I don't see how it's used in this PR.

src/envoy/http/authn/http_filter.h
diemtvu: Probably just put definition for decodeData & decodeTrailer here to keep it clear that they are not needed for authentication process.

src/envoy/http/authn/http_filter.h
```diff
@@ -0,0 +1,57 @@
...
+  FilterHeadersStatus decodeHeaders(HeaderMap& headers, bool) override;
+  FilterDataStatus decodeData(Buffer::Instance&, bool) override;
+  FilterTrailersStatus decodeTrailers(HeaderMap&) override;
+  void setDecoderFilterCallbacks(
+      StreamDecoderFilterCallbacks& callbacks) override;
+
+ private:
+  // The callback funcion.
+  StreamDecoderFilterCallbacks* decoder_callbacks_;
+  Upstream::ClusterManager& cm_;
```
diemtvu: Do we need this?

src/envoy/http/authn/http_filter.h
```diff
@@ -0,0 +1,57 @@
...
+  // Store the config.
+  std::shared_ptr<const istio::authentication::v1alpha1::Policy> config_;
```
diemtvu: I'm not sure it's a good idea to use shared_ptr for the config. At least use unique_ptr instead as it shouldn't be owned by more than one. @qiwzhang may have better idea of what type (copy object, ref, pointer etc) should be used

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,94 @@
...
+    istio::authentication::v1alpha1::Policy* ptr =
+        new istio::authentication::v1alpha1::Policy();
+    google::protobuf::util::Status status =
+        Utils::ParseJsonMessage(config.asJsonString(), ptr);
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : Utils::ParseJsonMessage()");
+    if (status.ok()) {
+      return createFilter(ptr, context);
+    } else {
+      ENVOY_LOG(debug, "Utils::ParseJsonMessage() return value is NOT ok!");
+      return nullptr;
```
diemtvu: Does the caller expect null is a legit return? Is throwing exception here is acceptable? + @qiwzhang

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,94 @@
...
+    return createFilter(&policy, context);
+  }
+
+  ProtobufTypes::MessagePtr createEmptyConfigProto() override {
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+    return ProtobufTypes::MessagePtr{
+        new istio::authentication::v1alpha1::Policy};
+  }
+
+  std::string name() override { return "authN"; }
```
diemtvu: Make a constant string for name.

repositories.bzl
lei-tang: Done.

src/envoy/http/authn/BUILD
lei-tang: Done.

repositories.bzl
lei-tang: Done.

src/envoy/http/authn/BUILD
lei-tang: Removed test and binary.,

src/envoy/http/authn/BUILD
lei-tang: Removed api_proto_library.

src/envoy/http/authn/sample/authn.conf
lei-tang: Agree, done.

src/envoy/http/authn/sample/authn.conf
lei-tang: destinations field has been removed.

src/envoy/http/authn/sample/authn.conf
lei-tang: This jwt-auth filter has been removed.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done. Has been revised based on the discussion.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done. Has been revised based on the discussion.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done. Has been revised based on the discussion.

src/envoy/http/authn/BUILD
lei-tang: Done.

src/envoy/http/authn/http_filter.cc
lei-tang: Done, state_ has been removed.

src/envoy/http/authn/http_filter.h
lei-tang: Done, decoder_callbacks_ and cm_ have been removed.

src/envoy/http/authn/http_filter.h
lei-tang: shared_ptr has been removed based on discussion with Wayne.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done.

src/envoy/http/authn/http_filter.h
lei-tang: decodeData and decodeTrailer only does return FilterDataStatus::Continue.

src/envoy/http/authn/BUILD
```diff
@@ -0,0 +1,40 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+load(
+    "@envoy//bazel:envoy_build_system.bzl",
+    "envoy_cc_library",
+)
+
+envoy_cc_library(
+    name = "http_filter_factory",
```
qiwzhang: call it filter_lib

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  ProtobufTypes::MessagePtr createEmptyConfigProto() override {
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+    return ProtobufTypes::MessagePtr{
+        new istio::authentication::v1alpha1::Policy};
+  }
+
+  std::string name() override { return std::string(name_); }
+
```
qiwzhang: name_ is already string type, why do you neet casting?

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  HttpFilterFactoryCb createFilterFactoryFromProto(
+      const Protobuf::Message& proto_config, const std::string&,
+      FactoryContext& context) override {
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+
+    const istio::authentication::v1alpha1::Policy& policy =
+        dynamic_cast<const istio::authentication::v1alpha1::Policy&>(
+            proto_config);
```
qiwzhang: you need to copy the data over

policy_ = policy;

Should not use the the one pass-in to you

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Has been changed to throw exception based on the discussion with Wayne.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,96 @@
...
+    Upstream::ClusterManager& cm = context.clusterManager();
+
+    return
+        [&cm, &policy](Http::FilterChainFactoryCallbacks& callbacks) -> void {
+          callbacks.addStreamDecoderFilter(
+              std::make_shared<Http::AuthenticationFilter>(cm, policy));
+        };
+  }
+
+  const std::string name_ = "istio_authn";
```
diemtvu: Doesn't have to be member of the class. You can define it in a no-name namespace top of the file.

diemtvu: Also use naming convention for constant, like kFilterName.

src/envoy/http/authn/http_filter.h
diemtvu: I know. I mean it simpler enough to just put it right here in the header, so we don't have to think about it. Anyway, it's just  my personal preference, no big deal.

src/envoy/http/authn/BUILD
lei-tang: Done.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: casting has been removed.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done.

src/envoy/http/authn/http_filter.cc
```diff
@@ -0,0 +1,77 @@
...
+ * limitations under the License.
+ */
+
+#include "src/envoy/http/authn/http_filter.h"
+
+namespace Envoy {
+namespace Http {
+
+AuthenticationFilter::AuthenticationFilter(
+    Upstream::ClusterManager&,
```
diemtvu: Do we even need this here?

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -23,19 +23,24 @@ namespace Envoy {
 namespace Server {
 namespace Configuration {
 
+namespace {
+// The name for the mixer server cluster.
```
diemtvu: .. name for istio authentication filter.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done.

src/envoy/http/authn/http_filter.cc
lei-tang: ClusterManager has been removed.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -23,19 +23,24 @@ namespace Envoy {
...
     ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
 
     google::protobuf::util::Status status =
         Utils::ParseJsonMessage(config.asJsonString(), &policy_);
     ENVOY_LOG(debug, "Called AuthnFilterConfig : Utils::ParseJsonMessage()");
     if (status.ok()) {
-      return createFilter(policy_, context);
+      return createFilter(policy_);
     } else {
       ENVOY_LOG(debug, "Utils::ParseJsonMessage() return value is NOT ok!");
```
diemtvu: Use error or critical. May be even better to use PANIC macro instead.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Fixed.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Changed to use PANIC.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -42,7 +42,7 @@ class AuthnFilterConfig : public NamedHttpFilterConfigFactory,
     if (status.ok()) {
       return createFilter(policy_);
     } else {
-      ENVOY_LOG(debug, "Utils::ParseJsonMessage() return value is NOT ok!");
+      PANIC("Utils::ParseJsonMessage() return value is NOT ok!");
```
diemtvu: One more small request, can you also put status code and/or message to the panic text.

lei-tang: Done.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -42,7 +42,7 @@ class AuthnFilterConfig : public NamedHttpFilterConfigFactory,
     if (status.ok()) {
       return createFilter(policy_);
     } else {
-      PANIC("Utils::ParseJsonMessage() return value is NOT ok!");
+      PANIC("Utils::ParseJsonMessage() return value is: " + status.ToString());
       throw EnvoyException(
```
diemtvu: Miss this in the last review :). I don't think we need both panic and throw exception. Panic already do it, I think, though I'm not 100% sure. If you want to be fully certain, LOG error/critical and throw exception.

lei-tang: Based on the following defintion, PANIC is doing log only. So we should be OK.
 ENVOY_LOG_TO_LOGGER(Envoy::Logger::Registry::getLog(Envoy::Logger::Id::assert), critical,        \
                      "panic: {}", X);

lei-tang: I take it back. PANIC also does abort().

lei-tang: #define PANIC(X)                                                                                   \
  ENVOY_LOG_TO_LOGGER(Envoy::Logger::Registry::getLog(Envoy::Logger::Id::assert), critical,        \
                      "panic: {}", X);                                                             \
  abort();

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,98 @@
...
+                                          FactoryContext&) override {
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+
+    google::protobuf::util::Status status =
+        Utils::ParseJsonMessage(config.asJsonString(), &policy_);
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : Utils::ParseJsonMessage()");
+    if (status.ok()) {
+      return createFilter(policy_);
+    } else {
+      PANIC("Utils::ParseJsonMessage() return value is: " + status.ToString());
```
qiwzhang: not, panic,  a wrong config should not crash Envoy.   an invalid config will be just discarded by Envoy.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -0,0 +1,98 @@
...
+      FactoryContext&) override {
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+
+    const istio::authentication::v1alpha1::Policy& policy =
+        dynamic_cast<const istio::authentication::v1alpha1::Policy&>(
+            proto_config);
+
+    policy_ = policy;
+
+    return createFilter(policy_);
```
qiwzhang: not need to pass policy_ since it is a member already

src/envoy/http/authn/http_filter_factory.cc
lei-tang: I have revised it to LOG critical and throw exception.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: PANIC has been replaced with log critical.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done.

PR #1189, opened 2018-03-08
Add a POC for Istio authentication to call jwt_auth as is

lei-tang:
**What this PR does / why we need it**: This is a POC for Istio authentication policy to use jwt_auth for authenticating JWT defined in Istio authentication policy without changing the code of jwt_auth.
The key idea is to convert a JWT defined in Istio authentication policy to a JWT defined in jwt_auth and call jwt_auth as is to authenticate the JWT defined in the Istio authentication policy.

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/authn/jwt_authn_store.h
+++ b/src/envoy/http/authn/jwt_authn_store.h
@@ -67,6 +67,9 @@ class JwtAuthnFactoryStore : public Logger::Loggable<Logger::Id::config> {
 
   // Add a JWT config to the store.
   void addToStore(const ::istio::authentication::v1alpha1::Jwt &jwt) {
+    // TODO(lei-tang): it may be ok to use issuer as the key but need to make
+    // sure
+    // no cases of multiple Jwt have the same issuer
     if (jwt_store_.find(jwt) != jwt_store_.end()) {
       ENVOY_LOG(debug, "{}: AuthFilterConfig exists already", __FUNCTION__);
       return;
@@ -85,14 +88,16 @@ class JwtAuthnFactoryStore : public Logger::Loggable<Logger::Id::config> {
 
   // Store the JwtAuthStoreFactory objects in a map.
   // The key is JWT.
-  // Todo: may only need to use the issuer as the key.
+  // TODO(lei-tang): it may be ok to use issuer as the key but need to make sure
+  // no cases of multiple Jwt have the same issuer
   JwtToAuthFactoryMap jwt_store_{};
 
   // Store the JwtAuthStore objects in a map.
   // Generated from JwtAuthStoreFactory at run time due to the thread local
   // nature.
   // The key is JWT.
-  // Todo: may only need to use the issuer as the key.
+  // TODO(lei-tang): it may be ok to use issuer as the key but need to make sure
+  // no cases of multiple Jwt have the same issuer
   JwtToAuthStoreMap jwt_store_tls_{};
 };
 
--- a/src/envoy/http/authn/jwt_authn_utils.cc
+++ b/src/envoy/http/authn/jwt_authn_utils.cc
@@ -26,7 +26,8 @@ const std::string kJwtClusterName("example_issuer");
 
 // Get the Jwks URI for Envoy cluster
 const std::string getJwksUriEnvoyCluster() {
-  // Todo: Pilot needs to put a cluster field in the Istio authn JWT config.
+  // TODO(lei-tang): Pilot needs to put a cluster field in the Istio authn JWT
+  // config.
   // Before such field is added to the Istio authn JWT config,
   // it is temporarily hard-coded.
   return kJwtClusterName;
@@ -36,7 +37,7 @@ const std::string getJwksUriEnvoyCluster() {
 void convertJwtAuthFormat(
     const ::istio::authentication::v1alpha1::Jwt& jwt_authn,
     Http::JwtAuth::Config::AuthFilterConfig* proto_config) {
-  // Todo: when istio-authn::jwt diverges from jwt_auth::jwt,
+  // TODO(lei-tang): when istio-authn::jwt diverges from jwt_auth::jwt,
   // may need to convert more fields.
   auto jwt = proto_config->add_jwts();
   MessageUtil::jsonConvert(jwt_authn, *jwt);
--- a/src/envoy/http/authn/origin_authenticator.cc
+++ b/src/envoy/http/authn/origin_authenticator.cc
@@ -56,17 +56,6 @@ void OriginAuthenticator::run() {
     }
     return;
   }
-  // Note to explain the change (the note will be removed after the code review)
-  // The callback variable in runMethod() is a temporary variable and will be
-  // de-allocated after runMethod() returns.
-  // When the callback is later invoked on the reference of the temporary
-  // callback variable, it causes segmentation fault.
-  //
-  //  runMethod(credential_rule_.origins(0), [this](const Payload* payload,
-  //                                                bool success) {
-  //    onMethodDone(payload, success);
-  //  });
-  // Changed into the following:
   callbackForRunMethod.reset(new AuthenticatorBase::MethodDoneCallback(
       [this](const Payload* payload, bool success) {
         onMethodDone(payload, success);
--- a/src/envoy/http/authn/origin_authenticator.h
+++ b/src/envoy/http/authn/origin_authenticator.h
@@ -50,7 +50,9 @@ class OriginAuthenticator : public AuthenticatorBase {
   // been tried.
   int method_index_{0};
 
-  // Callback invoked by runMethod
+  // Callback invoked by runMethod, will deallocate the MethodDoneCallback
+  // object
+  // when out of scope.
   std::unique_ptr<AuthenticatorBase::MethodDoneCallback> callbackForRunMethod;
 };
 


src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -74,12 +80,48 @@ class AuthnFilterConfig : public NamedHttpFilterConfigFactory,
...
+      const ::istio::authentication::v1alpha1::OriginAuthenticationMethod& m =
+          policy.credential_rules()[0].origins()[0];
+      if (m.has_jwt()) {
+        // In POC, only the following fields are converted.
+        // Todo: may need to convert more fields if necessary
+        Http::JwtAuth::Config::JWT jwt;
+        jwt.set_issuer(m.jwt().issuer());
+        jwt.set_jwks_uri(m.jwt().jwks_uri());
+        jwt.set_jwks_uri_envoy_cluster(kJwtClusterName);
+        auto jwts = proto_config->add_jwts();
```
qiwzhang: I think two "JWT" proto are very close,  you can use Envoy  MessageUtil::jsonConvert to convert them

MessageUtil::jsonConvert(m.jwt(),  *proto_config->add_jwts());

lei-tang: Done.

diemtvu: As a mater of fact, I'm not sure if those protos will be kept aligned in the future. Doing manual re-construct the proto may be a safer in the long run.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -74,12 +80,48 @@ class AuthnFilterConfig : public NamedHttpFilterConfigFactory,
   std::string name() override { return kAuthnFactoryName; }
 
  private:
-  HttpFilterFactoryCb createFilter() {
+  // Convert istio-authn::jwt to jwt_auth::jwt in protobuf format.
+  void convertJwtAuthFormat(
+      istio::authentication::v1alpha1::Policy& policy,
```
diemtvu: This function should simply take policy.Jwt instead of the whole policy. A policy can have more than one Jwt, in both peers and/or origins.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -74,12 +80,48 @@ class AuthnFilterConfig : public NamedHttpFilterConfigFactory,
...
+        jwt.set_jwks_uri_envoy_cluster(kJwtClusterName);
+        auto jwts = proto_config->add_jwts();
+        jwts->CopyFrom(jwt);
+      }
+    }
+  }
+
+  HttpFilterFactoryCb createFilter(FactoryContext& context) {
     ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+    Http::JwtAuth::Config::AuthFilterConfig proto_config;
```
diemtvu: I'm not sure your direction with this. This jwt_auth should be created per Jwt method, when it actually activated. I don't think it should be initialized here. I would recommend to wait for my PR (https://github.com/istio/proxy/pull/1171) and do it as part of the validateJwt function.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -94,14 +94,11 @@ class AuthnFilterConfig : public NamedHttpFilterConfigFactory,
...
-        jwt.set_issuer(m.jwt().issuer());
-        jwt.set_jwks_uri(m.jwt().jwks_uri());
-        jwt.set_jwks_uri_envoy_cluster(kJwtClusterName);
-        auto jwts = proto_config->add_jwts();
-        jwts->CopyFrom(jwt);
+        // Todo: when istio-authn::jwt diverges from jwt_auth::jwt,
+        // may need to convert more fields.
+        MessageUtil::jsonConvert(m.jwt(), *proto_config->add_jwts());
+        proto_config->mutable_jwts()->Mutable(0)->set_jwks_uri_envoy_cluster(
+            kJwtClusterName);
```
qiwzhang: this is too log.  you should create a variable then

auto jwt = proto_config->add_jwts();

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -82,32 +82,27 @@ class AuthnFilterConfig : public NamedHttpFilterConfigFactory,
...
+    // In POC, only inspect the first credential_rule
+    if (policy_.credential_rules_size() > 0 &&
+        policy_.credential_rules()[0].origins_size() > 0) {
+      auto m = policy_.credential_rules()[0].origins()[0];
+      if (m.has_jwt()) {
+        convertJwtAuthFormat(m.jwt(), &proto_config);
+      }
+    }
 
     std::shared_ptr<Http::JwtAuth::JwtAuthStoreFactory> jwt_store_factory =
```
qiwzhang: you should not create jwt_store_factory if proto_config is empty.

lei-tang: Done.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: I have tried creating jwt_auth per Jwt in decodeHeaders() but it does not work because: with the current implementation of JwtAuthStoreFactory, JwtAuthStoreFactory must be created in the main thread (the thread that creates filter) and the only way to pass a JWT config to JwtAuthStoreFactory is when calling the constructor of JwtAuthStoreFactory.
Due to the above reason, the JWT config variable has to be placed in the thread that creates the filter, then be passed into JwtAuthStoreFactory through its constructor.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: When these protos diverge, we can change to manually re-construct. A to-do item has been added to the conversion function.

src/envoy/http/authn/http_filter_factory.cc
diemtvu: Hmm, then it would be a problem. You then will need to create all possible JwtAuthStore objects that might be needed for the policy and pass them all to the filter. Not sure what are the cost involve. Alternative is having a 'store' create at main thread, but the store is able to serve any jwt config. I haven't read through the whole jwt library myself so not sure what is the right approach. Anyway, let's discuss further with @qiwzhang tomorrow.

src/envoy/http/authn/http_filter.cc
```diff
@@ -78,6 +85,44 @@ FilterHeadersStatus AuthenticationFilter::decodeHeaders(HeaderMap&, bool) {
...
 
+  int origins_size = 0;
+  // In POC, only inspect the first credential_rule
+  if (config_.credential_rules_size() > 0 &&
+      config_.credential_rules()[0].origins_size() > 0) {
+    origins_size = config_.credential_rules()[0].origins_size();
+  }
+  ENVOY_LOG(debug, "AuthenticationFilter: {} config.origins_size()={}",
+            __func__, origins_size);
+  if (origins_size > 0) {
```
qiwzhang: this is not need:  jwt_store_factory is null

src/envoy/http/authn/http_filter.cc
```diff
@@ -14,23 +14,30 @@
...
 namespace Envoy {
 namespace Http {
 
 AuthenticationFilter::AuthenticationFilter(
-    const istio::authentication::v1alpha1::Policy& config)
-    : config_(config) {}
+    const istio::authentication::v1alpha1::Policy& config,
+    Upstream::ClusterManager& cm,
+    std::shared_ptr<Http::JwtAuth::JwtAuthStoreFactory> jwt_store_factory)
+    : config_(config), cm_(cm), jwt_store_factory_(jwt_store_factory) {}
```
qiwzhang: I suggest you pass in  

JwtAuthStore*

I don't like to pass in factory into the filter code.

lei-tang: Done.

src/envoy/http/authn/http_filter.cc
lei-tang: Done.

lei-tang:
@qiwzhang @diemtvu : Based on the approach discussed last time, I have added to the POC the support of passing multiple JWT configs. Please review it.

src/envoy/http/authn/http_filter.h
```diff
@@ -15,18 +15,25 @@
...
 // The authentication filter.
 class AuthenticationFilter : public StreamDecoderFilter,
-                             public Logger::Loggable<Logger::Id::filter> {
+                             public JwtAuth::JwtAuthenticator::Callbacks,
+                             public Logger::Loggable<Logger::Id::http> {
  public:
-  AuthenticationFilter(const istio::authentication::v1alpha1::Policy& config);
+  AuthenticationFilter(const istio::authentication::v1alpha1::Policy& config,
+                       Upstream::ClusterManager& cm,
+                       IstioAuthn::JwtMultiFactoryStore& jwt_store);
```
diemtvu: Should this be a map/vector of JwtAuthStore, instead of Factory?

src/envoy/http/authn/jwt_authn_store.h
```diff
@@ -0,0 +1,82 @@
...
+#include "src/envoy/http/jwt_auth/config.pb.h"
+
+namespace Envoy {
+namespace Http {
+namespace IstioAuthn {
+
+enum JwtStoreType { PEER_STORE = 0, ORIGIN_STORE = 1 };
+
+typedef std::map<JwtStoreType,
+                 std::vector<Envoy::Http::JwtAuth::JwtAuthStoreFactory>>
```
diemtvu: I think this should be JwtAuthStore, not factory.

src/envoy/http/authn/jwt_authn_store.h
```diff
@@ -0,0 +1,82 @@
...
+#include "src/envoy/http/jwt_auth/auth_store.h"
+#include "src/envoy/http/jwt_auth/config.pb.h"
+
+namespace Envoy {
+namespace Http {
+namespace IstioAuthn {
+
+enum JwtStoreType { PEER_STORE = 0, ORIGIN_STORE = 1 };
+
+typedef std::map<JwtStoreType,
```
diemtvu: This could just be hash_map<string, JwtAuthStore>. In other words, I don't think we need to separate the type; if config are the same, they can use the same auth store.

About key, I'm not sure we need to use the whole proto, or just issuer is good enough (with assumption that jwt config for the same issuer should be the same). Leave a TODO for more discussion on this.

src/envoy/http/authn/jwt_authn_store.h
```diff
@@ -0,0 +1,82 @@
...
+    ENVOY_LOG(debug, "{}: added a JwtAuthStoreFactory to the type {}",
+              __FUNCTION__, type);
+  }
+
+ private:
+  // Store the FactoryContext object reference
+  Server::Configuration::FactoryContext &context_;
+
+  // Store AuthFilterConfig as string
+  std::map<JwtStoreType, std::set<std::string>> config_{};
```
diemtvu: I don't think we need this if store is the map itself.

src/envoy/http/authn/http_filter.cc
```diff
@@ -14,23 +14,28 @@
...
 void AuthenticationFilter::onDestroy() {
   ENVOY_LOG(debug, "Called AuthenticationFilter : {}", __func__);
+  if (jwt_auth_ != nullptr) {
+    jwt_auth_->onDestroy();
+  }
 }
 
-FilterHeadersStatus AuthenticationFilter::decodeHeaders(HeaderMap&, bool) {
+FilterHeadersStatus AuthenticationFilter::decodeHeaders(HeaderMap& headers,
+                                                        bool) {
```
diemtvu: Please rebase so you can hook it up to the proper place.

src/envoy/http/authn/http_filter_factory.cc
```diff
@@ -74,12 +79,38 @@ class AuthnFilterConfig : public NamedHttpFilterConfigFactory,
...
+    jwt->set_jwks_uri_envoy_cluster(kJwtClusterName);
+  }
 
-    return [&](Http::FilterChainFactoryCallbacks& callbacks) -> void {
+  HttpFilterFactoryCb createFilter(FactoryContext& context) {
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+    std::shared_ptr<Http::IstioAuthn::JwtAuthnFactoryStore> jwt_factory_store =
+        std::make_shared<Http::IstioAuthn::JwtAuthnFactoryStore>(context);
+    // In POC, only inspect the first credential_rule
+    if (policy_.credential_rules_size() > 0 &&
```
diemtvu: You'll need to iterate through all credential rules and methods, as well as peers' methods.

src/envoy/http/authn/http_filter.h
lei-tang: Because JwtAuthStoreFactory has included the thread local related implementations whereas JwtAuthStore does not have thread local related implementations, this should be a map/vector of JwtAuthStoreFactory, instead of a map/vector of JwtAuthStore.

src/envoy/http/authn/jwt_authn_store.h
lei-tang: Because JwtAuthStoreFactory has included the thread local related implementations whereas JwtAuthStore does not have thread local related implementations, this should be a map of JwtAuthStoreFactory, instead of a map of JwtAuthStore.

src/envoy/http/authn/http_filter.h
diemtvu: Look at http_filter from jwt_auth (https://github.com/istio/proxy/blob/master/src/envoy/http/jwt_auth/http_filter.h#L31), it takes JwtAuthStore object (ref). JwtAuthStoreFactory is created in http_filter_factory, but is used only to pass the store(). I don't see why do you need to the whole (array of) factory here.

src/envoy/http/authn/jwt_authn_store.h
```diff
@@ -0,0 +1,82 @@
...
+      return;
+    }
+    if (config_.find(type) == config_.end()) {
+      // Add set of AuthFilterConfig as strings for the given type
+      config_[type] = std::set<std::string>();
+    }
+    // Add config_str to the set
+    config_[type].insert(config_str);
+    if (store_.find(type) == store_.end()) {
+      store_[type] = std::vector<Envoy::Http::JwtAuth::JwtAuthStoreFactory>();
```
qiwzhang: not need for line 59 to 62.  
when you do map[key].  it will create an empty value for you if key doesn't exist

src/envoy/http/authn/jwt_authn_store.h
qiwzhang: It prefers not to use "typedef".  it is easier to read code with exact type definition.

src/envoy/http/authn/http_filter.cc
```diff
@@ -78,6 +83,53 @@ FilterHeadersStatus AuthenticationFilter::decodeHeaders(HeaderMap&, bool) {
...
+  }
+
+  if (jwt_store_.find(IstioAuthn::ORIGIN_STORE) != jwt_store_.end() &&
+      jwt_store_[IstioAuthn::ORIGIN_STORE].size() > 0) {
+    num_config = jwt_store_[IstioAuthn::ORIGIN_STORE].size();
+  }
+  ENVOY_LOG(debug, "AuthenticationFilter: {} num_config is {}", __func__,
+            num_config);
+
+  if (num_config > 0) {
```
qiwzhang: The key part is how to find the correct config/store with each policy step.   Please rebase your code with Diem change and implement this in AuthenticatorBase class.   

It is too simple to only get the first JWT

src/envoy/http/authn/jwt_authn_store.h
lei-tang: Done, no separation of the types and TODO added.

src/envoy/http/authn/jwt_authn_store.h
lei-tang: Done.

src/envoy/http/authn/http_filter.cc
lei-tang: Will rebase.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Stay tuned. Will iterate through all credential rules and methods when rebasing the code.

src/envoy/http/authn/jwt_authn_store.h
lei-tang: Done.

src/envoy/http/authn/jwt_authn_store.h
lei-tang: Done.

src/envoy/http/authn/http_filter.cc
lei-tang: Stay tuned. Will update the code to iterate through the credentials and find correct JWT configs when rebasing the code.

lei-tang: @qiwzhang, @diemtvu I have rebased the code and made the changes mentioned in your review comments: 
-- The JWT authentication is now implemented in AuthenticatorBase class. 
-- Iterate through all credential rules and methods, as well as peers' methods.
-- Find the correct config/store with each policy step.

src/envoy/http/authn/http_filter_factory.cc
lei-tang: Done.

src/envoy/http/authn/http_filter.cc
lei-tang: Done.

src/envoy/http/authn/jwt_authn_utils.cc
```diff
@@ -0,0 +1,45 @@
...
+
+// Convert istio-authn::jwt to jwt_auth::jwt in protobuf format.
+void convertJwtAuthFormat(
+    const ::istio::authentication::v1alpha1::Jwt& jwt_authn,
+    Http::JwtAuth::Config::AuthFilterConfig* proto_config) {
+  // Todo: when istio-authn::jwt diverges from jwt_auth::jwt,
+  // may need to convert more fields.
+  auto jwt = proto_config->add_jwts();
+  MessageUtil::jsonConvert(jwt_authn, *jwt);
+  jwt->set_jwks_uri_envoy_cluster(kJwtClusterName);
```
diemtvu: There should be a function to construct the cluster name from config (e.g from jwks_uri and issuer etc). E.g here is how pilot did it for current jwt config https://github.com/istio/istio/blob/c863c92ff0384fa5f343c6575b4fe9095a249703/pilot/pkg/proxy/envoy/v1/mixer.go#L401

You can make a similar function here.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -76,12 +80,44 @@ void AuthenticatorBase::validateX509(
...
+
+  // Verify the JWT token, onDone() will be called when completed.
+  jwt_auth_->Verify(*filter_context()->headers(), this);
+}
+
+void AuthenticatorBase::onDone(const JwtAuth::Status& status) {
+  ENVOY_LOG(debug, "AuthenticatorBase::onDone is called with status {}",
+            int(status));
+  if (status != JwtAuth::Status::OK) {
+    (*jwt_done_callback_)(&payload_, false);
```
diemtvu: nit: you can send back nullptr if it fail.

src/envoy/http/authn/authenticator_base.h
```diff
@@ -28,16 +29,18 @@ namespace AuthN {
 // AuthenticatorBase is the base class for authenticator. It provides functions
 // to perform individual authentication methods, which can be used to construct
 // compound authentication flow.
-class AuthenticatorBase : public Logger::Loggable<Logger::Id::filter> {
+class AuthenticatorBase : public Logger::Loggable<Logger::Id::filter>,
+                          public JwtAuth::JwtAuthenticator::Callbacks {
```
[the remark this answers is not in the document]
diemtvu: On second thought, forget nested. Define and use that class in .cc only to avoid cluttering this header file.

src/envoy/http/authn/jwt_authn_utils.cc
diemtvu: To be clear, this is also for quick prototype. Later, pilot may need to put that information in the config. We'll need introduce new proto for that at some point.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -76,12 +80,44 @@ void AuthenticatorBase::validateX509(
...
-  // pass done_callback so that it would be trigger by jwt_authenticator.onDone.
-  done_callback(&payload, false);
+    const iaapi::Jwt& jwt,
+    const AuthenticatorBase::MethodDoneCallback& done_callback) {
+  Http::JwtAuth::Config::AuthFilterConfig proto_config;
+  Envoy::Http::Istio::AuthN::convertJwtAuthFormat(jwt, &proto_config);
+
+  std::string config_str;
+  proto_config.SerializeToString(&config_str);
+  if (jwt_store_.find(config_str) == jwt_store_.end()) {
```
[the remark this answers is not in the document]

src/envoy/http/authn/authenticator_base.h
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/authenticator_base.h
```diff
@@ -58,16 +61,35 @@ class AuthenticatorBase : public Logger::Loggable<Logger::Id::filter> {
...
   // Pointer to filter state. Do not own.
   FilterContext& filter_context_;
 
   const DoneCallback done_callback_;
+
+  // ClusterManager reference
+  Upstream::ClusterManager& cm_;
+
+  // The JwtAuthnStore reference
+  std::map<std::string, JwtAuth::JwtAuthStore*>& jwt_store_;
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/authenticator_base.h
```diff
@@ -58,16 +61,35 @@ class AuthenticatorBase : public Logger::Loggable<Logger::Id::filter> {
...
+
+  // The JWT authenticator object.
+  std::unique_ptr<Envoy::Http::JwtAuth::JwtAuthenticator> jwt_auth_;
+
+  // The MethodDoneCallback for JWT authentication.
+  const AuthenticatorBase::MethodDoneCallback* jwt_done_callback_;
+
+  // The payload object
+  Payload payload_;
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/jwt_authn_store.h
```diff
@@ -0,0 +1,78 @@
...
+  std::map<std::string, Envoy::Http::JwtAuth::JwtAuthStore *> &store() {
+    store_tls_.clear();
+    for (auto it = store_.begin(); it != store_.end(); it++) {
+      store_tls_[it->first] = &(it->second.store());
+    }
+    return store_tls_;
+  }
+
+  // Add an AuthFilterConfig to the store.
+  void addToStore(Envoy::Http::JwtAuth::Config::AuthFilterConfig &config) {
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/jwt_authn_utils.cc
lei-tang: Done.

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authenticator_base.h
```diff
@@ -58,16 +61,35 @@ class AuthenticatorBase : public Logger::Loggable<Logger::Id::filter> {
...
+  // interface.
+  void onDone(const JwtAuth::Status& status);
+
   // Pointer to filter state. Do not own.
   FilterContext& filter_context_;
 
   const DoneCallback done_callback_;
+
+  // ClusterManager reference
+  Upstream::ClusterManager& cm_;
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/peer_authenticator.h
```diff
@@ -17,6 +17,7 @@
 
 #include "authentication/v1alpha1/policy.pb.h"
 #include "src/envoy/http/authn/authenticator_base.h"
+#include "src/envoy/http/authn/jwt_authn_store.h"
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/origin_authenticator.h
```diff
@@ -17,6 +17,7 @@
 
 #include "authentication/v1alpha1/policy.pb.h"
 #include "src/envoy/http/authn/authenticator_base.h"
+#include "src/envoy/http/authn/jwt_authn_store.h"
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/origin_authenticator.cc
```diff
@@ -56,10 +56,22 @@ void OriginAuthenticator::run() {
     }
     return;
   }
-  runMethod(credential_rule_.origins(0),
-            [this](const Payload* payload, bool success) {
-              onMethodDone(payload, success);
-            });
+  // Note to explain the change (the note will be removed after the code review)
```
diemtvu: It's probably your JwtAuthenticatorCallbacks  keep a ref to the callback method. You can keep it by value instead, then this change is not necessary.

src/envoy/http/authn/jwt_authn_utils.cc
```diff
@@ -0,0 +1,49 @@
...
+namespace Istio {
+namespace AuthN {
+
+namespace {
+const std::string kJwtClusterName("example_issuer");
+}  // namespace
+
+// Get the Jwks URI for Envoy cluster
+const std::string getJwksUriEnvoyCluster() {
+  // Todo: Pilot needs to put a cluster field in the Istio authn JWT config.
```
diemtvu: use TODO(lei-tang): instead of Todo. here and all

src/envoy/http/authn/authenticator_base.h
```diff
@@ -68,6 +72,12 @@ class AuthenticatorBase : public Logger::Loggable<Logger::Id::filter> {
   FilterContext& filter_context_;
 
   const DoneCallback done_callback_;
+
+  // The JWT authenticator object.
+  std::unique_ptr<Envoy::Http::JwtAuth::JwtAuthenticator> jwt_auth_;
+
+  // The JwtAuthenticatorCallbacks object.
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/filter_context.h
```diff
@@ -54,13 +59,25 @@ class FilterContext : public Logger::Loggable<Logger::Id::filter> {
   // Accessor to connection
   const Network::Connection* connection() { return connection_; }
 
+  // Accessor to ClusterManager
+  JwtToAuthStoreMap& jwtToAuthStoreMap() { return jwt_store_map_; }
```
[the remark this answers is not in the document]
lei-tang: The name jwtToAuthStoreMap is more descriptive than jwt_store_map_ and is more consistent with the function naming conventions in this file.

src/envoy/http/authn/jwt_authn_utils.cc
lei-tang: Done.

src/envoy/http/authn/origin_authenticator.cc
lei-tang: Removed the notes and added an explanation that the unique_ptr callbackForRunMethod is used to deallocate the the MethodDoneCallback object when out of scope.

src/envoy/http/authn/jwt_authn_store.h
```diff
@@ -0,0 +1,102 @@
...
+    jwt_store_tls_.clear();
+    for (auto it = jwt_store_.begin(); it != jwt_store_.end(); it++) {
+      jwt_store_tls_[it->first] = &(it->second.store());
+    }
+    return jwt_store_tls_;
+  }
+
+  // Add a JWT config to the store.
+  void addToStore(const ::istio::authentication::v1alpha1::Jwt &jwt) {
+    if (jwt_store_.find(jwt) != jwt_store_.end()) {
```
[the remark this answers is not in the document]
lei-tang: Added the following TODO:
  void addToStore(const ::istio::authentication::v1alpha1::Jwt &jwt) {
    // TODO(lei-tang): it may be ok to use issuer as the key but need to make
    // sure no cases of multiple Jwt have the same issuer

src/envoy/http/authn/filter_context.h
```diff
@@ -54,13 +59,25 @@ class FilterContext : public Logger::Loggable<Logger::Id::filter> {
...
+
  private:
   // Pointer to the headers of the request.
   HeaderMap* headers_;
 
   // Pointer to network connection of the request.
   const Network::Connection* connection_;
 
+  // ClusterManager reference
+  Upstream::ClusterManager& cm_;
```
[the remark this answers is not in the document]
lei-tang: Done.

Issue #991, opened 2018-02-05
JWT_auth: when checking audience, not to check protocol scheme


qiwzhang:
Following audiences should be considered as matched

http://audience/
https://audience
audience

Basically, remove https://, and http:// and trailing slash /

when comparing audience.

PR #1191, opened 2018-03-08
Check audience no protocol scheme

JimmyCYJ:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #https://github.com/istio/proxy/issues/991

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/envoy/http/jwt_auth/pubkey_cache.h
+++ b/src/envoy/http/jwt_auth/pubkey_cache.h
@@ -106,7 +106,8 @@ class PubkeyCacheItem {
     // Point beg to first character after protocol scheme prefix in audience.
     if (aud.compare(0, kHTTPSchemePrefix.length(), kHTTPSchemePrefix) == 0) {
       beg = kHTTPSchemePrefix.length();
-    } else if (aud.compare(0, kHTTPSSchemePrefix.length(), kHTTPSSchemePrefix) == 0) {
+    } else if (aud.compare(0, kHTTPSSchemePrefix.length(),
+                           kHTTPSSchemePrefix) == 0) {
       beg = kHTTPSSchemePrefix.length();
     }
     // Point end to trailing slash in aud.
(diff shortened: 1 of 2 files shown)


src/envoy/http/jwt_auth/pubkey_cache.h
```diff
@@ -83,6 +98,25 @@ class PubkeyCacheItem {
   }
 
  private:
+  // Searches "http://" or "https://" prefix in aud, and stores the position
+  // after protocol scheme prefix in beg. Searches trailing slash in aud, and
+  // stores the position before trailing slash in end. Returns true if aud has
+  // characters other than prefix and suffix.
+  bool SanitizeAudience(const std::string& aud, int& beg, int& end) {
```
qiwzhang: I think it is better for this function to return a processed string.
If not modification, (for most case), just return the input string which is not data copy (as I understand std::string) which is fast.

src/envoy/http/jwt_auth/pubkey_cache.h
```diff
@@ -27,6 +27,10 @@ namespace JwtAuth {
 namespace {
 // Default cache expiration time in 5 minutes.
 const int kPubkeyCacheExpirationSec = 600;
+
+const std::string kHTTPSchemePrefix("http://");
```
qiwzhang: add comment per style guide for each global variable

JimmyCYJ: Done.

src/envoy/http/jwt_auth/pubkey_cache.h
JimmyCYJ: Done.

JimmyCYJ:
Please take a look. Thanks!

PR #1203, opened 2018-03-12
Pubkey kid alg optional

### LINKED ISSUE #1121: jwt-auth:  "kid" and "alg" should be optional for pubkey
According to https://tools.ietf.org/html/rfc7517#page-8.  "kid" and "alg" should be optional.  But they are required in the current code  at  https://github.com/istio/proxy/blob/master/src/envoy/auth/jwt.cc#L429

Need to fix it

JimmyCYJ:
**What this PR does / why we need it**:Change checking for "kid" and "alg" claims in public key to be an optional step.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1121

**Release note**:

```release-note
NONE
```


JimmyCYJ:
cla/google

PR #1205, opened 2018-03-14
Upgrade Envoy to newest on master - pre cluster-warming

andraxylia:
Need to move to newest on master to be ready to pick up cluster warming once it is ready.

PR contains the newest envoy SHA + changes needed because of Envoy API changes, please pay attention to report_data.h.

### DIFF
--- a/src/envoy/http/mixer/report_data.h
+++ b/src/envoy/http/mixer/report_data.h
@@ -30,11 +30,11 @@ const std::set<std::string> ResponseHeaderExclusives = {};
 }  // namespace
 
 class ReportData : public ::istio::control::http::ReportData {
-  const HeaderMap* headers_;
-  const RequestInfo::RequestInfo& info_;
+  const HeaderMap *headers_;
+  const RequestInfo::RequestInfo &info_;
 
  public:
-  ReportData(const HeaderMap* headers, const RequestInfo::RequestInfo& info)
+  ReportData(const HeaderMap *headers, const RequestInfo::RequestInfo &info)
       : headers_(headers), info_(info) {}
 
   std::map<std::string, std::string> GetResponseHeaders() const override {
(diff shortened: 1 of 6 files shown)


src/envoy/http/mixer/report_data.h
```diff
@@ -48,12 +48,12 @@ class ReportData : public ::istio::control::http::ReportData {
       ::istio::control::http::ReportData::ReportInfo* data) const override {
     data->received_bytes = info_.bytesReceived();
     data->send_bytes = info_.bytesSent();
-    data->duration =
-        std::chrono::duration_cast<std::chrono::nanoseconds>(info_.duration());
+    data->duration = std::chrono::duration_cast<std::chrono::nanoseconds>(
+        std::chrono::system_clock::now() - info_.startTime());
```
JimmyCYJ: Could you please change it back to info_.duration()? 
duration() returns the microseconds duration of the first byte received to the last byte sent.
And this duration is used to generate "response.duration" attribute, which is defined as "The amount of time the response took to generate."

andraxylia: duration() is no longer an API. Only startTime() is available.

lizan: Use startTimeMonotonic and `chrono::steady_clock` to calculate the duration.

andraxylia:
I updated to use requestComplete(0 for duration, PTAL.

src/envoy/http/mixer/report_data.h
```diff
@@ -45,15 +45,17 @@ class ReportData : public ::istio::control::http::ReportData {
   }
 
   void GetReportInfo(
-      ::istio::control::http::ReportData::ReportInfo* data) const override {
+      ::istio::control::http::ReportData::ReportInfo *data) const override {
     data->received_bytes = info_.bytesReceived();
     data->send_bytes = info_.bytesSent();
-    data->duration =
-        std::chrono::duration_cast<std::chrono::nanoseconds>(info_.duration());
+    data->duration = std::chrono::nanoseconds{0};
```
lizan: Use value_or() instead of these 4 lines.

andraxylia: Done.

PR #1208, opened 2018-03-15
Workaround for gRPC cluster delivered by CDS

lizan:

Workaround for https://github.com/envoyproxy/envoy/issues/2762 as it is blocking proxy update.
Fixes https://github.com/istio/istio/issues/4257

**Release note**:

```release-note
None
```

### DIFF
--- a/src/envoy/utils/mixer_control.cc
+++ b/src/envoy/utils/mixer_control.cc
@@ -36,6 +36,23 @@ class EnvoyTimer : public ::istio::mixerclient::Timer {
   Event::TimerPtr timer_;
 };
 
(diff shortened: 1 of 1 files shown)


saumoh:
@qiwzhang @mandarjog Don't you folks think there is a race condition under which the gRPC cluster may not be there if learned via CDS?
[2018-03-15 20:57:12.485][105][info][config] external/envoy/source/server/listener_manager_impl.cc:583] all dependencies initialized. starting workers
[2018-03-15 20:57:12.487][105][warning][upstream] external/envoy/source/server/lds_subscription.cc:68] lds: fetch failure: Unknown gRPC client cluster 'mixer_check_server'
wdyt?
cc @spikecurtis

PR #1215, opened 2018-03-16
Add ALTS support

lizan:

Early prototype of ALTS support.

**Release note**:

```release-note
None
```


src/envoy/alts/alts_socket.proto
```diff
@@ -0,0 +1,27 @@
+// Copyright 2017 Istio Authors
+//
```
JimmyCYJ: s/2017/2018, please update other comments as well.

src/envoy/alts/alts_socket.proto
```diff
@@ -0,0 +1,27 @@
...
+
+syntax = "proto3";
+
+package envoy.security.v2;
+
+import "google/protobuf/duration.proto";
+import "validate/validate.proto";
+
+message AltsSocket {
+  string handshaker_service = 1 [(validate.rules).string.min_bytes = 1];
```
JimmyCYJ: Could you add some comments to explain this field? Is it used specify tls/loas handshake?

src/envoy/alts/tsi_handshaker.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include "src/core/tsi/transport_security_interface.h"
+
+namespace Envoy {
+namespace Security {
+
+class TsiHandshakerCallbacks {
+ public:
+  virtual ~TsiHandshakerCallbacks() {}
+
```
JimmyCYJ: Could you add some comments to briefly introduce methods and handshake process?

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,278 @@
...
+
+void TsiSocket::setTransportSocketCallbacks(
+    Envoy::Network::TransportSocketCallbacks &callbacks) {
+  callbacks_ = &callbacks;
+
+  handshaker_ = handshaker_cb_(callbacks.connection().dispatcher());
+  handshaker_->setHandshakerCallbacks(*this);
+}
+
+std::string TsiSocket::protocol() const { return ""; }
```
qiwzhang: why protocol is empty?

qiwzhang:
BTW, could we add some tests ?

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,278 @@
...
+    ENVOY_CONN_LOG(debug, "TSI: unprotecting message size: {}",
+                   callbacks_->connection(), message_size);
+
+    while (message_size > 0) {
+      size_t unprotected_buffer_size_to_send = unprotected_buffer_size;
+      size_t processed_message_size = message_size;
+      status = tsi_frame_protector_unprotect(
+          frame_protector_, message_bytes, &processed_message_size,
+          unprotected_buffer, &unprotected_buffer_size_to_send);
+      if (status != TSI_OK) break;
```
JimmyCYJ: Can we add ENVOY_CONN_LOG for status != TSI_OK to make it easier to debug?

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,278 @@
...
+      return {action, 0, false};
+    }
+  }
+
+  if (handshake_complete_) {
+    size_t message_size = raw_read_buffer_.length();
+    auto *message_bytes = reinterpret_cast<unsigned char *>(
+        raw_read_buffer_.linearize(message_size));
+
+    unsigned char unprotected_buffer[4096];
```
JimmyCYJ: Is there any reason that we set buffer size to 4096?

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,278 @@
...
+      ASSERT(result == TSI_OK);
+      raw_write_buffer_.add(protected_buffer, protected_buffer_size_to_send);
+      message_bytes += processed_message_size;
+      message_size -= processed_message_size;
+
+      ENVOY_CONN_LOG(debug, "TSI: protecting message processed: {}",
+                     callbacks_->connection(), processed_message_size);
+
+      // Don't forget to flush.
+      if (message_size == 0) {
```
JimmyCYJ: Can we move this if () { do {} while () } out side the while loop? I think that would make it clearer. 
Maybe add ENVOY_CONN_LOG for debugging when result != TSI_OK and still_pending_size > 0?

src/envoy/alts/tsi_transport_socket.h
```diff
@@ -0,0 +1,96 @@
...
+
+class TsiSocketFactory : public Network::TransportSocketFactory {
+ public:
+  explicit TsiSocketFactory(HandshakerFactoryCb handshaker_factory);
+
+  bool implementsSecureTransport() const override;
+  Network::TransportSocketPtr createTransportSocket() const override;
+
+ private:
+  HandshakerFactoryCb handshaker_factory_;
```
diemtvu: Please be consistent with they way you name variable. In previous class, the similar variable is named handshaker_cb_. Personally, I like to use the class name as variable name to avoid any guessing.

src/envoy/alts/alts_socket_factory.cc
```diff
@@ -0,0 +1,108 @@
...
+  std::string handshaker_service = config.handshaker_service();
+
+  return std::make_unique<Security::TsiSocketFactory>([handshaker_service](
+      Event::Dispatcher &dispatcher) {
+    grpc_alts_credentials_options *options =
+        grpc_alts_credentials_client_options_create();
+
+    tsi_handshaker *handshaker = nullptr;
+
+    alts_tsi_handshaker_create(options, "target_name",
```
yangminzhu: Just curious what's the purpose of the hard-coded "target_name" here?

The [doc](https://github.com/grpc/grpc/blob/3287ceb06773ff3d7463a06c5f2dac742527135c/src/core/tsi/alts/handshaker/alts_tsi_handshaker.h#L48) says it's "the name of the endpoint that the channel is connecting to", but I don't quite understand how can we just use a hard-coded value?

yangminzhu:
Would it be possible to associate a high-level design doc in the PR description? Thanks!

src/envoy/alts/tsi_transport_socket.h
```diff
@@ -0,0 +1,96 @@
...
+    }
+    bool shouldDrainReadBuffer() override { return false; }
+    void setReadBufferReady() override {}
+    void raiseEvent(Network::ConnectionEvent) override {}
+
+   private:
+    TsiSocket& parent_;
+  };
+
+  Network::PostIoAction doHandshake();
```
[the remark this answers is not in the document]
lizan: I don't think either Google or Envoy style requires comment for member variables, they are pretty obvious from the naming, let me know if you want comment for specific member.

src/envoy/alts/tsi_transport_socket.h
```diff
@@ -0,0 +1,96 @@
...
+#include "common/network/raw_buffer_socket.h"
+#include "envoy/network/transport_socket.h"
+#include "src/envoy/alts/tsi_handshaker.h"
+
+namespace Envoy {
+namespace Security {
+
+typedef std::function<TsiHandshakerPtr(Event::Dispatcher&)> HandshakerFactoryCb;
+
+class TsiSocket : public Network::TransportSocket,
```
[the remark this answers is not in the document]
lizan: Done

src/envoy/alts/tsi_handshaker.h
```diff
@@ -0,0 +1,70 @@
...
+  }
+  void deferredDelete();
+
+ private:
+  static void onNextDone(tsi_result status, void* user_data,
+                         const unsigned char* bytes_to_send,
+                         size_t bytes_to_send_size,
+                         tsi_handshaker_result* handshaker_result);
+
+  tsi_handshaker* handshaker_{nullptr};
```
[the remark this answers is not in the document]
lizan: ditto

src/envoy/alts/tsi_handshaker.cc
```diff
@@ -0,0 +1,87 @@
...
+    onNextDone(status, callbacks_, bytes_to_send, bytes_to_send_size, result);
+  }
+  return status;
+}
+
+void TsiHandshaker::deferredDelete() {
+  if (calling_) {
+    delete_on_done_ = true;
+  } else {
+    dispatcher_.deferredDelete(Event::DeferredDeletablePtr{this});
```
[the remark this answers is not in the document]
lizan: Why not? defered delete is to avoid lifecycle issues between TsiSocket and TsiHandshaker.

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,278 @@
...
+    tsi_handshaker_result_extract_peer(handshaker_result, &peer);
+    ENVOY_CONN_LOG(debug, "TSI: Handshake successful: peer properties: {}",
+                   callbacks_->connection(), peer.property_count);
+    for (size_t i = 0; i < peer.property_count; ++i) {
+      ENVOY_CONN_LOG(debug, "  {}: {}", callbacks_->connection(),
+                     peer.properties[i].name,
+                     std::string(peer.properties[i].value.data,
+                                 peer.properties[i].value.length));
+    }
+    tsi_peer_destruct(&peer);
```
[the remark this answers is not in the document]
lizan: It will be used later for validating peer identities. I'd like to keep as it is.

src/envoy/alts/tsi_transport_socket.cc
lizan: ALTS has no knowledge whether the protocol above it is right now, so this is empty string. When ALTS has similar support of ALPN or NPN in its metadata, this will be populated from there.

src/envoy/alts/tsi_transport_socket.cc
lizan: Done

src/envoy/alts/tsi_transport_socket.cc
lizan: No particular, it's from sample code of TSI, I will tune later, added a comment

src/envoy/alts/tsi_transport_socket.cc
lizan: Done

src/envoy/alts/alts_socket.proto
```diff
@@ -0,0 +1,29 @@
...
+//   limitations under the License.
+
+syntax = "proto3";
+
+package envoy.security.v2;
+
+import "google/protobuf/duration.proto";
+import "validate/validate.proto";
+
+message AltsSocket {
```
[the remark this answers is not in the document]
lizan: yes, this will be eventually upstream to istio/api or envoyproxy/data-plane-api, in a meanwhile since this is simple enough, a copy or just craft the JSON in pilot should be enough

src/envoy/alts/tsi_transport_socket.h
lizan: Done

lizan:
Ping @qiwzhang tests will be added in follow up PRs.

src/envoy/alts/tsi_handshaker.cc
```diff
@@ -0,0 +1,88 @@
...
+  tsi_handshaker_result *result = nullptr;
+  tsi_result status =
+      tsi_handshaker_next(handshaker_,
+                          reinterpret_cast<const unsigned char *>(
+                              received.linearize(received_size)),
+                          received_size, &bytes_to_send, &bytes_to_send_size,
+                          &result, onNextDone, this);
+
+  if (status != TSI_ASYNC) {
+    onNextDone(status, callbacks_, bytes_to_send, bytes_to_send_size, result);
```
qiwzhang: It seems pass callback_ to second argument is wrong.   It is user_data,  it should be "TsiHandshaker"  according to line 26

lizan: done

src/envoy/alts/tsi_handshaker.cc
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  auto next_result = new TsiHandshakerCallbacks::NextResult{
+      status, std::move(to_send), handshaker_result};
+
+  handshaker->dispatcher_.post([handshaker, next_result]() {
+    ASSERT(handshaker->calling_);
+    handshaker->calling_ = false;
+    if (handshaker->delete_on_done_) {
+      handshaker->dispatcher_.deferredDelete(
```
qiwzhang: next_result is not freed in delete_on_done_ case

src/envoy/alts/alts_socket_factory.cc
```diff
@@ -0,0 +1,112 @@
...
+UpstreamAltsTransportSocketConfigFactory::createTransportSocketFactory(
+    const Protobuf::Message &message, TransportSocketFactoryContext &) {
+  auto config =
+      MessageUtil::downcastAndValidate<const envoy::security::v2::AltsSocket &>(
+          message);
+
+  std::string handshaker_service = config.handshaker_service();
+
+  return std::make_unique<Security::TsiSocketFactory>(
+      [handshaker_service](Event::Dispatcher &dispatcher) {
```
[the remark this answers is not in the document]
lizan: This cannot be capture by reference since handshaker_service will be destroyed when the lambda is invoked.

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,295 @@
...
+
+TsiSocket::~TsiSocket() {
+  ASSERT(!handshaker_);
+  tsi_frame_protector_destroy(frame_protector_);
+  frame_protector_ = nullptr;
+}
+
+void TsiSocket::setTransportSocketCallbacks(
+    Envoy::Network::TransportSocketCallbacks &callbacks) {
+  callbacks_ = &callbacks;
```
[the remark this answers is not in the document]
lizan: It is never same object, what do you mean?

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,295 @@
...
+              info, "TSI: protect flush message failure {}, closing connection",
+              callbacks_->connection(), result);
+          return {Network::PostIoAction::Close, 0, false};
+        }
+        raw_write_buffer_.add(protected_buffer, protected_buffer_size_to_send);
+      } while (still_pending_size > 0);
+    }
+
+    buffer.drain(buffer.length() - message_size);
+    ASSERT(result == TSI_OK);
```
qiwzhang: you will hit this assert if you "break" in line 230

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    // Don't forget to flush.
+    if (message_size == 0) {
+      ENVOY_CONN_LOG(debug, "TSI: protecting message flush: {}",
+                     callbacks_->connection(), message_size);
+      size_t still_pending_size;
+      do {
+        size_t protected_buffer_size_to_send = protected_buffer_size;
+        result = tsi_frame_protector_protect_flush(
```
qiwzhang: this frame protector code really looks ugly.   Maybe you should wrap it with c++ object now

src/envoy/alts/tsi_transport_socket.cc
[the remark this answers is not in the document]
diemtvu: ah, nvm, I miss read the other call is on raw_buffer_socket object.

script/pre-commit
```diff
@@ -38,7 +38,7 @@ if [[ ! -x "${CLANG_FORMAT}" ]]; then
 fi
 
 CLANG_FORMAT_VERSION="$(${CLANG_FORMAT} -version | cut -d ' ' -f 3)"
-CLANG_FORMAT_VERSION_REQUIRED="3.8.0"
+CLANG_FORMAT_VERSION_REQUIRED="5.0.0"
```
qiwzhang: I thought this version change is already in.    If no, can we do it in a separate PR.

lizan: I think you're on old diff...

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,295 @@
...
+#include "src/envoy/alts/tsi_transport_socket.h"
+
+#include "common/common/assert.h"
+#include "common/common/enum_to_int.h"
+#include "envoy/event/dispatcher.h"
+
+namespace Envoy {
+namespace Security {
+
+TsiSocket::TsiSocket(HandshakerFactory handshaker_factory)
```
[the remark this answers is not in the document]
lizan: yes, `std::function`s are generally passed by value rather than reference

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,295 @@
...
+}
+
+std::string TsiSocket::protocol() const { return ""; }
+
+Network::PostIoAction TsiSocket::doHandshake() {
+  ASSERT(!handshake_complete_);
+  ENVOY_CONN_LOG(debug, "TSI: doHandshake", callbacks_->connection());
+
+  if (handshaker_next_calling_) {
+    ENVOY_CONN_LOG(debug, "TSI: doHandshake next is pending, wait...",
```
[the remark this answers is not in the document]
lizan: Done

src/envoy/alts/tsi_transport_socket.cc
lizan: Done

lizan:
@qiwzhang @diemtvu @yangminzhu PTAL again. Thanks!

yangminzhu:
/lgtm 
I'm not familiar with the Envoy code so please wait for other reviewers' LGTM before submit, Thanks!

src/envoy/alts/tsi_transport_socket.h
```diff
@@ -0,0 +1,111 @@
...
+};
+
+/**
+ * An implementation of Network::TransportSocketFactory for TsiSocket
+ */
+class TsiSocketFactory : public Network::TransportSocketFactory {
+ public:
+  explicit TsiSocketFactory(HandshakerFactory handshaker_factory);
+
+  bool implementsSecureTransport() const override;
```
diemtvu: inline?

src/envoy/alts/tsi_transport_socket.cc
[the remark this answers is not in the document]
diemtvu: Add a TODO if so.

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -0,0 +1,207 @@
...
+    const unsigned char *unused_bytes;
+    size_t unused_byte_size;
+
+    status = tsi_handshaker_result_get_unused_bytes(
+        handshaker_result, &unused_bytes, &unused_byte_size);
+    ASSERT(status == TSI_OK);
+    if (unused_byte_size > 0) {
+      raw_read_buffer_.add(unused_bytes, unused_byte_size);
+    }
+    ENVOY_CONN_LOG(debug, "TSI: Handshake successful: unused_bytes: {}",
```
diemtvu: Would be nice to document why/what are these unused byte buffer.

src/envoy/alts/tsi_frame_protector.cc
```diff
@@ -0,0 +1,79 @@
...
+  return TSI_OK;
+}
+
+tsi_result TsiFrameProtector::unprotect(Buffer::Instance &input,
+                                        Buffer::Instance &output) {
+  ASSERT(frame_protector_);
+
+  // TODO(lizan): Tune the buffer size.
+  unsigned char unprotected_buffer[4096];
+  size_t unprotected_buffer_size = sizeof(unprotected_buffer);
```
diemtvu: just curious, 4096 is chosen so it's large enough for processed message? What happen if it's not?

src/envoy/alts/tsi_frame_protector.cc
```diff
@@ -0,0 +1,79 @@
...
+  while (input.length() > 0) {
+    auto *message_bytes =
+        reinterpret_cast<unsigned char *>(input.linearize(input.length()));
+    size_t unprotected_buffer_size_to_send = unprotected_buffer_size;
+    size_t processed_message_size = input.length();
+    tsi_result result = tsi_frame_protector_unprotect(
+        frame_protector_.get(), message_bytes, &processed_message_size,
+        unprotected_buffer, &unprotected_buffer_size_to_send);
+    if (result != TSI_OK) {
+      ASSERT(result != TSI_INVALID_ARGUMENT && result != TSI_UNIMPLEMENTED);
```
diemtvu: What's special about these 2 status?

lizan: Those two should never happen, usually means we're using TSI in a wrong way, so just `ASSERT`ing here.

src/envoy/alts/tsi_transport_socket.h
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * @param handshaker_factory a function to initiate a TsiHandshaker
+   */
+  explicit TsiSocket(HandshakerFactory handshaker_factory);
+  virtual ~TsiSocket();
+
+  // Network::TransportSocket
+  void setTransportSocketCallbacks(
+      Envoy::Network::TransportSocketCallbacks& callbacks) override;
+  std::string protocol() const override;
```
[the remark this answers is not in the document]
lizan: it might be, you have to extract metadata from the negotiation part. There is no reason to inline virtual methods in headers, same for below.

src/envoy/alts/tsi_transport_socket.cc
lizan: It is clear from tsi_handshaker doc, the bytes not used for handshake.

src/envoy/alts/tsi_frame_protector.cc
lizan: In each loop, the buffer is added to output, and re-used in next loop, so just more loop.

lizan:
This PR itself is a prototype, it is large enough so tests will be in follow up PRs.

diemtvu:
Generally, I prefer seeing tests with new code, even for prototype (in fact, I find it even more useful as you are going to change the code / fix bug more often during prototyping). Anyway, I'm not going to block you on this, and hope more tests will be added. Meanwhile, please add a short README to mention this is POC and what are known features/limitations so other don't make any wrong assumption. Thanks.

/lgtm

src/envoy/alts/tsi_transport_socket.cc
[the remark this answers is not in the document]
lizan: done

src/envoy/alts/tsi_frame_protector.cc
```diff
@@ -0,0 +1,79 @@
...
+      return result;
+    }
+    output.add(unprotected_buffer, unprotected_buffer_size_to_send);
+    input.drain(processed_message_size);
+  }
+
```
qiwzhang: do we need to call unproect_flush()?

src/envoy/alts/tsi_frame_protector.cc
```diff
@@ -0,0 +1,79 @@
...
+TsiFrameProtector::TsiFrameProtector(tsi_frame_protector *frame_protector)
+    : frame_protector_(frame_protector) {}
+
+tsi_result TsiFrameProtector::protect(Buffer::Instance &input,
+                                      Buffer::Instance &output) {
+  ASSERT(frame_protector_);
+
+  // TODO(lizan): tune size later
+  unsigned char protected_buffer[4096];
```
qiwzhang: can we define a const for 4096?

src/envoy/alts/tsi_frame_protector.cc
lizan: No, you need full protected frame to unprotect.

PR #1216, opened 2018-03-17
Validate JWTs in Istio authentication policy using the output of jwt-auth

### LINKED ISSUE #1214: Create AuthN filter to perform authentication.
The filter config should contain the peer/origin/principal logic as described by the policy (in the early version, it could be the original policy itself). 

The filter will be put behind 'validation' filters (e.g Jwt), which are responsible to verify certificate and extract claims. It will simply consume the output from those filters (which are stored in request header), and follow credential logic from policy to construct request principal (or reject request if data missing, per policy spec).

Parent issue: https://github.com/istio/istio/issues/1948

lei-tang:
**What this PR does / why we need it**:  POC: validate JWTs in Istio authentication policy using the output of jwt-auth

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1214 

**Release note**:

```release-note
```


src/envoy/http/authn/authenticator_base.cc
```diff
@@ -71,17 +70,48 @@ void AuthenticatorBase::validateX509(
...
+    const AuthenticatorBase::MethodDoneCallback& done_callback) {
   Payload payload;
-  // TODO (diemtvu/lei-tang): construct jwt_authenticator and call Verify;
-  // pass done_callback so that it would be trigger by jwt_authenticator.onDone.
-  done_callback(&payload, false);
+  Envoy::Http::HeaderMap& header = *filter_context()->headers();
+  const Network::Connection* connection = filter_context()->connection();
+  std::map<std::string, std::string> jwt_payload;
+
+  Mixer::CheckData check_data(header, connection);
```
diemtvu: You shouldn't call this mixer function here. The AuthN filter is generate data that would be used to construct "check" attributes by mixer. In future, the jwt payload might not be even part of the check data, as it will read data from authn payload.

You can clone the GetJWTPayload, or refactor it to some common library. I don't think it need anything special from 'CheckData'.  As mentioned above, mixer might not need to use this function after authN filter is in used, so either option is fine.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -71,17 +70,48 @@ void AuthenticatorBase::validateX509(
...
     const iaapi::Jwt&,
-    const AuthenticatorBase::MethodDoneCallback& done_callback) const {
+    const AuthenticatorBase::MethodDoneCallback& done_callback) {
   Payload payload;
-  // TODO (diemtvu/lei-tang): construct jwt_authenticator and call Verify;
-  // pass done_callback so that it would be trigger by jwt_authenticator.onDone.
-  done_callback(&payload, false);
+  Envoy::Http::HeaderMap& header = *filter_context()->headers();
+  const Network::Connection* connection = filter_context()->connection();
+  std::map<std::string, std::string> jwt_payload;
```
diemtvu: JwtPayload proto has a map<string, string> for (raw) claims. You can use it instead of this separate map. The raw claims need to be copied to payload message anyway.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -71,17 +70,48 @@ void AuthenticatorBase::validateX509(
...
+
+  Mixer::CheckData check_data(header, connection);
+  bool ret = check_data.GetJWTPayload(&jwt_payload);
+  if (!ret || jwt_payload.size() <= 0) {
```
diemtvu: use .empty()

src/envoy/http/authn/filter_context.cc
```diff
@@ -47,6 +47,18 @@ void FilterContext::setOriginResult(const Payload* payload) {
   // it's ok just to check jwt payload.
   if (payload != nullptr && payload->has_jwt()) {
     *result_.mutable_origin() = payload->jwt();
+    if (payload->jwt().user().size() > 0) {
```
diemtvu: Are you sure these debug prints are needed?

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -71,17 +70,48 @@ void AuthenticatorBase::validateX509(
...
+  if (jwt_payload.count("aud") > 0) {
+    payload->mutable_jwt()->add_audiences(jwt_payload["aud"]);
+  }
+  // Extract authorized presenter (azp)
+  if (jwt_payload.count("azp") > 0) {
+    payload->mutable_jwt()->set_presenter(jwt_payload["azp"]);
+  }
+}
+
 void AuthenticatorBase::validateJwt(
```
diemtvu: Please add unit tests for this.

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/filter_context.cc
lei-tang: These debug logs are removed.

src/envoy/http/authn/authenticator_base.cc
lei-tang: Working on it.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -71,17 +70,49 @@ void AuthenticatorBase::validateX509(
...
+    ::google::protobuf::Map< ::std::string, ::std::string>& jwt_payload,
+    Payload* payload) {
+  // Extract attributes from JWT payload
+  // Extract user
+  if (jwt_payload.count("iss") > 0 && jwt_payload.count("sub") > 0) {
+    payload->mutable_jwt()->set_user(jwt_payload["iss"] + "/" +
+                                     jwt_payload["sub"]);
+  }
+  // Extract audience
+  if (jwt_payload.count("aud") > 0) {
```
diemtvu: Do you know how multiple audiences are stored jwt output? Do we need to parse this? If if not supported, add a note for that here. + @qiwzhang

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,62 @@
...
+    ::google::protobuf::Map< ::std::string, ::std::string>* payload) {
+  const HeaderEntry* entry =
+      headers.get(JwtAuth::JwtAuthenticator::JwtPayloadKey());
+  if (!entry) {
+    return false;
+  }
+  std::string value(entry->value().c_str(), entry->value().size());
+  std::string payload_str = JwtAuth::Base64UrlDecode(value);
+  // Return an empty string if Base64 decode fails.
+  if (payload_str.empty()) {
```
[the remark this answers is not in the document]
lei-tang: Yes, this check is equivalent to the decode failure.  If it is for an empty input, then the JWT is empty, which is a failure (the required JWT is not present).

src/envoy/http/authn/authenticator_base.cc
lei-tang: Added a note that the aud field is taken from the jwt-auth produced claim as is. Wayne knows more about how multiple audiences are stored/formatted by jwt-auth. I think Istio authn does not need to parse the aud field because it is jwt-auth that validates the JWT audiences and other JWT attributes whereas Istio authn simply passes these attributes to next consumers.

src/envoy/http/authn/authn_utils.h
```diff
@@ -0,0 +1,38 @@
...
+namespace Istio {
+namespace AuthN {
+
+// AuthnUtils class provides utility functions used for authentication.
+class AuthnUtils : public Logger::Loggable<Logger::Id::filter> {
+ public:
+  // Retrieve the JWT payload from HTTP headers into the output payload map
+  static bool GetJWTPayloadFromHeaders(
+      const HeaderMap& headers,
+      ::google::protobuf::Map< ::std::string, ::std::string>* payload);
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -71,17 +70,49 @@ void AuthenticatorBase::validateX509(
...
+
+  bool ret = AuthnUtils::GetJWTPayloadFromHeaders(header, jwt_payload);
+  if (!ret || jwt_payload->empty()) {
+    ENVOY_LOG(debug, "AuthenticatorBase: {} no such JWT.", __func__);
+    done_callback(&payload, false);
+  } else {
+    ENVOY_LOG(debug, "AuthenticatorBase: {}(): a valid JWT is found.",
+              __func__);
+    // Extract attributes from JWT payload
+    extractJwtPayload(*jwt_payload, &payload);
```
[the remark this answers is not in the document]

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,62 @@
...
+    ENVOY_LOG(error, "Invalid {} header, invalid base64: {}",
+              JwtAuth::JwtAuthenticator::JwtPayloadKey().get(), value);
+    return false;
+  }
+  try {
+    auto json_obj = Json::Factory::loadFromString(payload_str);
+    json_obj->iterate(
+        [claims](const std::string& key, const Json::Object& obj) -> bool {
```
diemtvu: I feel this is missing smt. I know this is copied from CheckData, but the jwt implementation to extract payload is like this (note the special treatment for "aud"), and afaik, that payload is the same payload the jwt filter put back to the output (it simply strip the header and signature off the original jwt). Please check with @qiwzhang  on this.

https://github.com/istio/proxy/blob/0e2a8fc267887c3f5862fd202edc35720cef55c3/src/envoy/http/jwt_auth/jwt.cc#L312

src/envoy/http/authn/authenticator_base_test.cc
```diff
@@ -124,6 +124,16 @@ TEST_F(AuthenticatorBaseTest,
 
 // TODO: more tests for Jwt.
 
+TEST_F(AuthenticatorBaseTest, ValidateJwtWithNoJwtInHeader) {
```
diemtvu: Please add test for non trivial cases.

src/envoy/http/authn/BUILD
```diff
@@ -73,6 +74,20 @@ envoy_cc_library(
...
+    name = "authn_utils_lib",
+    srcs = [
+        "authn_utils.cc",
+    ],
+    hdrs = [
+        "authn_utils.h",
+    ],
+    repository = "@envoy",
+    deps = [
+        "//src/envoy/http/jwt_auth:jwt_authenticator_lib",
```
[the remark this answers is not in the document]
lei-tang: This dependency on JwtAuth::JwtAuthenticator::JwtPayloadKey() is to get the key of the header containing the jwt-auth output. We may get the key of the header through the jwt-auth library (the current implementation) or get the key of the header from the jwt-auth config file if the jwt-auth config file has such a field. If getting the key of the header from the jwt-auth config file, the dependency will be on the jwt-auth protobuf.

src/envoy/http/authn/authn_utils.cc
lei-tang: This missing state for handling the exception is intentional. When (*claims)[key] = obj.asString() is called, if the obj type is not string, an exception is thrown. But this should not be treated as error (i.e., return false) since other objects still need to be added to the claims.

src/envoy/http/authn/authn_utils.h
```diff
@@ -0,0 +1,38 @@
...
+namespace Envoy {
+namespace Http {
+namespace Istio {
+namespace AuthN {
+
+// AuthnUtils class provides utility functions used for authentication.
+class AuthnUtils : public Logger::Loggable<Logger::Id::filter> {
+ public:
+  // Retrieve the JWT payload from HTTP headers into the output payload map
+  static bool GetJWTPayloadFromHeaders(
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -71,17 +70,49 @@ void AuthenticatorBase::validateX509(
...
+    const AuthenticatorBase::MethodDoneCallback& done_callback) {
   Payload payload;
-  // TODO (diemtvu/lei-tang): construct jwt_authenticator and call Verify;
-  // pass done_callback so that it would be trigger by jwt_authenticator.onDone.
-  done_callback(&payload, false);
+  Envoy::Http::HeaderMap& header = *filter_context()->headers();
+
+  bool ret = AuthnUtils::GetJWTPayloadFromHeaders(
+      header, payload.mutable_jwt()->mutable_claims());
+  if (!ret || payload.mutable_jwt()->mutable_claims()->empty()) {
```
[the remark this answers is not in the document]

src/envoy/http/authn/http_filter_integration_test.cc
```diff
@@ -32,6 +54,7 @@ class AuthenticationFilterIntegrationTest
 
  protected:
   Http::TestHeaderMapImpl default_request_headers_;
+  Http::TestHeaderMapImpl request_headers_with_sec_istio_auth_userinfo;
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/http_filter_integration_test.cc
```diff
@@ -70,9 +93,82 @@ TEST_P(AuthenticationFilterIntegrationTest, SourceMTlsFail) {
   EXPECT_TRUE(response_->complete());
   EXPECT_STREQ("401", response_->headers().Status()->value().c_str());
 }
+//
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/authenticator_base_test.cc
lei-tang: Done.

src/envoy/http/authn/authn_utils.cc
diemtvu: 'aud' could be array.

src/envoy/http/authn/BUILD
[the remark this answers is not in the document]
diemtvu: If only that, then remove it and hard coded the location (to match with the one in jwt filter impl) for now.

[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/authn_utils.cc
lei-tang: Added the handling of the case when 'aud' may be an array of strings or a single string.

src/envoy/http/authn/sample/authn.conf
```diff
@@ -30,24 +30,36 @@
               }
             ],
             "filters": [
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/authenticator_base.cc
lei-tang: Added the handling of the case when aud may be an array of strings or a single string

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,97 @@
...
+                  ENVOY_LOG(error,
+                            "{}: an exception is thrown when get audience as "
+                            "string. The exception is: {}",
+                            __FUNCTION__, e.what());
+                }
+              }
+              std::string aud_str;
+              // If aud_vector has more than 1 items, the result audiences are
+              // delimited by ','
+              for (size_t i = 0; i < aud_vector.size(); i++) {
```
diemtvu: Why do you want to concat aud data? It may destroy the usability of that field later.
On a closer look at the purpose of this function, it's better that you combine the buildAttribute in to this function, and make function output JwtPayload directly.
For those special attribute (iss, aud, presenter), write them to the designate location, for the rest, put into claims map (it's ok to ignore non string claims for now)

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -71,17 +72,59 @@ void AuthenticatorBase::validateX509(
...
-  // TODO (lei-tang): Adding other attributes (i.e ip) to payload if desire.
+  // TODO (lei-tang): Adding other attributes (i.e ip) to payload if needed.
   done_callback(&payload, true);
 }
 
+// Build JWT attributes from JWT payload
+// TODO (lei-tang): Build more attributes/claims to payload if needed.
+void buildJwtAttributes(JwtPayload* payload) {
+  ::google::protobuf::Map< ::std::string, ::std::string>& claims =
+      *payload->mutable_claims();
```
qiwzhang: why do you need mutable_claims?   you are NOT modifying it.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -71,17 +72,59 @@ void AuthenticatorBase::validateX509(
...
+  } else {
+    if (payload.mutable_jwt()->mutable_claims()->empty()) {
+      ENVOY_LOG(error, "AuthenticatorBase: {} there is no JWT claims.",
+                __func__);
+      done_callback(nullptr, false);
+      return;
+    }
+    ENVOY_LOG(debug, "AuthenticatorBase: {}(): a valid JWT is found.",
+              __func__);
+    buildJwtAttributes(payload.mutable_jwt());
```
qiwzhang: Do you need to fail if required claims are not there?

src/envoy/http/authn/authenticator_base_test.cc
```diff
@@ -123,6 +127,30 @@ TEST_F(AuthenticatorBaseTest,
...
+  iaapi::Jwt jwt;
+  Http::TestHeaderMapImpl request_headers_with_jwt{
+      {kSecIstioAuthUserInfoHeaderKey, kSecIstioAuthUserinfoHeaderValue}};
+  FilterContext filter_context{&request_headers_with_jwt, &connection_};
+  MockAuthenticatorBase authenticator{&filter_context};
+
+  authenticator.validateJwt(jwt, [](const Payload* payload, bool success) {
+    // When there is a verified JWT in the HTTP header, validateJwt()
+    // should return non-nullptr and success.
+    EXPECT_TRUE(payload != nullptr);
```
qiwzhang: it is better to also verify payload data.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,97 @@
...
+    const HeaderMap& headers, const LowerCaseString& jwt_payload_key,
+    ::google::protobuf::Map< ::std::string, ::std::string>* claims) {
+  const HeaderEntry* entry = headers.get(jwt_payload_key);
+  if (!entry) {
+    ENVOY_LOG(error, "{}: no JwtPayloadKey entry {} in the header",
+              __FUNCTION__, jwt_payload_key.get());
+    return false;
+  }
+  std::string value(entry->value().c_str(), entry->value().size());
+  std::string payload_str = Base64::decode(value);
```
qiwzhang: you need to use Base64URLDecode()  which is different with Base64::decode()

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,97 @@
...
+              for (size_t i = 0; i < aud_vector.size(); i++) {
+                aud_str += aud_vector[i];
+                if (i != aud_vector.size() - 1) {
+                  aud_str += ",";
+                }
+              }
+              (*claims)[key] = aud_str;
+            } else {
+              // will throw execption if value type is not string.
+              (*claims)[key] = obj.asString();
```
qiwzhang: why not use asJsonString() so it will work for "aud" too.

src/envoy/http/authn/authenticator_base.cc
lei-tang: JWT verification is handled by jwt-auth filter, so the required claims have been verified by jwt-auth filter. If the verified JWT is non-empty, validateJwt() regards the JWT has been verified and outputs the verified claims from the jwt-auth filter.

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authenticator_base.cc
lei-tang: This function has been removed.

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc
lei-tang: The 'aud' string array has been changed to bed outputed to the JwtPayload proto message.
message JwtPayload {
  .....
  repeated string audiences = 2;
}

src/envoy/http/authn/authenticator_base_test.cc
lei-tang: Done.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -25,6 +24,8 @@ namespace Http {
 namespace Istio {
 namespace AuthN {
 namespace {
+const std::string kJwtHeaderKey = "sec-istio-auth-userinfo";
```
diemtvu: Add comment for this constant, also note that it would be removed once jwt filter support output parameters.

src/envoy/http/authn/authenticator_base_test.cc
```diff
@@ -123,6 +141,37 @@ TEST_F(AuthenticatorBaseTest,
...
+    EXPECT_TRUE(success);
+    EXPECT_STREQ(kJwtUserExpected.c_str(), payload->jwt().user().c_str());
```
diemtvu: Use  EXPECT_TRUE with TestUtility::protoEqual to test all data in proto.

src/envoy/http/authn/http_filter_integration_test.cc
```diff
@@ -70,9 +92,79 @@ TEST_P(AuthenticationFilterIntegrationTest, SourceMTlsFail) {
...
+  const Envoy::Http::HeaderString& header_value =
+      upstream_request_->headers()
+          .get(Envoy::Http::LowerCaseString(kSecIstioAuthnPayloadHeaderKey))
+          ->value();
+  std::string value_base64(header_value.c_str(), header_value.size());
+  const std::string value = Base64::decode(value_base64);
+  Envoy::Http::Istio::AuthN::Result result;
+  bool parse_ret = result.ParseFromString(value);
+  EXPECT_TRUE(parse_ret);
+  EXPECT_TRUE(result.has_origin());
```
diemtvu: EXPECT_TRUE and TestUtility::protoEqual

src/envoy/http/authn/authenticator_base_test.cc
lei-tang: Done.  TestUtility::protoEqual() uses SerializeAsString() and the output is non-deterministic.  Thus, MessageDifferencer::Equals() is used, in which "Equality is defined as  all fields within the two messages being set to the same value. Primitive fields and strings are compared by value while embedded messages/groups are compared as if via a recursive call. "

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/http_filter_integration_test.cc
lei-tang: Done.

src/envoy/http/authn/testdata/envoy_origin_jwt.conf
```diff
@@ -30,24 +30,36 @@
               }
             ],
             "filters": [
+              {
+                "type": "decoder",
+                "name": "jwt-auth",
```
diemtvu: Do you need this? I don't think you use this file at all.

lei-tang: Yes, I use this config file to do non-mocked test.

lei-tang: When running tests with this config file, in one terminal, I run "bazel-bin/src/envoy/envoy -c src/envoy/http/authn/testdata/envoy_origin_jwt.conf". In another terminal, I run: curl --header "Authorization: Bearer $token" http://localhost:9090/echo -d "hello world"

diemtvu: I don't see it use in this PR. Also, it sounds like you have to link with jwt_auth lib in order to do that. It doesn't sound right.

lei-tang: This file has been removed.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+namespace Http {
+namespace Istio {
+namespace AuthN {
+
+bool AuthnUtils::GetJWTPayloadFromHeaders(
+    const HeaderMap& headers, const LowerCaseString& jwt_payload_key,
+    istio::authn::JwtPayload* payload) {
+  const HeaderEntry* entry = headers.get(jwt_payload_key);
+  if (!entry) {
+    ENVOY_LOG(error, "{}: no JwtPayloadKey entry {} in the header",
```
qiwzhang: should not be an error.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+              __FUNCTION__, jwt_payload_key.get());
```
qiwzhang: can remove __FUNCTION__ too

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+          try {
+            aud_vector = obj.getStringArray(key, true);
+          } catch (Json::Exception& e) {
+            // Try as string
+            try {
+              auto audience = obj.asString();
+              aud_vector.push_back(audience);
+            } catch (Json::Exception& e) {
+              ENVOY_LOG(error,
+                        "{}: an exception is thrown when get audience as "
```
qiwzhang: don't mention exception in the log.   just mention "aud" field value type is not string or string array.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+            }
+          }
+          for (size_t i = 0; i < aud_vector.size(); i++) {
+            payload->add_audiences(aud_vector[i]);
+          }
+        } else {
+          // will throw execption if value type is not string.
+          // In current implementation, only string objects are extracted into
+          // claims.
+          (*claims)[key] = obj.asString();
```
qiwzhang: so "aud" don't need to add to "claims" map?  who will use claims field?  will they expect that?

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+      payload->mutable_claims();
+  try {
+    auto json_obj = Json::Factory::loadFromString(payload_str);
+    ENVOY_LOG(debug, "{}: json object is {}", __FUNCTION__,
+              json_obj->asJsonString());
+    json_obj->iterate([claims, &payload](const std::string& key,
+                                         const Json::Object& obj) -> bool {
+      ENVOY_LOG(debug, "{}: key is {}", __FUNCTION__, key);
+      try {
+        if (key == "aud") {
```
qiwzhang: lambda function is suppose to be small.  this is too big.  could you write a separate function for this.

src/envoy/http/authn/authn_utils.cc
qiwzhang: How about using asJsonString() to take care all types.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+          for (size_t i = 0; i < aud_vector.size(); i++) {
+            payload->add_audiences(aud_vector[i]);
+          }
+        } else {
+          // will throw execption if value type is not string.
+          // In current implementation, only string objects are extracted into
+          // claims.
+          (*claims)[key] = obj.asString();
+        }
+      } catch (...) {
```
qiwzhang: not sure what case will be catched here?

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+      }
+      return true;
+    });
+  } catch (...) {
+    ENVOY_LOG(error, "Invalid {} header, invalid json: {}",
```
qiwzhang: What will be catched here?

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+              jwt_payload_key.get(), payload_str);
+    return false;
+  }
+
+  if (payload->mutable_claims()->empty()) {
```
qiwzhang: why do you need to call mutable_?

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -27,6 +26,12 @@ namespace Http {
 namespace Istio {
 namespace AuthN {
 namespace {
+// The HTTP header from which to get the verified Jwt result.
+// It is currently hard-coded. After jwt-auth has a
+// parameter for this header, the hardcoded parameter will
+// be removed.
+const std::string kJwtHeaderKey = "sec-istio-auth-userinfo";
```
qiwzhang: you can define LowerCaseString here, not to construct LowerCaseString everytime it is used.

src/envoy/http/authn/authenticator_base_test.cc
```diff
@@ -124,6 +129,51 @@ TEST_F(AuthenticatorBaseTest,
...
+  iaapi::Jwt jwt;
+  Http::TestHeaderMapImpl request_headers_with_jwt{
+      {kSecIstioAuthUserInfoHeaderKey, kSecIstioAuthUserinfoHeaderValue}};
+  FilterContext filter_context{&request_headers_with_jwt, &connection_};
+  MockAuthenticatorBase authenticator{&filter_context};
+
+  authenticator.validateJwt(jwt, [](const Payload* payload, bool success) {
+    Payload expected_payload;
+    google::protobuf::util::JsonParseOptions options;
+    JsonStringToMessage(
```
qiwzhang: not to call this inside lambda.

src/envoy/http/authn/authenticator_base_test.cc
```diff
@@ -124,6 +129,51 @@ TEST_F(AuthenticatorBaseTest,
...
+        )",
+        &expected_payload, options);
+
+    // When there is a verified JWT in the HTTP header, validateJwt()
+    // should return non-nullptr and success.
+    EXPECT_TRUE(payload != nullptr);
+    EXPECT_TRUE(success);
+    // Note: TestUtility::protoEqual() uses SerializeAsString() and the output
+    // is non-deterministic.  Thus, MessageDifferencer::Equals() is used.
+    EXPECT_TRUE(google::protobuf::util::MessageDifferencer::Equals(
```
qiwzhang: At the top of file,  do

using google::protobuf::util::MessageDifferencer:

src/envoy/http/authn/authn_utils.h
```diff
@@ -0,0 +1,39 @@
...
+namespace Envoy {
+namespace Http {
+namespace Istio {
+namespace AuthN {
+
+// AuthnUtils class provides utility functions used for authentication.
+class AuthnUtils : public Logger::Loggable<Logger::Id::filter> {
+ public:
+  // Retrieve the JWT payload from the HTTP header into the output payload map
+  static bool GetJWTPayloadFromHeaders(const HeaderMap& headers,
```
qiwzhang: It is weird a class with only one static function.

src/envoy/http/authn/authn_utils.cc
lei-tang: Per proxy/src/istio/authn/context.proto, 'aud' is a repeated string in JwtPayload and it is not in the claims map. Per the examples in proxy/src/istio/authn/context.proto, the values are plain strings. If there are requirements to encoded them into json strings, we can modify later.

diemtvu: I'm ok with this for now. As I am working on the mixer part, I think we haven't handle multiple aud case properly neither. We can also keep all claims as json string, just in case, but it could be done in the next PR.

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc
lei-tang: It is intended to call all kinds of exceptions.

src/envoy/http/authn/authn_utils.cc
lei-tang: It is intended to call all kinds of exceptions.

src/envoy/http/authn/authn_utils.cc
lei-tang: Changed it to claims().

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authenticator_base_test.cc
lei-tang: Done.

src/envoy/http/authn/authenticator_base_test.cc
lei-tang: Done.

src/envoy/http/authn/authn_utils.h
lei-tang: One more function has been added to this class.

src/envoy/http/authn/authenticator_base.cc
qiwzhang: Am I looking at the wrong version?   The type is still std::string.

src/envoy/http/authn/authn_utils.h
```diff
@@ -0,0 +1,47 @@
...
+                                       const LowerCaseString& jwt_payload_key,
+                                       istio::authn::JwtPayload* payload);
+
+ private:
+  // The json iterator callback function
+  static bool JsonIteratorCallback(
```
qiwzhang: This should be inside annoymous namespace inside the cc file

src/envoy/http/authn/authenticator_base.cc
lei-tang: Just upload the change, please take a look.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,112 @@
...
+  }
+  ::google::protobuf::Map< ::std::string, ::std::string>* claims =
+      payload->mutable_claims();
+  try {
+    auto json_obj = Json::Factory::loadFromString(payload_str);
+    ENVOY_LOG(debug, "{}: json object is {}", __FUNCTION__,
+              json_obj->asJsonString());
+    json_obj->iterate([claims, payload](const std::string& key,
+                                        const Json::Object& obj) -> bool {
+      return JsonIteratorCallback(key, obj, claims, payload);
```
qiwzhang: if claims is part of payload,  
claims =     payload->mutable_claims();

why you want to pass both?

can you just pass payload?

src/envoy/http/authn/authn_utils.h
lei-tang: Done.

qiwzhang:
@diemtvu 

Sorry,  I felt this code will be wasted too.

If JWT validation is done in other filter,  this filter simply just verifies results based on the config.   It doesn't need to make any remote call,  so it doesn't need to do any call-back, or async call. 
In this case, this filter can be simplified as following:

in decodeHeader()  {

     check_config,
     check_JWT result,  
     and yes or not.
}

diemtvu:
I agree, but let's not change the framework now until we are 100% sure we will go with this direction. It can be done in separate PR to simplify the flow. For now, I don't think there is any gain to rewrite the flow. There will be a lot of iterations a head, including not using the policy as is but some simpler/specific envoy-facing API.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,123 @@
...
+};
+
+bool JsonIterCallback::JsonIteratorCallback(const std::string& key,
+                                            const Envoy::Json::Object& obj,
+                                            istio::authn::JwtPayload* payload) {
+  ENVOY_LOG(debug, "{}: key is {}", __FUNCTION__, key);
+  ::google::protobuf::Map< ::std::string, ::std::string>* claims =
+      payload->mutable_claims();
+  try {
+    if (key == "aud") {
```
diemtvu: Sorry I miss this usecase. We do want to keep all original claims in 'claims' (may be as JsonString to be compatible with all types). Then for audiences and presenters (and user), parse the object again to construct the payload properly.

You can make change that in the following up PR.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -0,0 +1,123 @@
...
+namespace {
+class JsonIterCallback : public Logger::Loggable<Logger::Id::filter> {
+ public:
+  // The json iterator callback function
+  static bool JsonIteratorCallback(const std::string&,
+                                   const Envoy::Json::Object&,
+                                   istio::authn::JwtPayload* payload);
+};
+
+bool JsonIterCallback::JsonIteratorCallback(const std::string& key,
```
diemtvu: Please have sufficient tests for this to cover single and mutlple aud values.

lei-tang: Will add more tests in next PR.

src/envoy/http/authn/authn_utils.cc
lei-tang: Will change that in the following up PR.

lei-tang: Done, in the PR #1284.

src/envoy/http/authn/authn_utils.cc
lei-tang: Done, in the PR #1284.

PR #1219, opened 2018-03-20
Add support for v1 config.

### LINKED ISSUE #1217: Migrate "v2" config to "v1" in Mixer fitler config
Currently,  it is using "v2" which is confusing.   Decided to migrate it back to "v1" with following steps

1) change proxy to support both "v1" and "v2" fields.  if this is version x
2) After that is widely deployed,  change Pilot to "s/v2/v1", also release node that side-car proxy has to be at lease version x.  Please fiile an issue in istio.
3) after a while, remove "v2" in the proxy.

JimmyCYJ:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1217 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/envoy/utils/config.cc
+++ b/src/envoy/utils/config.cc
@@ -36,8 +36,12 @@ const std::string kDefaultMixerClusterName("mixer_server");
 // ReadConfig() finds config from |json| that matches version |config_version|,
 // and parses config into |message|. Returns true if config is read and parsed
 // successfully.
-bool ReadConfig(const Json::Object &json, Message *message,
-                const std::string &config_version) {
+bool ReadConfig(const Json::Object &json, const std::string &config_version,
+                Message *message) {
+  if (!json.hasObject(config_version)) {
+    return false;
+  }
+
   std::string config_str = json.getObject(config_version)->asJsonString();
   Status status = ParseJsonMessage(config_str, message);
   auto &logger = Logger::Registry::getLog(Logger::Id::config);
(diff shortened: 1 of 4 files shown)


src/envoy/utils/config.cc
```diff
@@ -28,10 +28,31 @@ namespace Utils {
...
+const std::string kV1Config("v1");
 
 // The name for the mixer server cluster.
 const std::string kDefaultMixerClusterName("mixer_server");
 
+// ReadConfig() finds config from |json| that matches version |config_version|,
+// and parses config into |message|. Returns true if config is read and parsed
+// successfully.
+bool ReadConfig(const Json::Object &json, Message *message,
+                const std::string &config_version) {
```
qiwzhang: per style guide,  output argument should be the last one.

src/envoy/utils/config.cc
```diff
@@ -47,19 +68,14 @@ bool ReadV2Config(const Json::Object &json, Message *message) {
...
-  auto &logger = Logger::Registry::getLog(Logger::Id::config);
-  if (status.ok()) {
-    ENVOY_LOG_TO_LOGGER(logger, info, "V2 mixer client config: {}",
-                        message->DebugString());
-    return true;
+  return ReadConfig(json, message, kV2Config);
+}
+
+bool ReadV1Config(const Json::Object &json, Message *message) {
+  if (!json.hasObject(kV1Config)) {
```
qiwzhang: this checking can be part of ReadConfig() too.

src/envoy/utils/config.cc
JimmyCYJ: Thanks for pointing that out. Fixed.

src/envoy/utils/config.cc
JimmyCYJ: Done.

PR #1221, opened 2018-03-21
Move authn context to istio/authn for sharing between filters.

diemtvu:
Move authn context proto to istio in order to use them in other filters (e.g mixer filter to consume authn output for check and report)

```NONE
```


src/istio/authn/BUILD
```diff
@@ -0,0 +1,28 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+load(
+    "@envoy//bazel:envoy_build_system.bzl",
+    "envoy_proto_library",
+)
+
+envoy_proto_library(
+    name = "context_proto",
```
qiwzhang: The package name in the proto should be changed to match the folder path

diemtvu: done

PR #1237, opened 2018-03-22
Let tests build on macOS.

PiotrSikora:
Access logs in tests were changed from /dev/stdout to /dev/null,
because opening /dev/stdout results in "Permission Denied" error
on macOS when running as an unprivileged user.


mandarjog:
This is awesome !

ldemailly:
maybe /dev/fd/1 ?

PiotrSikora:
@ldemailly it doesn't matter, tests don't look at those files anyway.

PR #1239, opened 2018-03-22
Change mixer filter to use authn result data if available

diemtvu:
The new authn filter will be placed before mixer filter in the filter change. This filter will consolidate identities and attributes from x509 and JWTs (if any) and output the authenticated result into header. Mixer should consume these information for the check request.

For backward compatible (i.e when authn filter is not in used), the code still fallback to fetch attributes directly from X509 and Jwt payload as of today, when result data is not found. This fallback will be removed once authn filter released.

Issue: https://github.com/istio/istio/issues/4337

### DIFF
--- a/include/istio/control/http/check_data.h
+++ b/include/istio/control/http/check_data.h
@@ -82,6 +82,8 @@ class CheckData {
   virtual bool GetJWTPayload(
       std::map<std::string, std::string> *payload) const = 0;
 
+  // If the request has authentication result in header, parses data into the
+  // output result; returns true if success. Otherwise, returns false.
   virtual bool GetAuthenticationResult(istio::authn::Result *result) const = 0;
 };
 
--- a/include/istio/control/http/check_data.h
+++ b/include/istio/control/http/check_data.h
@@ -19,6 +19,8 @@
 #include <map>
 #include <string>
 
+#include "src/istio/authn/context.pb.h"
+
 namespace istio {
 namespace control {
 namespace http {
@@ -79,6 +81,8 @@ class CheckData {
   // string map, and return true. Otherwise return false.
   virtual bool GetJWTPayload(
       std::map<std::string, std::string> *payload) const = 0;
+
+  virtual bool GetAuthenticationResult(istio::authn::Result *result) const = 0;
 };
 
 // An interfact to update request HTTP headers with Istio attributes.
--- a/src/istio/control/http/attributes_builder.cc
+++ b/src/istio/control/http/attributes_builder.cc
@@ -75,9 +75,9 @@ void AttributesBuilder::ExtractAuthAttributes(CheckData *check_data) {
         builder.AddString(AttributeName::kRequestAuthUser, origin.user());
       }
       if (!origin.audiences().empty()) {
-        // TODO(diemtvu): this should be send as repeated field once mixer support string_list
-        // (https://github.com/istio/istio/issues/2802)
-        // For now, just use the first value.
+        // TODO(diemtvu): this should be send as repeated field once mixer
+        // support string_list (https://github.com/istio/istio/issues/2802) For
+        // now, just use the first value.
         builder.AddString(AttributeName::kRequestAuthAudiences,
                           origin.audiences(0));
       }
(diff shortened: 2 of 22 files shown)


src/istio/control/tcp/attributes_builder.cc
```diff
@@ -32,6 +32,8 @@ void AttributesBuilder::ExtractCheckAttributes(CheckData* check_data) {
     builder.AddInt64(AttributeName::kSourcePort, source_port);
   }
 
+  // TODO(diemtvu): add TCP authn filter similar to http case, and use authn
+  // result output here instead.
```
qiwzhang: I think we don't need authn for tcp.  not need to add this TODO

src/istio/control/http/attributes_builder.cc
```diff
@@ -57,11 +58,46 @@ void AttributesBuilder::ExtractRequestHeaderAttributes(CheckData *check_data) {
...
+    }
+    if (!authn_result.peer_user().empty()) {
+      builder.AddString(AttributeName::kSourceUser, authn_result.peer_user());
+    }
+    if (authn_result.has_origin()) {
+      const auto &origin = authn_result.origin();
+      if (!origin.user().empty()) {
+        builder.AddString(AttributeName::kRequestAuthUser, origin.user());
+      }
+      if (!origin.audiences().empty()) {
```
qiwzhang: audiences is a repeated field, you need to check its _size().  I think

src/istio/control/http/attributes_builder.cc
```diff
@@ -57,11 +58,46 @@ void AttributesBuilder::ExtractRequestHeaderAttributes(CheckData *check_data) {
...
+        // TODO(diemtvu): what to do with multiple aud ?. For now, just send the
+        // first one.
+        builder.AddString(AttributeName::kRequestAuthAudiences,
```
qiwzhang: we can join them with ","

src/istio/control/tcp/attributes_builder.cc
diemtvu: I can remove TODO, but I would like to align them somehow. Here, for TCP, we try to get user id from x509 directly, which seem weird.

src/istio/control/http/attributes_builder_test.cc
```diff
@@ -317,6 +319,70 @@ TEST(AttributesBuilderTest, TestCheckAttributes) {
...
+
+  ClearContextTime(AttributeName::kRequestTime, &request);
+
+  std::string out_str;
+  TextFormat::PrintToString(request.attributes, &out_str);
+  GOOGLE_LOG(INFO) << "===" << out_str << "===";
+
+  Attributes expected_attributes;
+  ASSERT_TRUE(
+      TextFormat::ParseFromString(kCheckAttributes, &expected_attributes));
```
qiwzhang: Did know see the new "kCheckAttributes" string defined?  are you using the old one?

include/istio/control/http/check_data.h
```diff
@@ -79,6 +81,8 @@ class CheckData {
   // string map, and return true. Otherwise return false.
   virtual bool GetJWTPayload(
       std::map<std::string, std::string> *payload) const = 0;
+
+  virtual bool GetAuthenticationResult(istio::authn::Result *result) const = 0;
```
qiwzhang: please add comment

src/istio/control/http/attributes_builder.cc
diemtvu: empty also works for vector.

src/istio/control/http/attributes_builder.cc
diemtvu: Possibly. Though I want to review what is the right approach here. In google, the attribute_context mentioned audiences as repeated string, but looks like mixer take a single string at the moment, I don't know should we change that behavior.

src/istio/control/http/attributes_builder_test.cc
diemtvu: Yes. every thing should be the same. We only swap the source of data.

include/istio/control/http/check_data.h
diemtvu: Done

PR #1242, opened 2018-03-23
Check quota cache hit

### LINKED ISSUE #1181: Mixer client should Report check.cache_hit, quota.cache_hit attributes
Mixer client has access to this information but does not include it in the report call.

If included it can be added as dimension to metrics.

Also rename
`check.error_code`  to `mixer.check_status`

That way all present and future mixer info is in the `mixer.*` namespace.

JimmyCYJ:
**What this PR does / why we need it**: When Check() is enabled, Mixer filter sends attributes check.cache_hit and quota.cache_hit in Report() call. When Check() is disabled, Mixer filter does not send these attributes.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1181 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/include/istio/mixerclient/check_response.h
+++ b/include/istio/mixerclient/check_response.h
@@ -13,8 +13,8 @@
  * limitations under the License.
  */
 
-#ifndef PROXY_CHECK_RESPONSE_H
-#define PROXY_CHECK_RESPONSE_H
+#ifndef ISTIO_MIXERCLIENT_CHECK_RESPONSE_H
+#define ISTIO_MIXERCLIENT_CHECK_RESPONSE_H
 
 #include "google/protobuf/stubs/status.h"
 
@@ -23,22 +23,18 @@ namespace mixerclient {
 
 // The CheckResponseInfo holds response information in detail.
 struct CheckResponseInfo {
-  CheckResponseInfo()
-      : is_check_cache_hit(false),
-        is_quota_cache_hit(false),
-        response_status(::google::protobuf::util::Status::UNKNOWN) {}
-
   // Whether this check response is from cache.
-  bool is_check_cache_hit;
+  bool is_check_cache_hit{false};
 
   // Whether this quota response is from cache.
-  bool is_quota_cache_hit;
+  bool is_quota_cache_hit{false};
 
   // The check and quota response status.
-  ::google::protobuf::util::Status response_status;
+  ::google::protobuf::util::Status response_status{
+      ::google::protobuf::util::Status::UNKNOWN};
 };
 
 }  // namespace mixerclient
 }  // namespace istio
 
-#endif  // PROXY_CHECK_RESPONSE_H
+#endif  // ISTIO_MIXERCLIENT_CHECK_RESPONSE_H
--- a/include/istio/mixerclient/check_response.h
+++ b/include/istio/mixerclient/check_response.h
@@ -1,0 +1,42 @@
+/* Copyright 2018 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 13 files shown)


include/istio/mixerclient/check_response.h
```diff
@@ -0,0 +1,44 @@
...
+
+// The CheckResponseInfo holds response information in detail.
+struct CheckResponseInfo {
+  CheckResponseInfo()
+      : is_check_cache_hit(false),
+        is_quota_cache_hit(false),
+        response_status(::google::protobuf::util::Status::UNKNOWN) {}
+
+  // Whether this check response is from cache.
+  bool is_check_cache_hit;
```
qiwzhang: can you use is_check_cache_hit{false};   
you can do this for status too.  so constructor init can be avoid.

include/istio/mixerclient/environment.h
```diff
@@ -28,6 +29,10 @@ namespace mixerclient {
 // Uses UNAVAILABLE status code to indicate network failure.
 using DoneFunc = std::function<void(const ::google::protobuf::util::Status&)>;
```
qiwzhang: can we move this DoneFunction to "control"

src/istio/mixerclient/client_impl.cc
```diff
@@ -53,14 +54,19 @@ MixerClientImpl::~MixerClientImpl() {}
...
     const std::vector<::istio::quota_config::Requirement> &quotas,
-    TransportCheckFunc transport, DoneFunc on_done) {
+    TransportCheckFunc transport, CheckDoneFunc on_done) {
   ++total_check_calls_;
 
   std::unique_ptr<CheckCache::CheckResult> check_result(
       new CheckCache::CheckResult);
   check_cache_->Check(attributes, check_result.get());
+
+  std::unique_ptr<CheckResponseInfo> check_response_info(new CheckResponseInfo);
```
qiwzhang: not need to new,  just allocate it from stack

src/istio/mixerclient/client_impl.cc
```diff
@@ -117,11 +125,16 @@ CancelFunc MixerClientImpl::Check(
        on_done](const Status &status) {
         raw_check_result->SetResponse(status, *request_copy, *response);
         raw_quota_result->SetResponse(status, *request_copy, *response);
+        CheckResponseInfo check_response_info;
+        check_response_info.is_check_cache_hit = false;
+        check_response_info.is_quota_cache_hit = false;
```
qiwzhang: not need to set default values

src/istio/mixerclient/client_impl.cc
```diff
@@ -117,11 +125,16 @@ CancelFunc MixerClientImpl::Check(
...
         if (on_done) {
           if (!raw_check_result->status().ok()) {
-            on_done(raw_check_result->status());
+            check_response_info.response_status = raw_check_result->status();
+            on_done(check_response_info);
           } else {
-            on_done(raw_quota_result->status());
+            check_response_info.response_status = raw_quota_result->status();
+            on_done(check_response_info);
```
qiwzhang: combine above,  to only call on_done() once,   just setting status is different

include/istio/mixerclient/environment.h
JimmyCYJ: As discussed, we will keep DoneFunc in this file.

include/istio/mixerclient/check_response.h
JimmyCYJ: Done.

src/istio/mixerclient/client_impl.cc
JimmyCYJ: Done.

src/istio/mixerclient/client_impl.cc
JimmyCYJ: Done. Thanks.

JimmyCYJ:
Please take a look. Thanks!

PR #1261, opened 2018-03-23
Add the allow_missing_or_failed option to jwt-auth

### LINKED ISSUE #1212: Add params to Jwt filter to pass even when some Jwt rules failed to validate
When chaining with AuthN filter, we want to make request reject decision by the AuthN filter, instead of Jwt filter, as it need access to the policy logic.

The param could be in the AuthFilterConfig, assuming each Jwt rules, when running validation doesn't reject request by itself (i.e Jwt filter will run each Jwt rules in sequence).

Parent issue: https://github.com/istio/istio/issues/1948

lei-tang:
**What this PR does / why we need it**: Add the allow_missing_or_failed option to jwt-auth so jwt-auth won't return 401 when jwt is missing or the jwt authentication fails.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1212 

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/jwt_auth/config.proto
+++ b/src/envoy/http/jwt_auth/config.proto
@@ -191,4 +191,11 @@ message AuthFilterConfig {
   //      path_exact: /healthz
   //
   repeated HttpPattern bypass_jwt = 3;
+
+  // From
+  // https://github.com/envoyproxy/data-plane-api/blob/11c750628bcaeeeca3f72f89fe5c99921aa727b1/envoy/config/filter/http/jwt_authn/v2alpha/config.proto
+  // If true, the request is allowed if JWT is missing or JWT verification
+  // fails. Default is false, a request without JWT or failed JWT verification
+  // is not allowed.
+  bool allow_missing_or_failed = 4;
 }
--- a/src/envoy/http/jwt_auth/config.proto
+++ b/src/envoy/http/jwt_auth/config.proto
@@ -197,5 +197,7 @@ message AuthFilterConfig {
   // If true, the request is allowed if JWT is missing or JWT verification
   // fails. Default is false, a request without JWT or failed JWT verification
   // is not allowed.
+  // Note: in current implementation, if token is not there, no rejection (401
+  // code) is sent.
   bool allow_missing_or_failed = 4;
 }
--- a/src/envoy/http/jwt_auth/http_filter.cc
+++ b/src/envoy/http/jwt_auth/http_filter.cc
@@ -28,7 +28,7 @@ namespace Http {
 
 JwtVerificationFilter::JwtVerificationFilter(Upstream::ClusterManager& cm,
                                              JwtAuth::JwtAuthStore& store)
-    : jwt_auth_(cm, store) {}
+    : jwt_auth_(cm, store), store_(store) {}
 
 JwtVerificationFilter::~JwtVerificationFilter() {}
 
@@ -61,7 +61,8 @@ void JwtVerificationFilter::onDone(const JwtAuth::Status& status) {
   if (state_ == Responded) {
     return;
   }
-  if (status != JwtAuth::Status::OK) {
+  if (status != JwtAuth::Status::OK &&
+      !store_.config().allow_missing_or_failed()) {
     state_ = Responded;
     // verification failed
     Code code = Code(401);  // Unauthorized
--- a/src/envoy/http/jwt_auth/http_filter.cc
+++ b/src/envoy/http/jwt_auth/http_filter.cc
@@ -28,7 +28,8 @@ namespace Http {
 
 JwtVerificationFilter::JwtVerificationFilter(Upstream::ClusterManager& cm,
                                              JwtAuth::JwtAuthStore& store)
-    : jwt_auth_(cm, store), store_(store) {}
+    : jwt_auth_(cm, store),
+      allow_missing_or_failed_(store.config().allow_missing_or_failed()) {}
 
 JwtVerificationFilter::~JwtVerificationFilter() {}
 
@@ -61,8 +62,7 @@ void JwtVerificationFilter::onDone(const JwtAuth::Status& status) {
   if (state_ == Responded) {
     return;
   }
-  if (status != JwtAuth::Status::OK &&
-      !store_.config().allow_missing_or_failed()) {
+  if (status != JwtAuth::Status::OK && !allow_missing_or_failed_) {
     state_ = Responded;
     // verification failed
     Code code = Code(401);  // Unauthorized
--- a/src/envoy/http/jwt_auth/http_filter.cc
+++ b/src/envoy/http/jwt_auth/http_filter.cc
@@ -28,8 +28,7 @@ namespace Http {
 
 JwtVerificationFilter::JwtVerificationFilter(Upstream::ClusterManager& cm,
                                              JwtAuth::JwtAuthStore& store)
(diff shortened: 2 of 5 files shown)


src/envoy/http/jwt_auth/http_filter.h
```diff
@@ -51,6 +51,8 @@ class JwtVerificationFilter : public StreamDecoderFilter,
   StreamDecoderFilterCallbacks* decoder_callbacks_;
   // The auth object.
   JwtAuth::JwtAuthenticator jwt_auth_;
+  // The auth store object, which can be used to get the config
```
diemtvu: try to keep only what you need. It avoid any unintentional uses and make it easy to know the purpose of each variables. In this case, you can just need one boolean, or at most, ref to the config.

src/envoy/http/jwt_auth/http_filter.cc
```diff
@@ -61,7 +62,7 @@ void JwtVerificationFilter::onDone(const JwtAuth::Status& status) {
   if (state_ == Responded) {
     return;
   }
-  if (status != JwtAuth::Status::OK) {
+  if (status != JwtAuth::Status::OK && !allow_missing_or_failed_) {
```
qiwzhang: JwtAuthentictor should be the one to use the config.   You can put this code in 

https://github.com/istio/proxy/blob/master/src/envoy/http/jwt_auth/jwt_authenticator.cc#L230

lei-tang: Done.

lei-tang:
Test added.

lei-tang:
Will make a similar change to the one in api (https://github.com/istio/api/blob/master/mixer/v1/config/client/auth.proto) after this PR is done.

src/envoy/http/jwt_auth/config.proto
```diff
@@ -191,4 +191,11 @@ message AuthFilterConfig {
   //      path_exact: /healthz
   //
   repeated HttpPattern bypass_jwt = 3;
+
+  // From
+  // https://github.com/envoyproxy/data-plane-api/blob/11c750628bcaeeeca3f72f89fe5c99921aa727b1/envoy/config/filter/http/jwt_authn/v2alpha/config.proto
+  // If true, the request is allowed if JWT is missing or JWT verification
+  // fails. Default is false, a request without JWT or failed JWT verification
+  // is not allowed.
+  bool allow_missing_or_failed = 4;
```
diemtvu: Btw, I noticed that this implementation does NOT reject if token is not there (at least from my observation when testing the policy). Not very important for our purpose, but you may want to rename and/or change the comment here to reflect the correct impact.

src/envoy/http/jwt_auth/jwt_authenticator_test.cc
```diff
@@ -107,6 +107,13 @@ const char kExampleConfig[] = R"(
 }
 )";
 
+// An example JSON config with allow_missing_or_failed option enabled
+const char kExampleConfigWithAllowMissingOrFailed[] = R"(
+{
```
diemtvu: You test config have no jwt spec, what use cases can this cover?

lei-tang: Done: added comments to explain the test case (when jwt is missing, status in the callback should be OK).

src/envoy/http/jwt_auth/config.proto
lei-tang: I have changed the comment based on your observation.

src/envoy/http/jwt_auth/jwt_authenticator_test.cc
```diff
@@ -472,7 +472,8 @@ TEST_F(JwtAuthenticatorTest, TestMissedJWT) {
 }
 
 TEST_F(JwtAuthenticatorTest, TestAllowMissingOrFailedIsTrue) {
-  // When JWT is missing from the config, the status should be OK
+  // In this test, when JWT is missing, the status should still be OK
```
diemtvu: The test case you want is when jwt spec is available, and request contain invalid jwt, but request will be accepted. This test is ok to add, but not for that purpose

lei-tang: Done: added another test when the request contains an invalid JWT but still returns OK because allow_missing_or_failed is true.

src/envoy/http/jwt_auth/jwt_authenticator_test.cc
```diff
@@ -464,6 +471,34 @@ TEST_F(JwtAuthenticatorTest, TestMissedJWT) {
...
+
+  // Empty headers.
+  auto headers = TestHeaderMapImpl{};
+  auth_->Verify(headers, &mock_cb_);
+}
+
+TEST_F(JwtAuthenticatorTest, TestInValidJwtWhenAllowMissingOrFailedIsTrue) {
+  // In this test, when JWT is invalid, the status should still be OK
+  // because allow_missing_or_failed is true.
+  SetupConfig(kExampleConfigWithAllowMissingOrFailed);
```
diemtvu: Not enough. The config have no jwt spec so no jwt validation actually run. I believe even without your flag, request will still be accepted (that a good test case to add, btw).

You need a config like the kExampleConfig, but with your flag, and setup the test similar to TestInvalidJwt (and/or all those that not returns OK). But this time, should see status is OK.

If you want thorough, can also use TestWithParams and contruct example config and expected output according to the flag value.

lei-tang: Done: changed the example JSON config to contain a good JWT config and allow_missing_or_failed = true. With allow_missing_or_failed == true and an invalid JWT, the status should still be OK.

lei-tang: Also changed the missing JWT test case to use a config with JWT and allow_missing_or_failed == true.

diemtvu: Your kExampleConfigWithAllowMissingOrFailed doesn't have any jwt spec still. Do I missing smt?

lei-tang: kExampleConfigWithAllowMissingOrFailed has been removed and the tests use
kExampleConfigWithJwtAndAllowMissingOrFailed, which has JWT spec.

src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -227,7 +227,11 @@ bool JwtAuthenticator::OkToBypass() {
 void JwtAuthenticator::DoneWithStatus(const Status& status) {
   ENVOY_LOG(debug, "Jwt authentication completed with: {}",
             JwtAuth::StatusToString(status));
```
JimmyCYJ: Could you please modify the ENVOY_LOG here to add allow_missing_or_failed information, so that it would be easier for debugging when allow_missing_or_failed is true and status is not Status::OK?

JimmyCYJ:
I will approve when this PR is updated. Thanks.

src/envoy/http/jwt_auth/jwt_authenticator.cc
lei-tang: Done.

PR #1284, opened 2018-03-26
Add aud to claims and check it in the tests

lei-tang:
**What this PR does / why we need it**: Add aud to JWT claims and check it in the tests

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/authn/BUILD
+++ b/src/envoy/http/authn/BUILD
@@ -28,6 +28,7 @@ envoy_cc_library(
     name = "authenticator",
     srcs = [
         "authenticator_base.cc",
+        "authn_utils.cc",
         "filter_context.cc",
         "mtls_authentication.cc",
         "origin_authenticator.cc",
@@ -35,6 +36,7 @@ envoy_cc_library(
     ],
     hdrs = [
         "authenticator_base.h",
+        "authn_utils.h",
         "filter_context.h",
         "mtls_authentication.h",
         "origin_authenticator.h",
@@ -42,8 +44,8 @@ envoy_cc_library(
     ],
     repository = "@envoy",
     deps = [
-        ":authn_utils_lib",
         "//external:authentication_policy_config_cc_proto",
+        "//src/envoy/http/jwt_auth:jwt_lib",
         "//src/envoy/utils:utils_lib",
         "//src/istio/authn:context_proto",
     ],
@@ -64,22 +66,6 @@ envoy_cc_library(
         "//external:authentication_policy_config_cc_proto",
         "//src/envoy/utils:authn_lib",
         "//src/envoy/utils:utils_lib",
-        "//src/istio/authn:context_proto",
-        "@envoy//source/exe:envoy_common_lib",
-    ],
-)
-
-envoy_cc_library(
-    name = "authn_utils_lib",
-    srcs = [
-        "authn_utils.cc",
-    ],
-    hdrs = [
-        "authn_utils.h",
-    ],
-    repository = "@envoy",
-    deps = [
-        "//src/envoy/http/jwt_auth:jwt_lib",
         "//src/istio/authn:context_proto",
         "@envoy//source/exe:envoy_common_lib",
     ],
--- a/src/envoy/http/authn/BUILD
+++ b/src/envoy/http/authn/BUILD
@@ -119,6 +119,17 @@ envoy_cc_test(
 )
 
 envoy_cc_test(
+    name = "authn_utils_test",
+    srcs = ["authn_utils_test.cc"],
+    repository = "@envoy",
+    deps = [
+        ":authenticator",
+        ":test_utils",
+        "@envoy//test/test_common:utility_lib",
+    ],
+)
+
+envoy_cc_test(
     name = "peer_authenticator_test",
     srcs = ["peer_authenticator_test.cc"],
     repository = "@envoy",
--- a/src/envoy/http/authn/authenticator_base_test.cc
+++ b/src/envoy/http/authn/authenticator_base_test.cc
@@ -35,12 +35,8 @@ namespace {
 
 const std::string kSecIstioAuthUserInfoHeaderKey = "sec-istio-auth-userinfo";
 const std::string kSecIstioAuthUserinfoHeaderValue =
-    "eyJpc3MiOiI2Mjg2NDU3NDE4ODEtbm9hYml1MjNmNWE4bThvdmQ4dWN2Njk4bGo3OH"
-    "Z2MGxAZGV2ZWxvcGVyLmdzZXJ2aWNlYWNjb3VudC5jb20iLCJzdWIiOiI2Mjg2NDU3"
-    "NDE4ODEtbm9hYml1MjNmNWE4bThvdmQ4dWN2Njk4bGo3OHZ2MGxAZGV2ZWxvcGVyLm"
-    "dzZXJ2aWNlYWNjb3VudC5jb20iLCJhdWQiOiJib29rc3RvcmUtZXNwLWVjaG8uY2xv"
-    "dWRlbmRwb2ludHNhcGlzLmNvbSIsImlhdCI6MTUxMjc1NDIwNSwiZXhwIjo1MTEyNz"
-    "U0MjA1fQ==";
+    "eyJpc3MiOiJpc3N1ZXJAZm9vLmNvbSIsInN1YiI6InN1YkBmb28uY29tIiwiYXVkIj"
+    "oiYXVkMSIsImlhdCI6MTUxMjc1NDIwNSwiZXhwIjo1MTEyNzU0MjA1fQ==";
 
 class MockAuthenticatorBase : public AuthenticatorBase {
  public:
@@ -152,19 +148,18 @@ TEST_F(AuthenticatorBaseTest, ValidateJwtWithJwtInHeader) {
   JsonStringToMessage(
       R"({
              "jwt": {
-               "user": "628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com/628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com",
-               "audiences": ["bookstore-esp-echo.cloudendpointsapis.com"],
+               "user": "issuer@foo.com/sub@foo.com",
+               "audiences": ["aud1"],
                "presenter": "",
                "claims": {
-                 "aud": "bookstore-esp-echo.cloudendpointsapis.com",
-                 "iss": "628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com",
-                 "sub": "628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com"
+                 "aud": "aud1",
+                 "iss": "issuer@foo.com",
+                 "sub": "sub@foo.com"
                }
              }
            }
         )",
       &expected_payload, options);
-
   authenticator.validateJwt(
       jwt, [&expected_payload](const Payload* payload, bool success) {
         // When there is a verified JWT in the HTTP header, validateJwt()
--- a/src/envoy/http/authn/authenticator_base_test.cc
+++ b/src/envoy/http/authn/authenticator_base_test.cc
@@ -156,6 +156,7 @@ TEST_F(AuthenticatorBaseTest, ValidateJwtWithJwtInHeader) {
                "audiences": ["bookstore-esp-echo.cloudendpointsapis.com"],
                "presenter": "",
                "claims": {
+                 "aud": "bookstore-esp-echo.cloudendpointsapis.com",
                  "iss": "628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com",
                  "sub": "628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com"
                }
--- a/src/envoy/http/authn/authenticator_base_test.cc
+++ b/src/envoy/http/authn/authenticator_base_test.cc
@@ -14,6 +14,7 @@
  */
 
 #include "src/envoy/http/authn/authenticator_base.h"
+#include "common/common/base64.h"
 #include "common/protobuf/protobuf.h"
 #include "gmock/gmock.h"
 #include "src/envoy/http/authn/test_utils.h"
@@ -35,8 +36,15 @@ namespace {
 
 const std::string kSecIstioAuthUserInfoHeaderKey = "sec-istio-auth-userinfo";
 const std::string kSecIstioAuthUserinfoHeaderValue =
-    "eyJpc3MiOiJpc3N1ZXJAZm9vLmNvbSIsInN1YiI6InN1YkBmb28uY29tIiwiYXVkIj"
-    "oiYXVkMSIsImlhdCI6MTUxMjc1NDIwNSwiZXhwIjo1MTEyNzU0MjA1fQ==";
+    R"(
+     {
+       "iss": "issuer@foo.com",
+       "sub": "sub@foo.com",
+       "aud": "aud1",
+       "non-string-will-be-ignored": 1512754205,
+       "some-other-string-claims": "some-claims-kept"
+     }
+   )";
 
 class MockAuthenticatorBase : public AuthenticatorBase {
  public:
@@ -137,10 +145,20 @@ TEST_F(AuthenticatorBaseTest, ValidateJwtWithNoJwtInHeader) {
   });
 }
 
+Http::TestHeaderMapImpl CreateTestHeaderMap(const std::string& header_key,
+                                            const std::string& header_value) {
+  // The base64 encoding is done through Base64::encode().
+  // If the test input has special chars, may need to use the counterpart of
+  // Base64UrlDecode().
+  std::string value_base64 =
+      Base64::encode(header_value.c_str(), header_value.size());
+  return Http::TestHeaderMapImpl{{header_key, value_base64}};
+}
+
 TEST_F(AuthenticatorBaseTest, ValidateJwtWithJwtInHeader) {
   iaapi::Jwt jwt;
-  Http::TestHeaderMapImpl request_headers_with_jwt{
-      {kSecIstioAuthUserInfoHeaderKey, kSecIstioAuthUserinfoHeaderValue}};
+  Http::TestHeaderMapImpl request_headers_with_jwt = CreateTestHeaderMap(
+      kSecIstioAuthUserInfoHeaderKey, kSecIstioAuthUserinfoHeaderValue);
   FilterContext filter_context{&request_headers_with_jwt, &connection_};
   MockAuthenticatorBase authenticator{&filter_context};
   Payload expected_payload;
@@ -154,7 +172,8 @@ TEST_F(AuthenticatorBaseTest, ValidateJwtWithJwtInHeader) {
                "claims": {
                  "aud": "aud1",
                  "iss": "issuer@foo.com",
-                 "sub": "sub@foo.com"
+                 "sub": "sub@foo.com",
+                 "some-other-string-claims": "some-claims-kept"
                }
              }
            }
--- a/src/envoy/http/authn/authn_utils.cc
+++ b/src/envoy/http/authn/authn_utils.cc
@@ -22,68 +22,29 @@ namespace Http {
 namespace Istio {
 namespace AuthN {
 namespace {
-class JsonIterCallback : public Logger::Loggable<Logger::Id::filter> {
- public:
-  // The json iterator callback function for extracting Jwt claims
-  static bool JsonIteratorCallbackForJwtClaims(
-      const std::string&, const Envoy::Json::Object&,
-      istio::authn::JwtPayload* payload);
-  // The json iterator callback function for exracting Jwt audience
-  static bool JsonIteratorCallbackForJwtAudience(
-      const std::string&, const Envoy::Json::Object&,
-      istio::authn::JwtPayload* payload);
-};
-
-bool JsonIterCallback::JsonIteratorCallbackForJwtClaims(
-    const std::string& key, const Envoy::Json::Object& obj,
-    istio::authn::JwtPayload* payload) {
-  ENVOY_LOG(debug, "{}: key is {}", __FUNCTION__, key);
-  ::google::protobuf::Map< ::std::string, ::std::string>* claims =
-      payload->mutable_claims();
-  // In current implementation, only string objects are extracted into
-  // claims. If call obj.asJsonString(), will get "panic: not reached" from
-  // json_loader.cc.
-  try {
-    // Try as string, will throw execption if object type is not string.
-    std::string obj_str = obj.asString();
-    (*claims)[key] = obj_str;
-  } catch (Json::Exception& e) {
-  }
-  return true;
-}
-
-bool JsonIterCallback::JsonIteratorCallbackForJwtAudience(
-    const std::string& key, const Envoy::Json::Object& obj,
-    istio::authn::JwtPayload* payload) {
-  ENVOY_LOG(debug, "{}: key is {}", __FUNCTION__, key);
-  // Only extract audience
-  if (key != "aud") {
-    return true;
-  }
+// Extract JWT audience into the JwtPayload.
+// Return true if "aud" is extracted. Otherwise, return false.
+bool ExtractJwtAudience(const Envoy::Json::Object& obj,
+                        istio::authn::JwtPayload* payload) {
+  const std::string key = "aud";
   // "aud" can be either string array or string.
   // First, try as string, will throw execption if object type is not string.
   try {
-    std::string obj_str = obj.asString();
+    std::string obj_str = obj.getString(key);
     // Save "aud" to the payload
     payload->add_audiences(obj_str);
     return true;
   } catch (Json::Exception& e) {
     // Not convertable to string
   }
-  // Next, try as string array, read it as empty array if doesn't exist.
+  // Next, try as string array
   try {
-    // TODO (lei-tang): confirm the following code is a correct way to extract
-    // the string array.
-    std::string json_str = "{\"" + key + "\":" + obj.asJsonString() + "}";
-    Envoy::Json::ObjectSharedPtr json_ptr =
-        Envoy::Json::Factory::loadFromString(json_str);
-    std::vector<std::string> aud_vector = json_ptr->getStringArray(key, true);
+    std::vector<std::string> aud_vector = obj.getStringArray(key);
     for (size_t i = 0; i < aud_vector.size(); i++) {
       payload->add_audiences(aud_vector[i]);
     }
   } catch (Json::Exception& e) {
-    ENVOY_LOG(error, "aud field type is not string or string array");
-    ENVOY_LOG(error, "{}", e.what());
+    return false;
   }
   // return true to proceed to the next iteration.
   return true;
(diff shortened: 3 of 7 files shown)


src/envoy/http/authn/authn_utils.cc
```diff
@@ -56,9 +58,10 @@ bool JsonIterCallback::JsonIteratorCallback(const std::string& key,
         payload->add_audiences(aud_vector[i]);
       }
     } else {
-      // will throw execption if value type is not string.
+      // Will throw execption if value type is not string.
       // In current implementation, only string objects are extracted into
-      // claims.
+      // claims. If call obj.asJsonString(), will get "panic: not reached" from
+      // json_loader.cc.
       (*claims)[key] = obj.asString();
```
diemtvu: It would be cleaner to do this in a separate try block on top. Then the second part is just to extract aud. It's avoid nested try catch as well as easy to see all string claims are kept.

lei-tang:
Working on authn_utils_test, will update later.

lei-tang:
@diemtvu authn_utils_test added.

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -0,0 +1,71 @@
...
+const std::string kSecIstioAuthUserInfoHeaderKey = "sec-istio-auth-userinfo";
+const std::string kSecIstioAuthUserinfoHeaderValue =
+    "eyJpc3MiOiI2Mjg2NDU3NDE4ODEtbm9hYml1MjNmNWE4bThvdmQ4dWN2Njk4bGo3OH"
+    "Z2MGxAZGV2ZWxvcGVyLmdzZXJ2aWNlYWNjb3VudC5jb20iLCJzdWIiOiI2Mjg2NDU3"
+    "NDE4ODEtbm9hYml1MjNmNWE4bThvdmQ4dWN2Njk4bGo3OHZ2MGxAZGV2ZWxvcGVyLm"
+    "dzZXJ2aWNlYWNjb3VudC5jb20iLCJhdWQiOiJib29rc3RvcmUtZXNwLWVjaG8uY2xv"
+    "dWRlbmRwb2ludHNhcGlzLmNvbSIsImlhdCI6MTUxMjc1NDIwNSwiZXhwIjo1MTEyNz"
+    "U0MjA1fQ==";
+
+TEST(AuthnUtilsTest, GetJwtPayloadFromHeaderTest) {
```
diemtvu: It would be better to test the JsonIteratorCallback directly, with more thorough input cases (claims with string, array, and other primitive types etc)

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -0,0 +1,71 @@
...
+  JwtPayload payload, expected_payload;
+  Http::TestHeaderMapImpl request_headers_with_jwt{
+      {kSecIstioAuthUserInfoHeaderKey, kSecIstioAuthUserinfoHeaderValue}};
+  google::protobuf::util::JsonParseOptions options;
+  JsonStringToMessage(
```
diemtvu: You can use proto TextFormat::ParseFromString to create expected output message.

src/envoy/http/authn/authn_utils_test.cc
lei-tang: JsonIteratorCallback() is in an anonymous namespace, which is inaccessible from authn_utils_test.cc. Moreover, the const Envoy::Json::Object& parameter in JsonIteratorCallback(...const Envoy::Json::Object&...) is of type Field, which is also in an anonymous namespace and is only creatable in the original cc file.

diemtvu: Then you should create more test data to test other scenario (e.g aud as array, non string claims). I also don't like the base64 input; it hides all the information to check the test itself.

Alternatively, I suggest to add this function to the AuthUtils (as private static), and use mock  Json object for test.

lei-tang: Done.

src/envoy/http/authn/authn_utils_test.cc
lei-tang: Done.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -36,32 +36,41 @@ bool JsonIterCallback::JsonIteratorCallback(const std::string& key,
   ENVOY_LOG(debug, "{}: key is {}", __FUNCTION__, key);
   ::google::protobuf::Map< ::std::string, ::std::string>* claims =
       payload->mutable_claims();
+
+  // In current implementation, only string objects are extracted into
+  // claims. If call obj.asJsonString(), will get "panic: not reached" from
+  // json_loader.cc.
   try {
```
diemtvu: Keep try-catch block short, and avoid nested. Simplify this code to smt like:

try {
  std::string obj_str = obj.asString();
  (*claims)[key] = obj_str;
} catch (Json::Exception&) {
  // Ignore non-string claims.
}

if (key == "aud") {
  if (claims.find(key) != claims.end()) {
    payload->add_audiences(claims[key]);
  } else {
    try {
       aud_vector = obj.getStringArray(key, true);
       ....
    } catch (Json::Exception& e) {
      ENVOY_LOG(error, "aud field type is not string nor string array: {}", e.what());
    }
  }

src/envoy/http/authn/authn_utils.cc
```diff
@@ -36,32 +36,41 @@ bool JsonIterCallback::JsonIteratorCallback(const std::string& key,
...
       try {
-        aud_vector = obj.getStringArray(key, true);
-      } catch (Json::Exception& e) {
-        // Try as string
-        try {
-          auto audience = obj.asString();
-          aud_vector.push_back(audience);
-        } catch (Json::Exception& e) {
-          ENVOY_LOG(error, "aud field type is not string or string array");
+        std::string json_str = "{\"" + key + "\":" + obj.asJsonString() + "}";
```
diemtvu: Please confirm this implementation with Wayne or someone (may be lookup cloudendpoints repo to see how it was done). Or leave a TODO (and create a tracking issue for this to remember) if you unsure. This new change doesn't feel right to me.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -36,32 +36,41 @@ bool JsonIterCallback::JsonIteratorCallback(const std::string& key,
...
+        Envoy::Json::ObjectSharedPtr json_ptr =
+            Envoy::Json::Factory::loadFromString(json_str);
+        std::vector<std::string> aud_vector =
+            json_ptr->getStringArray(key, true);
+        for (size_t i = 0; i < aud_vector.size(); i++) {
+          ENVOY_LOG(debug, "aud_vector {} is: {}", i, aud_vector[i]);
```
diemtvu: Even with debug log, please keep only essential (to debug program). Seeing the value of aud claims here doesn't provide any useful information to see why jwt fail or not fail, and bloat the log a lot.

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -0,0 +1,164 @@
...
+      request_headers_with_jwt, LowerCaseString(kSecIstioAuthUserInfoHeaderKey),
+      &payload);
+  EXPECT_TRUE(ret);
+  EXPECT_TRUE(MessageDifferencer::Equals(expected_payload, payload));
+}
+
+TEST(AuthnUtilsTest, GetJwtPayloadFromHeaderWithNoAudTest) {
+  JwtPayload payload, expected_payload;
+  std::string value_base64 =
+      Base64::encode(kSecIstioAuthUserInfoHeaderWithNoAudValue.c_str(),
```
diemtvu: Can we use this one? I though jwt need to use base64url encode/decode?

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -0,0 +1,164 @@
...
+namespace Http {
+namespace Istio {
+namespace AuthN {
+namespace {
+
+const std::string kSecIstioAuthUserInfoHeaderKey = "sec-istio-auth-userinfo";
+const std::string kSecIstioAuthUserinfoHeaderValue =
+    R"(
+     {
+       "iss": "628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com",
```
diemtvu: Use fake, short id like foo, bar etc.

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc
lei-tang: Added the TODO. @qiwzhang : can you review the code mentioned in Diem's comment, which extracts the string array from aud?

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils_test.cc
lei-tang: There is no base64url encode function in istio/proxy repository; the existing code use base64 encode function to do the encoding. If there is no special chars in the input string that needed to be handled by base64url, base64 encode generates the same encoding result.

src/envoy/http/authn/authn_utils_test.cc
lei-tang: Done.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -91,11 +114,18 @@ bool AuthnUtils::GetJWTPayloadFromHeaders(
...
               json_obj->asJsonString());
+    // Extract claims
     json_obj->iterate(
         [payload](const std::string& key, const Json::Object& obj) -> bool {
-          return JsonIterCallback::JsonIteratorCallback(key, obj, payload);
+          return JsonIterCallback::JsonIteratorCallbackForJwtClaims(key, obj,
+                                                                    payload);
+        });
+    // Extract audience
+    json_obj->iterate(
```
diemtvu: Why do you need to iterate for aud? Just try getStringArray or getString with key = 'aud' directly. You can also check the claims map for "aud" first. If found, just use it as single aud, otherwise try getStringArray. So that, don't need to try getString for aud.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -24,45 +24,68 @@ namespace AuthN {
...
-  // The json iterator callback function
-  static bool JsonIteratorCallback(const std::string&,
-                                   const Envoy::Json::Object&,
-                                   istio::authn::JwtPayload* payload);
+  // The json iterator callback function for extracting Jwt claims
+  static bool JsonIteratorCallbackForJwtClaims(
+      const std::string&, const Envoy::Json::Object&,
+      istio::authn::JwtPayload* payload);
+  // The json iterator callback function for exracting Jwt audience
+  static bool JsonIteratorCallbackForJwtAudience(
```
diemtvu: Doesn't need this function. In fact, even the other can be lambda now, given it's so much simpler.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -24,45 +24,68 @@ namespace AuthN {
...
+      istio::authn::JwtPayload* payload);
 };
 
-bool JsonIterCallback::JsonIteratorCallback(const std::string& key,
-                                            const Envoy::Json::Object& obj,
-                                            istio::authn::JwtPayload* payload) {
+bool JsonIterCallback::JsonIteratorCallbackForJwtClaims(
+    const std::string& key, const Envoy::Json::Object& obj,
+    istio::authn::JwtPayload* payload) {
   ENVOY_LOG(debug, "{}: key is {}", __FUNCTION__, key);
```
diemtvu: Please remove this debug, and many similar. This make debugging actually harder as you need to dig through so many logs.

src/envoy/http/authn/authn_utils_test.cc
diemtvu: Put a comment for that. Make these lines 100 - 104 into a function that return HeaderMap (or set the header map in the test fixture) so that it's easier to change the encode in the future if need to.

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -0,0 +1,165 @@
...
+namespace Http {
+namespace Istio {
+namespace AuthN {
+namespace {
+
+const std::string kSecIstioAuthUserInfoHeaderKey = "sec-istio-auth-userinfo";
+const std::string kSecIstioAuthUserinfoHeaderValue =
+    R"(
+     {
+       "iss": "foo@developer.gserviceaccount.com",
```
diemtvu: Make domains and aud shorter too. Use some simple schema to make it easy to see what is what in the test expect statatement (e.g iss = "issuer@foo.com", sub = "sub@foo.com", aud = "aud1" etc)

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -0,0 +1,165 @@
...
+       "sub": "foo@developer.gserviceaccount.com",
+       "aud": "bookstore-esp-echo.cloudendpointsapis.com",
+       "iat": 1512754205,
```
diemtvu: If these claims doesn't have any special meaning for the test, name them something like "non-string-will-be-ignored" (and add a "some-other-string-claims" to see they are kept).

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils_test.cc
lei-tang: Done.

src/envoy/http/authn/authn_utils_test.cc
lei-tang: Done.

lei-tang:
@diemtvu Done. Cite: "Put a comment for that. Make these lines 100 - 104 into a function that return HeaderMap (or set the header map in the test fixture) so that it's easier to change the encode in the future if need to."

src/envoy/http/authn/authn_utils_test.cc
lei-tang: Done.

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -0,0 +1,180 @@
...
+       {
+         "iss": "issuer@foo.com",
+         "sub": "sub@foo.com",
+         "aud": ["aud1", "aud2"],
+         "non-string-will-be-ignored": 1512754205,
+         "some-other-string-claims": "some-claims-kept"
+       }
+     )";
+
+Http::TestHeaderMapImpl GetTestHeaderMap(std::string header_key,
```
diemtvu: If you not plan to use this function with key different than kSecIstioAuthUserInfoHeaderKey, then drop header_key params

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -0,0 +1,180 @@
...
+                                         std::string header_value) {
```
diemtvu: const std::string&

src/envoy/http/authn/authn_utils_test.cc
diemtvu: Create may be a better name.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -22,47 +22,31 @@ namespace Http {
...
-                                            const Envoy::Json::Object& obj,
-                                            istio::authn::JwtPayload* payload) {
-  ENVOY_LOG(debug, "{}: key is {}", __FUNCTION__, key);
-  ::google::protobuf::Map< ::std::string, ::std::string>* claims =
-      payload->mutable_claims();
+// Extract JWT audience into the JwtPayload.
+// Return true if "aud" is extracted. Otherwise, return false.
+bool ExtractJwtAudience(const Envoy::Json::Object& obj,
+                        istio::authn::JwtPayload* payload) {
+  const std::string key = "aud";
```
diemtvu: static const, or even constexpr

src/envoy/http/authn/authn_utils.cc
```diff
@@ -22,47 +22,31 @@ namespace Http {
...
-
-bool JsonIterCallback::JsonIteratorCallback(const std::string& key,
-                                            const Envoy::Json::Object& obj,
-                                            istio::authn::JwtPayload* payload) {
-  ENVOY_LOG(debug, "{}: key is {}", __FUNCTION__, key);
-  ::google::protobuf::Map< ::std::string, ::std::string>* claims =
-      payload->mutable_claims();
+// Extract JWT audience into the JwtPayload.
+// Return true if "aud" is extracted. Otherwise, return false.
+bool ExtractJwtAudience(const Envoy::Json::Object& obj,
```
diemtvu: Do you need to return bool ?

src/envoy/http/authn/authn_utils.cc
```diff
@@ -22,47 +22,31 @@ namespace Http {
...
+                        istio::authn::JwtPayload* payload) {
+  const std::string key = "aud";
+  // "aud" can be either string array or string.
+  // First, try as string, will throw execption if object type is not string.
+  try {
```
diemtvu: You can check claims["aud"] instead, as if it a string, it should be there already. It saves a try catch, which typically more expensive.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -22,47 +22,31 @@ namespace Http {
...
-      for (size_t i = 0; i < aud_vector.size(); i++) {
-        payload->add_audiences(aud_vector[i]);
-      }
-    } else {
-      // will throw execption if value type is not string.
-      // In current implementation, only string objects are extracted into
-      // claims.
-      (*claims)[key] = obj.asString();
+    std::vector<std::string> aud_vector = obj.getStringArray(key);
+    for (size_t i = 0; i < aud_vector.size(); i++) {
```
diemtvu: Use c11 syntax:

for (const auto& aud : aud_vector) {
   payload->add...
}

src/envoy/http/authn/authn_utils.cc
```diff
@@ -91,14 +75,24 @@ bool AuthnUtils::GetJWTPayloadFromHeaders(
...
+            // Try as string, will throw execption if object type is not string.
+            (*claims)[key] = obj.asString();
+          } catch (Json::Exception& e) {
+          }
+          return true;
         });
-
+    // Extract audience
+    ExtractJwtAudience(*json_obj, payload);
   } catch (...) {
```
diemtvu: Move this catch before iterate.

src/envoy/http/authn/BUILD
```diff
@@ -118,6 +118,17 @@ envoy_cc_test(
     ],
 )
 
+envoy_cc_test(
+    name = "authn_utils_test",
+    srcs = ["authn_utils_test.cc"],
+    repository = "@envoy",
+    deps = [
+        ":authenticator",
```
diemtvu: authn_util_lib

src/envoy/http/authn/authn_utils_test.cc
lei-tang: This function is used at three locations.

src/envoy/http/authn/authn_utils_test.cc
lei-tang: Done.

src/envoy/http/authn/authn_utils_test.cc
lei-tang: Done.

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc
lei-tang: Returning bool provides more information, i.e., whether audience is extracted. Though currently such information is not used, it makes this function more future-proof.

src/envoy/http/authn/authn_utils.cc
lei-tang: Using checking claims["aud"] instead assumes that ExtractJwtAudience() is called after extracting claims. Such calling order requirement tends to be error prone.

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc
diemtvu: There is no point to have the feature if you don't use it yet, it creates confusion for reader. Plus, the function is in the file scope only, so it's not much a hassle to change it when you need it.

src/envoy/http/authn/BUILD
lei-tang: #include "src/envoy/http/authn/authenticator_base.h" is required in authn_utils_test.cc and it is in ":authenticator"

src/envoy/http/authn/authn_utils.cc
lei-tang: Done: changed returning bool to void.

src/envoy/http/authn/BUILD
diemtvu: Why do you need that?
Anyway, if they have such a closed coupling, then maybe remove auth_utils lib, and add those files into authenticator directly.

diemtvu: I can't see why you need authentication_base.h for auth_utils_test still. It's not a big deal, but would be nice to make sure you include only what you need.

lei-tang: Done: authentication_base.h removed.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -22,51 +22,36 @@ namespace Http {
...
-      }
-      for (size_t i = 0; i < aud_vector.size(); i++) {
-        payload->add_audiences(aud_vector[i]);
-      }
-    } else {
-      // will throw execption if value type is not string.
-      // In current implementation, only string objects are extracted into
-      // claims.
-      (*claims)[key] = obj.asString();
+    std::string obj_str = obj.getString(key);
```
diemtvu: I forgot to check one of my previous comment. Why don't you want to check claims instead? try-catch is typically more expensive and we should try to avoid it when possible.

src/envoy/http/authn/BUILD
diemtvu: [optional] then you can break the library again :)

src/envoy/http/authn/authn_utils.cc
lei-tang: Using checking claims["aud"] instead assumes that ExtractJwtAudience() is called after extracting claims. Such calling order assumption tends to be error prone.

diemtvu: All calls in this file scope and simple enough to be tested thoroughly (as a matter of fact, doesn't really need a separate function in the first place). I would not sacrifice performance for that.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -87,21 +72,32 @@ bool AuthnUtils::GetJWTPayloadFromHeaders(
...
+          // Try as string, will throw execption if object type is not string.
+          (*claims)[key] = obj.asString();
+        } catch (Json::Exception& e) {
+        }
+        return true;
+      });
+  // Extract audience
+  ExtractJwtAudience(*json_obj, payload);
   if (payload->claims().empty()) {
     ENVOY_LOG(error, "{}: there is no JWT claims.", __func__);
```
JimmyCYJ: Is this log misleading if there are JWT claims but we fail to parse them as string and throw an exception at line 94?
Would it be necessary to distinguish errors between empty claim and failure of parsing claim?

src/envoy/http/authn/authenticator_base_test.cc
```diff
@@ -156,6 +156,7 @@ TEST_F(AuthenticatorBaseTest, ValidateJwtWithJwtInHeader) {
                "audiences": ["bookstore-esp-echo.cloudendpointsapis.com"],
                "presenter": "",
                "claims": {
+                 "aud": "bookstore-esp-echo.cloudendpointsapis.com",
                  "iss": "628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com",
                  "sub": "628645741881-noabiu23f5a8m8ovd8ucv698lj78vv0l@developer.gserviceaccount.com"
```
JimmyCYJ: We can avoid using these long claim values by generating a JWT payload for test. 
There are examples in proxy/src/envoy/http/jwt_auth/jwt_test.cc, and scripts in proxy/src/envoy/http/jwt_auth/tools/.
I am okay to leave it as is for now, but if we are going to add more tests, then it is better to keep these claims short for good readability.

JimmyCYJ:
LGTM, only a few nits.

src/envoy/http/authn/authenticator_base_test.cc
lei-tang: Done: long claim values are shortened.

src/envoy/http/authn/authn_utils.cc
lei-tang: Done: the log has been removed.

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc
```diff
@@ -22,51 +22,34 @@ namespace Http {
...
+// The JWT audience key name
+static const std::string kJwtAudienceKey = "aud";
 
-bool JsonIterCallback::JsonIteratorCallback(const std::string& key,
-                                            const Envoy::Json::Object& obj,
-                                            istio::authn::JwtPayload* payload) {
-  ENVOY_LOG(debug, "{}: key is {}", __FUNCTION__, key);
-  ::google::protobuf::Map< ::std::string, ::std::string>* claims =
-      payload->mutable_claims();
+// Extract JWT audience into the JwtPayload.
```
diemtvu: nit: line can be longer (100 chars, I think)

src/envoy/http/authn/authn_utils.cc
diemtvu: Good catch. I think we don't even need this check. 

If JWT filter output empty payload as its authentication result (it should not, but assume it does), then we should still consume the payload normally. The authN output will have no claims (nor principal, audiences etc), but it would be authZ to decide to reject request or not.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -22,51 +22,34 @@ namespace Http {
...
+// This function should to be called after
+// the claims are extracted into the payload.
+void ExtractJwtAudience(const Envoy::Json::Object& obj,
+                        istio::authn::JwtPayload* payload) {
+  const std::string& key = kJwtAudienceKey;
+  // "aud" can be either string array or string.
+  // First, try as string
+  if (payload->claims().count(key) > 0) {
```
qiwzhang: this code has dependency on the code in line 92.  it is better not to since it has "obj"
Please change this to use "obj" directly to check if "aud" as string.

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -0,0 +1,176 @@
...
+using google::protobuf::util::MessageDifferencer;
+using istio::authn::JwtPayload;
+
+namespace Envoy {
+namespace Http {
+namespace Istio {
+namespace AuthN {
+namespace {
+
+const std::string kSecIstioAuthUserInfoHeaderKey = "sec-istio-auth-userinfo";
```
qiwzhang: User LowerCaseString here

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils.cc

src/envoy/http/authn/authn_utils_test.cc
lei-tang: Done.

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -26,7 +26,7 @@ namespace Istio {
 namespace AuthN {
 namespace {
 
-const std::string kSecIstioAuthUserInfoHeaderKey = "sec-istio-auth-userinfo";
+const LowerCaseString kSecIstioAuthUserInfoHeaderKey("sec-istio-auth-userinfo");
```
qiwzhang: you can remove all these cast when calling GetJWTPayloadFromHeaders

lei-tang: Done.

PR #1322, opened 2018-03-29
Sanitize JWT verification result in HTTP headers

### LINKED ISSUE #1321: Sanitize JWT verification result in HTTP headers
To defend the attack of injecting forged JWT verification results in HTTP headers, sanitize the JWT verification result when jwt_auth filter starts.

lei-tang:
**What this PR does / why we need it**: To defend the attack of injecting forged JWT verification results in HTTP headers, sanitize the JWT verification result when jwt_auth filter starts.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1321

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/jwt_auth/http_filter.cc
+++ b/src/envoy/http/jwt_auth/http_filter.cc
@@ -25,6 +25,14 @@
 
 namespace Envoy {
 namespace Http {
+namespace {
+// This function sanitizes the HTTP headers to remove the existing
+// JWT verification results specified by the key.
+void SanitizeJwtVerificationResultInHeaders(const LowerCaseString key,
+                                            HeaderMap& headers) {
+  headers.remove(key);
+}
+}  // namespace
 
 JwtVerificationFilter::JwtVerificationFilter(Upstream::ClusterManager& cm,
                                              JwtAuth::JwtAuthStore& store)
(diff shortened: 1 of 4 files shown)


src/envoy/http/jwt_auth/http_filter.cc
```diff
@@ -25,6 +25,14 @@
 
 namespace Envoy {
 namespace Http {
+namespace {
+// This function sanitizes the HTTP headers to remove the existing
+// JWT verification results specified by the key.
+void SanitizeJwtVerificationResultInHeaders(const LowerCaseString key,
```
qiwzhang: I felt not need to have this as a separate function, it only has one line, and only called in one place.

lei-tang: I thought about that too. But I think using a function makes the code easier to understand and makes it easier to extend the sanitation code.

diemtvu: +1 on remove the function. Have to scroll up to see what it does hurt readability too. Add function back when you really need to do more than just headers.remove()

src/envoy/http/jwt_auth/http_filter.cc
```diff
@@ -43,6 +51,12 @@ FilterHeadersStatus JwtVerificationFilter::decodeHeaders(HeaderMap& headers,
   state_ = Calling;
   stopped_ = false;
 
+  // Sanitize the JWT verification result in the HTTP headers
+  // TODO (lei-tang): when the JWT verification result is in a configurable
+  // header, need to sanitize based on the configuration.
+  SanitizeJwtVerificationResultInHeaders(
```
diemtvu: Add unit test (in jwt_authenticator_test.cc) for this.

lei-tang: This change is not visible to JwtAuthenticator.

diemtvu: Sorry, I thought the change is in the jwt_authenticator. Ignore this then.

On a related note, there should be unit tests for http_filter. Some extra tasks to think about :)

PR #1330, opened 2018-03-30
Jwt-auth use envoy proto

qiwzhang:

Jwt-auth filter to directly use config.proto defined in envoyproxy/data-plane-api

**Release note**:

```release-note
None
```


src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -206,21 +207,11 @@ void JwtAuthenticator::VerifyKey(const PubkeyCacheItem& issuer_item) {
 }
 
 bool JwtAuthenticator::OkToBypass() {
```
diemtvu: What this bypass logic is about? Is this removed from the new API, or you just delay the implementation? If the former, just remove this function completely. Otherwise, add some comment why it's not needed yet. In fact, if it's not needed for common use cases, delete it still. We can wait to have it back with upstream version.

src/envoy/http/jwt_auth/http_filter_factory.cc
```diff
@@ -28,7 +30,7 @@ class JwtVerificationFilterConfig : public NamedHttpFilterConfigFactory {
   HttpFilterFactoryCb createFilterFactory(const Json::Object& config,
                                           const std::string&,
                                           FactoryContext& context) override {
-    Http::JwtAuth::Config::AuthFilterConfig proto_config;
+    JwtAuthentication proto_config;
```
diemtvu: Can we support both proto format? I.e move the ConvertProto from mixer to here, and try to parse json config to both proto type (only one would success, I guess)

It would be safer as we will need to sync with pilot changes (code change + proxy SHA update)? If it cause too much trouble, then we can plan to submit all PRs together.

qiwzhang: EndUserSpec is only used by Mixer.  Unless you want Pilot to use EndUserSpec to pass filter config.
Beside, I don't want this module to deal with EndUserSpec.

diemtvu: The new authn policy already insert jwt filter (not the one under mixer) using EUSpec proto. While it's still in private testing, I try to not breaking it neither. The least we should do is prepare changes in both sides and try to submit together (I'll do it today for pilot change anyway).

With respect to EUC in mixer, after this changes, I think we should start removing it (proto and pilot) to steer user to use authn policy.

PR #1352, opened 2018-03-30
use bazel --batch for release script

lizan:

Follow up on #1223, bazel server consumes more memory while it doesn't provide benefits in CI.
This will make postsubmit more stable.

**Release note**:
```release-note
None
```

### DIFF
--- a/script/release-binary
+++ b/script/release-binary
@@ -36,7 +36,7 @@ gsutil stat "${DST}/${BINARY_NAME}" \
   || echo 'Building a new binary.'
 
 # Build the release binary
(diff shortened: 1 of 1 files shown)


lizan:
cc @ldemailly @mandarjog per slack convo

lizan:
@ldemailly pasted the bazel help to PR description.

bazel usually run client/server, server keeps incremental build caches in memory to allow faster rebuilds. Though it doesn't make sense for CI under docker since we're not reusing them in subsequent CI builds. With --batch it saves some memory, so we're seeing less OOM.

ldemailly:
ty

PR #1364, opened 2018-04-05
Remove JWT headers consumed by Istio authn and mixer filters

### LINKED ISSUE #1363: Remove internal headers used to pass data between filters
From  jwt_authn:  we use "forward_payload_header" for each JWT token to forward Jwt Payload to istio_authn filter.

istio_authn should remove these headers.

from istio_authn filter, we use "sec-istio-authn-payload" to pass results to mixer filter.

mixer fitler should remove this header.

lei-tang:
**What this PR does / why we need it**: This PR removes the JWT headers after they have been consumed by Istio authn and mixer filters.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1363 

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/authn/http_filter.cc
+++ b/src/envoy/http/authn/http_filter.cc
@@ -74,6 +74,12 @@ void AuthenticationFilter::onPeerAuthenticationDone(bool success) {
 }
 void AuthenticationFilter::onOriginAuthenticationDone(bool success) {
   ENVOY_LOG(debug, "{}: success = {}", __func__, success);
+  // After Istio authn, the JWT headers consumed by Istio authn should be
+  // removed.
+  for (auto const iter : filter_config_.jwt_output_payload_locations()) {
+    filter_context_->headers()->remove(LowerCaseString(iter.second));
+  }
+
   if (success) {
     // Put authentication result to headers.
     if (filter_context_ != nullptr) {
--- a/src/envoy/http/authn/http_filter_integration_test.cc
+++ b/src/envoy/http/authn/http_filter_integration_test.cc
@@ -137,7 +137,7 @@ TEST_P(AuthenticationFilterIntegrationTest, CheckConsumedJwtHeadersAreRemoved) {
   const Envoy::Http::LowerCaseString header_location(
       "location-to-read-jwt-result");
   const std::string jwt_header =
-     R"(
+      R"(
      {
        "iss": "issuer@foo.com",
        "sub": "sub@foo.com",
--- a/src/envoy/http/authn/http_filter_integration_test.cc
+++ b/src/envoy/http/authn/http_filter_integration_test.cc
@@ -133,6 +133,48 @@ TEST_P(AuthenticationFilterIntegrationTest, CheckValidJwtPassAuthentication) {
   EXPECT_STREQ("200", response_->headers().Status()->value().c_str());
 }
 
+TEST_P(AuthenticationFilterIntegrationTest, CheckConsumedJwtHeadersAreRemoved) {
+  const Envoy::Http::LowerCaseString header_location(
+      "location-to-read-jwt-result");
+  const std::string jwt_header =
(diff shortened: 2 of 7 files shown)


src/envoy/http/mixer/filter.cc
```diff
@@ -122,6 +121,9 @@ FilterHeadersStatus Filter::decodeHeaders(HeaderMap& headers, bool) {
       [this](const Status& status) { completeCheck(status); });
   initiating_call_ = false;
 
+  // Remove Istio authentication header after Check()
+  Envoy::Utils::Authentication::ClearResultInHeader(&headers);
```
qiwzhang: The right place to remove should be in 
completeCheck()

qiwzhang: Can we have test for this removing?

lei-tang: It has been placed inside completeCheck(). As there is no testing for the mixer filter, per the discussion, this code in the mixer filter is not tested.

src/envoy/http/authn/http_filter.cc
```diff
@@ -74,6 +74,12 @@ void AuthenticationFilter::onPeerAuthenticationDone(bool success) {
 }
 void AuthenticationFilter::onOriginAuthenticationDone(bool success) {
   ENVOY_LOG(debug, "{}: success = {}", __func__, success);
+  // After Istio authn, the JWT headers consumed by Istio authn should be
```
diemtvu: Not always true, if peer authn fail, it will not reach origin auth part. Maybe you should put this in onDestroy (or alike) function.

src/envoy/http/authn/http_filter_integration_test.cc
```diff
@@ -133,6 +133,48 @@ TEST_P(AuthenticationFilterIntegrationTest, CheckValidJwtPassAuthentication) {
   EXPECT_STREQ("200", response_->headers().Status()->value().c_str());
 }
 
+TEST_P(AuthenticationFilterIntegrationTest, CheckConsumedJwtHeadersAreRemoved) {
```
diemtvu: It's a default behavior, so it's not necessary to crate separate test for this. You should add the check in all other tests to verify it is removed, regardless of request is accepted or rejected

src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -195,6 +195,9 @@ void JwtAuthenticator::VerifyKey(const PubkeyCacheItem& issuer_item) {
     return;
   }
 
+  // kJwtPayloadKey is preserved for backward compatibility.
+  // Please refer to https://github.com/istio/istio/issues/4744
```
diemtvu: Simply state: TODO(lei-tang): remove this backward compatibility. Tracking issue: ...

(the issue is for all tasks related to remove the old EUC policy, feel free to add note to that to remember).

src/envoy/http/authn/http_filter.cc
lei-tang: If peer authn fails,  rejectRequest("Peer authentication failed.") is called, the http client receives a 401 code, and the request will not proceed.

src/envoy/http/authn/http_filter_integration_test.cc
lei-tang: If the http request is rejected, the processing of the http request stops and the request will not be propagated to the downstream filters/backends. Thus, the removal is unnecessary as the entire http request is dropped.

src/envoy/http/jwt_auth/jwt_authenticator.cc
lei-tang: Done: https://github.com/istio/proxy/pull/1367

PR #1367, opened 2018-04-05
Add the support of the allow_tls option for Istio authn

### LINKED ISSUE #1365: mTLS authentication to consider allowTls flag
MutualTls params for authn policy now have `allowTls`, which is used to setup server to accept both mTLS and TLS connection. For TLS connection, there will be no x509, hence it cannot extract principal. However, it should accept request in that case.

lei-tang:
**What this PR does / why we need it**: This PR adds the support of the allow_tls option for Istio authn.
When allow_tls is true, the client certificate does not have to be present. If it is present, it will be extracted after the authentication.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1365 

**Release note**:

```release-note
```


src/envoy/http/authn/peer_authenticator.cc
```diff
@@ -47,7 +47,11 @@ void PeerAuthenticator::runMethod(const iaapi::PeerAuthenticationMethod& method,
                                   const MethodDoneCallback& done_callback) {
   switch (method.params_case()) {
     case iaapi::PeerAuthenticationMethod::ParamsCase::kMtls:
-      validateX509(method.mtls(), done_callback);
+      if (!method.mtls().allow_tls()) {
```
diemtvu: It shouldn't fork here, better inside the validateX509 (rename it if you want) and relax the condition for not having certificate.

If fork here, we need to keep both function in sync in the future, which I think is harder to maintain.

Plus, we want to keep this validateXXX is 1-1 mapping with each method supported by policy.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -61,6 +61,32 @@ void AuthenticatorBase::validateX509(
   done_callback(&payload, true);
 }
 
+void AuthenticatorBase::validateTls(
+    const iaapi::MutualTls&,
+    const AuthenticatorBase::MethodDoneCallback& done_callback) const {
+  // In TLS connection, a client certificate may not always be present.
+  // If the client certificate is present, extract its identity.
+  ENVOY_LOG(debug, "AuthenticatorBase: {} this connection allows TLS",
```
diemtvu: Please reduce this kind of debug message, we have too many. Delete old one when you come across as well.

qiwzhang:
@diemtvu @lei-tang please simply istio_authn filter first.   Please don't add any more new features to such over complicated code.  It is wasting everybody time.

diemtvu:
This feature is needed for @wattli, hopefully it is the last one before we do any thing else. Consider this is bug fix.

Btw, by simplify the code, do you mean to remove callback stuff. I'd like to wait of one release cycle or two to gather feedback if we don't need to combine filters. Though I won't object if we have cycle to do it now; not to hard to revert still.

src/envoy/http/authn/peer_authenticator.cc
lei-tang: Done.

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -36,6 +36,16 @@ AuthenticatorBase::~AuthenticatorBase() {}
 void AuthenticatorBase::done(bool success) const { done_callback_(success); }
 
 void AuthenticatorBase::validateX509(
+    const iaapi::MutualTls& mtls,
+    const AuthenticatorBase::MethodDoneCallback& done_callback) const {
+  if (mtls.allow_tls()) {
+    validateTls(mtls, done_callback);
```
diemtvu: I still don't think splitting function is needed. Simply check the flag in the place where you need to get principal. Smt like

```cc
   MtlsAuthentication mtls_authn(filter_context_.connection());
   if (!mtls_authn.IsTLS()) {
     done_callback(nullptr, false);
     return;	
   }
 	 
   Payload payload;
   const auto hasCert = mtls_authn.GetSourceUser(payload.mutable_x509()->mutable_user()));
   if (!hasCert && !mtls.allow_tls()) {
      done_callback(nullptr, false);
  }
```

Also, I find this MtlsAuthentication class is over-killed, all those functions are single line. Please remove it and just put them inline.

lei-tang: Separating the handling of mTLS and TLS makes the code easier to understand and less error prone. Will remove MtlsAuthentication class.

diemtvu: It doesn't. They are not deviate enough to be treated differently. Split them early will cause more maintenance cost to keep them in sync.

diemtvu: The first if can be just IsTLS() (the other one is covered by getsourceuser())

The rest is as my previous comment.

diemtvu: As matter of fact, even the first if can be removed completely, if adding a check for cert available in getsourceuser.

diemtvu: As matter of fact, even the first if can be removed completely, if adding a check for cert available in getsourceuser.

lei-tang: Can you provide your proposed validateX509() implementation in code?

diemtvu: Smt like this:

```cc
   if ( connection_ == nullptr || connection_->ssl() != nullptr)
     done_callback(nullptr, false);
     return;	
   }
 	 
   Payload payload;
   const auto hasCert = 
       connection_->ssl()->peerCertificatePresented() &&
       mtls_authn.GetSourceUser(payload.mutable_x509()->mutable_user()));

   if (!hasCert && !mtls.allow_tls()) {
      done_callback(nullptr, false);
  }
  done_callback(&payload, true);
```

lei-tang: The above code has the following problems:
1. when connection_->ssl() is non-null, the above code calls done_callback(nullptr, false) and returns. A fix for this problem is to change "connection_->ssl() != nullptr" into "connection_->ssl() == nullptr".

2. GetSourceUser() can return true while the user returned is empty, which is not handled by this code. A fix for this problem is to change GetSourceUser() to return false when the user is empty.

3. when !hasCert && !mtls.allow_tls(), the above code will first call done_callback(nullptr, false) then call done_callback(&payload, true). To fix this problem, after done_callback(nullptr, false), the code should return.

I will fix the above problems and commit a new version.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -36,29 +36,28 @@ AuthenticatorBase::~AuthenticatorBase() {}
...
     return;
   }
 
   Payload payload;
-  if (!mtls_authn.GetSourceUser(payload.mutable_x509()->mutable_user())) {
-    done_callback(&payload, false);
-  }
+  std::string user;
+  bool has_user = connection->ssl()->peerCertificatePresented() &&
+                  Utils::GetSourceUser(connection, &user) && !user.empty();
```
diemtvu: It's not required to check if user is empty (previous impl didn't). If such check require, it's better as part of GetSourceUser.

May be we should leave it out for now, and review the GetSourceUser later (even the check for peerCert above can be part of that too, and make this code cleaner).

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -36,29 +36,28 @@ AuthenticatorBase::~AuthenticatorBase() {}
...
 
-  // TODO (lei-tang): Adding other attributes (i.e ip) to payload if needed.
-  done_callback(&payload, true);
+  if (!has_user && !mtls.allow_tls()) {
+    // mTLS and no source user
+    done_callback(nullptr, false);
+  } else {
+    // TODO (lei-tang): Adding other attributes (i.e ip) to payload if needed.
```
diemtvu: Remove this TODO for now, there is no other attributes in x509 in immediate future.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -36,29 +36,28 @@ AuthenticatorBase::~AuthenticatorBase() {}
...
+    payload.mutable_x509()->set_user(user);
```
diemtvu: [optional] as mentioned above, you can call GetSourceUser directly output to this to avoid addition line of code.

src/envoy/http/authn/authenticator_base.cc
lei-tang: Done: after discussion with Wayne and Jimmy, the empty result check is added as part of GetSourceUser().

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authenticator_base.cc
lei-tang: With current implementation, no writing to payload unless the result is valid.

diemtvu: GetSourceUser doesn't modify the output if it fails, nor this payload object is used in that case. Thus it should be safe to call GetSourceUser directly on payload.mutable_user().

PR #1372, opened 2018-04-08
Support Mixer HTTP filter to report sent.bytes and received.bytes attributes

### LINKED ISSUE #1370: Support Mixer HTTP filter to send attributes 'sent_bytes' and 'received_bytes'.
Support Mixer HTTP filter to send attributes 'sent_bytes' and 'received_bytes' for *total* network traffic for HTTP. These would be distinct from request.size and response.size because they would include headers, etc., in the calculation. 

cc @douglas-reid @qiwzhang

JimmyCYJ:
**What this PR does / why we need it**:Support Mixer HTTP filter to send attributes "sent.bytes" and "received.bytes" in Report() calls. "sent.bytes" records total response sent in bytes, including response headers, body, and trailers. "received.bytes" records total request received in bytes, including request headers, body, and trailers.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1370 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/src/istio/control/attribute_names.cc
+++ b/src/istio/control/attribute_names.cc
@@ -29,6 +29,7 @@ const char AttributeName::kRequestPath[] = "request.path";
 const char AttributeName::kRequestReferer[] = "request.referer";
 const char AttributeName::kRequestScheme[] = "request.scheme";
 const char AttributeName::kRequestSize[] = "request.size";
+const char AttributeName::kReceivedBytes[] = "received.bytes";
 const char AttributeName::kRequestTime[] = "request.time";
 const char AttributeName::kRequestUserAgent[] = "request.useragent";
 const char AttributeName::kRequestApiKey[] = "request.api_key";
@@ -37,6 +38,7 @@ const char AttributeName::kResponseCode[] = "response.code";
 const char AttributeName::kResponseDuration[] = "response.duration";
 const char AttributeName::kResponseHeaders[] = "response.headers";
 const char AttributeName::kResponseSize[] = "response.size";
+const char AttributeName::kSentBytes[] = "sent.bytes";
 const char AttributeName::kResponseTime[] = "response.time";
 
 // TCP attributes
--- a/src/istio/control/attribute_names.cc
+++ b/src/istio/control/attribute_names.cc
@@ -28,8 +28,8 @@ const char AttributeName::kRequestMethod[] = "request.method";
 const char AttributeName::kRequestPath[] = "request.path";
 const char AttributeName::kRequestReferer[] = "request.referer";
 const char AttributeName::kRequestScheme[] = "request.scheme";
-const char AttributeName::kRequestSize[] = "request.size";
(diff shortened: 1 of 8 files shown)


src/istio/control/attribute_names.cc
```diff
@@ -28,6 +28,7 @@ const char AttributeName::kRequestPath[] = "request.path";
 const char AttributeName::kRequestReferer[] = "request.referer";
 const char AttributeName::kRequestScheme[] = "request.scheme";
 const char AttributeName::kRequestSize[] = "request.size";
+const char AttributeName::kReceivedBytes[] = "received.bytes";
```
qiwzhang: Hi @douglas-reid I felt it is better to name these two new attributes as "request.total_size" and "response.total_size" so it consists with existing attributes of "request.size"

qiwzhang:
Please also update this file https://github.com/istio/api/blob/master/mixer/v1/global_dictionary.yaml for the new attributes you added

qiwzhang:
and this document too.
https://istio.io/docs/reference/config/mixer/attribute-vocabulary.html

qiwzhang:
Please make sure to change Pilot code for the filter type change.

src/istio/control/attribute_names.cc
douglas-reid: @qiwzhang `request.total_size` and `response.total_size` sgtm.

src/istio/control/attribute_names.cc
```diff
@@ -28,15 +28,17 @@ const char AttributeName::kRequestMethod[] = "request.method";
...
 const char AttributeName::kRequestTime[] = "request.time";
 const char AttributeName::kRequestUserAgent[] = "request.useragent";
 const char AttributeName::kRequestApiKey[] = "request.api_key";
 
 const char AttributeName::kResponseCode[] = "response.code";
 const char AttributeName::kResponseDuration[] = "response.duration";
 const char AttributeName::kResponseHeaders[] = "response.headers";
-const char AttributeName::kResponseSize[] = "response.size";
+const char AttributeName::kResponseBodySize[] = "response.size";
+const char AttributeName::kResponseTotleSize[] = "response.totle_size";
```
douglas-reid: sorry to be late with this comment, but should these attributes be `request.total_size` and `response.total_size` ?

JimmyCYJ: request.total_size is added at line 32. Prior to this PR, we already let Mixer filter to report request.size and response.size, and they are actually request body size and response body size. So I update the variable name (e.g. s/kResponseSize/kResponseBodySize) to make it less confusing.

douglas-reid: @JimmyCYJ I'm actually wondering about the difference between `totle` and `total` (and what the name of the attribute is).

JimmyCYJ: That is typo. I will fix it soon. Thanks!

PR #1374, opened 2018-04-10
Add peer validation to ALTS implementation.

yangminzhu:
**What this PR does / why we need it:**

- Added the peer validation to ALTS implementation.
- Manually tested with custom config and environment on GKE.

None

**Release note:**
None

### DIFF
--- a/src/envoy/alts/tsi_transport_socket.cc
+++ b/src/envoy/alts/tsi_transport_socket.cc
@@ -91,17 +91,23 @@ Network::PostIoAction TsiSocket::doHandshakeNextDone(
                      std::string(peer.properties[i].value.data,
                                  peer.properties[i].value.length));
     }
-    std::string err = "";
-    bool peer_validated = handshake_validator_(peer, err);
-    tsi_peer_destruct(&peer);
-    if (peer_validated) {
-      ENVOY_CONN_LOG(info, "TSI: Handshake validation succeeded.",
-                     callbacks_->connection());
+    if (handshake_validator_) {
+      std::string err = "";
+      bool peer_validated = handshake_validator_(peer, err);
+      if (peer_validated) {
+        ENVOY_CONN_LOG(info, "TSI: Handshake validation succeeded.",
+                       callbacks_->connection());
+      } else {
+        ENVOY_CONN_LOG(warn, "TSI: Handshake validation failed: {}",
+                       callbacks_->connection(), err);
+        tsi_peer_destruct(&peer);
+        return Network::PostIoAction::Close;
+      }
     } else {
-      ENVOY_CONN_LOG(warn, "TSI: Handshake validation failed: {}",
-                     callbacks_->connection(), err);
-      return Network::PostIoAction::Close;
+      ENVOY_CONN_LOG(info, "TSI: Handshake validation skipped.",
+                     callbacks_->connection());
     }
+    tsi_peer_destruct(&peer);
 
     const unsigned char *unused_bytes;
     size_t unused_byte_size;


JimmyCYJ:
Please fix format error in proxy-presubmit test. You can run ./script/check-style to fix that.
And add PR description back. e.g. https://github.com/istio/proxy/pull/1019

src/envoy/alts/tsi_transport_socket.h
```diff
@@ -99,13 +106,15 @@ class TsiSocket : public Network::TransportSocket,
  */
 class TsiSocketFactory : public Network::TransportSocketFactory {
  public:
-  explicit TsiSocketFactory(HandshakerFactory handshaker_factory);
+  explicit TsiSocketFactory(HandshakerFactory handshaker_factory,
+                            HandshakeValidator handshake_validator);
```
JimmyCYJ: I think explicit is not needed anymore, as there are more than one argument.

src/envoy/alts/tsi_transport_socket.h
```diff
@@ -35,7 +37,8 @@ class TsiSocket : public Network::TransportSocket,
   /**
    * @param handshaker_factory a function to initiate a TsiHandshaker
    */
-  explicit TsiSocket(HandshakerFactory handshaker_factory);
+  explicit TsiSocket(HandshakerFactory handshaker_factory,
+                     HandshakeValidator handshake_validator);
```
JimmyCYJ: I think explicit is not needed anymore.

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -82,14 +85,23 @@ Network::PostIoAction TsiSocket::doHandshakeNextDone(
...
                    callbacks_->connection(), peer.property_count);
-    // TODO(lizan): validate peers
     for (size_t i = 0; i < peer.property_count; ++i) {
       ENVOY_CONN_LOG(debug, "  {}: {}", callbacks_->connection(),
                      peer.properties[i].name,
                      std::string(peer.properties[i].value.data,
                                  peer.properties[i].value.length));
     }
+    std::string err = "";
+    bool peer_validated = handshake_validator_(peer, err);
```
JimmyCYJ: Do we need to check if handshake_validator_ is nullptr?

src/envoy/alts/alts_socket_factory.cc
```diff
@@ -29,6 +29,36 @@ namespace Envoy {
...
+      for (int j = 0; j < targets.size(); ++j) {
+        if (value.compare(targets[j]) == 0) {
+          return true;
+        }
+      }
+    }
+  }
+
+  err = "Couldn't find peer's service account in targets: ";
+  for (int i = 0; i < targets.size(); ++i) {
```
qiwzhang: For easy debugging, it is better to log both peers and targets.

src/envoy/alts/tsi_transport_socket.h
yangminzhu: Done.

src/envoy/alts/tsi_transport_socket.h
yangminzhu: Thanks for the review, also updated the comments.

src/envoy/alts/tsi_transport_socket.cc
yangminzhu: I'm assuming it's used the same as handshake_factory which should always be provided by the caller. If the caller do not want to validate peer, they can pass in a lambda that always return true.

src/envoy/alts/alts_socket_factory.cc
yangminzhu: The peers will be logged in src/envoy/alts/tsi_transport_socket.cc just before the handshake_validator is called.

src/envoy/alts/tsi_transport_socket.cc
yangminzhu: After a second thought, I think it doesn't harm to add such check, so updated accordingly.

src/envoy/alts/tsi_transport_socket.h
```diff
@@ -24,6 +24,8 @@ namespace Envoy {
 namespace Security {
 
 typedef std::function<TsiHandshakerPtr(Event::Dispatcher&)> HandshakerFactory;
+typedef std::function<bool(const tsi_peer& peer, std::string& err)>
```
lizan: Add a comment for how they are used. err is an output param, right?

src/envoy/alts/tsi_transport_socket.cc
```diff
@@ -82,13 +85,28 @@ Network::PostIoAction TsiSocket::doHandshakeNextDone(
...
                    callbacks_->connection(), peer.property_count);
-    // TODO(lizan): validate peers
     for (size_t i = 0; i < peer.property_count; ++i) {
       ENVOY_CONN_LOG(debug, "  {}: {}", callbacks_->connection(),
                      peer.properties[i].name,
                      std::string(peer.properties[i].value.data,
                                  peer.properties[i].value.length));
     }
+    if (handshake_validator_) {
+      std::string err = "";
```
lizan: no need `= ""`

src/envoy/alts/alts_socket_factory.cc
```diff
@@ -29,6 +29,36 @@ namespace Envoy {
 namespace Server {
 namespace Configuration {
 
+using ::google::protobuf::RepeatedPtrField;
+
+// Returns true if the peer's service account is found in targets, otherwise
+// returns false and fills out err with an error message.
+static bool doValidate(const tsi_peer &peer,
+                       const RepeatedPtrField<std::string> &targets,
```
lizan: pre-process `targets` to a `unordered_set` then you don't have to do O(N^2) validation below.

src/envoy/alts/example.yaml
```diff
@@ -19,6 +19,9 @@ static_resources:
         name: alts
         config:
           handshaker_service: "169.254.169.254:8080"
+          # Please add the actual service account used in your environment, otherwise the ALTS
+          # connection will be closed due to validation failure.
+          target_service_accounts: ["test-service-account"]
```
lizan: Can you rename `target_service_accounts` to `peer_service_accountss`? Thanks!

PR #1383, opened 2018-04-10
Change target_service_accounts to peer_service_accounts in ALTS config proto.

yangminzhu:
- Change the target_service_accounts to peer_service_accounts in config proto, address some other comments.
- Skip validation if peer_service_accounts is empty.
- Manually tested with custom config (both validation success and fail) and environment on GKE.

**Which issue this PR fixes**:
Per request in #1374 .

None.

**Release note**:
None

### DIFF
--- a/src/envoy/alts/alts_socket_factory.cc
+++ b/src/envoy/alts/alts_socket_factory.cc
@@ -31,30 +31,27 @@ namespace Configuration {
 
 using ::google::protobuf::RepeatedPtrField;
 
-// Returns true if the peer's service account is found in targets, otherwise
+// Returns true if the peer's service account is found in peers, otherwise
 // returns false and fills out err with an error message.
 static bool doValidate(const tsi_peer &peer,
-                       const RepeatedPtrField<std::string> &targets,
+                       std::unordered_set<std::string> peers,
                        std::string &err) {
   for (size_t i = 0; i < peer.property_count; ++i) {
     std::string name = std::string(peer.properties[i].name);
     std::string value = std::string(peer.properties[i].value.data,
                                     peer.properties[i].value.length);
-    if (name.compare(TSI_ALTS_SERVICE_ACCOUNT_PEER_PROPERTY) == 0) {
-      for (int j = 0; j < targets.size(); ++j) {
-        if (value.compare(targets[j]) == 0) {
-          return true;
-        }
-      }
+    if (name.compare(TSI_ALTS_SERVICE_ACCOUNT_PEER_PROPERTY) == 0 &&
+        peers.count(value) > 0) {
+      return true;
     }
   }
 
-  err = "Couldn't find peer's service account in targets: ";
-  for (int i = 0; i < targets.size(); ++i) {
-    if (i != 0) {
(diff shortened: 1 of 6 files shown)


src/envoy/alts/alts_socket_factory.cc
```diff
@@ -97,9 +103,8 @@ UpstreamAltsTransportSocketConfigFactory::createTransportSocketFactory(
         return std::make_unique<Security::TsiHandshaker>(handshaker,
                                                          dispatcher);
       },
-      [peers](const tsi_peer &peer, std::string &err) {
-        return doValidate(peer, peers, err);
-      });
+      // Skip validation if peers is empty.
+      peers.empty() ? empty_validator : actual_validator);
```
lizan: you should be able to use `nullptr` here instead of empty_validator.

yangminzhu: It seems I have to use static_case to cast the nullptr to have the same type as actual_validator. I prefer not to use static_cast so refactored to not to use empty_validator.

src/envoy/alts/alts_socket_factory.cc
```diff
@@ -31,30 +31,27 @@ namespace Configuration {
 
 using ::google::protobuf::RepeatedPtrField;
 
-// Returns true if the peer's service account is found in targets, otherwise
+// Returns true if the peer's service account is found in peers, otherwise
 // returns false and fills out err with an error message.
 static bool doValidate(const tsi_peer &peer,
-                       const RepeatedPtrField<std::string> &targets,
+                       std::unordered_set<std::string> peers,
```
lizan: const reference

src/envoy/alts/alts_socket_factory.cc
```diff
@@ -31,30 +31,27 @@ namespace Configuration {
...
     std::string value = std::string(peer.properties[i].value.data,
                                     peer.properties[i].value.length);
-    if (name.compare(TSI_ALTS_SERVICE_ACCOUNT_PEER_PROPERTY) == 0) {
-      for (int j = 0; j < targets.size(); ++j) {
-        if (value.compare(targets[j]) == 0) {
-          return true;
-        }
-      }
+    if (name.compare(TSI_ALTS_SERVICE_ACCOUNT_PEER_PROPERTY) == 0 &&
+        peers.count(value) > 0) {
```
lizan: peers.find(value) != peers.end()

src/envoy/alts/alts_socket_factory.cc
```diff
@@ -31,30 +31,27 @@ namespace Configuration {
...
+      return true;
     }
   }
 
-  err = "Couldn't find peer's service account in targets: ";
-  for (int i = 0; i < targets.size(); ++i) {
-    if (i != 0) {
+  err = "Couldn't find peer's service account in peer_service_accounts: ";
+  for (auto it = peers.cbegin(); it != peers.cend(); ++it) {
```
lizan: Use absl::StrJoin?

yangminzhu: Thanks, that's much better for now. I didn't know that before.

PR #1400, opened 2018-04-11
Populate source.principal attribute with the same value as source.user

diemtvu:
This PR adds `source.principal` attribute with the same value as `source.user`. `source.user` will be removed in later PR to complete the migration.

Issue: https://github.com/istio/istio/issues/4689


qiwzhang:
some tests in istio repo under mixer/test/clients may fail.  please fix them after updating proxy sha to istio

PR #1465, opened 2018-04-19
Fix a bug that Istio authn filter blocks downstream filters

### LINKED ISSUE #1464: Fix a bug that Istio authn filter blocks downstream filters
Fix a bug that the Istio authn filter blocks downstream filters. With this bug, the processing of a HTTP request will not proceed to the filters after the Istio authn filter.

lei-tang:
**What this PR does / why we need it**: Fix a bug that the Istio authn filter blocks downstream filters. With this bug, the processing of a HTTP request will not proceed to the filters after the Istio authn filter.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1464

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/authn/http_filter.cc
+++ b/src/envoy/http/authn/http_filter.cc
@@ -77,7 +77,7 @@ FilterHeadersStatus AuthenticationFilter::decodeHeaders(HeaderMap& headers,
     Utils::Authentication::SaveResultToHeader(
         filter_context_->authenticationResult(), filter_context_->headers());
   }
-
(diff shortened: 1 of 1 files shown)


diemtvu:
do we still need to keep the state variable (was my comment in the previous refactor PR too)?

lei-tang:
I suggest creating a new issue to clean up/remove the state variable in Istio authn filter.

PR #1473, opened 2018-04-19
Change Mixer HTTP filter back to StreamDecoderFilter.

JimmyCYJ:
**What this PR does / why we need it**: Change Mixer HTTP filter back to StreamDecoderFilter.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -121,7 +121,7 @@ void Filter::ReadPerRouteConfig(
 
 FilterHeadersStatus Filter::decodeHeaders(HeaderMap& headers, bool) {
   ENVOY_LOG(debug, "Called Mixer::Filter : {}", __func__);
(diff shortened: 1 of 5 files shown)


src/envoy/http/mixer/filter.cc
```diff
@@ -259,8 +229,10 @@ void Filter::log(const HeaderMap* request_headers,
     CheckData check_data(*request_headers, nullptr);
     handler_->ExtractRequestAttributes(&check_data);
   }
-  ReportData report_data(response_headers, request_info, sent_bytes_,
-                         received_bytes_);
+  response_total_size_ =
```
qiwzhang: we don't need response_total_size as member variable. 
please add comment
// response trailer header is not counted to response total size.

JimmyCYJ: Done. Thanks.

Issue #1373, opened 2018-04-09
Support to use public key that is set directly in jwt config.


diemtvu:
Current implementation, public key is fetched from the jwksURI. We may want to make pilot responsible for this task, and send the public key in the jwt filter config (will need a proper design for pilot about caching and key rotation). Anyway, we will need proxy to be able to use the key.

PR #1476, opened 2018-04-19
Support inline jwks from config

qiwzhang:

Support sending jwks in the config.

Issue: https://github.com/istio/proxy/issues/1373
**Release note**:
```release-note
None
```

### DIFF
--- a/src/envoy/http/jwt_auth/pubkey_cache.h
+++ b/src/envoy/http/jwt_auth/pubkey_cache.h
@@ -18,6 +18,8 @@
 #include <chrono>
 #include <unordered_map>
 
+#include "common/common/base64.h"
+#include "common/common/logger.h"
 #include "envoy/config/filter/http/jwt_authn/v2alpha/config.pb.h"
 #include "src/envoy/http/jwt_auth/jwt.h"
 
@@ -36,7 +38,7 @@ const std::string kHTTPSSchemePrefix("https://");
 }  // namespace
 
 // Struct to hold an issuer cache item.
-class PubkeyCacheItem {
+class PubkeyCacheItem : public Logger::Loggable<Logger::Id::filter> {
  public:
   PubkeyCacheItem(
       const ::envoy::config::filter::http::jwt_authn::v2alpha::JwtRule&
@@ -45,6 +47,18 @@ class PubkeyCacheItem {
     // Convert proto repeated fields to std::set.
     for (const auto& aud : jwt_config_.audiences()) {
       audiences_.insert(SanitizeAudience(aud));
+    }
+
+    if (jwt_config_.has_local_jwks() &&
+        jwt_config_.local_jwks().specifier_case() ==
+            ::envoy::api::v2::core::DataSource::kInlineString) {
+      Status status =
+          SetKey(Base64::decode(jwt_config_.local_jwks().inline_string()));
+      if (status != Status::OK) {
+        ENVOY_LOG(warn, "Invalid inline jwks for issuer: {}, jwks: {}",
+                  jwt_config_.issuer(),
+                  jwt_config_.local_jwks().inline_string());
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/envoy/http/jwt_auth/pubkey_cache.h
```diff
@@ -46,6 +48,18 @@ class PubkeyCacheItem {
     for (const auto& aud : jwt_config_.audiences()) {
       audiences_.insert(SanitizeAudience(aud));
     }
+
+    if (jwt_config_.has_local_jwks() &&
+        jwt_config_.local_jwks().specifier_case() ==
+            ::envoy::api::v2::core::DataSource::kInlineString) {
+      Status status =
+          SetKey(Base64::decode(jwt_config_.local_jwks().inline_string()));
```
diemtvu: So do you expect the key is base64 encoded?

quanjielin: does it mean pilot side needs encoding the key ?

qiwzhang: Yes,  Envoy is doing a lot of JsonConvert for filter config.   We could not just put json as value here

lizan: Please don't do this, if you want inline them with base64 in YAML/JSON, just use `inline_bytes`, by proto/json spec they are base64 encoded. Don't mix them up.

src/envoy/http/jwt_auth/pubkey_cache.h
```diff
@@ -46,6 +48,18 @@ class PubkeyCacheItem {
     for (const auto& aud : jwt_config_.audiences()) {
       audiences_.insert(SanitizeAudience(aud));
     }
+
+    if (jwt_config_.has_local_jwks() &&
```
lizan: Instead of the whole block here, just use [`Envoy::Config::Datastore::read`](https://github.com/envoyproxy/envoy/blob/master/source/common/config/datasource.h#L16) and you're good with all case.

qiwzhang: Done

src/envoy/http/jwt_auth/pubkey_cache.h
```diff
@@ -84,12 +98,17 @@ class PubkeyCacheItem {
     pubkey_ = std::move(pubkey);
 
     expiration_time_ = std::chrono::steady_clock::now();
-    if (jwt_config_.remote_jwks().has_cache_duration()) {
-      const auto& duration = jwt_config_.remote_jwks().cache_duration();
-      expiration_time_ += std::chrono::seconds(duration.seconds()) +
-                          std::chrono::nanoseconds(duration.nanos());
+    if (jwt_config_.has_remote_jwks()) {
```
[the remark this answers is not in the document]
qiwzhang: Done

src/envoy/http/jwt_auth/pubkey_cache.h
qiwzhang: Done

src/envoy/http/jwt_auth/pubkey_cache.h
```diff
@@ -46,6 +49,17 @@ class PubkeyCacheItem {
     for (const auto& aud : jwt_config_.audiences()) {
       audiences_.insert(SanitizeAudience(aud));
     }
+
+    auto inline_jwks = Config::DataSource::read(jwt_config_.local_jwks(), true);
+    if (!inline_jwks.empty()) {
+      Status status = SetKey(Base64::decode(inline_jwks),
```
lizan: Not Base64::decode?

qiwzhang: I think it is still better to base64 encod it.  Here is my concern, Envoy is using JsonConvert() to convert Struct or Protobuf to other config proto.   If we don't base64 encode jwks,  jwks is a JSON,  it will break.  It is safer to base64 encode the jwks.

lizan: No please dont, it won't break JSON encoding, a string will be escaped correctly. If you want base64 encoded one, just use inline_bytes, which is by proto spec. You don't want to load base64 encoded from DataSource file. please revert.

PR #1480, opened 2018-04-19
Support per-route config with new Envoy method

qiwzhang:

Envoy has a new perFilterConfig method.  Change Proxy code to use the new method

**Release note**:

```release-note
None
```

### DIFF
--- a/src/envoy/http/mixer/filter_factory.cc
+++ b/src/envoy/http/mixer/filter_factory.cc
@@ -14,6 +14,7 @@
  */
 
 #include "common/config/utility.h"
+#include "common/protobuf/utility.h"
 #include "envoy/json/json_object.h"
 #include "envoy/registry/registry.h"
 #include "envoy/server/filter_config.h"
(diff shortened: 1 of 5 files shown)


src/envoy/http/mixer/filter_factory.cc
```diff
@@ -52,9 +53,15 @@ class MixerConfigFactory : public NamedHttpFilterConfigFactory {
...
     return ProtobufTypes::MessagePtr{new HttpClientConfig};
   }
-  ProtobufTypes::MessagePtr createEmptyRouteConfigProto() override {
-    return ProtobufTypes::MessagePtr{new ServiceConfig};
+
+  Router::RouteSpecificFilterConfigConstSharedPtr
+  createRouteSpecificFilterConfig(const ProtobufWkt::Struct& config) override {
+    auto obj = std::make_shared<Http::Mixer::PerRouteServiceConfig>();
+    MessageUtil::jsonConvert(config, obj->config);
+    obj->hash = std::to_string(MessageUtil::hash(obj->config));
```
rshriram: why do you need the hash? you convert into a typed object. So I see no point in hashing

qiwzhang: This is the easiest way to hook up per-route config in MixerClient.  Ideally, it is this object (https://github.com/istio/proxy/blob/master/src/istio/control/http/service_context.h#L30) that needed to be created for per route config.  But it is deep inside MixerClient,  MixerClient other objects are per-thread,  it is not easy to mix per-thread objects with a global object created at config time. 

Well, it requires some refractory of MixerClient in order not to use hash.

PR #1562, opened 2018-04-26
#Performance issue in mixer Check cache about signature computing

### LINKED ISSUE #1531: Performance issue in mixer Check cache
hi, I found some performance issue in the code of mixer cache.

In check_cache.cc, about line 100, method Check():

```c++
for (const auto &it : referenced_map_) {
    const Referenced &reference = it.second;
    std::string signature;
    if (!reference.Signature(attributes, "", &signature)) {
      continue;
    }
......
}
```
Here we compute the signature in the for loop? So, if we have many entries in referenced_map, we have to go through and call the signature many times, and almost all of these call return false except last one.

Then in this Signature() method, in referenced.cc, we do two things in this method:

1. check if match or not
2. if match, compute the signature

From the logic in Check() method we get that this method should always return false.

So let see the code in Signature() 😀 

```c++
utils::MD5 hasher;

  for (std::size_t i = 0; i < exact_keys_.size(); ++i) {
    const auto &key = exact_keys_[i];
    const auto it = attributes_map.find(key.name);
    // if an "exact" attribute not present, return false for mismatch.
    if (it == attributes_map.end()) {
      return false;
    }

    hasher.Update(it->first);
    hasher.Update(kDelimiter, kDelimiterLength);

    const Attributes_AttributeValue &value = it->second;
    switch (value.value_case()) {
      case Attributes_AttributeValue::kStringValue:
        hasher.Update(value.string_value());
        break;
utils::MD5 hasher;

  for (std::size_t i = 0; i < exact_keys_.size(); ++i) {
    const auto &key = exact_keys_[i];
    const auto it = attributes_map.find(key.name);
    // if an "exact" attribute not present, return false for mismatch.
    if (it == attributes_map.end()) {
      return false;
    }

    hasher.Update(it->first);
    hasher.Update(kDelimiter, kDelimiterLength);

    const Attributes_AttributeValue &value = it->second;
    switch (value.value_case()) {
      case Attributes_AttributeValue::kStringValue:
        hasher.Update(value.string_value());
        break;
```

In the implementation, we call the hasher.Update() and check in the same for loop! And the hasher.Update() will finally invoke the openssh/md5.cc to compute the md5, this method is very heavy!

So if this method finally return false, then all of these hasher.Update() will waste our cpu and memory and no benefit. And this hasher.Update() will be invoked many times in Signature() method, also Signature() method maybe be invoked in many times in Check() method. And Check() method will be invoked in every traffic.

I suggest to improve this Signature method, to do the exact key check loop two times:

1. first loop, just check if match, don't do the hasher.Update()
2. Only if the result if true in the first loop, we will do the second loop for exact keys to do the hasher.Update and calculate the signature.

If you agree with this improvement, I will submit a PR soon.

skyao:

Fix the bug (or performance issue) in the signature computing in mixer Check cache.

> Please review this PR and issue, It may be a BIG BIG performance issue.

`fixes #1531`

Please help to check if this line is OK:

```c++
if (value.value_case() == Attributes_AttributeValue::kStringMapValue) {
}
```
I'm sorry that I'm not a c++ developer so I don't know should I used == or equals() method. 
Correct it if some mistaken!

### DIFF
--- a/src/istio/mixerclient/referenced.cc
+++ b/src/istio/mixerclient/referenced.cc
@@ -153,8 +153,8 @@ bool Referenced::Signature(const Attributes &attributes,
     } while (true);
   }
 
-  utils::MD5 hasher;
-
+  // First loop for exact_keys
+  // DON‘T DO Signature in this loop!
   for (std::size_t i = 0; i < exact_keys_.size(); ++i) {
     const auto &key = exact_keys_[i];
     const auto it = attributes_map.find(key.name);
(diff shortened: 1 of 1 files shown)


qiwzhang:
It seems there is some code format error.

Files not formatted: src/istio/mixerclient/referenced.cc

Could you run script/check-style

src/istio/mixerclient/referenced.cc
```diff
@@ -153,8 +153,8 @@ bool Referenced::Signature(const Attributes &attributes,
     } while (true);
   }
 
-  utils::MD5 hasher;
-
+  // First loop for exact_keys
+  // DON‘T DO Signature in this loop!
   for (std::size_t i = 0; i < exact_keys_.size(); ++i) {
```
lizan: Can you factor out this loop pattern out to a method? There is already 3 loops (one for absence_keys, two for exact_keys) of similar pattern.

skyao:
hi, all,

I'm sorry that I'm NOT a c++ developer and I don't have development env to complete this PR, especially for the code format. I have check every line by my eyes but it still reports "no formatted"

Can some body help to check if this fix is OK and do left code format or more code refactory? You can consider my code as "fake code" and do your fix.

qiwzhang:
https://github.com/istio/proxy/pull/1581 to replace this PR

PR #1593, opened 2018-04-28
Create child spans for mixer check calls

bianpengyuan:


bianpengyuan:
Now tracing looks like this:
screen shot 2018-04-27 at 11 53 33 pm

cc @douglas-reid @kyessenov

mandarjog:
This seems odd. The labels say “report” while calling istio-policy, should say check?

bianpengyuan:
@mandarjog Yes, that is a separated problem. I am going to fix it as well, and I think the problem is in policy's sidecar config?

kyessenov:
@mandarjog Yes, that's a bug. Check/Report are swapped: https://github.com/istio/istio/blob/master/pilot/docker/envoy_telemetry.yaml.tmpl

kyessenov:
PR out https://github.com/istio/istio/pull/5283

kyessenov:
Why is there two Check calls on the first hop?

bianpengyuan:
@kyessenov One from Ingress and one from productpage sidecar.

douglas-reid:
Reminder: These PRs should be against the 0.8 release branch, instead of master. Please re-open against the correct branch (as master is now locked).

douglas-reid:
Should ingress itself also not have a span?

kyessenov:
Yes, ingress tracing has issues, too, I think. My expectation is that the
check should be on a span that reaches ingress service, not the outbound
call from ingress. That's not how it looks on the chart. Additionally, we
should probably try a gateway as well.

bianpengyuan:
Close this one with another one open in 0.8 branch: #1610

PR #1610, opened 2018-04-28
Create a child span for mixer check call

bianpengyuan:


rshriram:
Plese see this:
get the active span from here: https://github.com/envoyproxy/envoy/blob/master/include/envoy/http/filter.h#L135
and spawn a child with this:
https://github.com/envoyproxy/envoy/blob/master/include/envoy/tracing/http_tracer.h#L100

bianpengyuan:
@rshriram @qiwzhang Updated. Ptal.

bianpengyuan:
@kyessenov @douglas-reid @mandarjog Just FYI the new trace will look like this:
screenshot from 2018-05-01 11-57-21
and an individual check span:
screenshot from 2018-05-01 11-57-51
The report/check name problem will be resolved by istio/istio#5335. If we want to separate ingress out of the first hop, [start_child_span](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/http/router/v2/router.proto) needs to be true in ingress router config.

kyessenov:
The trace looks awesome. Exactly what I would want to see from it (latency
between hops).

bianpengyuan:
@kyessenov envoy grpc client library manages span creation: https://github.com/envoyproxy/envoy/blob/HEAD/source/common/grpc/google_async_client_impl.cc#L412

bianpengyuan:
I can test the option and see if that works.

kyessenov:
That would mean that would be done from within mixerfilter in ingress? I'm
not positive that's a good idea to treat ingress specially. Besides, there
is now ingressgateway, which can act as a middle proxy. I think what's
happening here is that ingress initiates the trace, so we're missing the
client of the mesh in the picture.

douglas-reid:
If we use `start_child_span`, what does the picture look like? Do we have an easy way to see an example?

In these particular traces, the ultimate client is outside the mesh (we haven't tried showing a trace in a backend like stackdriver where the caller could have added spans in the distributed trace as well, for instance). 

Having ingress not shown at all, when it introduces an additional hop (and its own routing latency) feels a bit odd to me, but I'm honestly not sure what is expected. I would certainly expect the gateway to show up, especially if routed to via some ingress.

rshriram:
Nm ignore. Thanks for doing it the clean way.

mandarjog:
@geeknoid this PR is ready to merge.
@hklai the merge bot is indefinitely spamming.

kyessenov:
@geeknoid the issue is https://github.com/istio/issues/issues/306 (and other instances, this was debugged for awhile). I think it's pretty important, since without the fix, we get incorrect traces (a regression from 0.7).

yutongz:
Once again, please DO NOT change the setting of a robot. A robot need right access to do its job.

wx20180502-093913 2x

hklai:
It might have been changed in the attempt to stop the merge from merging to master, which is now locked due to code orange.

Let's just disable the merge bot also for istio/proxy for now (already done that for istio/istio).

yutongz:
Please contract @istio/test-infra-hackers for the help and behavior change of istio robots.

yutongz:
Disabled merge robot

douglas-reid:
@bianpengyuan @kyessenov what are the chances we can get the updated proxy image with these changes in now as well?

PR #1733, opened 2018-05-04
Use mTLS mode enum to decide authentication result.

diemtvu:
We want to support different authentication mode in order to support mTLS migrations.

Issue: https://github.com/istio/istio/issues/5059


src/envoy/http/authn/authenticator_base.cc
```diff
@@ -34,16 +34,33 @@ AuthenticatorBase::~AuthenticatorBase() {}
...
+  // - STRICT: must be TLS with valid certificate.
+  switch (mtls.mode()) {
+    case iaapi::MutualTls::PERMISSIVE:
+      return true;
+    case iaapi::MutualTls::TLS_PERMISSIVE:
+      return connection->ssl() != nullptr;
+    case iaapi::MutualTls::STRICT:
+      return has_user;
+    default:
+      ENVOY_LOG(warn, "Invalid mTLS mode {}",
```
qiwzhang: use NOT_REACHED 

defined in common/common/assert.h

diemtvu: Done

PR #1751, opened 2018-05-15
Optimize Referenced::Signature function

qiwzhang:

Use extra loop to optimize signature calculation.

**Release note**:

```release-note
None
```

### DIFF
--- a/src/istio/mixerclient/referenced.cc
+++ b/src/istio/mixerclient/referenced.cc
@@ -117,11 +117,7 @@ bool Referenced::Fill(const Attributes &attributes,
 bool Referenced::Signature(const Attributes &attributes,
                            const std::string &extra_key,
                            std::string *signature) const {
-  if (!CheckAbsenceKeys(attributes)) {
-    return false;
-  }
-
-  if (!CheckExactKeys(attributes)) {
+  if (!CheckAbsenceKeys(attributes) || !CheckExactKeys(attributes)) {
     return false;
   }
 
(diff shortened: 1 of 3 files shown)


src/istio/mixerclient/referenced.cc
```diff
@@ -117,8 +117,20 @@ bool Referenced::Fill(const Attributes &attributes,
 bool Referenced::Signature(const Attributes &attributes,
                            const std::string &extra_key,
                            std::string *signature) const {
-  const auto &attributes_map = attributes.attributes();
+  if (!CheckAbsenceKeys(attributes)) {
```
lizan: Combine two ifs

src/istio/mixerclient/referenced.cc
```diff
@@ -142,27 +154,64 @@ bool Referenced::Signature(const Attributes &attributes,
...
     const auto it = attributes_map.find(key.name);
-    // if an "exact" attribute not present, return false for mismatch.
+    // If an "exact" attribute not present, return false for mismatch.
     if (it == attributes_map.end()) {
       return false;
     }
 
+    const Attributes_AttributeValue &value = it->second;
+    if (value.value_case() == Attributes_AttributeValue::kStringMapValue) {
+      std::string map_key = key.map_key;
```
lizan: const &

src/istio/mixerclient/referenced.h
```diff
@@ -49,6 +49,17 @@ class Referenced {
   std::string DebugString() const;
 
  private:
+  // Return true if all absent keys are not in the attributes.
+  bool CheckAbsenceKeys(const ::istio::mixer::v1::Attributes &attributes) const;
```
JimmyCYJ: s/CheckAbsenceKeys/CheckAbsentKeys ?

src/istio/mixerclient/referenced.cc
qiwzhang: Could not,  map_key is changed later.

src/istio/mixerclient/referenced.h
qiwzhang: Done

PR #1752, opened 2018-05-15
Make release-binary script takes argument for gcs destination

SpringUpenn:
Hi folks,

To istio/proxy git on borg release, release-binary needs to take gcs bucket parameter, it has no impact in OSS istio/proxy release, besides, for internal proxy release script, it can pass another destination instead of gs://istio-build/proxy.

This PR is tested locally which can build proxy outputs with my own GCB account.

Thanks for your review.

### DIFF
--- a/script/release-binary
+++ b/script/release-binary
@@ -17,13 +17,30 @@
 ################################################################################
 #
 set -ex
+# The bucket name to store proxy binary
+DST="gs://istio-build/proxy"
+
+function usage() {
+  echo "$0
+    -d  The bucket name to store proxy binary (optional)"
+  exit 1
+}
+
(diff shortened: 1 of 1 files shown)


script/release-binary
```diff
@@ -18,13 +18,28 @@
...
+  case "${arg}" in
+    d) DST="${OPTARG}";;
+    *) usage;;
+  esac
+done
+
+echo "Destination bucket: $DST"
+
+# The bucket name to store proxy binary
+# DST="gs://istio-build/proxy"
```
qiwzhang: Can we keep 

DST="gs://istio-build/proxy"

as default for istio

if -d flag is not set, DST will use the default

PR #1757, opened 2018-05-18
update envoy sha for sni

ijsnellf:
Updates Envoy SHA so that SNI is available. SNI support was merged here: envoyproxy/envoy#3217

I am not sure about the update process for istio/proxy, so please double check this PR.

**Release note**:

```release-note
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -38,7 +38,7 @@ git_repository(
 )
 
 # When updating envoy sha manually please update the sha in istio.deps file also
-ENVOY_SHA = "2b2c299144600fb9e525d21aabf39bf48e64fb1f"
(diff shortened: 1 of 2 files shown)


lizan:
It looks like it is due to the change to method signature of `sendLocalReply`, so the code change should be easy. Though if you're not familiar with the code base, someone from proxy can handle it.

ijsnellf:
I can give it a try, but it would probably make more sense for someone with familiarity with the code base to make the changes.

rlenglet:
#1758 did those changes but was closed before merging.

qiwzhang:
Replace it with https://github.com/istio/proxy/pull/1759

PR #1759, opened 2018-05-21
Release-0.8: Update envoy sha to bb6762a058

qiwzhang:
**Release note**:
```release-note
None
```

PR #1761, opened 2018-05-21
Update envoy sha to bb6762a058

qiwzhang:

**Release note**:

```release-note
None
```


qiwzhang:
It is blocked by the same integration test failures as https://github.com/istio/proxy/pull/1759

PR #1763, opened 2018-05-21
Add debug symbol to release image for easy debug

qiwzhang:

This is essential to debug Envoy crash in production.

**Release note**:
```release-note
None
```

### DIFF
--- a/tools/bazel.rc
+++ b/tools/bazel.rc
@@ -49,7 +49,6 @@ test --test_env=HEAPCHECK=normal --test_env=PPROF_PATH
 
 # Release builds
 build:release -c opt
-build:release --strip=always
 
 # Add compile option for all C++ files
 build --cxxopt -Wnon-virtual-dtor


hklai:
Since this crash is not blocking 0.8.0, can we do this in master (or in your own branch) instead?

ldemailly:
I agree that stripping is very premature, I think I commented as much in the PR that introduced that

qiwzhang:
@hklai it is better to ad this back in 0.8 branch.  This is for any future Envoy crashes in 0.8 release.

ldemailly:
?

qiwzhang:
Will build 3 artifacts:  debug-proxy,  release-proxy (no debug symbols) and release-proxy-with-symbol 
the second one will be used in docker image,  if crash, the last one can be used for debugging.  it will be pushed to gs://istio-build/proxy

ldemailly:
ic, thanks for doing this!

qiwzhang:
https://github.com/istio/proxy/pull/1767

PR #1765, opened 2018-05-22
UID override in EDS

kyessenov:
This PR allows overriding `destination.uid` per host in EDS response.

### DIFF
--- a/src/envoy/utils/utils.cc
+++ b/src/envoy/utils/utils.cc
@@ -70,6 +70,20 @@ bool GetIpPort(const Network::Address::Ip* ip, std::string* str_ip, int* port) {
   return false;
 }
 
+bool GetDestinationUID(const envoy::api::v2::core::Metadata& metadata,
+                       std::string* uid) {
+  const auto filter_it = metadata.filter_metadata().find("mixer");
+  if (filter_it == metadata.filter_metadata().end()) {
+    return false;
+  }
+  const auto fields_it = filter_it->second.fields().find("destination.uid");
+  if (fields_it == filter_it->second.fields().end()) {
+    return false;
+  }
+  *uid = fields_it->second.string_value();
+  return true;
+}
+
 bool GetSourceUser(const Network::Connection* connection, std::string* user) {
   if (connection) {
     Ssl::Connection* ssl = const_cast<Ssl::Connection*>(connection->ssl());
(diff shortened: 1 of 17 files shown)


src/envoy/utils/utils.cc
```diff
@@ -70,6 +70,20 @@ bool GetIpPort(const Network::Address::Ip* ip, std::string* str_ip, int* port) {
   return false;
 }
 
+bool GetDestinationUID(const envoy::api::v2::core::Metadata& metadata,
+                       std::string* uid) {
+  const auto filter_it = metadata.filter_metadata().find("mixer");
```
qiwzhang: can we use the same constant string in the mixer filter factory and here?

src/envoy/utils/utils.cc
```diff
@@ -70,6 +70,20 @@ bool GetIpPort(const Network::Address::Ip* ip, std::string* str_ip, int* port) {
...
+  if (filter_it == metadata.filter_metadata().end()) {
+    return false;
+  }
+  const auto fields_it = filter_it->second.fields().find("destination.uid");
```
qiwzhang: can we define a constant string for this metadata name?

src/istio/control/http/attributes_builder_test.cc
```diff
@@ -465,6 +470,8 @@ TEST(AttributesBuilderTest, TestReportAttributes) {
   Attributes expected_attributes;
   ASSERT_TRUE(
       TextFormat::ParseFromString(kReportAttributes, &expected_attributes));
+  (*expected_attributes.mutable_attributes())[AttributeName::kDestinationUID]
+      .set_string_value("pod1.ns2");
```
qiwzhang: can we just add this attribute to kReportAttributes?

kyessenov: There's unexpected use of the kReportAttributes in auth tests. This keeps the change contained to the tests of Report.

src/envoy/utils/utils.cc
kyessenov: Done

src/envoy/utils/utils.cc
kyessenov: There's no shared filter factory between http/tcp, while this applies to both. I defined a new constant for this purpose.

PR #1774, opened 2018-05-23
Per-host metadata for mixerfilter

kyessenov:
Adds an override for `destination.uid` for the mixer filter EDS metadata.

### DIFF
--- a/src/envoy/utils/utils.cc
+++ b/src/envoy/utils/utils.cc
@@ -30,7 +30,7 @@ const std::string kSPIFFEPrefix("spiffe://");
 const std::string kPerHostMixer("mixer");
 
 // Attribute field for per-host data override
-const std::string kMetadataDestinationUID("destination.uid");
+const std::string kMetadataDestinationUID("uid");
 
 }  // namespace
 
--- a/src/envoy/utils/utils.cc
+++ b/src/envoy/utils/utils.cc
@@ -17,6 +17,7 @@
 #include "mixer/v1/attributes.pb.h"
 
 using ::google::protobuf::Message;
+using ::google::protobuf::Struct;
 using ::google::protobuf::util::Status;
 
 namespace Envoy {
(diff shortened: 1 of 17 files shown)


kyessenov:
Note: this is just a repeat of the old PR against master.

Per route config is not sufficient, since the upstream host is not selected yet. Upstream host selection is done after routing, but prior to logging, and logging/reporting is where this is used.

qiwzhang:
I am confused.  logging/report is done after the request is replied by the proxy.  it is the last step.  Route should have been selected.  Actually Route has been selected with for Check() too.  That is how we use per_route config for both Check/Report.

src/envoy/utils/utils.cc
```diff
@@ -70,6 +76,21 @@ bool GetIpPort(const Network::Address::Ip* ip, std::string* str_ip, int* port) {
   return false;
 }
 
+bool GetDestinationUID(const envoy::api::v2::core::Metadata& metadata,
+                       std::string* uid) {
+  const auto filter_it = metadata.filter_metadata().find(kPerHostMixer);
+  if (filter_it == metadata.filter_metadata().end()) {
+    return false;
+  }
+  const auto fields_it =
```
qiwzhang: can we do following to make this code more readable

const Struct& struct_pb = filter_it->second;

src/envoy/utils/utils.cc
```diff
@@ -70,6 +76,21 @@ bool GetIpPort(const Network::Address::Ip* ip, std::string* str_ip, int* port) {
...
+      filter_it->second.fields().find(kMetadataDestinationUID);
+  if (fields_it == filter_it->second.fields().end()) {
+    return false;
+  }
+  *uid = fields_it->second.string_value();
```
qiwzhang: may need to check second.type to make sure it is string case

src/envoy/utils/utils.cc
```diff
@@ -26,6 +26,12 @@ namespace {
 
 const std::string kSPIFFEPrefix("spiffe://");
 
+// Per-host opaque data field
+const std::string kPerHostMixer("mixer");
+
+// Attribute field for per-host data override
+const std::string kMetadataDestinationUID("destination.uid");
```
qiwzhang: Please change it to "uid".

Add comment as
// This is same as defining following proto
//  message PerHostData {
//        string   uid = 1;
// }

src/envoy/utils/utils.cc
kyessenov: I think this is fine even if it is not a string. It is a user error to put other values, and this defaults to an empty string.

src/envoy/utils/utils.cc
kyessenov: Done

PR #1776, opened 2018-05-23
Add forwarded_attributes to per route config

kyessenov:
Implements the API change https://github.com/istio/api/pull/500


kyessenov:
PTAL. This is a repeat of the old PR against master.

Issue #1789, opened 2018-05-29
Frequent segmentation faults


mattatcha:
Seems that segfaults only happen while proxy is handling http traffic.
 
Running istio/istio@d91d44ead27d9b4018ab5997ef7ec0f0a4621d9d on kubernetes.

Raw backtrace
```
[libprotobuf INFO src/istio/mixerclient/check_cache.cc:155] Add a new Referenced for check cache: Absence-keys: Exact-keys: context.protocol, destination.service, destination.uid, source.ip, source.uid,
[libprotobuf INFO src/istio/mixerclient/check_cache.cc:155] Add a new Referenced for check cache: Absence-keys: Exact-keys: context.protocol, destination.service, destination.uid, source.ip, source.uid,
[2018-05-29 18:12:00.841][38][critical][backtrace] bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] Caught Segmentation fault, suspect faulting address 0x0
[2018-05-29 18:12:00.858][38][critical][backtrace] bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:90] Backtrace obj</usr/local/bin/envoy> thr<38> (use tools/stack_decode.py):
[2018-05-29 18:12:00.858][38][critical][backtrace] bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:107] thr<38> #0 0x4db088
[2018-05-29 18:12:00.858][38][critical][backtrace] bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:107] thr<38> #1 0x4e69d2
[2018-05-29 18:12:00.858][38][critical][backtrace] bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:107] thr<38> #2 0x4e6d21
[2018-05-29 18:12:00.858][38][critical][backtrace] bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:107] thr<38> #3 0x4e6d87
[2018-05-29 18:12:00.858][38][critical][backtrace] bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:107] thr<38> #4 0x4e6e00
[...]

PR #1790, opened 2018-05-30
Not to flush out batched report in destructor

qiwzhang:

This will fix proxy crash:  https://github.com/istio/proxy/issues/1789

**Release note**:
```release-note
None
```


kyessenov:
Would it be easy to add a regression test?

lizan:
Does this means we will lost some reports? There should be alternative to do this instead of destructor.

qiwzhang:
Yes,  https://github.com/istio/proxy/issues/1792 to track it.

PR #1797, opened 2018-05-31
Route directive

kyessenov:
Implementation of the route directive.
API change: https://github.com/istio/api/pull/510

Manually tested with the following set-up.

envoy config:
```yaml
node:
  id: test
admin:
  access_log_path: /dev/stdout
  address:
    socket_address:
      address: 127.0.0.1
      port_value: 8001
static_resources:
  clusters:
    - name: mixer_server
      connect_timeout: 1s
      type: STATIC
      http2_protocol_options: {}
      hosts:
        - socket_address:
            address: 127.0.0.1
            port_value: 9091
    - name: echo
      connect_timeout: 1s
      type: STATIC
      hosts:
        - socket_address:
            address: 127.0.0.1
            port_value: 8080
  listeners:
    - name: echo
      address:
        socket_address:
          address: 127.0.0.1
          port_value: 8000
      filter_chains:
        - filters:
          - name: envoy.http_connection_manager
            config:
              codec_type: AUTO
              stat_prefix: http
              generate_request_id: true
              http_filters:
                - name: mixer
                  config:
                    default_destination_service: echo
                    service_configs:
                      echo:
                        disable_report_calls: true
                    transport:
                      network_fail_policy:
                        policy: FAIL_CLOSE
                - name: envoy.router
              route_config:
                name: echo
                virtual_hosts:
                  - domains:
                    - '*'
                    name: echo
                    routes:
                      - match:
                          prefix: /
                        route:
                          cluster: echo
                          timeout: 0s
```

Mixer check response:
```golang
&mixerpb.CheckResponse{
                Precondition: mixerpb.CheckResponse_PreconditionResult{
                        ReferencedAttributes: mixerpb.ReferencedAttributes{
                                Words: []string{"request.method"},
                                AttributeMatches: []mixerpb.ReferencedAttributes_AttributeMatch{{
                                        Name:      -1,
                                        Condition: mixerpb.EXACT,
                                }},
                        },
                        ValidDuration: time.Minute,
                        Status:        rpc.Status{Code: int32(rpc.OK)},
                        ValidUseCount: 10000000,
                        RouteDirective: &mixerpb.RouteDirective{
                                UpdateRequestHeaders: []mixerpb.HeaderOperation{{
                                        Name:  ":path",
                                        Value: "/path",
                                }, {
                                        Name:  ":authority",
                                        Value: "echo",
                                }, {
                                        Name:  ":method",
                                        Value: "POST",
                                }, {
                                        Name:  "x-istio-request",
                                        Value: "value",
                                }, {
                                        Name:      "x-istio-request",
                                        Value:     "value2",
                                        Operation: mixerpb.APPEND,
                                }, {
                                        Name:      "user-agent",
                                        Value:     "value2",
                                        Operation: mixerpb.REMOVE,
                                }},
                                UpdateResponseHeaders: []mixerpb.HeaderOperation{{
                                        Name:  "x-istio-response",
                                        Value: "value",
                                }, {
                                        Name:      "x-istio-response",
                                        Value:     "value2",
                                        Operation: mixerpb.APPEND,
                                }},
                        },
                },
```

Making a request:
```bash
 curl -v 127.0.0.1:8000
* Rebuilt URL to: 127.0.0.1:8000/
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: curl/7.47.0
> Accept: */*
>
< HTTP/1.1 200 OK
< content-length: 0
< date: Tue, 05 Jun 2018 23:03:40 GMT
< x-envoy-upstream-service-time: 0
< x-istio-response: value
< x-istio-response: value2
< server: envoy
<
* Connection #0 to host 127.0.0.1 left intact
```

Backend echo:
```bash
POST /path HTTP/1.1 127.0.0.1:55792
Accept: */*
X-Forwarded-Proto: http
X-Request-Id: 77f19556-8ea2-4a71-b4e9-4ef2b6bbb2bc
X-Istio-Request: value
X-Istio-Request: value2
Content-Length: 0
Host: echo
```

### DIFF
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -166,7 +166,7 @@ FilterTrailersStatus Filter::decodeTrailers(HeaderMap& trailers) {
 }
 
 FilterHeadersStatus Filter::encodeHeaders(HeaderMap& headers, bool) {
-  ENVOY_LOG(debug, "Called Mixer::Filter : {}", __func__);
+  ENVOY_LOG(debug, "Called Mixer::Filter : {} {}", __func__, state_);
   if (state_ == Calling) {
     return FilterHeadersStatus::StopIteration;
   }
@@ -175,13 +175,14 @@ FilterHeadersStatus Filter::encodeHeaders(HeaderMap& headers, bool) {
     for (auto const iter : route_directive_.update_response_headers()) {
       switch (iter.operation()) {
         case ::istio::mixer::v1::HeaderOperation_Operation_REPLACE:
-          headers.setReference(LowerCaseString(iter.name()), iter.value());
+          headers.remove(LowerCaseString(iter.name()));
+          headers.addCopy(LowerCaseString(iter.name()), iter.value());
           break;
         case ::istio::mixer::v1::HeaderOperation_Operation_REMOVE:
           headers.remove(LowerCaseString(iter.name()));
           break;
         case ::istio::mixer::v1::HeaderOperation_Operation_APPEND:
-          headers.addReference(LowerCaseString(iter.name()), iter.value());
+          headers.addCopy(LowerCaseString(iter.name()), iter.value());
           break;
         default:
           PANIC("unreachable header operation");
@@ -220,7 +221,6 @@ void Filter::completeCheck(const CheckResponseInfo& info) {
   // Remove Istio authentication header after Check() is completed
   if (nullptr != headers_) {
     Envoy::Utils::Authentication::ClearResultInHeader(headers_);
-    headers_ = nullptr;
   }
 
   // This stream has been reset, abort the callback.
@@ -253,22 +253,24 @@ void Filter::completeCheck(const CheckResponseInfo& info) {
   for (auto const iter : route_directive_.update_request_headers()) {
     switch (iter.operation()) {
       case ::istio::mixer::v1::HeaderOperation_Operation_REPLACE:
-        headers_->setReference(LowerCaseString(iter.name()), iter.value());
+        headers_->remove(LowerCaseString(iter.name()));
+        headers_->addCopy(LowerCaseString(iter.name()), iter.value());
         break;
       case ::istio::mixer::v1::HeaderOperation_Operation_REMOVE:
         headers_->remove(LowerCaseString(iter.name()));
         break;
       case ::istio::mixer::v1::HeaderOperation_Operation_APPEND:
-        headers_->addReference(LowerCaseString(iter.name()), iter.value());
+        headers_->addCopy(LowerCaseString(iter.name()), iter.value());
         break;
       default:
         PANIC("unreachable header operation");
     }
   }
 
+  headers_ = nullptr;
+
   if (!initiating_call_) {
     decoder_callbacks_->continueDecoding();
-    encoder_callbacks_->continueEncoding();
   }
 }
 
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -166,19 +166,19 @@ FilterTrailersStatus Filter::decodeTrailers(HeaderMap& trailers) {
 }
 
 void Filter::UpdateHeaders(
-    HeaderMap* headers, const ::google::protobuf::RepeatedPtrField<
+    HeaderMap& headers, const ::google::protobuf::RepeatedPtrField<
                             ::istio::mixer::v1::HeaderOperation>& operations) {
   for (auto const iter : operations) {
     switch (iter.operation()) {
       case ::istio::mixer::v1::HeaderOperation_Operation_REPLACE:
-        headers->remove(LowerCaseString(iter.name()));
-        headers->addCopy(LowerCaseString(iter.name()), iter.value());
+        headers.remove(LowerCaseString(iter.name()));
+        headers.addCopy(LowerCaseString(iter.name()), iter.value());
         break;
       case ::istio::mixer::v1::HeaderOperation_Operation_REMOVE:
-        headers->remove(LowerCaseString(iter.name()));
+        headers.remove(LowerCaseString(iter.name()));
         break;
       case ::istio::mixer::v1::HeaderOperation_Operation_APPEND:
-        headers->addCopy(LowerCaseString(iter.name()), iter.value());
+        headers.addCopy(LowerCaseString(iter.name()), iter.value());
         break;
       default:
         PANIC("unreachable header operation");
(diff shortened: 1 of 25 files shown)


kyessenov:
@qiwzhang looking for initial feedback how to update the mixerclient. It's unfamiliar codebase.

src/istio/mixerclient/client_impl.cc
```diff
@@ -70,6 +70,8 @@ CancelFunc MixerClientImpl::Check(
     return nullptr;
   }
 
+  check_response_info.route_directive = check_result->route_directive();
```
qiwzhang: this should be in line 67

kyessenov:
@qiwzhang @lizan PTAL

src/envoy/http/mixer/filter.cc
```diff
@@ -192,6 +237,38 @@ void Filter::completeCheck(const Status& status) {
...
+    ENVOY_LOG(debug, "Mixer::Filter direct response");
+    state_ = Responded;
+    decoder_callbacks_->sendLocalReply(
+        Code(route_directive_.direct_response_code()),
+        route_directive_.direct_response_body(), nullptr);
+    return;
+  }
+
+  // handle request header operations
+  for (auto const iter : route_directive_.update_request_headers()) {
```
qiwzhang: can we share a function for this one with update_response_headers?

src/envoy/http/mixer/filter.cc
```diff
@@ -164,26 +165,70 @@ FilterTrailersStatus Filter::decodeTrailers(HeaderMap& trailers) {
   return FilterTrailersStatus::Continue;
 }
 
+FilterHeadersStatus Filter::encodeHeaders(HeaderMap& headers, bool) {
+  ENVOY_LOG(debug, "Called Mixer::Filter : {} {}", __func__, state_);
+  if (state_ == Calling) {
+    return FilterHeadersStatus::StopIteration;
```
kyessenov: cc @PiotrSikora : I have a question here about the filter state machine. What is the order between encoding and decoding of headers? It seems that header encoding never kicks in prior to check completing. Calling continueEncoding on check complete leads to a segfault (assert failure). So it's not clear to me how to get out of `StopIteration` state here...

lizan: This can never happen. So if you are in this state it is certainly a bug, so just do `ASSERT(state_ == Complete)`.

In the `continueEncoding` is a member of encoder_callbacks, which is not supposed to be called in decoder path (check complete). In decoder path, we stops at `decoderHeader` until check is complete, so the request won't go to upstream until check is done, hence the encoder stream would never be created at that moment.

kyessenov: Done. Put an assert for complete/responded states.

src/envoy/http/mixer/filter.cc
kyessenov: Done

kyessenov:
@qiwzhang ASAN test flaked in LRU test. Seems unrelated.

src/envoy/http/mixer/filter.cc
```diff
@@ -165,29 +165,33 @@ FilterTrailersStatus Filter::decodeTrailers(HeaderMap& trailers) {
   return FilterTrailersStatus::Continue;
 }
 
+void Filter::UpdateHeaders(
+    HeaderMap* headers, const ::google::protobuf::RepeatedPtrField<
```
qiwzhang: per envoy styple guide,  use & instead of raw pointer

src/envoy/http/mixer/filter.cc
```diff
@@ -164,26 +165,74 @@ FilterTrailersStatus Filter::decodeTrailers(HeaderMap& trailers) {
...
+  if (state_ == Complete) {
+    // handle response header operations
+    UpdateHeaders(&headers, route_directive_.update_response_headers());
+  }
+  return FilterHeadersStatus::Continue;
+}
+
+FilterDataStatus Filter::encodeData(Buffer::Instance&, bool) {
+  ENVOY_LOG(debug, "Called Mixer::Filter : {}", __func__);
+  if (state_ == Calling) {
```
qiwzhang: Same as encodeHeaders,  we can assert the state_ as Complete or Respond.

src/envoy/http/mixer/filter.cc
```diff
@@ -164,26 +165,74 @@ FilterTrailersStatus Filter::decodeTrailers(HeaderMap& trailers) {
...
+    return FilterDataStatus::StopIterationAndWatermark;
+  }
+  return FilterDataStatus::Continue;
+}
+
+FilterTrailersStatus Filter::encodeTrailers(HeaderMap&) {
+  ENVOY_LOG(debug, "Called Mixer::Filter : {}", __func__);
+  if (state_ == Calling) {
```
qiwzhang: Same comment as encodeData()

src/istio/mixerclient/check_cache_test.cc
```diff
@@ -49,16 +49,18 @@ class CheckCacheTest : public ::testing::Test {
     CheckResponse ok_response;
     ok_response.mutable_precondition()->set_valid_use_count(1000);
     // Just to calculate signature
-    EXPECT_ERROR_CODE(Code::NOT_FOUND, cache_->Check(attributes_, FakeTime(0)));
+    EXPECT_ERROR_CODE(Code::NOT_FOUND,
+                      cache_->Check(attributes_, FakeTime(0), nullptr));
     // set to the cache
```
qiwzhang: Please add a test case to test caching of route_directive.

src/envoy/http/mixer/filter.cc
kyessenov: Filter is caching a pointer `headers_` to the request header map. Any idea why it is a pointer, not a reference?

qiwzhang: bcs you only have "headers" at decodeHeader(), not at constructor.

kyessenov: how should I cast a pointer to a reference here?

src/envoy/http/mixer/filter.cc
kyessenov: done

src/envoy/http/mixer/filter.cc
kyessenov: done

src/istio/mixerclient/check_cache_test.cc
kyessenov: done

src/envoy/http/mixer/filter.cc
qiwzhang: *headers_  will cast to reference.

PR #1806, opened 2018-06-08
performance improvement: if the request matches one of the referenced, it won't match others.

skyao:

This is performance improvement: if the request matches one of the referenced, it won't match others. So we can directly exit the for loop and return NOT_FOUND.

No issue.

### DIFF
--- a/src/istio/mixerclient/check_cache.cc
+++ b/src/istio/mixerclient/check_cache.cc
@@ -122,6 +122,8 @@ Status CheckCache::Check(const Attributes &attributes, Tick time_now,
         result->route_directive_ = elem->route_directive();
       }
       return elem->status();
+    } else {
+      return Status(Code::NOT_FOUND, "");
     }
   }
 


skyao:
@qiwzhang This is a similar issue in  mixer client cache.

src/istio/mixerclient/check_cache.cc
```diff
@@ -122,6 +122,8 @@ Status CheckCache::Check(const Attributes &attributes, Tick time_now,
         result->route_directive_ = elem->route_directive();
       }
       return elem->status();
+    } else {
+      return Status(Code::NOT_FOUND, "");
```
qiwzhang: No,  It is not correct.   For example,  An request may be able to generate signature from Referenced a and b.  But its previous result was cached by the signature from b,   If we return not found here,  it will not be able to find its cached item.

skyao: As I understand, a request should only match ONE Referenced in referenced_map_ of mixer check cache.

Suppose we have some mixer adapters which care about three attributes x/y/z, the mixer will response with referenced attributes depending on the actual attributes in the requests (if x,y,z present in requests or not, we assume we had sent all kinds of following cases ):

- "exact_keys=x,y,z, absence_keys="
- "exact_keys=y,z, absence_keys=x"
- "exact_keys=x,z, absence_keys=y"
- "exact_keys=x,y, absence_keys=z"
- "exact_keys=x, absence_keys=y,z"
- "exact_keys=y, absence_keys=x,z"
- "exact_keys=z, absence_keys=x,y"
- "exact_keys=, absence_keys=x,y,z"

All above will be saved as Referenced in referenced_map_ . So When we loops referenced_map_ in Check() meth and reference.Signature() returns true, it means that the request matches one of above saved referenced.

    Status CheckCache::Check(const Attributes &attributes, Tick time_now) {
      ......
      std::lock_guard<std::mutex> lock(cache_mutex_);
      for (const auto &it : referenced_map_) {
        const Referenced &reference = it.second;
        std::string signature;
        if (!reference.Signature(attributes, "", &signature)) {
          continue;
        }
    
        CheckLRUCache::ScopedLookup lookup(cache_.get(), signature);
        if (lookup.Found()) {
          return elem->status();
        }
      }
    
    }

if we have some attributes such as "x=1,y=2,z=3,m=0,n=0,p=1,q=1" in the request, this request will match Referenced with "exact_keys=x,y,z, absence_keys=". It will never match other Referenced. Also if another request contains  attributes such as "x=1,y=2,m=0,n=0,p=1,q=1" , it should only match Referenced with "exact_keys=x,y, absence_keys=z"

That's why I think if reference.Signature() return true, we should break the loop of referenced_map_ directly.

I think we shouldn't have similar saved referenced:

- "exact_keys=x,y,z, absence_keys="
- "exact_keys=m,n absence_keys="

if the mixer adapters care about x,y,z and m,n, they should set referenced attributes in response like this:

- "exact_keys=x,y,z, absence_keys=m,n"
- "exact_keys=m,n absence_keys=x,y,z"

skyao: The most important is: 

An request may be able to generate signature from Referenced a and b.

Do we have any use case for this?

qiwzhang: My understanding is:  it is possible we will have referenced like following:

* "exact_keys=x,y,z, absence_keys="
* "exact_keys=m,n absence_keys="

So we could not bail out after a Referenced signature not found in the cache.

Hi @geeknoid and @mandarjog  What do you think?

geeknoid: [I gave a response to this issue on istio-dev@ earlier, but I believe I was incorrect]

We have three states for each attribute:

  - Must be present with value X
  - Must be absent
  - Not relevant

Using Wayne's example, we can make it more explicit:

  "exact_keys=x,y,z, absence_keys=, dont_care_keys=m,n,*"
  "exact_keys=m,n absence_keys=, dont_care_keys=x,y,z,*"

So if you have a request that contains all attributes x,y,z,m,n, it will be a match for both cases.

skyao: Let's talk about this in istio-dev.

https://groups.google.com/forum/#!topic/istio-dev/GheKGIbokKc

I copied yours answers into it.

@qiwzhang , @geeknoid

skyao: close this PR since we have some special case as mentioned in 

https://groups.google.com/forum/#!topic/istio-dev/GheKGIbokKc

PR #1809, opened 2018-06-08
Fix invalid assertion for mixerfilter

kyessenov:
mixerfilter decode may never be called if a prior filter interrupts the chain.

### DIFF
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -189,24 +189,14 @@ void Filter::UpdateHeaders(
 
 FilterHeadersStatus Filter::encodeHeaders(HeaderMap& headers, bool) {
   ENVOY_LOG(debug, "Called Mixer::Filter : {} {}", __func__, state_);
-  ASSERT(state_ == Complete || state_ == Responded);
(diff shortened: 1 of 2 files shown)


src/envoy/http/mixer/filter.cc
```diff
@@ -189,26 +189,16 @@ void Filter::UpdateHeaders(
 
 FilterHeadersStatus Filter::encodeHeaders(HeaderMap& headers, bool) {
   ENVOY_LOG(debug, "Called Mixer::Filter : {} {}", __func__, state_);
-  ASSERT(state_ == Complete || state_ == Responded);
+  // Init state is possible if a filter prior to mixerfilter interrupts the
+  // filter chain
+  ASSERT(state_ == NotStarted || state_ == Complete || state_ == Responded);
```
qiwzhang: My suggestion is to remove this ASSERT.

kyessenov: I think it's useful to know that Calling is not happening. That looks like the only invariant in the control flow here, that "Check must be called before proceeding". I removed all the other asserts.

Issue #1805, opened 2018-06-07
Extract grpc protocol, status, message


qiwzhang:
This is from https://github.com/istio/istio/issues/5429

proxy should set attribute context.protocol  as grpc  if request.headers["content-type"] == "application/grpc"

add two new attributes to extract grpc status
response.grpc-status
response.grpc-message

kyessenov:
Make sure to cover `application/grpc+json`

zainul-ma:
@qiwzhang and @kyessenov for implement two things above , should i waiting for version 1.0 or change some other code in v.0.8 ?

qiwzhang:
It will be in 1.0

PR #1810, opened 2018-06-09
Add grpc status, set grpc protocol

qiwzhang:

For https://github.com/istio/proxy/issues/1805

Check content type to set context.protocol correctly.
Extract grpc status and grpc message attributes from response.tailers. 

**Release note**:
```release-note
None
```

### DIFF
--- a/src/envoy/http/mixer/report_data.h
+++ b/src/envoy/http/mixer/report_data.h
@@ -31,6 +31,7 @@ const std::set<std::string> ResponseHeaderExclusives = {};
 
 class ReportData : public ::istio::control::http::ReportData {
   const HeaderMap *headers_;
+  const HeaderMap *trailers_;
   const RequestInfo::RequestInfo &info_;
   uint64_t response_total_size_;
   uint64_t request_total_size_;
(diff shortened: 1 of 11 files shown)


src/envoy/http/mixer/report_data.h
```diff
@@ -84,6 +90,19 @@ class ReportData : public ::istio::control::http::ReportData {
     }
     return false;
   }
+
+  bool GetGrpcStatus(GrpcStatus *status) const override {
```
JimmyCYJ: IIRC, Sometimes when response body is empty, then trailers_ will be empty too. In this case grpc status and message are sent in response headers.

qiwzhang: Thanks.  updated the code to get grpc-status from both trailers and response headers.

src/envoy/http/mixer/report_data.h
```diff
@@ -84,6 +104,13 @@ class ReportData : public ::istio::control::http::ReportData {
...
+    // Check trailer first.
+    // If not response body, grpc-status is in response headers.
+    return ExtractGrpcStatus(trailers_, status) ||
+           ExtractGrpcStatus(headers_, status);
```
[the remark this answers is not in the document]
qiwzhang: No, they are always together.  Please see: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

PR #1813, opened 2018-06-12
Update envoy

kyessenov:


WORKSPACE
```diff
@@ -29,16 +29,14 @@ bind(
     actual = "//external:ssl",
 )
 
-# We need newer gRPC than Envoy has for ALTS, this could be removed once Envoy picks
-# newer gRPC with ALTS support. (likely v1.11).
 git_repository(
     name = "com_github_grpc_grpc",
     remote = "https://github.com/grpc/grpc.git",
-    commit = "c50405364a194a0e4931153cbe329662d90530bc",  # Mar 28, 2018
+    commit = "bec3b5ada2c5e5d782dff0b7b5018df646b65cb0",  # May 15, 2018
```
PiotrSikora: Nit: comment should say `v1.12.0`, since it's tagged release and not `master`.

PiotrSikora: Actually, I think the whole `git_repository` block can be removed, since it looks that it was added only to override Envoy's gRPC.

PR #1814, opened 2018-06-13
pre-serialize attributes_for_mixer_proxy

kyessenov:
Also fixes the stale pointer dereference.

### DIFF
--- a/src/envoy/http/mixer/control.cc
+++ b/src/envoy/http/mixer/control.cc
@@ -33,16 +33,8 @@ Control::Control(const Config& config, Upstream::ClusterManager& cm,
                  [this](::istio::mixerclient::Statistics* stat) -> bool {
                    return GetStats(stat);
                  }) {
-  if (!config_.config_pb()
-           .transport()
-           .attributes_for_mixer_proxy()
-           .attributes()
-           .empty()) {
-    config_.config_pb()
-        .transport()
-        .attributes_for_mixer_proxy()
-        .SerializeToString(&serialized_forward_attributes_);
-  }
+  Utils::SerializeForwardedAttributes(config_.config_pb().transport(),
+                                      &serialized_forward_attributes_);
 
   ::istio::control::http::Controller::Options options(config_.config_pb());
 
--- a/src/envoy/http/mixer/control.cc
+++ b/src/envoy/http/mixer/control.cc
@@ -46,10 +46,9 @@ Control::Control(const Config& config, Upstream::ClusterManager& cm,
 
   ::istio::control::http::Controller::Options options(config_.config_pb());
 
-  Utils::CreateEnvironment(
-      dispatcher, random, *check_client_factory_, *report_client_factory_,
-      serialized_forward_attributes_,
-      &options.env);
+  Utils::CreateEnvironment(dispatcher, random, *check_client_factory_,
+                           *report_client_factory_,
+                           serialized_forward_attributes_, &options.env);
 
   controller_ = ::istio::control::http::Controller::Create(options);
 }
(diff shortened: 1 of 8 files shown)


kyessenov:
/hold
still segfaulting...

kyessenov:
/hold cancel

confirmed this fixes the segfault

src/envoy/http/mixer/control.cc
```diff
@@ -45,9 +45,21 @@ Control::Control(const Config& config, Upstream::ClusterManager& cm,
...
+  if (!config_.config_pb()
+           .transport()
+           .attributes_for_mixer_proxy()
+           .attributes()
+           .empty()) {
+    config_.config_pb()
+        .transport()
+        .attributes_for_mixer_proxy()
+        .SerializeToString(&serialized_forward_attributes);
+  }
```
qiwzhang: This is bad for performance.  It is done per request, per Check call.  Could you do following
1) similar to check_client_factory and report_client_factory define here https://github.com/istio/proxy/blob/master/src/envoy/http/mixer/control.h#L54

add the serialized_forward_attributes string at Control class.

2) then just pass its reference to all other paces.

kyessenov: Done

src/envoy/utils/mixer_control.h
```diff
@@ -31,8 +31,8 @@ void CreateEnvironment(Event::Dispatcher &dispatcher,
                        Runtime::RandomGenerator &random,
                        Grpc::AsyncClientFactory &check_client_factory,
                        Grpc::AsyncClientFactory &report_client_factory,
-                       ::istio::mixerclient::Environment *env,
-                       const Attributes &forward_attributes);
+                       std::string serialized_forward_attributes,
```
qiwzhang: please use const &

src/envoy/utils/grpc_transport.h
```diff
@@ -62,7 +62,7 @@ class GrpcTransport : public Grpc::TypedAsyncRequestCallbacks<ResponseType>,
 
   Grpc::AsyncClientPtr async_client_;
   ResponseType* response_;
-  const ::istio::mixer::v1::Attributes& forward_attributes_;
+  std::string serialized_forward_attributes_;
```
qiwzhang: I think you can use const &,  not need to copy the data

src/envoy/http/mixer/control.cc
```diff
@@ -33,21 +33,30 @@ Control::Control(const Config& config, Upstream::ClusterManager& cm,
...
+  if (!config_.config_pb()
+           .transport()
+           .attributes_for_mixer_proxy()
+           .attributes()
+           .empty()) {
+    config_.config_pb()
+        .transport()
+        .attributes_for_mixer_proxy()
+        .SerializeToString(&serialized_forward_attributes_);
+  }
```
qiwzhang: Can we define a function to share this code with tcp?   similar to Utils::CreateEnvironment()

src/envoy/utils/mixer_control.h
kyessenov: TCP control doesn't own the string though? Do you want me to store it there?

qiwzhang: Yes.

src/envoy/http/mixer/control.cc
kyessenov: Done

src/envoy/utils/grpc_transport.h
kyessenov: Done

src/envoy/utils/mixer_control.h
kyessenov: Done

PR #1819, opened 2018-06-18
Distinguish different filter config when creating authn filter.

incfly:
**Which issue this PR fixes**: https://github.com/istio/istio/issues/6361

**Release note**:
Each authn filter has its own configuration.

With a proxy build included change, I can get multiplexing work when different port have different policies.

### DIFF
--- a/src/envoy/http/authn/http_filter.h
+++ b/src/envoy/http/authn/http_filter.h
@@ -31,7 +31,8 @@ class AuthenticationFilter : public StreamDecoderFilter,
                              public Logger::Loggable<Logger::Id::filter> {
  public:
   AuthenticationFilter(
-      const istio::envoy::config::filter::http::authn::v2alpha1::FilterConfig& config);
+      const istio::envoy::config::filter::http::authn::v2alpha1::FilterConfig&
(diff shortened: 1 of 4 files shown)


incfly:
@wattli @diemtvu @qiwzhang

incfly:
@qiwzhang Interesting. So the filter is created on each request, then for sure each filter shoudn't do this proto copy paste.

But the current implementation, if you look carefully, is globally the authn filter factory class only hold one filter config. This cause problem because `AuthenticatorFilter::createFilterFactory` can be invoked with different config param, thus we need to return filter with different config. 

This cause us a problem for multiplexing.

I looked at the mixer filter implementation it seems they introduce another control class and use that as shared_ptr. Do you suggest us to follow that here? Or is there a standard Envoy approach to deal with this?

https://github.com/istio/proxy/blob/master/src/envoy/tcp/mixer/filter_factory.cc#L60

incfly:
@qiwzhang I changed back to const reference in AuthnFilter and use shared_ptr in `createFilterFactory`. PTAL, thanks!

PR #1824, opened 2018-06-20
Update istio/api dependency to be the latest in release-1.0.0-snapshot.0.

incfly:
Fortunately, the SHA `2d2fd0bdcf2ce26f5d96615e05974913b3d260c5` is already present in istio/api:release-1.0-snapshot.0 branch, proof:  https://github.com/istio/api/commits/release-1.0.0-snapshot.0, search "2d2fd0"

istio/istio#5091

### DIFF
--- a/istio.deps
+++ b/istio.deps
@@ -4,7 +4,7 @@
 		"name": "ISTIO_API",
 		"repoName": "api",
 		"file": "repositories.bzl",
-		"lastStableSHA": "f403f2ff3a7ee656a1bfbf23e66966a633160300",
+		"lastStableSHA": "2d2fd0bdcf2ce26f5d96615e05974913b3d260c5",
 	},
 	{
 		"_comment": "",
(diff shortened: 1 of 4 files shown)


incfly:
@hklai

hklai:
I think this should be at c6ad574a1cd532463cd805f948bef90d03564f60 which is the latest according to https://github.com/istio/api/commits/release-1.0.0-snapshot.0.

Why 2d2fd0bdcf2ce26f5d96615e05974913b3d260c5?

incfly:
Okay, istio/istio also have that SHA, https://github.com/istio/istio/blob/release-1.0.0-snapshot.0/Gopkg.lock#L1323

Changed to c6ad574a1cd532463cd805f948bef90d03564f60

Interesting topic though. I think there supposed to be some automatic mechanism to ensure the consistency.

PR #1825, opened 2018-06-21
authn: save attributes to RequestInfo.metadata for later use in RBAC filter.

yangminzhu:
Support to pass the auth attributes in metadata for later use in RBAC filter.

See istio/istio#6377

Signed-off-by: Yangmin Zhu <ymzhu@google.com>

### DIFF
--- a/include/istio/utils/BUILD
+++ b/include/istio/utils/BUILD
@@ -34,3 +34,11 @@ cc_library(
     ],
     visibility = ["//visibility:public"],
 )
+
+cc_library(
+    name = "attribute_names_lib",
+    hdrs = [
+        "attribute_names.h",
+    ],
+    visibility = ["//visibility:public"],
+)
--- a/include/istio/utils/BUILD
+++ b/include/istio/utils/BUILD
@@ -36,7 +36,7 @@ cc_library(
 )
 
 cc_library(
-    name = "attribute_names_lib",
+    name = "attribute_names_header",
     hdrs = [
         "attribute_names.h",
     ],
--- a/src/istio/control/http/attributes_builder.cc
+++ b/src/istio/control/http/attributes_builder.cc
@@ -29,6 +29,12 @@ namespace {
 // The gRPC content types.
 const std::set<std::string> kGrpcContentTypes{
     "application/grpc", "application/grpc+proto", "application/grpc+json"};
+
+// Helper function to set a key/value pair into Struct.
+static void setKeyValue(::google::protobuf::Struct &data, std::string key,
+                        std::string value) {
+  (*data.mutable_fields())[key].set_string_value(value);
+}
 }  // namespace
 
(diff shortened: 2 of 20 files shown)


src/istio/control/http/attributes_builder.cc
```diff
@@ -62,6 +68,44 @@ void AttributesBuilder::ExtractRequestHeaderAttributes(CheckData *check_data) {
   }
 }
 
+void AttributesBuilder::SaveAuthAttributesToStruct(
```
qiwzhang: It is better to move this function to https://github.com/istio/proxy/blob/master/src/envoy/utils/authn.h#L29

yangminzhu: I tried that but it couldn't access the attributes names (See [here](https://github.com/istio/proxy/blob/624540603b7ebb206335d5cb375faa4d8ffcfbdc/src/istio/control/BUILD#L28)), I don't want to change the visibility of those attributes names so I added this in attributes_builder.cc, also it's closer to the `ExtractAuthAttributes()` function that makes it clear we are setting the right names.

qiwzhang: I see. Then I like to have following changes to expose attribute_names

1) move attribute_name.h to include/istio/utils/
2) create a separate library as attribute_names in src/istio/util  and set its visibility as public

yangminzhu: Done.

include/istio/utils/BUILD
```diff
@@ -34,3 +34,11 @@ cc_library(
     ],
     visibility = ["//visibility:public"],
 )
+
+cc_library(
+    name = "attribute_names_lib",
```
qiwzhang: attribute_names_header

src/istio/control/http/request_handler_impl_test.cc
```diff
@@ -403,8 +403,9 @@ TEST_F(RequestHandlerImplTest, TestDefaultApiKey) {
                           TransportCheckFunc transport,
                           CheckDoneFunc on_done) -> CancelFunc {
         auto map = attributes.attributes();
-        EXPECT_EQ(map[AttributeName::kRequestApiKey].string_value(),
-                  "test-api-key");
+        EXPECT_EQ(
+            map[istio::utils::AttributeName::kRequestApiKey].string_value(),
```
qiwzhang: nit:  if a module is under istio namaespace,  you can remove namespace prefix "istio::"

include/istio/utils/BUILD
yangminzhu: Done.

src/istio/control/http/request_handler_impl_test.cc
yangminzhu: Done.

PR #1843, opened 2018-07-10
Add connection requested server name attribute to TCP read filter

vadimeisenbergibm:

Adds SNI attribute to the TCP read filter, to be used in telemetry reports and policy checks.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: implements #https://github.com/istio/istio/issues/6810

**Release note**:

```
SNI report and check in TCP filter
```

### DIFF
--- a/include/istio/control/http/check_data.h
+++ b/include/istio/control/http/check_data.h
@@ -47,6 +47,9 @@ class CheckData {
   // Returns true if connection is mutual TLS enabled.
   virtual bool IsMutualTLS() const = 0;
 
+  // Get requested server name, SNI in case of TLS
+  virtual std::string GetRequestedServerName() const = 0;
+
   // These headers are extracted into top level attributes.
   // This is for standard HTTP headers.  It supports both HTTP/1.1 and HTTP2
   // They can be retrieved at O(1) speed by environment (Envoy).
--- a/src/envoy/tcp/mixer/filter.h
+++ b/src/envoy/tcp/mixer/filter.h
@@ -53,6 +53,7 @@ class Filter : public Network::Filter,
   bool GetSourceIpPort(std::string* str_ip, int* port) const override;
   bool GetSourceUser(std::string* user) const override;
   bool IsMutualTLS() const override;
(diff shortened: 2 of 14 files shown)


src/istio/control/tcp/attributes_builder.cc
```diff
@@ -51,6 +51,9 @@ void AttributesBuilder::ExtractCheckAttributes(CheckData* check_data) {
   builder.AddBool(utils::AttributeName::kConnectionMtls,
                   check_data->IsMutualTLS());
 
+  builder.AddString(utils::AttributeName::kConnectionRequestedServerName,
```
kyessenov: can you please check whether the string is empty first? we normally don't send unset values to mixer.

src/envoy/tcp/mixer/filter.h
```diff
@@ -53,6 +53,7 @@ class Filter : public Network::Filter,
   bool GetSourceIpPort(std::string* str_ip, int* port) const override;
   bool GetSourceUser(std::string* user) const override;
   bool IsMutualTLS() const override;
+  std::string GetRequestedServerName() const override;
```
kyessenov: Make sure to apply the same changes to http filter in `src/envoy/http/mixer`

vadimeisenbergibm:
@kyessenov I have applied your comments.

vadimeisenbergibm:
@qiwzhang could you please review this PR?

include/istio/control/http/check_data.h
```diff
@@ -47,6 +47,9 @@ class CheckData {
   // Returns true if connection is mutual TLS enabled.
   virtual bool IsMutualTLS() const = 0;
 
+  // Get requested server name, SNI in case of TLS
+  virtual std::string GetRequestedServerName() const = 0;
```
qiwzhang: Since there are cases data is not available,   the coding style is to return a bool.  It should be:

// return true if data is valid.
bool GetRequestedServerName(std::string* name);

PR #1849, opened 2018-07-12
Add report of tcp connection open event

### LINKED ISSUE #1835: Open connection.event will not be reported
Though three connection events are added by https://github.com/istio/proxy/pull/1371, connection `open` event will never be reported.

With current logic `open` is recorded when check is called for a TCP connection. However report won't be called until either periodic report triggered or connection closing, which rewrites connection event to either `continue` or `close`. 

Should we make a report call at onNewConnection?

@qiwzhang @JimmyCYJ

bianpengyuan:
Add a report call after tcp check finishes and before report timer starts, which record tcp connection open event. fix #1835

### DIFF
--- a/src/istio/control/tcp/request_handler_impl.cc
+++ b/src/istio/control/tcp/request_handler_impl.cc
@@ -55,16 +55,17 @@ CancelFunc RequestHandlerImpl::Check(CheckData* check_data,
 
 // Make remote report call.
 void RequestHandlerImpl::Report(ReportData* report_data) {
-  Report(report_data, /* is_final_report */ true);
+  Report(report_data, /* is_first_report */ false, /* is_final_report */ true);
 }
 
-void RequestHandlerImpl::Report(ReportData* report_data, bool is_final_report) {
+void RequestHandlerImpl::Report(ReportData* report_data, bool is_first_report,
+                                bool is_final_report) {
   if (!client_context_->enable_mixer_report()) {
     return;
   }
 
   AttributesBuilder builder(&request_context_);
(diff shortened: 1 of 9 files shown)


bianpengyuan:
@qiwzhang

src/istio/control/tcp/request_handler_impl.cc
```diff
@@ -55,16 +55,17 @@ CancelFunc RequestHandlerImpl::Check(CheckData* check_data,
 
 // Make remote report call.
 void RequestHandlerImpl::Report(ReportData* report_data) {
```
qiwzhang: I think we can remove this function. It was there when mixer_client library were at different repo.

bianpengyuan:
Ideally we would like to report open connection OnNewConnection, but it won't have principal information in auth case. So report open connection when check is ok.

@qiwzhang Let me know if you want me to make that boolean args into enum to improve readability.

qiwzhang:
Yeap,  I think making it to enum is better for readability.  We only need:  first,  intermediate, and last report, is that right?

bianpengyuan:
@qiwzhang Done, ptal.

src/istio/control/tcp/request_handler_impl.cc
bianpengyuan: Sure, removed.

bianpengyuan:
@hklai Can you help to merge this? Connection.event is documented as having three possible values while right now connection open event is not even reported. This PR fixes the bug. The risk is low. It has no thing to do with upgrading, no new attributes type and only one report call added to async Report buffer.

@douglas-reid @gargnupur

PR #1861, opened 2018-07-18
Going back to 3b05bff

rshriram:

### DIFF
--- a/istio.deps
+++ b/istio.deps
@@ -11,6 +11,6 @@
 		"name": "ENVOY_SHA",
 		"repoName": "envoyproxy/envoy",
 		"file": "WORKSPACE",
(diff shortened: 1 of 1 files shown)


rshriram:
and thank you for the daily words of wisdom

kyessenov:
@rshriram In istio repo, place your experimental envoy into `~/go/out/linux...` directory and run all the tests locally. The build is not checking that dependencies match, so it will take whatever envoy you throw at it. You then go debug the memory leak locally, instead of using shared test infra.

rshriram:
Oh! Thats a nice idea.. Will try.. We should document it in readme or something

rshriram:
/hold
We finally found the memory leak and fixed it. I also validated with Istio e2e tests. Waiting for envoy PR to be merged.

PR #1884, opened 2018-07-26
destination.principal derivation fix

kyessenov:
There is an untested dependency between mixer and auth filters. If both are present (as configured by istio), then destination principal was not derived.

This fix moves the derivation to the top, so it doesn't matter whether or not auth filter is present.

### DIFF
--- a/src/istio/control/http/attributes_builder.cc
+++ b/src/istio/control/http/attributes_builder.cc
@@ -67,9 +67,16 @@ void AttributesBuilder::ExtractRequestHeaderAttributes(CheckData *check_data) {
 }
 
 void AttributesBuilder::ExtractAuthAttributes(CheckData *check_data) {
+  utils::AttributesBuilder builder(&request_->attributes);
+
+  std::string destination_principal;
+  if (check_data->GetPrincipal(false, &destination_principal)) {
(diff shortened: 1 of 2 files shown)


kyessenov:
@geeknoid @douglas-reid : can we get this in? It turns out that auth filter before mixer filter was hijacking principal attribute derivation, meaning `destination.principal` was lost. We don't test combos of two filters AFAIK.

src/istio/control/http/attributes_builder.cc
```diff
@@ -67,9 +67,16 @@ void AttributesBuilder::ExtractRequestHeaderAttributes(CheckData *check_data) {
 }
 
 void AttributesBuilder::ExtractAuthAttributes(CheckData *check_data) {
+  utils::AttributesBuilder builder(&request_->attributes);
```
diemtvu: Shouldn't you need to update unit test for this, i.e remove this line
https://github.com/istio/proxy/blob/release-1.0/src/istio/control/http/attributes_builder_test.cc#L447

(I don't know why pre-submit tests didn't fail though)

kyessenov: Mock stubs were missing. Added and corrected the unit test.

PR #1890, opened 2018-07-30
Update Envoy SHA to cherry-pick fix for idle timer segfault.

PiotrSikora:
Cherry-pick https://github.com/envoyproxy/envoy/pull/3970.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -30,12 +30,12 @@ bind(
 )
 
 # When updating envoy sha manually please update the sha in istio.deps file also
-ENVOY_SHA = "280baee6dc45e48a4bf38ac03d32711fe5eaeee1"
+ENVOY_SHA = "5d2ac4fdcb455061f5c5ca757ecdea0a535d5d23"
 
 http_archive(
     name = "envoy",
     strip_prefix = "envoy-" + ENVOY_SHA,
-    url = "https://github.com/envoyproxy/envoy/archive/" + ENVOY_SHA + ".zip",
+    url = "https://github.com/istio/envoy/archive/" + ENVOY_SHA + ".zip",
 )
 
 load("@envoy//bazel:repositories.bzl", "envoy_dependencies")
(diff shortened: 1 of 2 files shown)


PiotrSikora:
Cherry-picked instead of rolling forward to minimize untested changes and to avoid picking https://github.com/envoyproxy/envoy/pull/3975, since it could require last minute changes in Pilot.

rshriram:
Please wait. I have already merged the latest changes in istio (for latest Envoy sha). We can’t cherry pick because we need the gRPC bug fix.

rshriram:
Sorry I spoke to soon. So we don’t need that fix because we don’t set the idle timer (ie disabled always). We set the stream_idle_timeout.

rkpagadala:
@PiotrSikora I bumped the API because it was merged in release-1.0 branch, we need all the code to be consistent when shipped. Also it was a doc fix only. I would prefer not to merge this.

lizan:
> Sorry I spoke to soon. So we don’t need that fix because we don’t set the idle timer (ie disabled always). We set the stream_idle_timeout. 

@rshriram If we set stream_idle_timeout, then it could trigger the segfault which fixed in envoyproxy/envoy#3970, when stream is created but envoy didn't get full request header, and idle timeout kicked in. @htuch correct me if I am wrong.

costinm:
We were supposed to cut 1.0 30 min ago. 

I think it is too late. If it crashes we'll need to cut 1.0.1 next week

htuch:
If you set it to zero, you are probably safe, otherwise you need to cherry
pick all follow-up PRs.

rshriram:
See pilot code. It’s always set to 0 (stream idle timeout). But we don’t set idle timeout. 

And when we allow users to pick idle timeouts in future, we will be upgrading Envoy as well. Ie pull in bug fix automatically.

PiotrSikora:
Superseded by #1894.

PR #1893, opened 2018-08-03
Using request dynamic metadata to pass data from JWT filter to authn filter.

diemtvu:

Previously, we used headers to exchange data between JWT filter and Authn filter (and authn filter and mixer filter, but this is not addressed in the PR). This has some overhead (e.g encode/decode data so it can be save in header, extra information needed in API to define location etc), and most of all, trouble to clean up all those added data properly.

In this PR, we change JWT filter to (always) output payload data to request info metadata instead. The entry is keyed by JWT issuer so no need location information in the config and simpler for look up.

* There is no simple way to verify data in request info for integration test. This will be addressed in the next PRs.
* Also, as mentioned, changes for mixer to use data in request info (even though it's already populated both places) will be done later.


src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -60,14 +60,14 @@ void JwtAuthenticator::Verify(HeaderMap& headers,
...
-    }
-  }
+  // for (const auto& rule : store_.config().rules()) {
+  //   if (!rule.forward_payload_header().empty()) {
+  //     ENVOY_LOG(debug, "Sanitize JWT authentication output header {}",
+  //               rule.forward_payload_header());
+  //     const LowerCaseString key(rule.forward_payload_header());
+  //     headers.remove(key);
+  //   }
+  // }
```
qiwzhang: No dead code please.   If code is not needed, please remove it.

src/envoy/http/authn/authenticator_base_test.cc
```diff
@@ -20,13 +20,17 @@
 #include "gmock/gmock.h"
 #include "src/envoy/http/authn/test_utils.h"
 #include "test/mocks/network/mocks.h"
+//#include "test/mocks/request_info/mocks.h"
```
qiwzhang: if not needed, please remove it, don't just comment it out

src/envoy/http/authn/filter_context.cc
```diff
@@ -71,6 +72,35 @@ void FilterContext::setPrincipal(const iaapi::PrincipalBinding& binding) {
...
+                                  std::string* payload) const {
+  // auto location_iter =
+  //     filter_config_.jwt_output_payload_locations().find(issuer);
+  // if (location_iter == filter_config_.jwt_output_payload_locations().end()) {
+  //   ENVOY_LOG(warn, "No JWT payload output location is found for the issuer
+  //   {}", issuer); return false;
+  // }
+  const auto filter_it =
+      dynamic_metadata_.filter_metadata().find(istio::utils::FilterName::kJwt);
+  if (filter_it == dynamic_metadata_.filter_metadata().end()) {
```
rshriram: Is this using the stuff that went into envoy recently (for exchanging info between filters)? If so, doesn't that require filters to declare dependencies ?

src/istio/utils/filter_names.cc
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+#include "include/istio/utils/filter_names.h"
+
+namespace istio {
+namespace utils {
+
+// TODO: using more standard naming, e.g istio.jwt, istio.authn
+const char FilterName::kJwt[] = "jwt-auth";
+const char FilterName::kAuthentication[] = "istio_authn";
```
rshriram: why can't we fix the filter names now? better now than late. 
Also, why not std::string ?

include/istio/utils/filter_names.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace istio {
+namespace utils {
+
+// These are name of filters that currently output data to dynamicMetadata (by
+// convention, under the the entry using filter name itself as key). Define them
+// here for easy access.
+struct FilterName {
+  static const char kJwt[];
+  static const char kAuthentication[];
```
qiwzhang: It seems these names are Envoy specific,  can we move this file under src/envoy?

src/envoy/http/authn/filter_context.h
```diff
@@ -56,8 +57,10 @@ class FilterContext : public Logger::Loggable<Logger::Id::filter> {
   // Returns the authentication result.
   const istio::authn::Result& authenticationResult() { return result_; }
 
-  // Accessor to headers.
-  HeaderMap* headers() { return headers_; }
+  // Acessor to dynamic metadata.
+  const envoy::api::v2::core::Metadata& dynamic_metadata() const {
```
qiwzhang: consistent name style:  dynamicMetadata?

src/envoy/http/authn/http_filter.cc
```diff
@@ -48,54 +49,44 @@ FilterHeadersStatus AuthenticationFilter::decodeHeaders(HeaderMap& headers,
...
 
   // Put authentication result to headers.
   if (filter_context_ != nullptr) {
-    // TODO(yangminzhu): Remove the header and only use the metadata to pass the
+    // TODO(diemvu): Remove the header and only use the metadata to pass the
     // attributes.
     Utils::Authentication::SaveResultToHeader(
-        filter_context_->authenticationResult(), filter_context_->headers());
-
+        filter_context_->authenticationResult(), &headers);
```
qiwzhang: so we are still using headers to pass data between authn and mixer?

include/istio/utils/filter_names.h
diemtvu: Done.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -64,21 +66,11 @@ bool AuthenticatorBase::validateX509(const iaapi::MutualTls& mtls,
...
-
-  auto iter =
-      filter_context()->filter_config().jwt_output_payload_locations().find(
-          jwt.issuer());
-  if (iter ==
-      filter_context()->filter_config().jwt_output_payload_locations().end()) {
-    ENVOY_LOG(warn, "No JWT payload header location is found for the issuer {}",
-              jwt.issuer());
-    return false;
+  std::string jwt_payload;
```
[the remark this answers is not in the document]
diemtvu: Which part do you feel the most complicated right now? The authn policy by itself have a complex logic (we want to support conditioning between multiple authn methods), and I don't see we can do much to simplify it, at least at the moment. Anyway, feel free to create issue and suggest improvement.

src/envoy/http/authn/authenticator_base_test.cc
diemtvu: Done

src/envoy/http/authn/filter_context.cc
```diff
@@ -71,6 +72,35 @@ void FilterContext::setPrincipal(const iaapi::PrincipalBinding& binding) {
...
 }
 
+bool FilterContext::getJwtPayload(const std::string& issuer,
+                                  std::string* payload) const {
+  // auto location_iter =
+  //     filter_config_.jwt_output_payload_locations().find(issuer);
+  // if (location_iter == filter_config_.jwt_output_payload_locations().end()) {
+  //   ENVOY_LOG(warn, "No JWT payload output location is found for the issuer
+  //   {}", issuer); return false;
+  // }
```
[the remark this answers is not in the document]
diemtvu: Done

src/envoy/http/authn/filter_context.cc
diemtvu: By dependencies do you mean having the right order in filter chain? Pilot code (i.e plugin) is responsible to do that properly. It's already the existing behavior.

src/envoy/http/authn/filter_context.h
diemtvu: Remove this function actually, as all needs is satisfied by getJwtPayload function below.

src/envoy/http/authn/http_filter.cc
diemtvu: Yes. As mentioned in PR description, I want to do it in different PRs to make the change size reasonable.

src/envoy/http/jwt_auth/jwt_authenticator.cc
diemtvu: Done

src/istio/utils/filter_names.cc
diemtvu: It requires accompany changes from Pilot, so I want to do it in a smaller PR with just that change.

diemtvu: for std::string, just a personal style. This is just a const literal, and I don't see the need to use string for it.

src/envoy/http/authn/filter_context.cc
```diff
@@ -71,6 +72,29 @@ void FilterContext::setPrincipal(const iaapi::PrincipalBinding& binding) {
   }
 }
 
+bool FilterContext::getJwtPayload(const std::string& issuer,
+                                  std::string* payload) const {
+  const auto filter_it =
+      dynamic_metadata_.filter_metadata().find(Utils::IstioFilterName::kJwt);
```
qiwzhang: Is this dependency data structure documented some where?  Is it already implemented by JWT filter?

diemtvu: It's implemented by (proxy) JWT filter. For upstream, I expect to have API defined in Envoy. Wdyt?

src/envoy/utils/filter_names.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <string>
+
+namespace Envoy {
+namespace Utils {
+
+// These are name of (Istio) filters that currently output data to
+// dynamicMetadata (by convention, under the the entry using filter name itself
+// as key). Define them here for easy access.
+struct IstioFilterName {
```
qiwzhang: can we define our constants following Envoy way?

https://github.com/envoyproxy/envoy/blob/master/source/extensions/filters/http/well_known_names.h

src/envoy/utils/authn.h
```diff
@@ -40,6 +41,11 @@ class Authentication : public Logger::Loggable<Logger::Id::filter> {
   static bool FetchResultFromHeader(const Http::HeaderMap& headers,
                                     istio::authn::Result* result);
 
+  // Returns a pointer to the authentication result from request info, if
+  // available. Otherwise, return nullptrl
+  static const ProtobufWkt::Struct* GetResultFromRequestInfo(
```
qiwzhang: can we not use Struct? just return the exact proto message?

src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -205,11 +195,11 @@ void JwtAuthenticator::VerifyKey(const PubkeyCacheItem& issuer_item) {
...
 
-  if (!issuer_item.jwt_config().forward_payload_header().empty()) {
-    const LowerCaseString key(
-        issuer_item.jwt_config().forward_payload_header());
-    headers_->addCopy(key, jwt_->PayloadStrBase64Url());
-  }
+  // TODO: can we save as proto or json object directly?
+  // User the issuer as the entry key for simplicity. The forward_payload_header
+  // field can be removed or replace by a boolean (to make `save` is
+  // conditional)
```
qiwzhang: We may have to do both.   I am afraid some backend app may expect payload if "forward_payload_header" is set.

src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -59,16 +59,6 @@ void JwtAuthenticator::Verify(HeaderMap& headers,
   headers_ = &headers;
   callback_ = callback;
 
-  // Sanitize the JWT verification result in the HTTP headers
-  for (const auto& rule : store_.config().rules()) {
-    if (!rule.forward_payload_header().empty()) {
-      ENVOY_LOG(debug, "Sanitize JWT authentication output header {}",
-                rule.forward_payload_header());
```
qiwzhang: This code should not be removed.

src/envoy/utils/authn.h
diemtvu: The authn result stored in request info is already converted into Struct (key-value pairs of attributes). It was implemented by @yangminzhu and use for RBAC. We will change mixer code to consume this struct and populate attributes from this.

src/envoy/http/authn/http_filter_integration_test.cc
```diff
@@ -131,118 +144,57 @@ TEST_P(AuthenticationFilterIntegrationTest, OriginJwtRequiredHeaderNoJwtFail) {
 }
 
 TEST_P(AuthenticationFilterIntegrationTest, CheckValidJwtPassAuthentication) {
-  createTestServer(
-      "src/envoy/http/authn/testdata/envoy_origin_jwt_authn_only.conf",
-      {"http"});
+  config_helper_.addFilter(kAuthnFilterWithJwt);
+  config_helper_.addFilter(MakeHeaderToMetadataConfig());
```
qiwzhang: can we just real JWT filter?  It is integration test, should not use fake filter to reject fake data?

src/envoy/http/jwt_auth/jwt_authenticator.cc
diemtvu: We didn't (want) expose forward_payload_header to user. It might be the case for the use of JWT filter alone, but not for the authn flow in Istio. Currently, these outputs are removed by authn filter, not to mention the header key (assigned by pilot) is a hash of issuer, so fairly unusable. On the other hand, they can always use the original JWT anyway.

src/envoy/http/jwt_auth/jwt_authenticator.cc
diemtvu: For the same reason as above. We will remove this forward_payload_header in our API.

src/envoy/http/authn/http_filter_integration_test.cc
diemtvu: Then I have to make this dependent on JWT filter. This integration test should be for this filter only, so we don't have to dig to much in other JWT filter change.

Btw, I am also working on adding another integration test, that set up the same filter chain we have for Istio (jwt + authn + mixer) which can cover the real end-to-end flow.

PR #1896, opened 2018-08-07
Add the groups claim to the attribute request.auth.groups

lei-tang:
**What this PR does / why we need it**: Add the groups claim to the attribute request.auth.groups.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes # https://github.com/istio/istio/issues/6984

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/authn/authn_utils.cc
+++ b/src/envoy/http/authn/authn_utils.cc
@@ -55,9 +55,9 @@ void ExtractJwtAudience(
 // Extract JWT groups into the JwtPayload.
 // This function should to be called after the claims are extracted.
 void ExtractJwtGroups(
-const Envoy::Json::Object& obj,
-const ::google::protobuf::Map< ::std::string, ::std::string>& claims,
-istio::authn::JwtPayload* payload) {
+    const Envoy::Json::Object& obj,
+    const ::google::protobuf::Map< ::std::string, ::std::string>& claims,
+    istio::authn::JwtPayload* payload) {
   const std::string& key = kJwtGroupsKey;
   // "groups" can be either string array or string.
   // First, try as string
--- a/src/envoy/http/authn/authn_utils.cc
+++ b/src/envoy/http/authn/authn_utils.cc
@@ -25,6 +25,9 @@ namespace {
 // The JWT audience key name
 static const std::string kJwtAudienceKey = "aud";
 
+// The JWT groups key name
+static const std::string kJwtGroupsKey = "groups";
+
 // Extract JWT audience into the JwtPayload.
 // This function should to be called after the claims are extracted.
 void ExtractJwtAudience(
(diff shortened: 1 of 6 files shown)


src/envoy/utils/authn.cc
```diff
@@ -74,6 +80,16 @@ void Authentication::SaveAuthAttributesToStruct(
       setKeyValue(data, istio::utils::AttributeName::kRequestAuthAudiences,
                   origin.audiences(0));
     }
+    if (!origin.groups().empty()) {
```
diemtvu: This is about to be removed. You need to add this logic to SaveAuthAttributesToStruct instead.

lei-tang: Have discussed it offline and it is already in SaveAuthAttributesToStruct.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -48,6 +51,30 @@ void ExtractJwtAudience(
...
   }
 }
+
+// Extract JWT groups into the JwtPayload.
+// This function should to be called after the claims are extracted.
+void ExtractJwtGroups(
+    const Envoy::Json::Object& obj,
+    const ::google::protobuf::Map< ::std::string, ::std::string>& claims,
+    istio::authn::JwtPayload* payload) {
+  const std::string& key = kJwtGroupsKey;
```
liminw: Can we extend this to any string array claims?

src/envoy/utils/authn.cc
```diff
@@ -34,6 +34,12 @@ static void setKeyValue(::google::protobuf::Struct& data, std::string key,
   (*data.mutable_fields())[key].set_string_value(value);
 }
 
+// Helper function to set a key list pair into Struct.
+static void setListValue(::google::protobuf::Struct& data, std::string key,
+                        ::google::protobuf::ListValue* value) {
+  (*data.mutable_fields())[key].set_allocated_list_value(value);
```
qiwzhang: are you sure of using set_allocated_xxx?  My understanding is it will own the "value" pointer,

src/envoy/utils/authn.cc
```diff
@@ -74,6 +80,16 @@ void Authentication::SaveAuthAttributesToStruct(
       setKeyValue(data, istio::utils::AttributeName::kRequestAuthAudiences,
                   origin.audiences(0));
     }
+    if (!origin.groups().empty()) {
+      std::vector<std::string> groups;
+      for(int i=0; i<origin.groups().size(); i++) {
+        groups.push_back(origin.groups(i));
+      }
```
qiwzhang: it is prefer to use
```
for (group :  origin.groups()) 

```

src/envoy/utils/authn.cc
lei-tang: You are reviewing an older version of the code. The current code does not have set_allocated_xxx.

src/envoy/http/authn/authn_utils.cc
lei-tang: The claim key (e.g., kJwtGroupsKey, kJwtAudienceKey)  can be extended as an argument of the extraction function to support extracting any claims with the key. However, it is not straightforward to make the following statements in these extraction functions to be generic for extracting any claim key. If you have a way to do it, please let me know.
payload->add_audiences(claims.at(key));
payload->add_groups(claims.at(key));

PR #1897, opened 2018-08-07
Add context.proxy_error_code to http report attributes

mandarjog:
Adds context.proxy_error_code = %RESPONSE_FLAGS% to report attributes.

### DIFF
--- a/src/istio/utils/attribute_names.cc
+++ b/src/istio/utils/attribute_names.cc
@@ -70,6 +70,7 @@ const char AttributeName::kConnectionEvent[] = "connection.event";
 // Context attributes
 const char AttributeName::kContextProtocol[] = "context.protocol";
 const char AttributeName::kContextTime[] = "context.time";
+const char AttributeName::kContextProxyErrorCode[] = "context.proxy_error_code";
 
 // Check error code and message.
 const char AttributeName::kCheckErrorCode[] = "check.error_code";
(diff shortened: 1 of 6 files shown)


src/istio/utils/attribute_names.cc
```diff
@@ -70,6 +70,7 @@ const char AttributeName::kConnectionEvent[] = "connection.event";
 // Context attributes
 const char AttributeName::kContextProtocol[] = "context.protocol";
 const char AttributeName::kContextTime[] = "context.time";
+const char AttributeName::kContextRoutingErrorReason[] = "context.routing_error_reason";
```
qiwzhang: 1)  should we call it request, instead of context.  It is about the request.?
2) response_flags may include errors other than route, such as rate limiting, fault injection.  any errors from envoy proxy.  so it is better not call it "routing_error_reason".

mandarjog: good point. `response.error_reason` ?

kyessenov:
This also applies to TCP. Make sure to do the same in TCP filter.

How about context.proxy_error_code? I think it's best to keep it internal
and context is the right home then.

qiwzhang:
Yeap,  tcp_proxy has it <https://github.com/envoyproxy/envoy/blob/master/source/common/tcp_proxy/tcp_proxy.h#L230> too

PR #1904, opened 2018-08-09
update ENVOY_SHA

quanjielin:


istio.deps
```diff
@@ -11,6 +11,6 @@
 		"name": "ENVOY_SHA",
 		"repoName": "envoyproxy/envoy",
 		"file": "WORKSPACE",
-		"lastStableSHA": "346059548e135199eb0b7f0006f3ef19e173bf79"
```
qiwzhang: you also need to change the one in WORKSPACE

quanjielin: thanks, done

Issue #1902, opened 2018-08-09
Add RBAC filter to istio_http_integration_test


diemtvu:
Minimum, verify the flow when RBAC reject and accept.

https://github.com/istio/proxy/blob/master/test/integration/istio_http_integration_test.cc

PR #1907, opened 2018-08-09
add rbac filter to istio http integration test.

yangminzhu:
**What this PR does / why we need it**: Add RBAC filter to test metadata passing.

Issue: https://github.com/istio/proxy/issues/1902


test/integration/istio_http_integration_test.cc
```diff
@@ -51,6 +52,22 @@ constexpr char kGoodToken[] =
     "XOxJ3Uc1MI0tz3uMiiZcyPV7sNCU4KRnemRIMHVOfuvHsU60_GhGbiSFzgPT"
     "Aa9WTltbnarTbxudb_YEOx12JiwYToeX0DCPb43W1tzIBxgm8NxUg";
 
+// Generate by gen-jwt.py as described in
+// https://github.com/istio/istio/blob/master/security/tools/jwt/samples/README.md
+// to generate token with another issuer.
+// `security/tools/jwt/samples/gen-jwt.py security/tools/jwt/samples/key.pem
+//  --expire=3153600000 --claims=foo:bar --iss "testing2@secure.istio.io"`
```
diemtvu: [optional] You may change the claims to make it more diverge. Even change issuer to smt totally different, so it's easy to spot in the log output.

yangminzhu: Done, updated the issuer to `testing-rbac`

yangminzhu:
@qiwzhang Could you review this change? Thank you.

test/integration/istio_http_integration_test.cc
```diff
@@ -184,6 +234,7 @@ class IstioHttpIntegrationTest : public HttpProtocolIntegrationTest {
 
   void SetUp() override {
     config_helper_.addFilter(MakeMixerFilterConfig());
+    config_helper_.addFilter(MakeRbacFilterConfig());
```
qiwzhang: Will this rbac filter impact other tests?   Is it possible only add this filter specific for your rbac test?

yangminzhu: It could have impact on other tests but if anything goes wrong, `TEST_P(IstioHttpIntegrationTest, RbacDeny)` should catch that.

Also note this file is already testing all filters together (jwt, authn, rbac and mixer), and I can't test RBAC filter without using authn and jwt filter, so I think it makes more sense to put it here.

qiwzhang: OK then

PR #1908, opened 2018-08-10
Not to send Authorization header to mixer

qiwzhang:

Not to send Authorization header to mixer

**Release note**:

```release-note
None
```

### DIFF
--- a/src/envoy/http/mixer/check_data.cc
+++ b/src/envoy/http/mixer/check_data.cc
@@ -15,6 +15,7 @@
 
 #include "src/envoy/http/mixer/check_data.h"
 #include "common/common/base64.h"
+#include "common/http/headers.h"
 #include "src/envoy/http/jwt_auth/jwt.h"
 #include "src/envoy/http/jwt_auth/jwt_authenticator.h"
 #include "src/envoy/utils/authn.h"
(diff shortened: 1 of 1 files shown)


kyessenov:
Can we scrape it only for telemetry? I am concerned that a third party
adapter might be affected if it uses authorization header.

jboeuf:
@liminw FYI.

Thanks much Wayne for the quick PR. Can we have a group of security sensitive headers that `Authorization` would be part of? That would make the solution more generic.

kyessenov:
/hold

Let me confirm that apigee adapter is not using this header in their Mixer adapter.

jboeuf:
I'll just copy-paste what I wrote on the related issue :).

My concern is mainly for telemetry / logging (a potentially big number of humans having access to auth tokens in logs).

For authorization adapters, I would much prefer for the authentication to be done on the proxy and that information being sent out (without the original auth token) to the mixer's authorization checks but I understand that this may be a disruptive change in the short term so I'm OK not filtering the security sensitive headers for check calls for now.

Does that help?

kyessenov:
Yes, thanks for the explanation, and that seems like the right approach. We'll need to work with existing authorization adapter developers in Mixer to move authentication bits into the proxy, and that might take time. I'm concerned removing them might disrupt their functionality as-is.

jboeuf:
@kyessenov understood that moving logic around will take time. I'm completely fine with the short term plan. Thanks much for pointing out the disruption risk!

PR #1923, opened 2018-08-15
set response flag for mixer filter

kyessenov:


src/envoy/http/mixer/filter.cc
```diff
@@ -221,6 +221,8 @@ void Filter::completeCheck(const CheckResponseInfo& info) {
     int status_code = ::istio::utils::StatusHttpCode(status.error_code());
     decoder_callbacks_->sendLocalReply(Code(status_code), status.ToString(),
                                        nullptr);
+    decoder_callbacks_->requestInfo().setResponseFlag(
+        RequestInfo::ResponseFlag::UnauthorizedExternalService);
```
mandarjog: Is suppose `fail policy` is used somewhere in `istio::utils::StatusHttpCode` or `sendLocalReply` ?

kyessenov: It's in gRPC transport used by mixerfilter. Fail-open means you will get status ok here.

mandarjog:
partially addresses https://github.com/istio/istio/issues/7959

src/envoy/http/mixer/filter.cc
mandarjog: thnx, can you update the description to `Set ResponseFlag`

kyessenov:
@sebastienvas something is wrong with prow build environment in proxy...

PR #1925, opened 2018-08-17
Authn uses protobuf.Struct to store claims and add list support for RBAC

### LINKED ISSUE #1920: Store JWT claims to a protobuf struct in the authn filter
Currently, the JWT claims are stored as a map<string, string> (https://github.com/istio/proxy/blob/master/src/istio/authn/context.proto#L37) . 
This issue is created to store JWT claims to a google.protobuf.Struct in the authn filter.  It will change the "claims" to proto struct, not individual entry under claims. For example, the 'group' entry can be a Value of type StringList, "iss" is value of type "String".

lei-tang:
1. Change authn to use protobuf.Struct to store claims
2. Add list support for RBAC

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1920 

**Release note**:

```release-note
```


lei-tang:
https://github.com/istio/istio/issues/6984

src/envoy/utils/authn.cc
```diff
@@ -57,25 +57,18 @@ void Authentication::SaveAuthAttributesToStruct(
...
-      for (int i = 0; i < origin.groups().size(); i++) {
-        value->add_values()->set_string_value(origin.groups(i));
-      }
-    }
     if (!origin.presenter().empty()) {
       setKeyValue(data, istio::utils::AttributeName::kRequestAuthPresenter,
                   origin.presenter());
     }
-    if (!origin.claims().empty()) {
+    if (!origin.claims().fields().empty()) {
```
diemtvu: I think you can simply set 
```
(*data.mutable_fields())
                   [istio::utils::AttributeName::kRequestAuthClaims]	                 =  orgin.claims();
```

src/istio/authn/context.proto
```diff
@@ -33,17 +35,13 @@ message JwtPayload {
   // id. Example 123456789012.my-svc.com
   string presenter = 3;
 
-  // Only raw JWT string claims are kept.
-  map<string, string> claims = 5;
+  // JWT claims stored as protobuf.Struct
```
diemtvu: Keep `comments that only string and stringlist are kept. String claims are kept string string list of one item`,

include/istio/utils/attributes_builder.h
```diff
@@ -99,11 +99,20 @@ class AttributesBuilder {
...
-      // Ignore all fields that are not string.
+      // Ignore all fields that are not string or string list.
       switch (field.second.kind_case()) {
         case google::protobuf::Value::kStringValue:
           (*entries)[field.first] = field.second.string_value();
           break;
+        case google::protobuf::Value::kListValue:
+          if (field.second.list_value().values_size() > 0) {
+            // Only uses the first item in the list as string
+            // TODO (lei-tang): the items in the list may be converted into a
```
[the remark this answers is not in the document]
lei-tang: Done.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -25,51 +26,21 @@ namespace {
...
-// This function should to be called after the claims are extracted.
-void ExtractJwtAudience(
-    const Envoy::Json::Object& obj,
-    const ::google::protobuf::Map< ::std::string, ::std::string>& claims,
-    istio::authn::JwtPayload* payload) {
-  const std::string& key = kJwtAudienceKey;
-  // "aud" can be either string array or string.
+// Extract JWT claim as a string list
+void ExtractStringList(const std::string& key, const Envoy::Json::Object& obj,
+                       std::vector<std::string>& list) {
```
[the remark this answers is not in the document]

src/envoy/http/authn/authn_utils.cc
```diff
@@ -25,51 +26,21 @@ namespace {
...
-
-// Extract JWT audience into the JwtPayload.
-// This function should to be called after the claims are extracted.
-void ExtractJwtAudience(
-    const Envoy::Json::Object& obj,
-    const ::google::protobuf::Map< ::std::string, ::std::string>& claims,
-    istio::authn::JwtPayload* payload) {
-  const std::string& key = kJwtAudienceKey;
-  // "aud" can be either string array or string.
+// Extract JWT claim as a string list
```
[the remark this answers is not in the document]

src/envoy/utils/authn.cc
lei-tang: Done.

src/istio/authn/context.proto
lei-tang: Done.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -89,37 +62,39 @@ bool AuthnUtils::ProcessJwtPayload(const std::string& payload_str,
...
-  // Extract audience
-  // ExtractJwtAudience() should be called after claims are extracted.
-  ExtractJwtAudience(*json_obj, payload->claims(), payload);
-  // ExtractJwtGroups() should be called after claims are extracted.
-  ExtractJwtGroups(*json_obj, payload->claims(), payload);
+  auto claims = payload->mutable_claims()->mutable_fields();
+  // Extract claims as string lists
+  json_obj->iterate([json_obj, claims](const std::string& key,
+                                       const Json::Object&) -> bool {
+    // In current implementation, only string/string list objects are extracted
```
yangminzhu: Instead of looping over the JSON object manually, why not use `ParseJsonMessage()` to parse the whole json string to the `claim` field? So that you could simplify the ProcessJwtPayload() a lot.

lei-tang: @yangminzhu The output of ParseJsonMessage() is of the type Message: 
Status ParseJsonMessage(const std::string& json, Message* output). Here the claims is of the type protobuf.Struct. Can you provide details of how to use ParseJsonMessage() to parse the json object into protobuf.Struct? One more thing, if the json object is parsed into the claims Struct as different types (e.g., strings, string lists, structs, and etc), how does the consumer of the claims interpret the type of a claim (since the type for a claim value is not only a string list anymore)?

yangminzhu: `Struct` is a subclass of `Message`, so it should be able to pass to `ParseJsonMessage()` directly.

I'm curious where are we using the `claims` field other than in `authn_utils.cc`? The type is changed and it's seems only used in this file directly?

@diemtvu WDYT?

lei-tang: The consumers of the claims protobuf.Struct include the producers of dynamic metadata and tests in the Istio proxy repo. If the claims become a map from string key to protobuf.Struct value, every consumer of the claims needs to add some code to verify the type of the claim value before accessing a claim. In contrast, with current implementation, the contract between the producer of the claims and the consumers is that a claim value must be of the string list type and a consumer directly reads a claim value as a string list without extra code to verify the type of a claim.

yangminzhu: The `claims`'s type is already changed to `Struct` so why it matters about how you construct it? If the constructed `Struct` doesn't have a string or string list as your implicit contract here, the consumer still get the same value as today when using the `.string_value()` method on the `Struct`. So there shouldn't have any changes on consumer side once its stored in dynamic metadata.

src/envoy/utils/authn.cc
```diff
@@ -57,26 +57,14 @@ void Authentication::SaveAuthAttributesToStruct(
...
-    if (!origin.claims().empty()) {
-      auto s = (*data.mutable_fields())
-                   [istio::utils::AttributeName::kRequestAuthClaims]
-                       .mutable_struct_value();
-      for (const auto& pair : origin.claims()) {
-        setKeyValue(*s, pair.first, pair.second);
-      }
+    if (!origin.claims().fields().empty()) {
+      *((*data.mutable_fields())
+            [istio::utils::AttributeName::kRequestAuthClaims]
```
qiwzhang: call setKeyValue?

include/istio/utils/attributes_builder.h
```diff
@@ -99,11 +99,25 @@ class AttributesBuilder {
...
       switch (field.second.kind_case()) {
         case google::protobuf::Value::kStringValue:
           (*entries)[field.first] = field.second.string_value();
           break;
+        case google::protobuf::Value::kListValue:
+          if (field.second.list_value().values_size() > 0) {
+            // The items in the list is converted into a
+            // comma separated string
+            std::string s;
+            for (int i = 0; i < field.second.list_value().values_size(); i++) {
```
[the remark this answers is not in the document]
lei-tang: "int i" here is used to avoid adding a comma to the last list item. Since the for loop here of six lines is only used at other places, it is not placed in a separate util function.

src/envoy/http/authn/authn_utils.cc
diemtvu: You can explore, but I do think ParseJsonMessage, even if it works with Struct, may has some limitation. For example, in our approach, we want to have consistent representation for string + string array. I'm ok with this custom conversion for now, and find a better utils later (or refactor them to a util function)

yangminzhu: Could you point me to one of the consumer side code that needs to change if you use `ProcessJwtPayload()`?

include/istio/utils/attributes_builder.h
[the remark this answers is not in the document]
diemtvu: May be using string join utils under https://github.com/envoyproxy/envoy/blob/master/source/common/common/utility.h#L322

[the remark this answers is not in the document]
lei-tang: join() input argument is a vector<std::string>: static std::string join(const std::vector<std::string>& source, const std::string& delimiter). To use join(), field.second needs to be converted to vector<std::string>. I have not found a one-liner to do the conversion.

src/envoy/http/authn/authn_utils.cc
lei-tang: The following is a few examples on the consumer side code that need to change if use ParseJsonMessage():
- In /usr/local/google/home/leitang/git/istio/proxy/src/envoy/utils/authn.cc: 
    payload->set_user(
        (*claims)["iss"].list_value().values().Get(0).string_value() + "/" +
        (*claims)["sub"].list_value().values().Get(0).string_value());
    payload->set_presenter(
        (*claims)["azp"].list_value().values().Get(0).string_value());

lei-tang: Another example on the consumer side code that need to change if use ParseJsonMessage():
  // Copy audience to the audience in context.proto
  if (claims->find(kJwtAudienceKey) != claims->end()) {
    for (const auto& v : (*claims)[kJwtAudienceKey].list_value().values()) {
      payload->add_audiences(v.string_value());
    }
  }

yangminzhu: These are just right in this function, I'm trying to see if there are any consumer code outside this file and I will be supersized if it is consuming the `claim` field the same way as the examples, i.e. getting string directly from the list.

Anyway, I'm ok if you would like to go with the current approach, we can discuss more and refactor later.

lei-tang:
https://github.com/istio/istio/issues/2802

PR #1928, opened 2018-08-21
support per-path JWT validation.

yangminzhu:
**What this PR does / why we need it**: Supports per-path JWT validation.

**Release note**:

```release-note
Support JWT validation based on request path.
```

### DIFF
--- a/src/envoy/http/authn/authn_utils.cc
+++ b/src/envoy/http/authn/authn_utils.cc
@@ -150,7 +150,11 @@ static bool matchRule(const char* const path,
   return true;
 }
 
-bool AuthnUtils::IsJwtTriggered(const char* const path, const iaapi::Jwt& jwt) {
+bool AuthnUtils::ShouldValidateJwtPerPath(const char* const path,
+                                          const iaapi::Jwt& jwt) {
+  // If the path is nullptr which shouldn't happen for a HTTP request or if
+  // there are no trigger rules at all, then simply return true as if there're
+  // no per-path jwt support.
   if (path == nullptr || jwt.trigger_rules_size() == 0) {
     return true;
   }
--- a/src/envoy/http/authn/authn_utils.cc
+++ b/src/envoy/http/authn/authn_utils.cc
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+#include <regex>
+
 #include "authn_utils.h"
 #include "common/json/json_loader.h"
 #include "google/protobuf/struct.pb.h"
@@ -96,6 +98,66 @@ bool AuthnUtils::ProcessJwtPayload(const std::string& payload_str,
   }
 
   return true;
+}
+
+bool AuthnUtils::MatchString(const char* str, const iaapi::StringMatch& match) {
+  if (str == nullptr) {
+    return false;
+  }
+  switch (match.match_type_case()) {
+    case iaapi::StringMatch::kExact: {
+      return match.exact().compare(str) == 0;
+    }
+    case iaapi::StringMatch::kPrefix: {
+      return StringUtil::startsWith(str, match.prefix());
+    }
+    case iaapi::StringMatch::kSuffix: {
+      return StringUtil::endsWith(str, match.suffix());
+    }
+    case iaapi::StringMatch::kRegex: {
+      return std::regex_match(str, std::regex(match.regex()));
+    }
+    default:
+      return false;
+  }
+}
+
+static bool matchRule(const char* path, const iaapi::Jwt_TriggerRule& rule) {
+  for (const auto& excluded : rule.excluded_paths()) {
+    if (AuthnUtils::MatchString(path, excluded)) {
+      // The rule is not matched if any of excluded_paths matched.
+      return false;
+    }
+  }
+
+  if (rule.included_paths_size() > 0) {
+    for (const auto& included : rule.included_paths()) {
+      if (AuthnUtils::MatchString(path, included)) {
+        // The rule is matched if any of included_paths matched.
+        return true;
+      }
+    }
+
+    // The rule is not matched if included_paths is not empty and none of them
+    // matched.
+    return false;
+  }
+
+  // The rule is matched if none of excluded_paths matched and included_paths is
+  // empty.
+  return true;
+}
+
+bool AuthnUtils::IsJwtTriggered(const char* path, const iaapi::Jwt& jwt) {
+  if (path == nullptr || jwt.trigger_rules_size() == 0) {
+    return true;
+  }
+  for (const auto& rule : jwt.trigger_rules()) {
+    if (matchRule(path, rule)) {
+      return true;
+    }
+  }
+  return false;
 }
 
 }  // namespace AuthN
--- a/src/envoy/http/authn/authn_utils.cc
+++ b/src/envoy/http/authn/authn_utils.cc
@@ -100,7 +100,8 @@ bool AuthnUtils::ProcessJwtPayload(const std::string& payload_str,
   return true;
 }
 
-bool AuthnUtils::MatchString(const char* str, const iaapi::StringMatch& match) {
+bool AuthnUtils::MatchString(const char* const str,
+                             const iaapi::StringMatch& match) {
   if (str == nullptr) {
     return false;
   }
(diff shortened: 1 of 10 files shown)


src/envoy/http/authn/origin_authenticator.cc
```diff
@@ -57,10 +58,29 @@ bool OriginAuthenticator::run(Payload* payload) {
...
+              "validation");
+  }
+
   for (const auto& method : policy_.origins()) {
-    if (validateJwt(method.jwt(), payload)) {
-      success = true;
-      break;
+    const auto& jwt = method.jwt();
+
+    if (AuthnUtils::IsJwtTriggered(request_path, jwt)) {
```
qiwzhang: Upstream have per-route or per-path requirement config. https://github.com/envoyproxy/envoy/blob/master/api/envoy/config/filter/http/jwt_authn/v2alpha/config.proto#L401

It is not clear to me how this implementation aligns with it.  If not, I think you have to re-do it with you use Envoy jwt_authn filter.

yangminzhu: This PR is for istio authn filter and the decision of if using a JWT is made in the authn filter. I think the Envoy jwt_authn is currently not used in proxy, right? Switching to it is a larger refactor that is independent of this PR.

After switched to the Envoy jwt_authn filter, we can update the implementation if we decided it's better to do this in jwt_authn filter instead of our proxy authn filter. There should be no user facing difference in two implementations.

@diemtvu WDYT?

qiwzhang: My point is:  have you looked at the Envoy jwt_authn per-path requirement config?   How easy is the current authn per-path config align with that config so it is easier to switch in the future? 

I just briefly looked at your config, e.g. inclusive, exclusive stuff.  Envoy config doesn't have that.  It will be hard to switch in the future.

liminw: I agree with @qiwzhang. It is better to align with upstream JWT filter now. Otherwise, we keep adding more logic in proxy code, which will have to be rewritten when we switch to upstream Envoy filter.

yangminzhu: @qiwzhang 
The current authn per-path semantic is a subset of the upstream JWT filter.
For a `included_path`, it could be converted to a `JwtRequirement` with `provider_name`, which means a specific path must have the specified JWT validation. 
For a `excluded_path`, it should be converted to a `JetRequirement` without provider, which means the path doesn't require jwt validation, e.g. `/health`.


@liminw The current authn per-path semantic is a subset of the upstream JWT filter. It takes time to switch to upstream Envoy filter and when that is done, we still need to convert the existing pilot code to generate corresponding config. For now, we don't need to modify pilot code and the only change we need is in this PR.

yangminzhu: I just had a second thought on this, it's okay for single JWT case but it turns out to be a problem when using multiple JWT with `included_paths` and `excluded_paths`. So I'll update the API design to reflect this.

diemtvu: I think I mentioned that (upstream jwt) to @yangminzhu at some point, though after some thoughts, I think it's ok as we may not to implement / use this feature in JWT filter. This will be used by authn_filter only, JWT filter still do validation regardless of the path. 

Being said that, I also realized that bundle this information inside JWT params may cause some confusion. On the other hand, to customers (who don't care or know about envoy implementation details), it may be ok. Thoughts?

yangminzhu: If it's okay to use this in authn_filter only, I think I would like to go with this way. The reason is if we really want to align with the upstream jwt config completely, we have to make some breaking changes (the user have to update their existing manually) to the current  [Policy](https://github.com/istio/api/blob/fac61dc6aaec82b5824918f379e29ff48afcd2cb/authentication/v1alpha1/policy.proto#L311).  And we tried our best to make this change backward compatible in the design doc.

The upstream jwt config focused on path first and then connect the path to a group of named JWTs, however, in proxy authn filter, we focused on JWT first, The JWT don't have name and we don't support any complex `and`/`or` logic.

@liminw  @qiwzhang WDYT?

src/envoy/http/authn/authn_utils.cc
```diff
@@ -123,6 +125,66 @@ bool AuthnUtils::ProcessJwtPayload(const std::string& payload_str,
   return true;
 }
 
+bool AuthnUtils::MatchString(const char* str, const iaapi::StringMatch& match) {
```
diemtvu: Use `const std::string&` or `const char* const`

yangminzhu:
@JimmyCYJ Could you help review this? I just talked with @diemtvu and we're okay to go with this way (regarding the concern from @qiwzhang), Thanks!

src/envoy/http/authn/origin_authenticator.cc
```diff
@@ -57,10 +58,29 @@ bool OriginAuthenticator::run(Payload* payload) {
...
+      ENVOY_LOG(debug, "Validating JWT: {}", jwt.DebugString());
```
JimmyCYJ: Maybe add request path into ENVOY_LOG as well?

src/envoy/http/authn/origin_authenticator_test.cc
```diff
@@ -219,6 +303,54 @@ TEST_P(OriginAuthenticatorTest, MultipleFail) {
...
+
+  createAuthenticator();
+  EXPECT_CALL(*authenticator_, validateJwt(_, _)).Times(0);
+
+  setPath("/bad");
+  EXPECT_TRUE(authenticator_->run(payload_));
+  EXPECT_TRUE(TestUtility::protoEqual(initial_result_,
+                                      filter_context_.authenticationResult()));
+}
+
```
JimmyCYJ: Add a test for the case that path is not existed in headers?

src/envoy/http/authn/authn_utils.h
```diff
@@ -33,6 +37,14 @@ class AuthnUtils : public Logger::Loggable<Logger::Id::filter> {
...
   static bool ProcessJwtPayload(const std::string& jwt_payload_str,
                                 istio::authn::JwtPayload* payload);
+
+  // Returns true if str is matched to match.
+  static bool MatchString(const char* const str,
+                          const iaapi::StringMatch& match);
+
+  // Returns true if the jwt should be validated. It will check if the request
+  // path is matched to the trigger rule in the jwt.
+  static bool IsJwtTriggered(const char* const path, const iaapi::Jwt& jwt);
```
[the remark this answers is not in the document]
yangminzhu: Done

src/envoy/http/authn/authn_utils.cc
```diff
@@ -98,6 +100,68 @@ bool AuthnUtils::ProcessJwtPayload(const std::string& payload_str,
...
+    return false;
+  }
+
+  // The rule is matched if none of excluded_paths matched and included_paths is
+  // empty.
+  return true;
+}
+
+bool AuthnUtils::IsJwtTriggered(const char* const path, const iaapi::Jwt& jwt) {
+  if (path == nullptr || jwt.trigger_rules_size() == 0) {
```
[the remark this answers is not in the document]
yangminzhu: Done.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -98,6 +100,68 @@ bool AuthnUtils::ProcessJwtPayload(const std::string& payload_str,
...
+    return true;
+  }
```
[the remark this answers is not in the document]
yangminzhu: The null check in `MatchString()` returns false as it's in the context of `MatchString()` which could be used by anyone other than `IsJwtTriggered()`. In this context, it's more natural to return a not matched if the passed in argument is null.

For here (line 153), it returns true as it should never happen that a path is null for a HTTP request, so there must be something badly wrong, and probably we just pretend there is no per-path stuff at all in this case, which is consistent to the current behavior.

src/envoy/http/authn/origin_authenticator.cc
yangminzhu: Done

src/envoy/http/authn/origin_authenticator_test.cc
yangminzhu: Done

yangminzhu:
@JimmyCYJ Comments addressed, PTAL, thank you

JimmyCYJ:
/lgtm 

Thanks!

PR #1946, opened 2018-08-27
mixer: clear route cache on header update

kyessenov:
Clear HTTP route table on header change. This allows mixer to re-route the traffic.

/assign @lizan 
/assign @PiotrSikora


src/envoy/http/mixer/filter.cc
```diff
@@ -245,6 +245,7 @@ void Filter::completeCheck(const CheckResponseInfo& info) {
   if (nullptr != headers_) {
     UpdateHeaders(*headers_, route_directive_.request_header_operations());
     headers_ = nullptr;
+    decoder_callbacks_->clearRouteCache();
```
lizan: clear only `request_header_operations` is not empty?

kyessenov: Done. Forgot to add this check.

PR #1954, opened 2018-08-30
Rename filter to follow naming convention.

diemtvu:

Use `istio.` prefix for filter name.

**Release note**:
This PR should be followed with change in [Pilot](https://github.com/istio/istio/blob/master/pilot/pkg/networking/plugin/authn/authentication.go#L42) when updating proxy SHA

### DIFF
--- a/src/envoy/http/authn/http_filter_integration_test.cc
+++ b/src/envoy/http/authn/http_filter_integration_test.cc
@@ -44,7 +44,7 @@ static const Http::TestHeaderMapImpl kSimpleRequestHeader{{
 static const char kJwtIssuer[] = "some@issuer";
 
 static const char kAuthnFilterWithJwt[] = R"(
-    name: istio_authn
+    name: istio.authn
     config:
       policy:
         origins:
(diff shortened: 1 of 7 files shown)


diemtvu:
PTAL

src/envoy/http/authn/http_filter_integration_test.cc
```diff
@@ -104,7 +104,7 @@ TEST_P(AuthenticationFilterIntegrationTest, EmptyPolicy) {
 
 TEST_P(AuthenticationFilterIntegrationTest, SourceMTlsFail) {
   config_helper_.addFilter(R"(
-    name: istio_authn
+    name: istio.authn
     config:
```
JimmyCYJ: Is it possible to cover both old name and new name? Once we deprecate the old name, we remove the test as well.

diemtvu:
Actually, it's still possible breaking stuff if we does not update pilot to use the new name right away :(. Making this fully backward compatible is a bit hairy (we will also need to change all filters to fallback to the old name when accessing filter dynamic metadata). On the other hand, change in pilot is just few lines and can be done right after this PR (together with SHA update). @kyessenov , do you think it's ok to do so? Otherwise, I can wait for your PR that merge proxy repo to istio, or not even bother to rename them.

kyessenov:
@diemtvu I am not an owner in either pilot or upgrade parts. You should probably talk to @costinm .

diemtvu:
postpone

PR #1957, opened 2018-09-04
Do not submit:  code change discussion.

qiwzhang:

**Release note**:

```release-note
```

### DIFF
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -19,7 +19,7 @@
 #include "include/istio/control/http/request_handler.h"
 #include "include/istio/mixerclient/client.h"
 #include "include/istio/utils/attribute_names.h"
-#include "include/istio/utils/attributes_builder.h"
+#include "include/istio/utils/local_attributes.h"
 #include "mixer/v1/config/client/client_config.pb.h"
 
 namespace istio {
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -72,7 +72,7 @@ class Controller {
   // * optional service config cache size.
   struct Options {
     Options(const ::istio::mixer::v1::config::client::HttpClientConfig& config,
-            const ::istio::utils::LocalAttributes& local_attributes)
+            const ::istio::utils::LocalAttributes* local_attributes)
         : config(config), local_attributes(local_attributes) {}
 
     // Mixer filter config
@@ -86,7 +86,7 @@ class Controller {
     int service_config_cache_size{};
 
     // local attributes
-    const ::istio::utils::LocalAttributes& local_attributes;
+    const ::istio::utils::LocalAttributes* local_attributes;
   };
 
   // The factory function to create a new instance of the controller.
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -72,8 +72,9 @@ class Controller {
   // * optional service config cache size.
   struct Options {
     Options(const ::istio::mixer::v1::config::client::HttpClientConfig& config,
-            const ::istio::utils::LocalAttributes* local_attributes)
-        : config(config), local_attributes(local_attributes) {}
+            std::unique_ptr<const ::istio::utils::LocalAttributes>&
+                local_attributes)
+        : config(config), local_attributes(std::move(local_attributes)) {}
 
     // Mixer filter config
     const ::istio::mixer::v1::config::client::HttpClientConfig& config;
@@ -86,7 +87,7 @@ class Controller {
     int service_config_cache_size{};
 
     // local attributes. not owned by options.
-    const ::istio::utils::LocalAttributes* local_attributes;
+    std::unique_ptr<const ::istio::utils::LocalAttributes> local_attributes;
   };
 
   // The factory function to create a new instance of the controller.
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -16,6 +16,7 @@
 #ifndef ISTIO_CONTROL_HTTP_CONTROLLER_H
 #define ISTIO_CONTROL_HTTP_CONTROLLER_H
 
+#include "envoy/local_info/local_info.h"
 #include "include/istio/control/http/request_handler.h"
 #include "include/istio/mixerclient/client.h"
 #include "mixer/v1/config/client/client_config.pb.h"
@@ -81,6 +82,8 @@ class Controller {
     // The LRU cache size for service config.
     // If not set or is 0 default value, the cache size is 1000.
     int service_config_cache_size{};
+
+    const ::envoy::LocalInfo::LocalInfo& local_info;
   };
 
   // The factory function to create a new instance of the controller.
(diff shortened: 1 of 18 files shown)


src/envoy/utils/mixer_control.cc
```diff
@@ -100,5 +102,67 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+     },
+     "build_version": "0/1.8.0-dev//RELEASE"
+    },
+
+    ==> uid: kubernetes://fortioclient-84469dc8d7-jbbxt.service-graph
+    reporter == uid
+    namespace
+    IP_Address only for inbound.
+**/
+const LocalAttributes GenerateLocalAttributes(
```
qiwzhang: The expected format should be clearly documented in the comment.  This is very error pron. Not sure if how often Pilot agent changes this format,   May have to add comment in the Pilot agent

src/envoy/utils/mixer_control.cc
```diff
@@ -100,5 +102,67 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+  ::istio::mixer::v1::Attributes ob;
+  auto &outbound = (*ob.mutable_attributes());
+  outbound[AttributeName::kSourceUID].set_string_value(uid);
+  outbound[AttributeName::kContextReporterUID].set_string_value(uid);
+  outbound[AttributeName::kSourceNamespace].set_string_value(ns);
+
+  ::istio::mixer::v1::Attributes fwd;
+  auto &forward = (*fwd.mutable_attributes());
+  forward[AttributeName::kSourceUID].set_string_value(uid);
+  return LocalAttributes(ib, ob, fwd);
```
qiwzhang: It is OK to return  the whole struct,  not need to use unique_ptr.   I think compiler will optimize it

src/envoy/utils/mixer_control.cc
```diff
@@ -100,5 +102,67 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+
+  ::istio::mixer::v1::Attributes ob;
+  auto &outbound = (*ob.mutable_attributes());
+  outbound[AttributeName::kSourceUID].set_string_value(uid);
+  outbound[AttributeName::kContextReporterUID].set_string_value(uid);
+  outbound[AttributeName::kSourceNamespace].set_string_value(ns);
+
+  ::istio::mixer::v1::Attributes fwd;
+  auto &forward = (*fwd.mutable_attributes());
+  forward[AttributeName::kSourceUID].set_string_value(uid);
```
qiwzhang: I did not see forward attributes is used in the diff.

include/istio/control/http/controller.h
```diff
@@ -81,6 +84,9 @@ class Controller {
     // The LRU cache size for service config.
     // If not set or is 0 default value, the cache size is 1000.
     int service_config_cache_size{};
+
+    // local attributes
+    const ::istio::utils::LocalAttributes* local_attributes;
```
qiwzhang: It should be rare to hold a raw pointer since ownership is not clear.   Please use unique_ptr

include/istio/utils/local_attributes.h
```diff
@@ -0,0 +1,44 @@
...
+#define ISTIO_UTILS_LOCAL_ATTRIBUTES_H
+
+#include "mixer/v1/attributes.pb.h"
+
+using ::istio::mixer::v1::Attributes;
+
+namespace istio {
+namespace utils {
+
+typedef struct LocalAttributes_t {
```
qiwzhang: No need for this typedef.  just

struct LocalAttributes {

src/envoy/utils/mixer_control.cc
```diff
@@ -100,5 +102,67 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+        << "GenerateLocalAttributes error len(split(longname, '.')) < 3: "
+        << longname;
+    return nullptr;
+  }
+
+  std::string ns = std::string(names[1].begin(), names[1].end());
+  std::string uid = "kubernetes://" + longname;
+
+  ::istio::mixer::v1::Attributes ib;
+  auto &inbound = (*ib.mutable_attributes());
```
qiwzhang: Can we use AttributeBuilder defined here: https://github.com/istio/proxy/blob/master/include/istio/utils/attributes_builder.h

src/envoy/utils/mixer_control.cc
```diff
@@ -100,5 +102,67 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+  auto &outbound = (*ob.mutable_attributes());
+  outbound[AttributeName::kSourceUID].set_string_value(uid);
+  outbound[AttributeName::kContextReporterUID].set_string_value(uid);
+  outbound[AttributeName::kSourceNamespace].set_string_value(ns);
+
+  ::istio::mixer::v1::Attributes fwd;
+  auto &forward = (*fwd.mutable_attributes());
+  forward[AttributeName::kSourceUID].set_string_value(uid);
+  return new LocalAttributes(ib, ob, fwd);
+}
```
qiwzhang: Please use unique_ptr

src/envoy/utils/mixer_control.cc
```diff
@@ -100,5 +102,137 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+  auto &forward = (*fwd.mutable_attributes());
+  forward[AttributeName::kSourceUID].set_string_value(local.uid);
+  return new LocalAttributes(ib, ob, fwd);
+}
+
+//
+// "sidecar~10.36.0.15~fortioclient-84469dc8d7-jbbxt.service-graph~service-graph.svc.cluster.local"
+//  --> {proxy_type}~{ip}~{node_name}.{node_ns}~{node_domain}
+bool extractInfo(localAttributesArgs *args, std::string nodeid) {
+  auto parts = StringUtil::splitToken(nodeid, "~");
```
kyessenov: this makes me very sad...

mandarjog: That is there only for compatibility with existing node names.

The other function uses node metadata.

kyessenov:
Is the goal here to stuff shared attributes into node metadata struct?
Or to move pilot plugin mixerfilter logic into envoy?

mandarjog:
The goal is for envoy to not need all this stuff in LDS configuration.
Node metadata is populated by proxy agent, which is then picked up by envoy and sent over to Mixer.

include/istio/utils/local_attributes.h
```diff
@@ -0,0 +1,44 @@
...
+                  const Attributes& forward)
+      : inbound(inbound), outbound(outbound), forward(forward) {}
+
+  // local inbound attributes
+  const Attributes inbound;
+
+  // local outbound attributes
+  const Attributes outbound;
+
+  // local forward attributes
```
kyessenov: I think we should just have a map from a string to a service config in the bootstrap (https://github.com/istio/api/blob/master/mixer/v1/config/client/client_config.proto#L52), and then add a string reference to service configs in LDS for defaulting. I don't like inbound/outbound stuff here, since it's not applicable to middle proxies.

PR #1961, opened 2018-09-06
Mixer Client uses Node metadata to populate Mixer attributes

mandarjog:
At present Pilot programs all attributes including proxy identity attributes like destination.uid / destination.namespace. This bloats the configuration and makes it difficult to cache.

This PR uses node metadata that is populated by proxy agent in boostrap to determine identity attributes.
It uses the old name parsing as a backup.

### DIFF
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -19,7 +19,7 @@
 #include "include/istio/control/http/request_handler.h"
 #include "include/istio/mixerclient/client.h"
 #include "include/istio/utils/attribute_names.h"
-#include "include/istio/utils/attributes_builder.h"
+#include "include/istio/utils/local_attributes.h"
 #include "mixer/v1/config/client/client_config.pb.h"
 
 namespace istio {
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -72,7 +72,7 @@ class Controller {
   // * optional service config cache size.
   struct Options {
     Options(const ::istio::mixer::v1::config::client::HttpClientConfig& config,
-            const ::istio::utils::LocalAttributes& local_attributes)
+            const ::istio::utils::LocalAttributes* local_attributes)
         : config(config), local_attributes(local_attributes) {}
 
     // Mixer filter config
@@ -86,7 +86,7 @@ class Controller {
     int service_config_cache_size{};
 
     // local attributes
-    const ::istio::utils::LocalAttributes& local_attributes;
+    const ::istio::utils::LocalAttributes* local_attributes;
   };
 
   // The factory function to create a new instance of the controller.
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -72,8 +72,9 @@ class Controller {
   // * optional service config cache size.
   struct Options {
     Options(const ::istio::mixer::v1::config::client::HttpClientConfig& config,
-            const ::istio::utils::LocalAttributes* local_attributes)
-        : config(config), local_attributes(local_attributes) {}
+            std::unique_ptr<const ::istio::utils::LocalAttributes>&
+                local_attributes)
+        : config(config), local_attributes(std::move(local_attributes)) {}
 
     // Mixer filter config
     const ::istio::mixer::v1::config::client::HttpClientConfig& config;
@@ -86,7 +87,7 @@ class Controller {
     int service_config_cache_size{};
 
     // local attributes. not owned by options.
-    const ::istio::utils::LocalAttributes* local_attributes;
+    std::unique_ptr<const ::istio::utils::LocalAttributes> local_attributes;
   };
 
   // The factory function to create a new instance of the controller.
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -16,6 +16,7 @@
 #ifndef ISTIO_CONTROL_HTTP_CONTROLLER_H
 #define ISTIO_CONTROL_HTTP_CONTROLLER_H
 
+#include "envoy/local_info/local_info.h"
 #include "include/istio/control/http/request_handler.h"
 #include "include/istio/mixerclient/client.h"
 #include "mixer/v1/config/client/client_config.pb.h"
@@ -81,6 +82,8 @@ class Controller {
     // The LRU cache size for service config.
     // If not set or is 0 default value, the cache size is 1000.
     int service_config_cache_size{};
+
+    const ::envoy::LocalInfo::LocalInfo& local_info;
   };
 
   // The factory function to create a new instance of the controller.
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -18,8 +18,9 @@
 
 #include "include/istio/control/http/request_handler.h"
 #include "include/istio/mixerclient/client.h"
-#include "mixer/v1/config/client/client_config.pb.h"
 #include "include/istio/utils/attribute_names.h"
+#include "include/istio/utils/attributes_builder.h"
+#include "mixer/v1/config/client/client_config.pb.h"
 
 namespace istio {
 namespace control {
@@ -71,11 +72,8 @@ class Controller {
   // * optional service config cache size.
   struct Options {
     Options(const ::istio::mixer::v1::config::client::HttpClientConfig& config,
-        const ::istio::mixer::v1::Attributes& local_inbound_attributes,
-        const ::istio::mixer::v1::Attributes& local_outbound_attributes,
-        const ::istio::mixer::v1::Attributes& local_forward_attributes)
-        : config(config), local_inbound_attributes(local_inbound_attributes), 
-          local_outbound_attributes(local_outbound_attributes), local_forward_attributes(local_forward_attributes) {}
+            const ::istio::utils::LocalAttributes& local_attributes)
+        : config(config), local_attributes(local_attributes) {}
 
     // Mixer filter config
     const ::istio::mixer::v1::config::client::HttpClientConfig& config;
@@ -87,17 +85,8 @@ class Controller {
     // If not set or is 0 default value, the cache size is 1000.
     int service_config_cache_size{};
 
+    // local attributes
     const ::istio::utils::LocalAttributes& local_attributes;
-
-    // local_inbound attributes
-    const ::istio::mixer::v1::Attributes& local_inbound_attributes;
-
-    // local_outbound attributes
-    const ::istio::mixer::v1::Attributes& local_outbound_attributes;
-
-    // local_forward attributes
-    const ::istio::mixer::v1::Attributes& local_forward_attributes;
-      
   };
 
   // The factory function to create a new instance of the controller.
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -16,10 +16,10 @@
 #ifndef ISTIO_CONTROL_HTTP_CONTROLLER_H
 #define ISTIO_CONTROL_HTTP_CONTROLLER_H
 
-#include "envoy/local_info/local_info.h"
 #include "include/istio/control/http/request_handler.h"
 #include "include/istio/mixerclient/client.h"
 #include "mixer/v1/config/client/client_config.pb.h"
+#include "include/istio/utils/attribute_names.h"
 
 namespace istio {
 namespace control {
@@ -70,8 +70,12 @@ class Controller {
   // * some functions provided by the environment (Envoy)
   // * optional service config cache size.
   struct Options {
-    Options(const ::istio::mixer::v1::config::client::HttpClientConfig& config)
-        : config(config) {}
+    Options(const ::istio::mixer::v1::config::client::HttpClientConfig& config,
+        const ::istio::mixer::v1::Attributes& local_inbound_attributes,
+        const ::istio::mixer::v1::Attributes& local_outbound_attributes,
+        const ::istio::mixer::v1::Attributes& local_forward_attributes)
+        : config(config), local_inbound_attributes(local_inbound_attributes), 
+          local_outbound_attributes(local_outbound_attributes), local_forward_attributes(local_forward_attributes) {}
 
     // Mixer filter config
     const ::istio::mixer::v1::config::client::HttpClientConfig& config;
@@ -83,7 +87,15 @@ class Controller {
     // If not set or is 0 default value, the cache size is 1000.
     int service_config_cache_size{};
 
-    const ::envoy::LocalInfo::LocalInfo& local_info;
+    // local_inbound attributes
+    const ::istio::mixer::v1::Attributes& local_inbound_attributes;
+
+    // local_outbound attributes
+    const ::istio::mixer::v1::Attributes& local_outbound_attributes;
+
+    // local_forward attributes
+    const ::istio::mixer::v1::Attributes& local_forward_attributes;
+      
   };
 
   // The factory function to create a new instance of the controller.
--- a/include/istio/control/http/controller.h
+++ b/include/istio/control/http/controller.h
@@ -87,6 +87,8 @@ class Controller {
     // If not set or is 0 default value, the cache size is 1000.
     int service_config_cache_size{};
 
+    const ::istio::utils::LocalAttributes& local_attributes;
+
     // local_inbound attributes
     const ::istio::mixer::v1::Attributes& local_inbound_attributes;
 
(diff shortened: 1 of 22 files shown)


include/istio/control/http/controller.h
```diff
@@ -81,6 +84,8 @@ class Controller {
     // The LRU cache size for service config.
     // If not set or is 0 default value, the cache size is 1000.
     int service_config_cache_size{};
+
+    const ::istio::utils::LocalNode local_node;
```
qiwzhang: Options is transit, just used to pass data.  so you should just store LocalNode& as reference here.

src/istio/utils/local_attributes.cc
```diff
@@ -0,0 +1,51 @@
...
+std::unique_ptr<const LocalAttributes> CreateLocalAttributes(
+    const LocalNode &local) {
+  ::istio::mixer::v1::Attributes inbound;
+  AttributesBuilder ib(&inbound);
+  ib.AddString(AttributeName::kDestinationUID, local.uid);
+  ib.AddString(AttributeName::kContextReporterUID, local.uid);
+  ib.AddString(AttributeName::kDestinationNamespace, local.ns);
+
+  if (!local.ip.empty()) {
+    // TODO: mjog check if destination.ip should be setup for inbound.
```
kyessenov: Let's not put IP here. We'll need to support multi-IP pods eventually.

test/integration/istio_http_integration_test.cc
```diff
@@ -249,6 +254,22 @@ class IstioHttpIntegrationTest : public HttpProtocolIntegrationTest {
     cleanupConnection(policy_connection_);
   }
 
+  ConfigHelper::ConfigModifierFunction addNodeMetadata() {
+    return [](envoy::config::bootstrap::v2::Bootstrap& bootstrap) {
+      ::google::protobuf::Struct meta;
+      MessageUtil::loadFromJson(
+          fmt::sprintf(R"({
+        "ISTIO_VERSION": "1.0.1",
```
qiwzhang: Do we have test to verify these local attributes ?

include/istio/utils/local_attributes.h
```diff
@@ -0,0 +1,54 @@
...
+
+  // local outbound attributes
+  const Attributes outbound;
+
+  // local forward attributes
+  const Attributes forward;
+};
+
+// LocalNode are used to extract information from envoy Node.
+struct LocalNode {
```
kyessenov: There should be nothing envoy specific under istio directory. Maybe get rid of this struct and just create LocalAttributes from envoy node in utils there?

src/istio/utils/local_attributes.cc
```diff
@@ -0,0 +1,51 @@
...
+#include "include/istio/utils/attribute_names.h"
+#include "include/istio/utils/attributes_builder.h"
+
+namespace istio {
+namespace utils {
+
+// create Local attributes object and return a pointer to it.
+// Should be freed by the caller.
+std::unique_ptr<const LocalAttributes> CreateLocalAttributes(
+    const LocalNode &local) {
```
qiwzhang: How about this

void CreateLocalAttributes(const LocalNode &node,   LocalAttributes* attr) {

then you can add attributes directly to the final object, not need to copy them.

src/envoy/utils/mixer_control.cc
```diff
@@ -17,9 +17,18 @@
...
 
 using ::istio::mixerclient::Statistics;
+using ::istio::utils::AttributeName;
+using ::istio::utils::LocalAttributes;
+using ::istio::utils::LocalNode;
 
 namespace Envoy {
 namespace Utils {
+
+const char NodeKey::kName[] = "NODE_NAME";
```
kyessenov: Can we just have UID and namespace? We don't know fully qualified UID yet, not until multi-cluster support lands in Mixer. Until then, we should treat UID as opaque, and not assume it include kubernetes:// and only has two parts.

include/istio/utils/local_attributes.h
mandarjog: LocalNode is a very portable concept. It has the few pieces of information that we need.
It is not evnoy specific.

src/envoy/utils/mixer_control.h
```diff
@@ -42,5 +48,40 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
     const std::string &cluster_name, Upstream::ClusterManager &cm,
     Stats::Scope &scope);
 
+std::unique_ptr<const LocalAttributes> CreateLocalAttributes(
+    const LocalNode &local);
+
+bool Extract(const envoy::api::v2::core::Node &node, LocalNode *args);
+
+inline bool ReadMap(
```
kyessenov: I don't think this is a correct use of metadata in envoy. It assumes that the key is some filter ID, and the list of values is inside the struct value. Take a look at `source/common/config/metadata.cc`.

src/istio/control/http/client_context.cc
```diff
@@ -14,25 +14,57 @@
...
+
+}  // namespace
 
 ClientContext::ClientContext(const Controller::Options& data)
     : ClientContextBase(data.config.transport(), data.env),
       config_(data.config),
-      service_config_cache_size_(data.service_config_cache_size) {}
+      service_config_cache_size_(data.service_config_cache_size),
+      outbound_(isOutbound(data.config)) {
+  local_attributes_ = CreateLocalAttributes(data.local_node);
```
qiwzhang: this local_attributes_ can be just directly embeded object.  not a unique_ptr.

ClientContext {

   LocalAttributes local_attributes_;
}

at Constructor 


  CreateLocalAttributes(local_node, &local_attributes_);
}

src/envoy/utils/mixer_control.h
```diff
@@ -42,5 +48,40 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+  const auto it = meta.find(key);
+  if (it != meta.end()) {
+    *val = it->second.string_value();
+    return true;
+  }
+  return false;
+}
+
+// NodeKey are node matadata keys that are expected to be set.
+struct NodeKey {
```
kyessenov: This should be a proto that is extracted from metadata for mixer filter (filter ID "mixer").

src/istio/control/http/service_context.cc
```diff
@@ -72,6 +74,8 @@ void ServiceContext::InjectForwardedAttributes(
     attributes.MergeFrom(service_config_->forward_attributes());
   }
 
+  client_context_->AddForwardAttributes(&attributes);
```
kyessenov: This should be the first call in the function. We want overriding happen in finer-grained config, not the other way around.

src/istio/control/http/service_context.cc
```diff
@@ -58,6 +58,8 @@ void ServiceContext::AddStaticAttributes(RequestContext *request) const {
   if (service_config_ && service_config_->has_mixer_attributes()) {
     request->attributes.MergeFrom(service_config_->mixer_attributes());
   }
+
+  client_context_->AddRequestAttributes(&request->attributes);
```
kyessenov: Same here: first node-level config, then filter level config, then route level config.

kyessenov:
@mandarjog There is not much testing in this repo. Have you been able to run istio integration tests against the build of envoy? There are a bunch of tests in `mixer/test/client` that will detect any regressions.

cc @PiotrSikora @lizan for review since I'll be OOO from tomorrow.

src/envoy/utils/mixer_control.h
```diff
@@ -42,5 +48,40 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
     const std::string &cluster_name, Upstream::ClusterManager &cm,
     Stats::Scope &scope);
 
+std::unique_ptr<const LocalAttributes> CreateLocalAttributes(
```
qiwzhang: Do you expect these functions to be used by multiple places? if no, just one place, please just move code to that place.

src/envoy/utils/mixer_control.h
```diff
@@ -42,5 +48,40 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+    const LocalNode &local);
+
+bool Extract(const envoy::api::v2::core::Node &node, LocalNode *args);
```
qiwzhang: ExtractNodeInfo?

src/envoy/utils/mixer_control.h
mandarjog: node metadata is actually `google.protobuf.struct` and it does not assume anything about a filter. This is too high level.

qiwzhang: ReadProtoMap

mandarjog: are you suggesting that this should be nested inside Mixer ?

src/envoy/utils/mixer_control.h
```diff
@@ -42,5 +48,40 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+    const std::string &key, std::string *val) {
+  const auto it = meta.find(key);
+  if (it != meta.end()) {
+    *val = it->second.string_value();
+    return true;
+  }
+  return false;
+}
+
+inline bool ReadMap(
```
qiwzhang: ReadAttributeMap

src/envoy/utils/mixer_control.h
qiwzhang: I suspect that these names are not shared by many places. If so, just defined inside .cc file and inside anonymous namespace

src/envoy/utils/mixer_control.cc
```diff
@@ -100,5 +109,70 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+  args->ns = ns;
+  args->uid = "kubernetes://" + longname;
+
+  return true;
+}
+
+// ExtractInfo depends on NODE_NAME, NODE_NAMESPACE, NODE_IP and optional
+// NODE_REG If work cannot be done, returns false.
+bool ExtractInfo(const envoy::api::v2::core::Node &node, LocalNode *args) {
+  const auto meta = node.metadata().fields();
```
qiwzhang: check 

if node.has_metadata() first

or 

if node.metadata().fields().empty()

src/envoy/utils/mixer_control.cc
```diff
@@ -100,5 +109,70 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+  std::string name;
+  if (!ReadMap(meta, NodeKey::kName, &name)) {
+    GOOGLE_LOG(ERROR) << "ExtractInfo  metadata missing " << NodeKey::kName
+                      << " " << node.metadata().DebugString();
+    return false;
+  }
+  std::string ns;
+  ReadMap(meta, NodeKey::kNamespace, &ns);
```
qiwzhang: check return value

src/envoy/utils/mixer_control_test.cc
```diff
@@ -0,0 +1,123 @@
...
+  EXPECT_OK(status) << status;
+  std::string val;
+
+  LocalNode largs;
+  largs.ip = "10.36.0.15";
+  largs.uid = "kubernetes://fortioclient-84469dc8d7-jbbxt.service-graph";
+  largs.ns = "service-graph";
+
+  auto la = GenerateLocalAttributes(node);
+  EXPECT_NE(la, nullptr);
```
qiwzhang: It doesn't make sense for expected to also call GenerateLocalAttributes().  You are not testing that function call.
My suggestion is: here just compare LoalNode result. 

Have a separate test to test GenerateLocalAttributes() call.

test/integration/istio_http_integration_test.cc
mandarjog: Yes. direct mixer attributes are now empty, and the integration test relies on attributes from bootstrap to work correctly.

src/envoy/utils/mixer_control.h
kyessenov: OK, yes, I assumed node proto is filter-aware. This is fine.

src/envoy/utils/mixer_control.cc
```diff
@@ -115,12 +113,12 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
 bool ExtractInfoCompat(const std::string &nodeid, LocalNode *args) {
   auto parts = StringUtil::splitToken(nodeid, "~");
   if (parts.size() < 3) {
-    GOOGLE_LOG(ERROR) << "ExtractInfoCompat error len(nodeid.split(~))<3: "
-                      << nodeid;
+    GOOGLE_LOG(ERROR)
```
qiwzhang: Don't use GOOGLE_LOG,  use Envoy log, take a look at this code:
https://github.com/istio/proxy/blob/master/src/envoy/utils/config.cc#L54

src/envoy/utils/mixer_control_test.cc
```diff
@@ -14,110 +14,113 @@
...
+     "cluster": "fortioclient",
+     %s
+     "build_version": "0/1.8.0-dev//RELEASE"
+    })",
+                      uid, meta);
+}
 
-  auto la = GenerateLocalAttributes(node);
-  EXPECT_NE(la, nullptr);
+void initTestLocalNode(LocalNode *lexp) {
```
qiwzhang: How about

LocalNode getExpectedLocalNode();

then

ASSERT_LOCAL_NODE(arg,  getExpectedLocalNode());

src/istio/control/http/service_context.cc
```diff
@@ -52,30 +52,30 @@ void ServiceContext::BuildParsers() {
 
 // Add static mixer attributes.
 void ServiceContext::AddStaticAttributes(RequestContext *request) const {
+  client_context_->AddRequestAttributes(&request->attributes);
```
qiwzhang: AddLocalNodeAttributes()?

src/istio/control/http/service_context.cc
```diff
@@ -52,30 +52,30 @@ void ServiceContext::BuildParsers() {
...
-
-  client_context_->AddRequestAttributes(&request->attributes);
 }
 
 // Inject a header that contains the static forwarded attributes.
 void ServiceContext::InjectForwardedAttributes(
     HeaderUpdate *header_update) const {
   Attributes attributes;
 
+  client_context_->AddForwardAttributes(&attributes);
```
qiwzhang: AddLocalNodeForwardAttribues()

include/istio/utils/local_attributes.h
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef ISTIO_UTILS_LOCAL_ATTRIBUTES_H
+#define ISTIO_UTILS_LOCAL_ATTRIBUTES_H
+
+#include "mixer/v1/attributes.pb.h"
+
+using ::istio::mixer::v1::Attributes;
```
qiwzhang: Per style guide,  could not use namespace "using" in the header file.  only in the cc file.

src/envoy/utils/mixer_control.h
```diff
@@ -42,5 +48,30 @@ Grpc::AsyncClientFactoryPtr GrpcClientFactoryForCluster(
...
+  const auto it = meta.find(key);
+  if (it != meta.end()) {
+    *val = it->second.string_value();
+    return true;
+  }
+
+  return false;
+}
+
+inline bool ReadAttributeMap(
```
qiwzhang: Are these two "read" functions used in other files?   If no,  prefer to not in the header file.  It is polluting .

src/envoy/utils/mixer_control_test.cc
```diff
@@ -0,0 +1,126 @@
...
+
+#define assertEqualLocalAttributes(laExpect, la)                               \
+  {                                                                            \
+    EXPECT_EQ((laExpect)->outbound.DebugString(),                              \
+              (la)->outbound.DebugString());                                   \
+    EXPECT_EQ((laExpect)->inbound.DebugString(), (la)->inbound.DebugString()); \
+    EXPECT_EQ((laExpect)->forward.DebugString(), (la)->forward.DebugString()); \
+  };
+
+#define ASSERT_LOCAL_NODE(lexp, la)  \
```
qiwzhang: these two macros can be inside anonymous name sapce too

src/envoy/utils/mixer_control.cc
```diff
@@ -55,6 +62,18 @@ class EnvoyGrpcAsyncClientFactory : public Grpc::AsyncClientFactory {
 
 }  // namespace
 
+inline bool ReadProtoMap(
```
qiwzhang: move inside anonymous  space

PR #1962, opened 2018-09-08
Fix a bug in origin authenticator that wrongly treats empty origin methods as pass

yangminzhu:
Fix a bug in origin authenticator that wrongly treats empty origin methods as pass

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: n/a

**Special notes for your reviewer**: n/a

### DIFF
--- a/src/envoy/http/authn/origin_authenticator.cc
+++ b/src/envoy/http/authn/origin_authenticator.cc
@@ -92,7 +92,7 @@ bool OriginAuthenticator::run(Payload* payload) {
 
   // If none of the JWT triggered, origin authentication will be ignored, as if
   // it is not defined.
-  return !triggered || success;
+  return (policy_.origins_size() > 0 && !triggered) || success;
 }
 
 }  // namespace AuthN
--- a/src/envoy/http/authn/origin_authenticator_test.cc
+++ b/src/envoy/http/authn/origin_authenticator_test.cc
@@ -41,6 +41,14 @@ namespace Istio {
 namespace AuthN {
 namespace {
 
+const char kZeroOriginMethodPolicyBindPeer[] = R"(
+  principal_binding: USE_PEER
+)";
+
+const char kZeroOriginMethodPolicyBindOrigin[] = R"(
+  principal_binding: USE_ORIGIN
+)";
+
 const char kSingleOriginMethodPolicy[] = R"(
   principal_binding: USE_ORIGIN
   origins {
@@ -206,6 +214,34 @@ TEST_P(OriginAuthenticatorTest, Empty) {
     initial_result_.set_principal("bar");
   }
   EXPECT_TRUE(TestUtility::protoEqual(initial_result_,
+                                      filter_context_.authenticationResult()));
+}
+
+TEST_P(OriginAuthenticatorTest, ZeroMethodFail) {
+  ASSERT_TRUE(Protobuf::TextFormat::ParseFromString(
+      kZeroOriginMethodPolicyBindOrigin, &policy_));
+  createAuthenticator();
+  EXPECT_FALSE(authenticator_->run(payload_));
+}
+
+TEST_P(OriginAuthenticatorTest, ZeroMethodPass) {
+  ASSERT_TRUE(Protobuf::TextFormat::ParseFromString(
+      kZeroOriginMethodPolicyBindPeer, &policy_));
+  createAuthenticator();
+
+  Result expected_result = TestUtilities::AuthNResultFromString(R"(
+      origin {
+        user: "bar"
+        presenter: "istio.io"
+      }
+    )");
+  if (set_peer_) {
+    expected_result.set_principal("bar");
+    expected_result.set_peer_user("bar");
+  }
+
+  EXPECT_TRUE(authenticator_->run(&jwt_extra_payload_));
+  EXPECT_TRUE(TestUtility::protoEqual(expected_result,
                                       filter_context_.authenticationResult()));
 }
 


src/envoy/http/authn/origin_authenticator_test.cc
```diff
@@ -209,6 +217,34 @@ TEST_P(OriginAuthenticatorTest, Empty) {
                                       filter_context_.authenticationResult()));
 }
```
JimmyCYJ: Please add comment describing each test.

src/envoy/http/authn/origin_authenticator.cc
```diff
@@ -92,7 +92,7 @@ bool OriginAuthenticator::run(Payload* payload) {
 
   // If none of the JWT triggered, origin authentication will be ignored, as if
   // it is not defined.
```
JimmyCYJ: Please update this comment.

src/envoy/http/authn/origin_authenticator.cc
```diff
@@ -92,7 +92,7 @@ bool OriginAuthenticator::run(Payload* payload) {
...
-  return !triggered || success;
+  return (policy_.origins_size() > 0 && !triggered) || success;
```
diemtvu: I think we can simplify the switch into 
```
if (policy_.origins_size() == 0 && policy_.principal_binding()==iaapi::PrincipalBinding::USE_ORIGIN) {
  // This should be rejected by validation during policy submitting. Log warning in case validation was
  // was not run properly.
  ENVOY_LOG(warn,
                   "Principal is binded to origin, but no method specified in "	                   "Principal is binded to origin, but no method specified in "
                   "policy {}",	                   "policy {}",
                   policy_.DebugString());	                   policy_.DebugString());
  return false;
}
```

And this return value is

```
// Always returns successful (true) if binding is USE_PEER, else 
// returns true if no jwt was triggered, or triggered and success.
policy_.principal_binding() == iaapi::PrincipalBinding::USE_PEER || !triggered || success;
```
(you may also move and rename the `success` variable to `triggered_success` to have clearer semantic)

yangminzhu: It looks like a change that have different behaviors, is it okay?
When binding is `USE_PEER` but origins not empty, before it may return false if origins validate fail. After (now) it always return true.

diemtvu: Ah, I miss read the code :). It's then even simpler, `return !triggered || success` should be equivalent, right?

yangminzhu: There is still a difference: 
When origins is empty and binding is `USE_PEER`, the success will be set to true in turn the setOriginResult() and setPrintipal() will be called.

How about just change the switch case to `return false` early and include the following logic in an else statement? (see my latest update)

src/envoy/http/authn/origin_authenticator_test.cc
yangminzhu: Done

src/envoy/http/authn/origin_authenticator.cc
yangminzhu: Updated the code, this doesn't need change anymore

src/envoy/http/authn/origin_authenticator.cc
diemtvu: My comment including change of the switch statement to `if` and return early. See the first comment for snippet.

yangminzhu: Just to confirm one last thing, when binding is `USE_ORIGIN`:
If origins is empty and , we return false
If origins is non empty but none of them triggered, we return true
This is the desired behavior, right?

diemtvu: Yes, given non triggered == non matching given path.

yangminzhu: Great! Could you approve if no more changes needed? Thanks.

src/envoy/http/authn/origin_authenticator.cc
```diff
@@ -31,68 +31,56 @@ OriginAuthenticator::OriginAuthenticator(FilterContext* filter_context,
     : AuthenticatorBase(filter_context), policy_(policy) {}
 
 bool OriginAuthenticator::run(Payload* payload) {
-  bool success = false;
+  bool triggered_success = false;
+  bool triggered = false;
 
   if (policy_.origins_size() == 0) {
```
diemtvu: This can be one if.

src/envoy/http/authn/origin_authenticator.cc
```diff
@@ -31,68 +31,56 @@ OriginAuthenticator::OriginAuthenticator(FilterContext* filter_context,
...
+      return false;
     }
-  }
-
-  bool triggered = false;
-  const char* request_path = nullptr;
-  if (filter_context()->headerMap().Path() != nullptr) {
-    request_path = filter_context()->headerMap().Path()->value().c_str();
-    ENVOY_LOG(debug, "Got request path {}", request_path);
   } else {
```
diemtvu: Don't need else.

yangminzhu: Ditto.

src/envoy/http/authn/origin_authenticator.cc
yangminzhu: I want to avoid getting the path when origins size is 0.

diemtvu: It doesn't make any different, does it? If the list is empty, the loop won't do anything thing, hence two booleans won't be changed.

yangminzhu: Right. I mean I want to avoid the line 51-58 that gets the path from the header. That part is not needed when the loop won't do anything.

yangminzhu: Removed the else part.

src/envoy/http/authn/origin_authenticator.cc
```diff
@@ -31,34 +31,21 @@ OriginAuthenticator::OriginAuthenticator(FilterContext* filter_context,
     : AuthenticatorBase(filter_context), policy_(policy) {}
 
 bool OriginAuthenticator::run(Payload* payload) {
-  bool success = false;
+  bool triggered_success = false;
```
diemtvu: [nit] Move these 2 lines after the first if to keep it closed to where it's used.

yangminzhu:
@JimmyCYJ  Could you approve if no more changes needed? Thanks.

src/envoy/http/authn/origin_authenticator.cc
```diff
@@ -31,34 +31,19 @@ OriginAuthenticator::OriginAuthenticator(FilterContext* filter_context,
...
-        ENVOY_LOG(warn,
-                  "Principal is binded to origin, but no method specified in "
-                  "policy {}",
-                  policy_.DebugString());
-        break;
-      case iaapi::PrincipalBinding::USE_PEER:
-        // On the other hand, it's ok to have no (origin) methods if
-        // rule USE_SOURCE
-        success = true;
-        break;
```
PiotrSikora: Previously, this code path would result in `filter_context()->setPrincipal(iaapi::PrincipalBinding::USE_PEER)` call, whereas after your changes the principal is never set in case of `iaapi::PrincipalBinding::USE_PEER`.

To be honest, I'm not sure if the previous code was correct, but there is a change here that I'm not sure if it's intentional.

yangminzhu: You mean when binding is `USE_PEER` and `policy_.origins_size() == 0`? In this case, we still call `setPrincipal()`, note I also changed line 76 to also check the `!triggered`

PiotrSikora: Oops, you're right. Thanks!

PR #1967, opened 2018-09-11
Mixer Client: Add support for TCP local attributes

mandarjog:
This PR is a follow on to the HTTP PR for local attributes.

1. Refactor: Move `AddLocalNodeAttributes` and `AddLocalNodeForwardAttribues` into `ClientContextBase`.
2. Use it from `http/clientContext` and `tcp/clientContext`
3. Move other supporting functions `outbound` computation to base class.


src/istio/control/client_context_base.h
```diff
@@ -17,24 +17,37 @@
...
 #include "include/istio/mixerclient/client.h"
+#include "include/istio/utils/attribute_names.h"
+#include "include/istio/utils/local_attributes.h"
 #include "mixer/v1/config/client/client_config.pb.h"
 #include "request_context.h"
 
 namespace istio {
 namespace control {
 
+bool isOutbound(const google::protobuf::Map<
```
qiwzhang: I suggest
1) move this isOutbound function to istio/util so it can be called by http and tcp
2) It takes "const ::istio::mixer::v1::Attributes&",  not its internal attributes map
3) ClientContextBase, take a "bool  is_outbound" in its constructor.   Or add "bool is_outbound" into LocalAttributes class since it is the one using it.

mandarjog: done.

src/istio/control/client_context_base.cc
```diff
@@ -69,11 +71,15 @@ ReportOptions GetReportOptions(const TransportConfig& config) {
 }  // namespace
 
 ClientContextBase::ClientContextBase(const TransportConfig& config,
-                                     const Environment& env) {
+                                     const Environment& env,
+                                     const bool& outbound,
```
qiwzhang: bool as a primary type, not need to do "const &"

src/istio/control/http/client_context.cc
```diff
@@ -19,51 +19,27 @@
 using ::istio::mixer::v1::Attributes_AttributeValue;
 using ::istio::mixer::v1::config::client::ServiceConfig;
 using ::istio::utils::AttributeName;
-using ::istio::utils::CreateLocalAttributes;
+using ::istio::utils::IsOutbound;
```
qiwzhang: Usually, we don't alias functions.  only classes

src/istio/utils/local_attributes.cc
```diff
@@ -49,5 +49,22 @@ bool SerializeForwardedAttributes(const LocalNode& local,
   return attributes.SerializeToString(serialized_forward_attributes);
 }
 
+static const char kReporterOutbound[] = "outbound";
```
qiwzhang: prefer not to use "static", instead put it inside anonymous namespace.

PR #1970, opened 2018-09-13
Remove delta update in Report batch

qiwzhang:

DeltaUpdate is comparing each attribute with the last Report in the Batch.  It is very slow, use a lot of cpu.  It compressed some bytes, but bandwidth is not the issue for Istio, cpu usage cause latency.
Hence remove the delta update.

Also reduce the batch size since the byte could not be compressed much.  Sending a huge batch report which may cause some requests to get delayed more (if these requests are in the queue). It will cause 99 percentitle latency to increase.

**Release note**:
```release-note
None
```


mandarjog:
@qiwzhang can you put before and after latency numbers here?

Do you want to keep it as an option until we are done with the performance work?

qiwzhang:
Hmm,  the test result shows:  for fortio-c=10, envoy-concurrency=1, report-only.

1) 50%tile: is faster from 2.2ms to 1.7ms
2) 99%tile: is slower from 5.4ms to 15.6ms if batch_numbers only reduced to 500 from 1000.  if reduces to 100:  from 5.4ms to 4.7ms

If max_number can be reduced to 100.  I think we can keep this change.   We like to have .5ms latency decrease for average request.  99% tile is good too.

I think qps to maxer will NOT be increased by 10 times.  Most report batches are flushed out by 1 second timer instead of max_numbers.

PR #1981, opened 2018-09-21
Not to add check attributes if check is disabled

qiwzhang:

Now Check attributes are extracted even check is disabled.  This is done in the request data path
It is better to move the operation to Report call which is outside if data-path. 

Also fixed:  if request is rejected before reaching mixer filter,   forward attributes are not extracted.

**Release note**:
```release-note
None
```

### DIFF
--- a/src/istio/control/http/request_handler_impl.cc
+++ b/src/istio/control/http/request_handler_impl.cc
@@ -29,26 +29,47 @@ namespace http {
 
 RequestHandlerImpl::RequestHandlerImpl(
     std::shared_ptr<ServiceContext> service_context)
-    : service_context_(service_context) {}
+    : service_context_(service_context),
+      check_attributes_added_(false),
+      forward_attributes_added_(false) {}
 
-void RequestHandlerImpl::ExtractRequestAttributes(CheckData* check_data) {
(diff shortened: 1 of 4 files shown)


src/istio/control/http/request_handler_impl.cc
```diff
@@ -29,26 +29,47 @@ namespace http {
...
+  AddForwardAttributes(check_data);
+  AddCheckAttributes(check_data);
+}
+
 CancelFunc RequestHandlerImpl::Check(CheckData* check_data,
                                      HeaderUpdate* header_update,
                                      TransportCheckFunc transport,
                                      CheckDoneFunc on_done) {
-  ExtractRequestAttributes(check_data);
+  AddForwardAttributes(check_data);
```
JimmyCYJ: Please add a comment explaining why we need to extract forward attributes here.

src/istio/control/http/request_handler_impl.h
```diff
@@ -42,11 +42,21 @@ class RequestHandlerImpl : public RequestHandler {
   void ExtractRequestAttributes(CheckData* check_data) override;
 
  private:
+  // Add Forward attributes, allow re-entry
```
JimmyCYJ: Do we need to state here and below that re-entry is a no op?

PR #1984, opened 2018-09-27
Attempt to fix bazel Server Crash

sebastienvas:
Ref https://github.com/bazelbuild/bazel/issues/4311

Using batch mode


rshriram:
still crashing :(

PR #1988, opened 2018-10-08
Fix memory leak at report batching

qiwzhang:

In report batching code.   it is using arena allocation for report button.  The arena was not freed for each sending report.

https://github.com/istio/istio/issues/9159

**Release note**:
```release-note
None
```


JimmyCYJ:
Thanks for the quick fix! @qiwzhang

PR #1993, opened 2018-10-15
Update dependencies and fix compile errors.

JimmyCYJ:
Signed-off-by: JimmyCYJ <jimmychen.0102@gmail.com>

**What this PR does / why we need it**:
Update Envoy SHA to 8607e912a1df840da1080b7b0d4a9ed6bc247c39 and fix dependencies as well as compile issues.

```release-note
NONE
```

PR #1995, opened 2018-10-16
Update HUB to gcr.io/istio

JimmyCYJ:
Signed-off-by: JimmyCYJ <jimmychen.0102@gmail.com>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: istio/ci:go1.10-bazel0.15-clang6.0
+      - image: gcr.io/istio/ci:go1.10-bazel0.15-clang6.0
     environment:
       - BAZEL_TEST_ARGS: "--test_env=ENVOY_IP_TEST_VERSIONS=v4only --test_output=all"
     resource_class: xlarge


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: istio/ci:go1.10-bazel0.15-clang6.0
+      - image: gcr.io/istio/ci:go1.10-bazel0.15-clang6.0
```
rkpagadala: should be
gcr.io/istio-testing/ci:go1.10-bazel0.15-clang6.0

qiwzhang: @rshriram It seems that we could not locate any Google people with the access.  Who owns the docker hub with admin permission to update the access list?  We like to add some Google test-infra people to the list too. I like to have access too as istio/proxy maintainer.

JimmyCYJ: @rshriram I was trying to update Envoy SHA, and found some build errors in circleCI [PR](https://github.com/istio/proxy/pull/1993), so I ask around and learn that I need to update circleCI image [PR](https://github.com/istio/proxy/pull/1994). We don't have write permission to the docker hub, could you add @qiwzhang, me and test-infra people to the access list? Is there any instruction to build images for circleCI?

JimmyCYJ:
Thanks a lot @mandarjog, for helping us to build and push the images.

PR #2002, opened 2018-10-26
Improve performance by removing MD5 for check cache keys

qiwzhang:

MD5 calculation uses a lot of CPU.   Just use concatenated string directly as cache key will result in bigger memory usage for cache.  But it will save CPU, and reduce latency. 

It is a trade-off between memory and cpu.  It is OK to use more memory in Istio/proxy.

**Release note**:
```release-note
None
```


include/istio/utils/concat_hash.h
```diff
@@ -0,0 +1,77 @@
...
+namespace istio {
+namespace utils {
+
+// This class concatenates multiple values into a string as hash
+class ConcatHash {
+ public:
+  ConcatHash(size_t reserve_size) { hash_.reserve(reserve_size); }
+
+  // Updates the context with data.
+  ConcatHash& Update(const void* data, size_t size) {
```
lizan: I think you can consolidate 3 of Update with string_view.

PR #2013, opened 2018-10-30
Send dynamic metadata as part of report calls

rshriram:
Envoy filters can add arbitrary attributes to dynamicMetadata. Its useful to log this as part of reports (envoy's gRPC access logs and normal access logs already print this info).

I am send the metadata as a dumb string (serialized form of envoy.api.v2.core.Metadata). Right now, there is no support in mixer for this, but an adapter wishing to utilize this can deserialize this info and log this info, thus providing the ability for third party codecs to log additional attributes as part of a connection.


rshriram:
temporary fix for https://github.com/istio/istio/issues/9613 . At the moment, this should unblock the internal use case I have to stream custom attributes generated by internal codecs to mixer without having to fork Istio mixer filter code. Would be nice to have a cleaner solution.

qiwzhang:
Only for tcp?   Just curious, why not Http?

rshriram:
Oh I can do the same for HTTP as well.

rshriram:
@qiwzhang updated to add dynamic metadata to http as well.

PR #2014, opened 2018-11-01
Use string map instead of opaque struct for dynamic metadata

rshriram:
Based on discussions here: https://github.com/istio/istio/issues/9613
The code will send all map entries, as attribute_name: string_map instead of
adding a new attribute.


include/istio/utils/attributes_builder.h
```diff
@@ -67,38 +67,38 @@ class AttributesBuilder {
...
+                               const google::protobuf::Struct &struct_map) {
     if (struct_map.fields().empty()) {
       return;
     }
     auto entries = (*attributes_->mutable_attributes())[key]
                        .mutable_string_map_value()
                        ->mutable_entries();
     entries->clear();
-    for (const auto& field : struct_map.fields()) {
+    for (const auto &field : struct_map.fields()) {
```
qiwzhang: Please add:  
if entries is still empty after the loop,  remove the key.  

I like to have a option for filters not to dump data to mixer.  Or another way,  if filters want to dump data to mixer, they have to say so.   We should not blindly dump everything to mixer.  I have concern about performance, and security

rshriram: This is only in the report path. Dynamic metadata is strictly for logging.
See https://github.com/envoyproxy/envoy/commit/700b7d09f400ada6fa5b2202d2588441b6f7cb66
In other words, even in Envoy, all dynamic metadata is already dumped on the wire, as is, just like static metadata. If filters store sensitive information, then there are bigger concerns including access logs, header values, etc. which are also being sent to mixer. So, lets have a separate discussion for that.

geeknoid: I agree with Shriram on the security front, I'm not concerned.

I do wonder about the end-to-end perf impact. Can we measure the cost on proxy and Mixer CPU usage?

rshriram: No matter what we do, there is no way to tell other external filters that they need to log things for Istio mixer. Second, dynamic metadata is not 1000s of kbs of data. Its very very small bits of info like name of resource, or some key claim type, etc. This is the *only* way the mixer has, if it wishes to be able to log info about other OSS codecs like mongo/redis/kafka in Envoy. Otherwise, we will still be stuck with opaque TCP attributes like bytes sent, received, etc., while people prefer to turn off mixer and use envoy gRPC access logging to get richer metrics. :)

rshriram:
@qiwzhang any other comments?

geeknoid:
How many attributes are we talking about? How big are the values for these? Do the values change at a high rate?

FYI @mandarjog since he has been doing a lot of perf work up and down this path

rshriram:
> How many attributes are we talking about? 
One per envoy codec. So, its utmost 1 extra attribute per connection, if that connection has a special codec like mongo/kafka/etc.
> How big are the values for these?
 Its just a string map. Very basic things that you log in access log. Think of this as access logs for mongo/kafka/redis/etc. You don't dump the entire protocol payload. Its just select info like topics, producer/consumer ids, etc..
> Do the values change at a high rate?
Answered by above. NO.

rshriram:
And if you want to optimize, you should be equally fair and cut down 30-40% of the attributes we are gathering today for http/tcp, including auth and rbac that are adding more attributes to the system.  What is important is a very relative measure :).

include/istio/utils/attributes_builder.h
rshriram: >logging is for debugging 
???
Access log is most certainly not debugging. In fact, this PR came from Google: https://github.com/envoyproxy/envoy/commit/700b7d09f400ada6fa5b2202d2588441b6f7cb66 and this is adding dynamic metadata to access logs streamed over gRPC. 
Remember, I can always choose to overload the mixer by sending very very large HTTP headers, auth keys, etc. or several 100s of headers. So, the argument about performance is moot. We can potentially agree on some common prefix but that requires discussion with envoy folks as well, things like mongo/redis that generate dynamic metadata need to be modified to use this agreed upon prefix. This will take time. Do I have to be wait until then?

PR #2017, opened 2018-11-03
Add a new TCP cluster rewrite filter

venilnoronha:
This PR adds a new TCP cluster rewrite filter which allows users to rewrite TCP cluster names obtained via TLS SNI by matching via regex configuration.

**Release note**:
```release-note
Added a new TCP cluster rewrite filter
```

/cc @rshriram @lizan

### DIFF
--- a/include/istio/mixerclient/environment.h
+++ b/include/istio/mixerclient/environment.h
@@ -18,7 +18,7 @@
 
 #include "check_response.h"
 #include "google/protobuf/stubs/status.h"
-#include "mixer/v1/service.pb.h"
+#include "mixer/v1/mixer.pb.h"
 #include "timer.h"
 
 namespace istio {
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -113,7 +113,7 @@ cc_library(
             actual = "@googletest_git//:googletest_prod",
         )
 
-ISTIO_API = "62c345bd6d6e4c2047dd2dee128b7413231be7b4"
+ISTIO_API = "6b9e3a501e6ef254958bf82f7b74c37d64a57a15"
 
 def mixerapi_repositories(bind=True):
     BUILD = """
@@ -192,6 +192,19 @@ cc_proto_library(
     ],
 )
 
+cc_proto_library(
+    name = "tcp_cluster_rewrite_config_cc_proto",
+    srcs = glob(
+        ["envoy/config/filter/network/tcp_cluster_rewrite/v2alpha1/*.proto", ],
+    ),
+    default_runtime = "//external:protobuf",
+    protoc = "//external:protoc",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//external:cc_gogoproto",
+    ],
+)
+
 filegroup(
     name = "global_dictionary_file",
     srcs = ["mixer/v1/global_dictionary.yaml"],
@@ -221,6 +234,10 @@ filegroup(
         native.bind(
             name = "jwt_auth_config_cc_proto",
             actual = "@mixerapi_git//:jwt_auth_config_cc_proto",
+        )
+        native.bind(
+            name = "tcp_cluster_rewrite_config_cc_proto",
+            actual = "@mixerapi_git//:tcp_cluster_rewrite_config_cc_proto",
         )
 
 load(":protobuf.bzl", "protobuf_repositories")
--- a/src/envoy/BUILD
+++ b/src/envoy/BUILD
@@ -29,6 +29,7 @@ envoy_cc_binary(
         "//src/envoy/http/jwt_auth:http_filter_factory",
         "//src/envoy/http/mixer:filter_lib",
         "//src/envoy/tcp/mixer:filter_lib",
+        "//src/envoy/tcp/tcp_cluster_rewrite:tcp_cluster_rewrite_lib",
         "@envoy//source/exe:envoy_main_entry_lib",
     ],
 )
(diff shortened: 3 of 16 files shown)


include/istio/mixerclient/environment.h
```diff
@@ -18,7 +18,7 @@
 
 #include "check_response.h"
 #include "google/protobuf/stubs/status.h"
-#include "mixer/v1/service.pb.h"
+#include "mixer/v1/mixer.pb.h"
```
rshriram: ???

repositories.bzl
```diff
@@ -192,6 +192,19 @@ cc_proto_library(
...
+    srcs = glob(
+        ["envoy/config/filter/network/tcp_cluster_rewrite/v2alpha1/*.proto", ],
+    ),
+    default_runtime = "//external:protobuf",
+    protoc = "//external:protoc",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//external:cc_gogoproto",
+    ],
+)
```
rshriram: Do we need this voodoo? How does the auth filter do it’s stuff ?

src/envoy/tcp/tcp_cluster_rewrite/config.cc
```diff
@@ -0,0 +1,73 @@
...
+namespace Tcp {
+namespace TcpClusterRewrite {
+
+Network::FilterFactoryCb
+TcpClusterRewriteFilterConfigFactory::createFilterFactory(
+    const Json::Object& config_json, Server::Configuration::FactoryContext&) {
+  v2alpha1::TcpClusterRewrite config_pb;
+  if (!Utils::ReadV2Config(config_json, &config_pb) &&
+      !Utils::ReadV1Config(config_json, &config_pb)) {
+    throw EnvoyException("Failed to parse JSON config");
```
rshriram: I think the v1 is not needed. You can simply throw exception from this function.

src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
```diff
@@ -0,0 +1,69 @@
...
+    }
+    ENVOY_CONN_LOG(trace, "tcp_cluster_rewrite: tcp proxy cluster name {}",
+                   read_callbacks_->connection(), cluster_name);
+
+    // Set the tcp_proxy cluster to the same value as the (rewritten) SNI. The
+    // data is mutable to allow other filters to change it.
+    read_callbacks_->connection().streamInfo().filterState().setData(
+        TcpProxy::PerConnectionCluster::Key,
+        std::make_unique<TcpProxy::PerConnectionCluster>(cluster_name),
+        StreamInfo::FilterState::StateType::Mutable);
```
rshriram: Kill all this. All we need is to rewrite the per connection cluster dynamic state, if set. Else it’s no op. The idea is to stack the SNI_cluster filter and this rewrite filter.

repositories.bzl
venilnoronha: This is similar to the auth and other filters.

src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
venilnoronha: Addressed in d633185.

rshriram:
@qiwzhang / @lizan any other comments?

src/envoy/tcp/tcp_cluster_rewrite/config.h
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Envoy {
+namespace Tcp {
+namespace TcpClusterRewrite {
+
+/**
+ * Config registration for the TCP cluster rewrite filter. @see
+ * NamedNetworkFilterConfigFactory.
+ */
+class TcpClusterRewriteFilterConfigFactory
```
lizan: use factory base

src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
```diff
@@ -0,0 +1,78 @@
...
+            .filterState()
+            .getDataReadOnly<TcpProxy::PerConnectionCluster>(
+                TcpProxy::PerConnectionCluster::Key)
+            .value();
+    ENVOY_CONN_LOG(trace,
+                   "tcp_cluster_rewrite: new connection with server name {}",
+                   read_callbacks_->connection(), cluster_name);
+
+    // Rewrite the cluster name prior to setting the tcp_proxy cluster name.
+    std::string final_cluster_name(absl::StrCat(cluster_name));
```
lizan: nit: you should be able to construct final_cluster_name without copying them with iterators, but perhaps fine for now.

src/envoy/tcp/tcp_cluster_rewrite/config.h
venilnoronha: That causes a compile error because we don't have `*.pb.validate.h` files here.

https://github.com/envoyproxy/envoy/blob/3e15c94901f7ce5ebca671e451efceec8053da86/source/extensions/filters/network/common/factory_base.h#L29

src/envoy/BUILD
```diff
@@ -29,6 +29,7 @@ envoy_cc_binary(
         "//src/envoy/http/jwt_auth:http_filter_factory",
         "//src/envoy/http/mixer:filter_lib",
         "//src/envoy/tcp/mixer:filter_lib",
+        "//src/envoy/tcp/tcp_cluster_rewrite:tcp_cluster_rewrite_lib",
```
lizan: config_lib? tcp_cluster_rewrite_lib doesn't have the static registration.

venilnoronha: Addressed in #2021.

PR #2026, opened 2018-11-07
Update Envoy SHA to 74de08a0

venilnoronha:
This PR updates the Envoy SHA to [74de08a0](https://github.com/envoyproxy/envoy/commit/74de08a0d4d31bd466639d25d681df5d290bb770) to bring in the new TCP RBAC filter to Istio.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -30,7 +30,7 @@ bind(
 )
 
 # When updating envoy sha manually please update the sha in istio.deps file also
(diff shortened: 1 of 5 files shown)


JimmyCYJ:
Once this update is checked in, could you update https://github.com/istio/istio/blob/master/istio.deps as well?

venilnoronha:
@JimmyCYJ will do.

PR #2040, opened 2018-11-20
Enable caching of dynamic metadata in mixer filter

venilnoronha:
This enables caching of dynamic metadata in the `onData` call in the `mixer` filter and returns the same during the `GetDynamicFilterState` call.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/src/envoy/tcp/mixer/filter.cc
+++ b/src/envoy/tcp/mixer/filter.cc
@@ -15,6 +15,7 @@
 
 #include "src/envoy/tcp/mixer/filter.h"
 #include "common/common/enum_to_int.h"
+#include "extensions/filters/network/well_known_names.h"
 #include "src/envoy/utils/utils.h"
 
 using ::google::protobuf::util::Status;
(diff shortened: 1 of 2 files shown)


src/envoy/tcp/mixer/filter.cc
```diff
@@ -78,6 +78,11 @@ Network::FilterStatus Filter::onData(Buffer::Instance &data, bool) {
                  filter_callbacks_->connection(), data.length());
   received_bytes_ += data.length();
 
+  cached_dynamic_metadata_ = filter_callbacks_->connection()
```
mandarjog: Can you explain why it is better to get this here, rather than the way it was done before?

rshriram: Envoy filters like mongo and mysql will clear the dynamic metadata everytime they start a new decoding cycle (from the onData() callback). They generate metadata for every query they see. But mixer sends reports periodically. So by the time it sends the report, the data in streamInfo.DynamicMetadata might get overwritten multiple times, to the point that we only have info about the last query and not every query since the last report. That said, this implementation above is wrong as it too is overwriting it..

venilnoronha: Addressed in 08fd80ef87e35e9ad52466d8ef2f7b97941d9eb1.

src/envoy/tcp/mixer/filter.cc
```diff
@@ -170,20 +208,20 @@ bool Filter::GetDestinationIpPort(std::string *str_ip, int *port) const {
...
   return false;
 }
+
 const ::google::protobuf::Map<std::string, ::google::protobuf::Struct>
     &Filter::GetDynamicFilterState() const {
-  return filter_callbacks_->connection()
-      .streamInfo()
-      .dynamicMetadata()
-      .filter_metadata();
+  return cached_filter_metadata_;
```
qiwzhang: It seems that we only cache MongoProxy metadata and assume its all field types are "list" types.  This is too specific, not general enough for Istio.   

This is really a hack to solve MongoProxy metadata problem.  Maybe Istio/proxy is not the place to solve this problem.   It should be solved in MongoProxy

If we have to solve this in Istio/proxy,  it should NOT break other users that want to log other metadata.

PR #2046, opened 2018-11-28
[WIP] add sni_verifier filter

vadimeisenbergibm:
This PR solves the security hole described in https://goo.gl/cQ29rX, "Enable egress policies per source workload". The solution is related to the following text:

> Set the outer SNI to be equal to the inner SNI in the sidecar proxy and then add a filter in the gateway’s proxy that will verify that the outer SNI and the inner SNI match.

This PR adds a filter "that will verify that the outer SNI and the inner SNI match".

Based on the TLS inspector of Envoy.

### DIFF
--- a/src/envoy/tcp/sni_verifier/sni_verifier.cc
+++ b/src/envoy/tcp/sni_verifier/sni_verifier.cc
@@ -81,7 +81,7 @@ Network::FilterStatus SniVerifierFilter::onData(Buffer::Instance& data, bool) {
   size_t lenToRead =
       (data.length() < freeSpaceInBuf) ? data.length() : freeSpaceInBuf;
   read_ += lenToRead;
-  parseClientHello(buf_, read_);
+  parseClientHello(*buf_, read_);
 
   return is_match_ ? Network::FilterStatus::Continue
                    : Network::FilterStatus::StopIteration;
--- a/src/envoy/tcp/sni_verifier/sni_verifier.cc
+++ b/src/envoy/tcp/sni_verifier/sni_verifier.cc
@@ -80,9 +80,10 @@ Network::FilterStatus SniVerifierFilter::onData(Buffer::Instance& data, bool) {
   size_t freeSpaceInBuf = config_->maxClientHelloSize() - read_;
   size_t lenToRead =
       (data.length() < freeSpaceInBuf) ? data.length() : freeSpaceInBuf;
+  data.copyOut(0, lenToRead, buf_.get() + read_);
   read_ += lenToRead;
-  parseClientHello(buf_.get(), read_);
 
+  parseClientHello(buf_.get(), read_);
   return is_match_ ? Network::FilterStatus::Continue
                    : Network::FilterStatus::StopIteration;
 }
--- a/src/envoy/tcp/sni_verifier/sni_verifier.cc
+++ b/src/envoy/tcp/sni_verifier/sni_verifier.cc
@@ -83,7 +83,8 @@ Network::FilterStatus Filter::onData(Buffer::Instance& data, bool) {
 
   auto start_handshake_data =
       restart_handshake_ ? buf_.get() : buf_.get() + read_;
-  auto handshake_size = restart_handshake_ ? read_ : data_to_read;
+  auto handshake_size =
+      restart_handshake_ ? read_ + data_to_read : data_to_read;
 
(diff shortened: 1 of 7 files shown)


src/envoy/tcp/sni_verifier/sni_verifier.cc
```diff
@@ -0,0 +1,161 @@
...
+        *out_alert = SSL_AD_USER_CANCELLED;
+        return SSL_TLSEXT_ERR_ALERT_FATAL;
+      });
+}
+
+bssl::UniquePtr<SSL> Config::newSsl() {
+  return bssl::UniquePtr<SSL>{SSL_new(ssl_ctx_.get())};
+}
+
+thread_local uint8_t SniVerifierFilter::buf_[Config::TLS_MAX_CLIENT_HELLO];
```
qiwzhang: Is it possible that two filter instances from two downstream connections are sharing this buffer so their data collapse?

src/envoy/tcp/sni_verifier/sni_verifier.cc
```diff
@@ -0,0 +1,161 @@
...
+
+SniVerifierFilter::SniVerifierFilter(const ConfigSharedPtr config)
+    : config_(config), ssl_(config_->newSsl()) {
+  RELEASE_ASSERT(sizeof(buf_) >= config_->maxClientHelloSize(), "");
+
+  SSL_set_app_data(ssl_.get(), this);
```
qiwzhang: Not sure when and how the callback is called.  If it maybe called outside of SSL_do_handshake() call, then it is dangerous, the filter may go away when it is called.   

If it is always called inside SSL_do_handshake(), then it is better to set the app_data and call_back right before the SSL_do_handshake() and reset it after.

src/envoy/tcp/sni_verifier/sni_verifier.cc
vadimeisenbergibm: @qiwzhang My understanding is that `thread_local` prevents it. This filter is modeled after Envoy's tls inspector, it also uses `thread_local` - https://github.com/envoyproxy/envoy/blob/master/source/extensions/filters/listener/tls_inspector/tls_inspector.cc#L63.

An explanation about `thread_local` on StackOverflow https://stackoverflow.com/a/11984017/553720

src/envoy/tcp/sni_verifier/sni_verifier_test.cc
```diff
@@ -0,0 +1,45 @@
...
+TEST(SniVerifier, ConfigTest) {
+  NiceMock<Server::Configuration::MockFactoryContext> context;
+  SniVerifierConfigFactory factory;
+
+  Network::FilterFactoryCb cb = factory.createFilterFactoryFromProto(
+      *factory.createEmptyConfigProto(), context);
+  Network::MockConnection connection;
+  EXPECT_CALL(connection, addReadFilter(_));
+  cb(connection);
+}
```
qiwzhang: Could you add some tests to call onData() to cover the ssl_handshake and callback flow.?

vadimeisenbergibm: yes, I am writing the tests now.

src/envoy/tcp/sni_verifier/sni_verifier.cc
htuch: Correct; one network filter instance per downstream connection, if you do `thread_local` you are will be sharing this buffer between them.

vadimeisenbergibm: @htuch and for a listener filter, like tls inspector, no such sharing will occur, will it?

htuch: @vadimeisenbergibm sharing will happen as long as the new connection being processesd in the listener filter is on the same worker thread.

vadimeisenbergibm: @htuch I see, now it is clear, thanks!

vadimeisenbergibm:
Closing in favor of https://github.com/istio/proxy/pull/2050.

@qiwzhang @lizan I resubmitting this PR as a PR to release-1.1 https://github.com/istio/proxy/pull/2050, this one I mistakenly created as a PR to master, sorry.

PR #2050, opened 2018-12-03
Add sni verifier filter

ronenschafferibm:
Resubmission of https://github.com/istio/proxy/pull/2046 into release-1.1.

This PR solves the security hole described in https://goo.gl/cQ29rX, "Enable egress policies per source workload". The solution is related to the following text:

> Set the outer SNI to be equal to the inner SNI in the sidecar proxy and then add a filter in the gateway’s proxy that will verify that the outer SNI and the inner SNI match.

This PR adds a filter "that will verify that the outer SNI and the inner SNI match".

Based on the TLS inspector of Envoy.

### DIFF
--- a/src/envoy/tcp/sni_verifier/sni_verifier.cc
+++ b/src/envoy/tcp/sni_verifier/sni_verifier.cc
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-#include "extensions/filters/network/network_level_sni_reader/network_level_sni_reader.h"
+#include "src/envoy/tcp/sni_verifier/sni_verifier.h"
 
 #include "envoy/buffer/buffer.h"
 #include "envoy/common/exception.h"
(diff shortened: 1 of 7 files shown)


vadimeisenbergibm:
cc @lizan

src/envoy/tcp/sni_verifier/sni_verifier.cc
```diff
@@ -0,0 +1,184 @@
...
+  SSL_set_app_data(ssl_.get(), this);
+  int ret = SSL_do_handshake(ssl_.get());
+
+  // reset the app data
+  SSL_set_app_data(ssl_.get(), nullptr);
+
+  // This should never succeed because an error is always returned from the SNI
+  // callback.
+  ASSERT(ret <= 0);
+  switch (SSL_get_error(ssl_.get(), ret)) {
```
qiwzhang: Do we need to clear SSL errors by calling ERR_clear_error()?

vadimeisenbergibm: @qiwzhang I have added it to be on the safe side in https://github.com/istio/proxy/pull/2050/commits/607ee02e0fe51e72d5e3327c54ef0d3765596bce.

PR #2052, opened 2018-12-07
Handle inflight request canceling properly

qiwzhang:

While reading Proxy code, found that in-flight request is not canceled properly.
cancel() should be called. Otherwise, if it is called, but memory is freed. it may cause crash

**Release note**:
```release-note
None
```


qiwzhang:
@lizan could you take a look?

qiwzhang:
Use https://github.com/istio/proxy/pull/2053 instead of this

PR #2063, opened 2018-12-13
Authenticate an exchanged token

### LINKED ISSUE #2062: Authenticate an exchanged token
A JWT token from a client may be exchanged to another token upon reaching the perimeter of a service mesh. Need to authenticate the exchanged token and extract the authenticated payload to be consumed by the authorization filter.

lei-tang:
**What this PR does / why we need it**: authenticate an exchanged token. 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2062 

**Release note**:

```release-note
```


src/envoy/http/authn/authn_utils.cc
```diff
@@ -27,6 +27,12 @@ namespace AuthN {
 namespace {
 // The JWT audience key name
 static const std::string kJwtAudienceKey = "aud";
+// The JWT issuer key name
+static const std::string kJwtIssuerKey = "iss";
+// The APToken issuer
+static const std::string kAPTokenIssuer = "https://cloud.google.com/iap";
```
quanjielin: remove this

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -68,7 +68,13 @@ bool AuthenticatorBase::validateX509(const iaapi::MutualTls& mtls,
 bool AuthenticatorBase::validateJwt(const iaapi::Jwt& jwt, Payload* payload) {
   std::string jwt_payload;
   if (filter_context()->getJwtPayload(jwt.issuer(), &jwt_payload)) {
-    return AuthnUtils::ProcessJwtPayload(jwt_payload, payload->mutable_jwt());
+    std::string payload_to_process = jwt_payload;
+    std::string original_payload;
+    if (AuthnUtils::ExtractOriginalPayload(jwt_payload, &original_payload)) {
```
quanjielin: we may require token exchange service to keep origin header of jwt, so that filter could restore original jwt payload (for back compatibility in case app process original jwt); this could be done later. /cc @diemtvu

src/envoy/http/jwt_auth/sample/APToken/APToken-example1.jwt
```diff
@@ -0,0 +1 @@
+eyJhbGciOiJSUzI1NiIsImtpZCI6IkRIRmJwb0lVcXJZOHQyenBBMnFYZkNtcjVWTzVaRXI0UnpIVV8tZW52dlEiLCJ0eXAiOiJKV1QifQ.eyJhdWQiOiJleGFtcGxlLWF1ZGllbmNlIiwiZW1haWwiOiJmb29AZ29vZ2xlLmNvbSIsImV4cCI6NDY5ODAyMzM3NywiaWF0IjoxNTQ0NDIzMzc3LCJpc3MiOiJodHRwczovL2Nsb3VkLmdvb2dsZS5jb20vaWFwIiwiaXN0aW9fYXR0cmlidXRlcyI6W3sic291cmNlLmlwIjoiMTI3LjAuMC4xIn1dLCJrZXkxIjpbInZhbDIiLCJ2YWwzIl0sIm9yaWdpbmFsX3BheWxvYWQiOnsiZW1haWwiOiJ1c2VyQGV4YW1wbGUuY29tIiwiaXNzIjoiaHR0cHM6Ly9hY2NvdW50cy5leGFtcGxlLmNvbSIsInN1YiI6ImV4YW1wbGUtc3ViamVjdCJ9LCJzdWIiOiJodHRwczovL2FjY291bnRzLmV4YW1wbGUuY29tLzEyMzQ1NTY3ODkwIn0.lY-bObNybFN8RxguZcgEd3TrC7-dSRH6VjXqkMvaN1WXz5xk8EEca0vlfkc2RRCZCCGDsSJiZOWOkV5c3RX5EAGcHppMFxuX2YZMQOoConuN_WCdp88gTKuICDUp_ZLi06nWOzqhIr4GG2QYDxSOMxTFqMl_qcJTKEcjpJM-pbhVTgbTb020C4_mFOEespff2Kh_ySzkZnlreddP_oUvUrKKoIHBf3M_ZKpivuSyyBjL52iFmsMJL4nmsk-xLIDuw5NeNyUGDa5j-EwiD3XvrFurrIAkKPT42E5Z5uomuw8h-zUoJt_jpdIynOOA2eyQXfXwsVnpOnMWDut66u92xQ
```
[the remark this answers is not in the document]
lei-tang: When I have time, I can create a separate PR for the readme of how this JWT is generated.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -100,6 +104,42 @@ bool AuthnUtils::ProcessJwtPayload(const std::string& payload_str,
   return true;
 }
 
+bool AuthnUtils::ExtractOriginalPayload(const std::string& token,
```
liminw: How do you decide when to extract original claims instead of normal claims? From the code, it seems that as long as "original_payload" claim exists, you think it is an AP token and extract from original payload? I don't think this is the right way to handle it. Admin should either specifically mention that this is AP token, or "iss" matches configured AP token issuer.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -100,6 +104,42 @@ bool AuthnUtils::ProcessJwtPayload(const std::string& payload_str,
...
+  if (json_obj->hasObject(kAPTokenOriginalPayload) == false) {
+    return false;
+  }
+
+  Envoy::Json::ObjectSharedPtr original_payload_obj;
+  try {
+    auto original_payload_obj = json_obj->getObject(kAPTokenOriginalPayload);
+    std::string iss1 = json_obj->getString(kJwtIssuerKey, "");
+    std::string iss2 = original_payload_obj->getString(kJwtIssuerKey, "");
+    // Token exchange makes the issuers of the APToken and the original JWT
```
liminw: This logic sounds like a hack. It basically says "if original_payload exists, and original issuer is not the same as new issuer, it is AP token". This is not how AP token is defined.

src/envoy/http/authn/authenticator_base.cc
liminw: Is the goal making AP token "transparent" to the service admin? I don't think this is right way of handling ap token. The service/namespace admin should be well aware that he/she is handling AP token, which is different from an external token.

src/envoy/http/authn/authn_utils.cc
lei-tang: If a token is an APToken, the original claims are extracted. Otherwise, normal claims are extracted. What token is considered an APToken is based on Quanjie's suggestion: if a token has an original_payload claim, and there is an issuer claim in the original_payload claim, and the issuer in the original_payload claim is different from the issuer of the token, the token is considered an APToken.  
The rationale of such code is that a token exchange makes the issuers of the APToken and the original JWT to be different (if the two issuers are the same, there is no need for the token exchange).The code to check that the issuers are different are on the lines 125~129.

lei-tang: The extracted original_payload will not be verified by a recipient of the APToken: the APToken issuer verified the original JWT and the recipient of the APToken verified the APToken. If an APToken is valid, the recipient regards the payload is valid.

liminw: @yangminzhu already gave an example why just checking "original_payload" (and issuers are different) is not enough to prove the token is indeed an AP token. My preferred way is to introduce APTokenConfig where users can specify their specific AP token properties (e.g., iss). If you want to work something out faster, an alternative is to standardize AP token header (e.g., x-ingress-authorization (TBD)). We can say that the token in the standardized header must be AP token.

src/envoy/http/jwt_auth/sample/APToken/key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
diemtvu: What are these key/cert file for. Please add a README if they are used to generate test data, and/or how to replace this file if need to.

src/envoy/http/jwt_auth/sample/APToken/aptoken-envoy.conf
```diff
@@ -0,0 +1,116 @@
+{
```
diemtvu: Does this file used in unit test? I don't see any code using this.

src/envoy/http/authn/authn_utils.cc
diemtvu: While agree with not doing this implicitly, I don't totally agree with Yangmin example. The original_payload (and all other claims in the outer jwt) are still verified by the outer jwt issuer, right

src/envoy/http/authn/authenticator_base.cc
lei-tang: Based on the discussion today, the code has been revised.

src/envoy/http/authn/authn_utils.cc
lei-tang: Based on the discussion today, the code has been revised.

src/envoy/http/authn/authn_utils.cc
lei-tang: Based on the discussion today, the code has been revised.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -27,6 +27,10 @@ namespace AuthN {
 namespace {
 // The JWT audience key name
 static const std::string kJwtAudienceKey = "aud";
+// The JWT issuer key name
+static const std::string kJwtIssuerKey = "iss";
+// The key name for the APToken original claims
+static const std::string kAPTokenOriginalPayload = "original_payload";
```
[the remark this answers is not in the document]
lei-tang: Based on the discussion today, original_payload has been changed to original_claims.

src/envoy/http/jwt_auth/sample/APToken/key.pem
lei-tang: These key/cert files have been removed. I have added a guide of sending an example exchanged token to the jwt-authn filter and the Istio authn filter, and observing that the example backend echoes back the request when the authentication succeeds.

src/envoy/http/jwt_auth/sample/APToken/aptoken-envoy.conf
lei-tang: This file is used in a guide I added.  I have added a guide of sending an example exchanged token to the jwt-authn filter and the Istio authn filter, and observing that the example backend echoes back the request when the authentication succeeds.

diemtvu: It's better to write a test. You can add them to test/integration. There is already few test setup examples.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -68,7 +73,21 @@ bool AuthenticatorBase::validateX509(const iaapi::MutualTls& mtls,
 bool AuthenticatorBase::validateJwt(const iaapi::Jwt& jwt, Payload* payload) {
   std::string jwt_payload;
   if (filter_context()->getJwtPayload(jwt.issuer(), &jwt_payload)) {
-    return AuthnUtils::ProcessJwtPayload(jwt_payload, payload->mutable_jwt());
+    std::string payload_to_process = jwt_payload;
+    std::string original_payload;
+    bool found = false;
+    for (auto h : jwt.jwt_headers()) {
```
diemtvu: For readability, make a helper function (say IsAPTokenIssuer) for this.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -68,7 +73,21 @@ bool AuthenticatorBase::validateX509(const iaapi::MutualTls& mtls,
...
+      if (kExchangedTokenHeaderName == h) {
+        found = true;
+        break;
+      }
+    }
+    if (found &&
+        AuthnUtils::ExtractOriginalPayload(jwt_payload, &original_payload)) {
```
diemtvu: If it is APToken (found), but cannot extract original payload, then probably should return false.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -68,7 +73,21 @@ bool AuthenticatorBase::validateX509(const iaapi::MutualTls& mtls,
 bool AuthenticatorBase::validateJwt(const iaapi::Jwt& jwt, Payload* payload) {
```
diemtvu: You should have new unit tests for this change.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -29,6 +29,11 @@ namespace Http {
 namespace Istio {
 namespace AuthN {
 
+namespace {
+// The default header name for an exchanged token
+static const std::string kExchangedTokenHeaderName = "x-ingress-authorization";
```
liminw: use "ingress-authorization".

src/envoy/http/authn/authenticator_base.cc
liminw: Instead of doing string copy, you can directly put "payload_to_process" in the function. 

AuthnUtils::ExtractOriginalPayload(jwt_payload, &payload_to_process)

src/envoy/http/authn/authn_utils.cc
liminw: @yangminzhu "xyz.com" is listed by the operator as a trusted issuer, and we also do signature verification to make sure the token is signed by "xyz.com". So whatever contained in the token, we can assume that it is trusted. 

"ingress-authorization" header is the default AP token location. The token put in there we can assume it is AP token. We will validate the token's signature to ensure that it is trusted. So I don't have concern on its security aspect.

src/envoy/http/authn/authenticator_base.cc
lei-tang: If put "payload_to_process" in ExtractOriginalPayload(), ExtractOriginalPayload() may change "payload_to_process" even when ExtractOriginalPayload() returns false. Therefore,  "payload_to_process" should not be put in ExtractOriginalPayload().

lei-tang: Limin mentioned that an APToken may not necessarily contain the original payload (i.e., original_claims). If an APToken must contain the original payload, we can use it as a way to tell whether a token is an APToken. Hence, no false is returned here.

src/envoy/http/jwt_auth/sample/APToken/aptoken-envoy.conf
lei-tang: A test has been added.

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authenticator_base.cc
lei-tang: Test added.

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -29,6 +29,24 @@ namespace Http {
...
 namespace AuthN {
 
+namespace {
+// The default header name for an exchanged token
+static const std::string kExchangedTokenHeaderName = "ingress-authorization";
+
+// Return whether the header for an exchanged token is found
+bool FindHeaderOfExchangedToken(const iaapi::Jwt& jwt) {
+  bool found = false;
+  for (auto h : jwt.jwt_headers()) {
```
diemtvu: If we make assumption on the location value, should we also make assumption/requirement that the Jwt spec should declare only one header location. It seems weird (and could be wrong) to define more than one for aptoken.
+ @liminw

src/envoy/http/authn/authenticator_base.cc
diemtvu: Is that so? Again, I feel wrong to while making assumption of the location, but not on the token format and it's behavior. Anyway, please add comment to explain the rationale.

src/envoy/http/authn/authenticator_base_test.cc
```diff
@@ -293,6 +309,34 @@ TEST_F(ValidateJwtTest, JwtPayloadAvailable) {
   EXPECT_TRUE(MessageDifferencer::Equals(expected_payload, *payload_));
 }
 
+TEST_F(ValidateJwtTest, OriginalPayloadOfExchangedToken) {
```
diemtvu: Please add a test for the case when original payload was not there. Any may be another test with original payload, but jwt was not in the intended location.

src/envoy/http/authn/authn_utils.cc
diemtvu: Do we still need to compare issuer?

src/envoy/http/authn/authn_utils.h
```diff
@@ -38,6 +38,12 @@ class AuthnUtils : public Logger::Loggable<Logger::Id::filter> {
   static bool ProcessJwtPayload(const std::string& jwt_payload_str,
                                 istio::authn::JwtPayload* payload);
 
+  // Parse the original_payload in an APToken.
```
diemtvu: nit Parse -> Parses, Return -> Returns

src/envoy/http/authn/sample/APToken/aptoken-envoy.conf
```diff
@@ -0,0 +1,118 @@
+{
```
diemtvu: Please add integration test to cover this setup, rather than giving sample.

src/envoy/http/authn/authenticator_base.cc
lei-tang: Done: added comments to explain the rationale.

src/envoy/http/authn/authenticator_base_test.cc
lei-tang: Done: added two more tests.

src/envoy/http/authn/authn_utils.h
lei-tang: Done.

src/envoy/http/authn/authn_utils.cc
lei-tang: I recommend comparing issuers because if an token exchange occurs, the issuer of the exchanged token should be different from the issuer of the original token. @quanjielin @liminw, what is your thought on comparing issuers?

src/envoy/http/authn/sample/APToken/aptoken-envoy.conf
lei-tang: aptoken-envoy.conf and the user guide added in this PR are for a developer that is interested in going through the APToken authentication flow on Istio proxy. The code change in this PR is on validateJwt(). Three new tests on validateJwt() have been added to test the code change. An integration test will duplicate what these three new tests have tested with respect to the code change on validateJwt().

diemtvu: The tests you added already simulate the payload input to authN filter. It doesn't cover the full interaction between jwt filter and authn filter (and even mixer) as you can have with 'integration' test (I admit, the name is a bit misleading; it's still unit test, but testing few filters chained together instead of one).

Having sample config add a little benefit to the maintenance in the long run (config might change, overhead of setting up runtime etc). Given the way to set up the test envoy exactly likes the one specified by this config is fairly simple, I would suggest convert this into a test.

src/envoy/http/authn/sample/APToken/guide.txt
```diff
@@ -0,0 +1,18 @@
+This is a guide of sending an example exchanged token to
```
diemtvu: To echo the previous point about adding test, why developer needs to follow this series of steps to maintain/debug the code if they can do so with a simple "go test" command? Not to mention, the instruction is not clear what significant one needs to look for to see if APToken was handled correctly.

lei-tang: The "go test" for running the Istio authn integration tests are in a different repository istio/istio under the path https://github.com/istio/istio/mixer/test/client. To add an integration test for authenticating an exchange token, this PR should be merged first, then the SHA in the istio.deps of the istio/istio proxy should be updated to the new SHA of istio-proxy, and finally an integration test for an exchanged token can be added. I have created an issue https://github.com/istio/istio/issues/10514 to add an integration test in the istio/istio repository and assigned it to myself.

src/envoy/http/authn/sample/APToken/aptoken-envoy.conf
lei-tang: The Istio authn integration tests that cover jwt filter, authn filter and mixer are in a different repository istio/istio under the path https://github.com/istio/istio/mixer/test/client. To add an integration test for authenticating an exchange token, this PR should be merged first, then the SHA in the istio.deps of the istio/istio proxy should be updated to the new SHA of istio-proxy, and finally an integration test for an exchanged token can be added under https://github.com/istio/istio/mixer/test/client. I have created an issue https://github.com/istio/istio/issues/10514 to add an integration test in the istio/istio repository and assigned it to myself.

src/envoy/http/authn/sample/APToken/guide.txt
diemtvu: I mean https://github.com/istio/proxy/tree/master/test/integration. The one in istio repo is a plus, but not really necessary.

src/envoy/http/authn/authenticator_base.cc
liminw: I think for the first version, it is ok to assume that AP token always carries original_claims. This may not be true going forward, but let's focus on this JWT token exchange use case for now.

src/envoy/http/authn/authn_utils.cc
liminw: I don't think it is a hard requirement that the issuers are different. We can probably drop the check here.

src/envoy/http/authn/sample/APToken/guide.txt
lei-tang: Done: integration tests have been added to https://github.com/istio/proxy/tree/master/test/integration.

src/envoy/http/authn/sample/APToken/aptoken-envoy.conf
lei-tang: Done: integration tests have been added to https://github.com/istio/proxy/tree/master/test/integration.

src/envoy/http/authn/authenticator_base_test.cc
```diff
@@ -293,6 +318,77 @@ TEST_F(ValidateJwtTest, JwtPayloadAvailable) {
...
+  jwt_.set_issuer("token-service");
+
+  (*dynamic_metadata_.mutable_filter_metadata())[Utils::IstioFilterName::kJwt]
+      .MergeFrom(
+          MessageUtil::keyValueStruct("token-service", kExchangedTokenPayload));
+
+  Payload expected_payload;
+  JsonStringToMessage(
+      R"({
+             "jwt": {
```
diemtvu: Why original_claims is not shown up in the output?

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -68,7 +86,25 @@ bool AuthenticatorBase::validateX509(const iaapi::MutualTls& mtls,
...
+      if (AuthnUtils::ExtractOriginalPayload(jwt_payload, &original_payload)) {
+        // When the header of an exchanged token is found and the token
+        // contains the claim of the original payload, the original payload
+        // is extracted and used as the token payload.
+        payload_to_process = original_payload;
+      } else {
+        // When the header of an exchanged token is found but the token
+        // does not contain the claim of the original payload, it
+        // is regarded as an invalid exchanged token.
+        ENVOY_LOG(error, "The token is invalid. The payload is {}",
```
diemtvu: Please make the error message self-explained. Change to smt like: "Expect exchanged-token with original payload claim. Received: ", ...

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -29,6 +29,24 @@ namespace Http {
 namespace Istio {
 namespace AuthN {
 
+namespace {
+// The default header name for an exchanged token
+static const std::string kExchangedTokenHeaderName = "ingress-authorization";
+
+// Return whether the header for an exchanged token is found
```
diemtvu: Return -> Returns

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -29,6 +29,24 @@ namespace Http {
...
+bool FindHeaderOfExchangedToken(const iaapi::Jwt& jwt) {
```
diemtvu: IsExchangedToken or IsAPToken may be a better name.

src/envoy/http/authn/authenticator_base.cc
diemtvu: Please change this per agreement. You can add check for  jwt_headers() size and compare the only one in there.

test/integration/exchanged_token_integration_test.cc
```diff
@@ -0,0 +1,418 @@
...
+    "https://accounts.example.com/example-subject";
+constexpr char kDestinationNamespace[] = "pod";
+constexpr char kDestinationUID[] = "kubernetes://dest.pod";
+constexpr char kSourceUID[] = "kubernetes://src.pod";
+constexpr char kTelemetryBackend[] = "telemetry-backend";
+constexpr char kPolicyBackend[] = "policy-backend";
+const std::string kHeaderForExchangedToken = "ingress-authorization";
+
+// Generates basic test request header.
+Http::TestHeaderMapImpl BaseRequestHeaders() {
```
diemtvu: I'm ok with this for now, but it would be nicer if you refactor these to a share library. Alternative (even better), you can have new test cases in the same test suit. They are not much different than the others.

liminw:
@lei-tang Based on conversation with Kurt today, AP token has the following format 
Ingress-Authorization: Istio <token> 

I think the current JWT filter expects "Bearer ". For AP token, it should be "Istio " instead. This will require some changes in JWT filter, I think.

liminw:
cc @kmbarry1

src/envoy/http/authn/authenticator_base_test.cc
lei-tang: Because original_claims is not a string or string list.

src/envoy/http/authn/authenticator_base.cc
lei-tang: FindHeaderOfExchangedToken() accurately describes what this function does.

src/envoy/http/authn/authenticator_base.cc

src/envoy/http/authn/authenticator_base.cc

test/integration/exchanged_token_integration_test.cc
lei-tang: Done: added a check for jwt_headers() size and compare the only one in there.

lei-tang:
@liminw due to time constraints, I suggest staying with the agreed solution of using ingress-authorization header for alpha. The new "Ingress-Authorization: Istio" can be put in for the next release after its design is finalized. @quanjielin, @diemtvu, @wenchenglu

liminw:
@lei-tang It's not a "new" header. We are still using "Ingress-Authorization" header. It's just instead of "Ingress-Authorization: Bearer <token>", the format is "Ingress-Authorization: Istio <token>". I do not expect this will require a big code change. Can you evaluate it? We must sync with Kurt on the expected header. Otherwise, it won't work end-to-end. cc @kmbarry1

JimmyCYJ:
@lei-tang could you create another local branch with 
```git checkout -b merge-authn-exchanged-token upstream/release-1.1```
Add your code and create another PR to release-1.1 branch?

I notice that you create this PR at 12/12, the release-1.1 branch might have diverged from master branch at that time. It would be risky to force current branch merge-authn-exchanged-token rebase to upstream/release-1.1. When updating Proxy SHA into istio/istio release-1.1 branch, this change would break pilot e2e tests.

File: src/envoy/http/authn/authn_utils.h [hunk 23]
```diff
@@ -38,6 +38,12 @@ class AuthnUtils : public Logger::Loggable<Logger::Id::filter> {
   static bool ProcessJwtPayload(const std::string& jwt_payload_str,
                                 istio::authn::JwtPayload* payload);
 
+  // Parses the original_payload in an APToken.
```
JimmyCYJ: Is APToken a standard term or an new name defined in a design doc? Could you elaborate a little more?

src/envoy/http/authn/authn_utils.h
```diff
@@ -38,6 +38,12 @@ class AuthnUtils : public Logger::Loggable<Logger::Id::filter> {
...
+  // Returns true if original_payload can be
+  // parsed successfully. Otherwise, return false.
```
JimmyCYJ: nit: returns false.

lei-tang:
As the master branch is locked, https://github.com/istio/proxy/pull/2070 is created to merge the code in this PR to the release-1.1 branch.

lei-tang:
The PR #2070 has been merged to the release-1.1 branch. The PR https://github.com/istio/proxy/pull/2071 is created to cherry-pick the PR #2070 commit from the release-1.1 branch to the master branch.

lei-tang:
#2078 has cherry-picked this PR from release 1.1 branch to the master branch.

PR #2068, opened 2018-12-15
make build work with bazel 0.20

mandarjog:
http_archive is not a native function since bazel 0.19, so it must be imported.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -15,6 +15,9 @@
 ################################################################################
 #
 
+# http_archive is not a native function since bazel 0.19
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
 load(
      "//:repositories.bzl",
      "googletest_repositories",


mandarjog:
Without this change
```
mjog@devinstance:/mnt/disks/sdb/go/src/istio.io/proxy$ bazel build //...
WARNING: The following rc files are no longer being read, please transfer their contents or import their path into one of the standard rc files:
/mnt/disks/sdb/go/src/istio.io/proxy/tools/bazel.rc
Starting local Bazel server and connecting to it...
INFO: Invocation ID: 28344aa4-6e97-43f1-9b60-b98974bcb3cc
ERROR: error loading package '': Encountered error while reading extension file 'bazel/repositories.bzl': no such package '@envoy//bazel': The native http_archive rule is deprecated. load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive") for a drop-in replacement.
```

PR #2073, opened 2019-01-04
support extract token from customer http header with prefix

quanjielin:
support extract token from customer http header with prefix like
Ingress-Authorization: Istio <token>

### DIFF
--- a/src/envoy/http/jwt_auth/token_extractor.cc
+++ b/src/envoy/http/jwt_auth/token_extractor.cc
@@ -82,9 +82,18 @@ void JwtTokenExtractor::Extract(
   for (const auto& header_it : header_maps_) {
     const HeaderEntry* entry = headers.get(header_it.first);
     if (entry) {
+      std::string token;
+      absl::string_view val = entry->value().getStringView();
+      if (val.find(" ") != absl::string_view::npos) {
+        // If the header value has prefix, trim the prefix.
+        absl::string_view prefix = val.substr(0, val.find(" "));
+        token = entry->value().c_str() + prefix.length() + 1;
+      } else {
+        token = std::string(entry->value().c_str(), entry->value().size());
+      }
+
       tokens->emplace_back(
-          new Token(std::string(entry->value().c_str(), entry->value().size()),
-                    header_it.second, false, &header_it.first));
+          new Token(token, header_it.second, false, &header_it.first));
       // Only take the first one.
       return;
     }
--- a/src/envoy/http/jwt_auth/token_extractor_test.cc
+++ b/src/envoy/http/jwt_auth/token_extractor_test.cc
@@ -154,6 +154,26 @@ TEST_F(JwtTokenExtractorTest, TestCustomHeaderToken) {
   EXPECT_FALSE(headers.get(LowerCaseString("token-header")));
 }
 
+TEST_F(JwtTokenExtractorTest, TestCustomHeaderTokenWithPrefix) {
+  auto headers = TestHeaderMapImpl{{"token-header", "istio jwt_token"}};
+  std::vector<std::unique_ptr<JwtTokenExtractor::Token>> tokens;
+  extractor_->Extract(headers, &tokens);
+  EXPECT_EQ(tokens.size(), 1);
+
+  EXPECT_EQ(tokens[0]->token(), "jwt_token");
+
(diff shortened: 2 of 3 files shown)


src/envoy/http/jwt_auth/token_extractor.cc
```diff
@@ -82,9 +82,18 @@ void JwtTokenExtractor::Extract(
   for (const auto& header_it : header_maps_) {
     const HeaderEntry* entry = headers.get(header_it.first);
     if (entry) {
+      std::string token;
+      absl::string_view val = entry->value().getStringView();
+      if (val.find(" ") != absl::string_view::npos) {
```
lizan: I think you can simplify this by using absl::StrSplit and take the last element from its result.

diemtvu: or use `rfind`, then substr(pos)

src/envoy/http/jwt_auth/token_extractor.cc
```diff
@@ -82,9 +82,18 @@ void JwtTokenExtractor::Extract(
   for (const auto& header_it : header_maps_) {
     const HeaderEntry* entry = headers.get(header_it.first);
     if (entry) {
+      std::string token;
```
diemtvu: Some comment why we want to do this would be good. May also create an issue to make the same change in upstream JWT ?

src/envoy/http/jwt_auth/token_extractor_test.cc
```diff
@@ -154,6 +154,25 @@ TEST_F(JwtTokenExtractorTest, TestCustomHeaderToken) {
...
 
+TEST_F(JwtTokenExtractorTest, TestCustomHeaderTokenWithPrefix) {
+  auto headers = TestHeaderMapImpl{{"token-header", "istio jwt_token"}};
+  std::vector<std::unique_ptr<JwtTokenExtractor::Token>> tokens;
+  extractor_->Extract(headers, &tokens);
+  EXPECT_EQ(tokens.size(), 1);
+
+  EXPECT_EQ(tokens[0]->token(), "jwt_token");
+
+  EXPECT_FALSE(tokens[0]->IsIssuerAllowed("issuer1"));
```
diemtvu: Does any of these EXPECT have something to do with your changes? I'm curious why they can have different values (i.e, where are these issuers come from? The token is just a fake one, no?)

src/envoy/http/jwt_auth/token_extractor.cc
quanjielin: thanks, updated.

src/envoy/http/jwt_auth/token_extractor.cc
quanjielin: it's added below ```// If the header value has prefix, trim the prefix.```

src/envoy/http/jwt_auth/token_extractor_test.cc
quanjielin: not really, merged to it to [TestCustomHeaderToken](https://github.com/istio/proxy/blob/release-1.1/src/envoy/http/jwt_auth/token_extractor_test.cc#L138), those issues are configured from [kExampleConfig](https://github.com/istio/proxy/blob/release-1.1/src/envoy/http/jwt_auth/token_extractor_test.cc#L31)

qiwzhang:
@lizan The different is big between JWT filter config in upstream Envoy and istio-proxy.   Istio has to change filter config when migrating to upstream filter.  So there is not much value in keeping the config consistent.

src/envoy/http/jwt_auth/token_extractor.cc
lizan: It is not simplified at all... at least cache the find result so don't do it twice.

quanjielin: address this comment in https://github.com/istio/proxy/pull/2075

liminw:
Thanks for getting this done! It looks good to me.

src/envoy/http/jwt_auth/token_extractor.cc
diemtvu: I mean it's unclear what does 'prefix' mean without context.

PR #2076, opened 2019-01-07
rename rbac permissive related attributes key due to envoy recent change

quanjielin:
https://github.com/istio/istio/issues/5345

recently envoy rename those attribute names -  
https://github.com/envoyproxy/envoy/blob/b1f2fbf46b4796c4e7f098f1496ac359a5964e45/docs/root/configuration/network_filters/rbac_filter.rst, incompatibility was caught in [mixer_client test PR](https://github.com/istio/istio/pull/10737)

### DIFF
--- a/src/envoy/http/mixer/report_data.h
+++ b/src/envoy/http/mixer/report_data.h
@@ -29,7 +29,7 @@ namespace Http {
 namespace Mixer {
 namespace {
 const std::string kRbacPermissivePolicyIDField = "shadow_effective_policy_id";
-const std::string kRbacPermissiveRespCodeField = "shadow_engine_result";
+const std::string kRbacPermissiveEngineResultField = "shadow_engine_result";
 
 // Set of headers excluded from response.headers attribute.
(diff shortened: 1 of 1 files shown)


src/envoy/http/mixer/report_data.h
```diff
@@ -28,8 +28,8 @@ namespace Envoy {
 namespace Http {
 namespace Mixer {
 namespace {
-const std::string kRbacPermissivePolicyIDField = "shadow_effective_policyID";
-const std::string kRbacPermissiveRespCodeField = "shadow_response_code";
+const std::string kRbacPermissivePolicyIDField = "shadow_effective_policy_id";
+const std::string kRbacPermissiveRespCodeField = "shadow_engine_result";
```
JimmyCYJ: Could you rename kRbacPermissiveRespCodeField to kRbacPermissiveEngineResultField or something similar, so that we keep the variable name consistent with the attribute.

quanjielin: renamed, PTAL.

PR #2078, opened 2019-01-09
Merge release-1.1 into master

hklai:


JimmyCYJ:
The macos test failure is weird. But I think it is fine as the test passes in release-1.1 recently.

PR #2089, opened 2019-01-24
POC: break batches on deletion

douglas-reid:

Proof-of-concept for discussion around how to handle delta updates with removals.

Issues related to compression with "missing" attributes.

Thoughts on approach?

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/istio/mixerclient/attribute_compressor.cc
+++ b/src/istio/mixerclient/attribute_compressor.cc
@@ -129,7 +129,14 @@ class BatchCompressorImpl : public BatchCompressor {
   BatchCompressorImpl(const GlobalDictionary& global_dict)
       : global_dict_(global_dict), dict_(global_dict) {}
 
+  bool CanAdd(const Attributes& attributes) override {
+    return attributes_used_.size() <= attributes.attributes().size();
+  }
+
   void Add(const Attributes& attributes) override {
+    for (const auto& it : attributes.attributes()) {
+      attributes_used_.insert(it.first);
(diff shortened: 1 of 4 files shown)


douglas-reid:
@geeknoid @mandarjog @kyessenov is something like this an alternative? is more needed here?

douglas-reid:
NOTE: this assumes that the sets will overlap. we could add more logic to do proper tracking / comparison of keys... this was just a quick PoC.

geeknoid:
This is exactly the kind of change I was suggesting. It's super small and localized and doesn't involve changing the protocol definition, and so avoid compat issues.

Me like.

kyessenov:
Fault injection is based on sampling. So assuming 50% fault injection rate, attributes are going to be removed every second bag on average. I feel like to we need to control variance here by making sure delta buckets are not too long and not too short. Otherwise, we're at risk of hard-to-control 99% percentile in performance. This would require much more complexity that doing it online per request.

douglas-reid:
Closing. Offline discussion summary:

- we upgrade proxies AFTER control plane so we don't need to support 1.1 proxy talking to 1.0 mixer.
- augmenting API and mixer client to set new field is less risky than intro'ing batch breaks and compression anew
- perf impact of non-compressed seems to be largely a wash (shifts CPU from proxy to mixer, making mixer CPU usage look worse, but making proxy look better by ~ same amount).

Decision: Augment API and make small client/server changes.

PR #2093, opened 2019-01-24
add mixer error details into metadata

lizan:

For istio/istio#10965, add the error details into metadata so it can be included in access log.

**Release note**:
```release-note
None
```

### DIFF
--- a/src/envoy/utils/utils.cc
+++ b/src/envoy/utils/utils.cc
@@ -139,5 +139,20 @@ Status ParseJsonMessage(const std::string& json, Message* output) {
   return ::google::protobuf::util::JsonStringToMessage(json, output, options);
 }
 
+void CheckResponseInfoToStreamInfo(
+    const istio::mixerclient::CheckResponseInfo& check_response,
+    StreamInfo::StreamInfo& stream_info) {
+  static std::string metadata_key = "istio.mixer";
+
+  if (!check_response.response_status.ok()) {
+    stream_info.setResponseFlag(StreamInfo::ResponseFlag::UnauthorizedExternalService);
+    ProtobufWkt::Struct metadata;
+    auto fields = *metadata.mutable_fields();
+    fields["status"].set_string_value(check_response.response_status.ToString());
+    stream_info.setDynamicMetadata(metadata_key, metadata);
+  }
+
+}
+
 }  // namespace Utils
 }  // namespace Envoy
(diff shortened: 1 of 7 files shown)


src/envoy/utils/utils.h
```diff
@@ -52,5 +53,9 @@ bool GetRequestedServerName(const Network::Connection* connection,
 ::google::protobuf::util::Status ParseJsonMessage(
     const std::string& json, ::google::protobuf::Message* output);
```
JimmyCYJ: Add short comment?

JimmyCYJ:
@lizan could you take a look at the macos test failure?

JimmyCYJ:
> @JimmyCYJ the macos failure is being addressed in #2094

Thanks for fixing this.

src/envoy/utils/utils.cc
```diff
@@ -139,5 +139,21 @@ Status ParseJsonMessage(const std::string& json, Message* output) {
...
+    StreamInfo::StreamInfo& stream_info) {
+  static std::string metadata_key = "istio.mixer";
+
+  if (!check_response.response_status.ok()) {
+    stream_info.setResponseFlag(
+        StreamInfo::ResponseFlag::UnauthorizedExternalService);
+    ProtobufWkt::Struct metadata;
+    auto& fields = *metadata.mutable_fields();
+    fields["status"].set_string_value(
+        check_response.response_status.ToString());
```
kyessenov: Why do we need the new attribute `istio.mixer` if we have `check.error_code` and `check.error_message` already? Can this be erased from reporting bag?

lizan: See https://github.com/istio/istio/issues/10965 for context, it is for putting descriptive message in access log. We can erase this from reporting bag, is report collecting all dynamic metadata?

kyessenov: I don't know, but it looks like all metadata is sent over. I don't think we want redundant data since it adds no value and has performance costs. Please delete this "istio.mixer" from reporting bag.

lizan: https://github.com/istio/proxy/issues/2096

wenchenglu:
@lizan could you link the corresponding 1.1 issue to this PR? thanks

wenchenglu:
never mind, I saw the issue now

PR #2105, opened 2019-02-06
Added client/server load test framework to find mixer faults.

duderino:
This is a load generator client + origin server I created to test the Mixer filter under various fault conditions using Envoy's client and server stacks.  This work falls under https://github.com/istio/istio/issues/8224

@PiotrSikora @jplevyak would love your feedback because it could be used for the wasm work and especially because this is the first >=C++11 code I've written

See test/integration/int_client_server_test.cc if you want to start with an example for context.

Another example that uses this framework to sandwich Envoy+Mixer filter between the load generator and multiple origin servers simulating Mixer servers can be found in https://github.com/istio/istio/issues/8224

### DIFF
--- a/test/integration/int_client.h
+++ b/test/integration/int_client.h
@@ -63,7 +63,7 @@ enum class ClientCallbackResult {
 };
 
 /**
- * Handle a non-termal connection event asynchronously.
+ * Handle a non-terminal connection event asynchronously.
  *
  * @param connection The connection with the event
  * @param state The state of the connection (connected or idle).
@@ -202,15 +202,11 @@ class Client : Envoy::Logger::Loggable<Envoy::Logger::Id::testing> {
   void releaseConnection(ClientConnection &connection);
 
   /**
-   * Asynchronously connect to a peer.  The same callback will be used for
-   * successful connections and subsequent disconnect/errors, so be prepared for
-   * it to be called multiple times with different ConnectionResults.
-   *
-   * @param socket_factory
-   * @param http_version
-   * @param address
-   * @param sockopts
-   * @param callback
+   * Asynchronously connect to a peer.  The connect_callback will be called on
+   * successful connection establishment and also on idle state, giving the
+   * caller the opportunity to reuse or close connections.  The close_callback
+   * will be called after the connection is closed, giving the caller the
+   * opportunity to cleanup additional resources, etc.
    */
   void connect(
       Envoy::Network::TransportSocketFactory &socket_factory,
@@ -311,7 +307,7 @@ class LoadGenerator : Envoy::Logger::Loggable<Envoy::Logger::Id::testing> {
   std::atomic<uint32_t> class_2xx_{0};
   std::atomic<uint32_t> class_4xx_{0};
   std::atomic<uint32_t> class_5xx_{0};
-  std::promise<bool> promise_;
+  std::promise<bool> promise_all_connections_closed_;
 };
 
 typedef std::unique_ptr<LoadGenerator> LoadGeneratorPtr;
--- a/test/integration/int_client.h
+++ b/test/integration/int_client.h
@@ -1,0 +1,320 @@
+/* Copyright 2019 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <future>
+#include "common/api/api_impl.h"
+#include "common/common/thread.h"
+#include "common/network/raw_buffer_socket.h"
+#include "common/stats/isolated_store_impl.h"
+#include "envoy/api/api.h"
+#include "envoy/event/dispatcher.h"
+#include "envoy/http/codec.h"
+#include "envoy/network/address.h"
+#include "envoy/thread/thread.h"
+#include "fmt/printf.h"
+#include "test/test_common/test_time.h"
+#include "test/test_common/utility.h"
+
+namespace Mixer {
+namespace Integration {
+enum class HttpVersion { HTTP1, HTTP2 };
+
+class ClientStream;
+class ClientConnection;
+class Client;
+typedef std::unique_ptr<ClientStream> ClientStreamPtr;
+typedef std::shared_ptr<ClientStream> ClientStreamSharedPtr;
+typedef std::unique_ptr<ClientConnection> ClientConnectionPtr;
+typedef std::shared_ptr<ClientConnection> ClientConnectionSharedPtr;
+typedef std::unique_ptr<Client> ClientPtr;
+typedef std::shared_ptr<Client> ClientSharedPtr;
+
+enum class ClientConnectionState {
+  CONNECTED,  // Connection established.  Non-Terminal.  Will be followed by one
+              // of the codes below.
+  IDLE,  // Connection has no active streams.  Non-Terminal.  Close it, use it,
+         // or put it in a pool.
+};
+
+enum class ClientCloseReason {
+  CONNECT_FAILED,  // Connection could not be established
+  REMOTE_CLOSE,    // Peer closed or connection was reset after it was
+                   // established.
+  LOCAL_CLOSE      // This process decided to close the connection.
+};
+
+enum class ClientCallbackResult {
+  CONTINUE,  // Leave the connection open
+  CLOSE      // Close the connection.
+};
+
+/**
+ * Handle a non-termal connection event asynchronously.
+ *
+ * @param connection The connection with the event
+ * @param state The state of the connection (connected or idle).
+ */
+typedef std::function<ClientCallbackResult(ClientConnection &connection,
+                                           ClientConnectionState state)>
+    ClientConnectCallback;
+
+/**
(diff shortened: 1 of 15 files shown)


test/integration/int_server.cc
```diff
@@ -0,0 +1,813 @@
...
+                         Envoy::Logger::Loggable<Envoy::Logger::Id::testing> {
+ public:
+  ServerStreamImpl(uint32_t id, ServerConnection &connection,
+                   ServerRequestCallback request_callback,
+                   Envoy::Http::StreamEncoder &stream_encoder)
+      : id_(id),
+        connection_(connection),
+        request_callback_(request_callback),
+        stream_encoder_(stream_encoder) {
+    // TODO do i have to do this? stream_encoder_.addCallbacks(*this);
```
jplevyak: If so then you should have annotated all the overrides with "override".

test/integration/int_server.cc
```diff
@@ -0,0 +1,813 @@
...
+    assert(nullptr == delay_timer.get());
+    if (delay_timer.get()) {
+      return;
+    }
+
+    delete_after_callback_ = false;
+    response_status_ = status;
+    response_body_ = Envoy::Grpc::Common::serializeBody(message);
+    delay_timer = connection_.dispatcher().createTimer([this, delay]() {
+      ENVOY_LOG(
```
jplevyak: You can share this code with a common lambda

test/integration/int_server.cc
```diff
@@ -0,0 +1,813 @@
...
+                connection_.name(), connection_.id(), id_);
+      stream_encoder_.encodeHeaders(
+          Envoy::Http::TestHeaderMapImpl{{":status", "200"}}, false);
+      Envoy::Buffer::InstancePtr serialized_response =
+          Envoy::Grpc::Common::serializeBody(message);
+      stream_encoder_.encodeData(*serialized_response, false);
+      stream_encoder_.encodeTrailers(Envoy::Http::TestHeaderMapImpl{
+          {"grpc-status", std::to_string(static_cast<uint32_t>(status))}});
+      delete_after_callback_ = true;
+      return;
```
jplevyak: It isn't clear to me why you are setting this here but not below.  (see below)

test/integration/int_server.cc
```diff
@@ -0,0 +1,813 @@
...
+          connection_.name(), connection_.id(), id_,
+          static_cast<long int>(delay.count()));
+      stream_encoder_.encodeHeaders(
+          Envoy::Http::TestHeaderMapImpl{{":status", "200"}}, false);
+      stream_encoder_.encodeData(*response_body_, false);
+      stream_encoder_.encodeTrailers(Envoy::Http::TestHeaderMapImpl{
+          {"grpc-status",
+           std::to_string(static_cast<uint32_t>(response_status_))}});
+      delay_timer->disableTimer();
+      connection_.removeStream(id_);
```
jplevyak: Why do you not need to call the request_callback_ here (i.e. set delete_after_callback) ?

test/integration/int_server.cc
```diff
@@ -0,0 +1,813 @@
...
+    } break;
+    default:
+      ENVOY_LOG(error,
+                "ServerConnection({}:{}) doesn't support http type %d, "
+                "defaulting to HTTP1",
+                name_, id_, static_cast<int>(http_type) + 1);
+      http_connection_ =
+          std::make_unique<Envoy::Http::Http1::ServerConnectionImpl>(
+              network_connection, *this, Envoy::Http::Http1Settings());
+  }
```
jplevyak: break

test/integration/int_server.h
```diff
@@ -0,0 +1,451 @@
...
+
+ private:
+  ServerCallbackHelper(const ServerCallbackHelper &) = delete;
+  void operator=(const ServerCallbackHelper &) = delete;
+
+  ServerAcceptCallback accept_callback_;
+  ServerRequestCallback request_callback_;
+  ServerCloseCallback close_callback_;
+
+  std::atomic<uint32_t> accepts_{0};
```
jplevyak: I think the google idiom of using Mutex and Mutex::Await along with lock annotations is simpler and less error prone.  Then you don't need to explicitly have a std::condition_variable or use atomics.

test/integration/int_client_server_test.cc
```diff
@@ -0,0 +1,379 @@
...
+  Envoy::Http::HeaderMapPtr request{
+      new Envoy::Http::TestHeaderMapImpl{{":method", "GET"},
+                                         {":path", "/"},
+                                         {":scheme", "http"},
+                                         {":authority", "host"}}};
+  load_generator.run(connections_to_initiate, requests_to_send,
+                     std::move(request));
+
+  // wait until the server has closed all connections created by the client
+  server_callbacks.wait(load_generator.connectSuccesses());
```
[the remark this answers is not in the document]
duderino: This waits for the server to complete every connection that passed the TCP 3-way handshake and the load generator (TCP client) has that info.   

If we don't wait for that event, then the assertions that follow will be flakey

test/integration/int_client.h
```diff
@@ -0,0 +1,320 @@
...
+  std::atomic<uint32_t> connect_failures_{0};
+  std::atomic<uint32_t> connect_successes_{0};
+  std::atomic<uint32_t> responses_received_{0};
+  std::atomic<uint32_t> response_timeouts_{0};
+  std::atomic<uint32_t> local_closes_{0};
+  std::atomic<uint32_t> remote_closes_{0};
+  std::atomic<uint32_t> class_2xx_{0};
+  std::atomic<uint32_t> class_4xx_{0};
+  std::atomic<uint32_t> class_5xx_{0};
+  std::promise<bool> promise_;
```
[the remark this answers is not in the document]
duderino: ACK, will rename

test/integration/int_client.h
```diff
@@ -0,0 +1,320 @@
...
+   */
+  void releaseConnection(uint32_t id);
+
+  /**
+   * For internal use
+   */
+  void releaseConnection(ClientConnection &connection);
+
+  /**
+   * Asynchronously connect to a peer.  The same callback will be used for
```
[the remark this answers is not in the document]
duderino: Thanks, code evolved but the comment didn't.  Will fix

test/integration/int_server.cc
```diff
@@ -0,0 +1,815 @@
...
+          ++remote_closes_;
+          break;
+        case ServerCloseReason::LOCAL_CLOSE:
+          ++local_closes_;
+          break;
+      }
+
+      close_callback(connection, reason);
+      std::unique_lock<std::mutex> lock(mutex_);
+      condvar_.notify_one();
```
duderino: @jplevyak if I use absl::Mutext instead of the mutex+condvar pair, this block becomes:

```
        absl::MutexLock lock(&mutex_);
        switch (reason) {
          case ServerCloseReason::REMOTE_CLOSE:
            ++remote_closes_;
            break;
          case ServerCloseReason::LOCAL_CLOSE:
            ++local_closes_;
            break;
        }
      }
 
      close_callback(connection, reason);
```

It may look a bit cleaner, but now close_callback is called under lock when before it wasn't.  Alternatively I could move the counter increments after the close_callback is called, but that too is new behavior.  And I definitely don't want to call release the lock / do the equivalent of the notification before the callback.  That would make the corresponding wait() return prematurely.

My head is far enough out of this code that I'd rather not introduce a substantive change with the 1.1 deadline looming.  Can I have a pass?  Would be happy to revisit post 1.1.

PiotrSikora: I think you can just scope it, so that lock is going to be released before the callback, i.e
```
{
    absl::MutexLock lock(&mutex_);
    
    switch (reason) {
      case ServerCloseReason::REMOTE_CLOSE:
        ++remote_closes_;
        break;
      case ServerCloseReason::LOCAL_CLOSE:
        ++local_closes_;
        break;
    }
}

close_callback(connection, reason);
```

test/integration/int_server.cc
duderino: Will do

test/integration/int_server.cc
duderino: Good catch.   The whole delete_after_callback branching can be removed safely without ill effect

test/integration/int_server.cc
duderino: I'm pretty sure all overrides were annotated with 'override', but deleting the comment since its no longer relevant.  I added it before I had tested with http2 and was concerned that my working http1 code was going to break here for http2

test/integration/int_server.cc
duderino: But I the lock going out of scope combines both an unlock and a notify/signal.  That signal is now sent before the close_callback which will introduce a race condition.  The close_callback() may still be doing something when unit test thinks everything has finished.... so the counters will be off and the test may flake

duderino: Read this:  https://abseil.io/about/design/mutex

test/integration/int_server.h
duderino: Resolved because the discussion moved to another conversation close to the tricky part of the code

test/integration/int_server.cc
silentdai: +1 for @PiotrSikora 
@duderino The current code doesn't guard you from
```close_callback() may still be doing something when unit test thinks everything has finished```
A sequence could be 
1. Thread 1: ++remote_close  // suppose this is the last one
1. Thread 2: server_callbacks.wait() returns as the number of closed matches // unit test exits
1. Thread 1: close_callback()  // oops!

duderino: I see, good call.  Will fix

duderino: done.  thank you @silentdai for the help

duderino:
@jplevyak @PiotrSikora OK I think I've addressed all the code comments.  Thank you very much for your review.  Anything else?

PiotrSikora:
You need to merge origin/release-1.1 to sync with test infra.

duderino:
@PiotrSikora @jplevyak lgtm?

PR #2106, opened 2019-02-06
[WIP] Add networking attributes and set encoding semantics

douglas-reid:
- Sets the `RepeatedAttributesSemantics` for `ReportRequests` (needed to address deletions in batches)
- Adds handling for `VirtualService` and `DestinationRule` metadata (converts them into attributes for use in telemetry)
- Adds attribute debugging statements (removing a long-standing TODO).

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*:
- https://github.com/istio/istio/issues/11151

Please provide feedback on the following:
- ~~Should we use `ReportData` instead of `CheckData` for the new attribute extraction? I initially hoped that some of the info might be available for forwarding to servers, but that doesn't appear to be happening~~ (now uses `ReportData`). 

NOTE: This PR is blocked on a decision around the naming of the attributes for the networking data.

**Release note**:

```release-note
NONE
```


src/envoy/http/mixer/check_data.cc
```diff
@@ -38,9 +38,12 @@ const std::set<std::string> RequestHeaderExclusives = {
...
 CheckData::CheckData(const HeaderMap& headers,
-                     const envoy::api::v2::core::Metadata& metadata,
-                     const Network::Connection* connection)
-    : headers_(headers), metadata_(metadata), connection_(connection) {
+                     const envoy::api::v2::core::Metadata& dynamic_metadata,
+                     const Network::Connection* connection,
+                     const envoy::api::v2::core::Metadata& route_metadata,
+                     const envoy::api::v2::core::Metadata& cluster_metadata)
+    : headers_(headers), dynamic_metadata_(dynamic_metadata), connection_(connection),
+    route_metadata_(route_metadata), cluster_metadata_(cluster_metadata) {
```
kyessenov: Cluster metadata is interesting for Check. I think if a re-route happens in a subsequent filter (or using route directive), then this selection of cluster is likely to be wrong.

src/envoy/http/mixer/filter.cc
```diff
@@ -72,7 +72,9 @@ FilterHeadersStatus Filter::decodeHeaders(HeaderMap& headers, bool) {
   initiating_call_ = true;
   CheckData check_data(headers,
                        decoder_callbacks_->streamInfo().dynamicMetadata(),
-                       decoder_callbacks_->connection());
+                       decoder_callbacks_->connection(),
+                       decoder_callbacks_->streamInfo().routeEntry()->metadata(),
+                       decoder_callbacks_->clusterInfo()->metadata());
```
kyessenov: Can a route not have an upstream, e.g. direct response action? If so, this needs to be null-protected.

src/envoy/http/mixer/check_data.cc
douglas-reid: Moved to Report.

src/envoy/http/mixer/filter.cc
douglas-reid: Moved to report data and added some checks. PTAL.

src/istio/control/http/attributes_builder.cc
```diff
@@ -274,6 +276,17 @@ void AttributesBuilder::ExtractReportAttributes(ReportData *report_data) {
   }
 
   builder.FlattenMapOfStringToStruct(report_data->GetDynamicFilterState());
+
+  std::string vs;
+  if (report_data->FindVirtualService(&vs)) {
+    // TODO: add name, namespace attributes
```
kyessenov: Can that be derived at the recipient side on-demand instead? Would save on bytes transferred.

kyessenov:
Change looks good to me from a high level perspective.
I expect to see similar thing done in a TCP filter in a follow-up.
I'll leave C++ nit-picking to experts, since I'm not at good at that.

/assign @PiotrSikora 
/assign @lizan

duderino:
@douglas-reid can this me resurrected against master?

douglas-reid:
@duderino this had some important fixes. @mandarjog did the "Sets the RepeatedAttributesSemantics for ReportRequests (needed to address deletions in batches)" work ever get done in `master` or in `release-1.1` ?

PR #2107, opened 2019-02-07
Fix a SIGSEGV when circuit breaker to upstream policy server is hit, fix a memory leak when client disconnects during upstream policy check, expose mixer client logging to Envoy, and add more mixer client counters.

duderino:

This is a partial fix for https://github.com/istio/istio/issues/8224.

I created a mixer fault test framework using an Envoy-based load generator and
upstream server.  See https://github.com/istio/proxy/pull/2105.   There will also be a future PR that I will link here that applies that test framework to Mixer.   These tests found two issues related to memory management in the MixerClientImpl.

Issue 1: If the upstream client disconnects during an upstream check to
the policy server, memory is leaked. Example:

```
Leak of 375808 bytes in 367 objects allocated from:
	@ 1fef7cf google::protobuf::internal::ArenaImpl::NewBlock
	@ 1ff011e google::protobuf::internal::ArenaImpl::SerialArena::AllocateAlignedFallback
	@ 1ff09b6 google::protobuf::internal::ArenaImpl::SerialArena::AllocateAligned
	@ 1fefb78 google::protobuf::internal::ArenaImpl::AllocateAligned
	@ 61544f google::protobuf::Arena::CreateMaybeMessage
	@ 55c835 google::protobuf::MessageLite::CreateMaybeMessage
	@ 55b973 istio::mixer::v1::CheckRequest::mutable_attributes
	@ 55a0aa istio::mixerclient::MixerClientImpl::Check
```
Issue 2: If too many requests are sent to the upstream policy server a
circuit breaker may be hit (by default this limits to 1024 concurrent
requests). The MixerClientImpl will free memory prematurely and SIGSEGV
with a stack trace like:

```
[2019-01-26 23:15:47.525][28][debug][pool] [external/envoy/source/common/http/http2/conn_pool.cc:79] max requests overflow
[2019-01-26 23:15:47.525][28][debug][router] [external/envoy/source/common/router/router.cc:532] [C0][S6223237052999230964] upstream reset
[2019-01-26 23:15:47.526][28][debug][http] [external/envoy/source/common/http/async_client_impl.cc:93] async http request response headers (end_stream=true):
':status', '200'
'content-type', 'application/grpc'
'grpc-status', '14'
'grpc-message', 'upstream connect error or disconnect/reset before headers'
'x-envoy-overloaded', 'true'

[2019-01-26 23:15:47.526][28][debug][grpc] [src/envoy/utils/grpc_transport.cc:78] Check failed with code: 14, upstream connect error or disconnect/reset before headers
[2019-01-26 23:15:47.526][28][warning][filter] [src/istio/mixerclient/client_impl.cc:136] Mixer policy check transport error: UNAVAILABLE:upstream connect error or disconnect/reset before headers
[2019-01-26 23:15:47.526][28][debug][filter] [src/istio/mixerclient/client_impl.cc:142] Mixer policy check status: UNAVAILABLE:upstream connect error or disconnect/reset before headers
[2019-01-26 23:15:47.526][28][debug][filter] [src/envoy/http/mixer/filter.cc:152] Called Mixer::Filter : check complete UNAVAILABLE:upstream connect error or disconnect/reset before headers
[2019-01-26 23:15:47.526][28][debug][filter] [src/envoy/http/mixer/filter.cc:132] Called Mixer::Filter : encodeHeaders 3
[2019-01-26 23:15:47.526][28][debug][http] [external/envoy/source/common/http/conn_manager_impl.cc:1234] [C2230][S9774838916162700467] encoding headers via codec (end_stream=false):
':status', '503'
'content-length', '69'
'content-type', 'text/plain'
'date', 'Sat, 26 Jan 2019 23:15:46 GMT'
'server', 'envoy'

[2019-01-26 23:15:47.526][28][debug][filter] [src/envoy/http/mixer/filter.cc:207] Called Mixer::Filter : onDestroy state: 3
[2019-01-26 23:15:47.526][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:125] Caught Segmentation fault, suspect faulting address 0x8
[2019-01-26 23:15:47.526][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:94] Backtrace thr<28> obj</usr/lib/x86_64-linux-gnu/libstdc++.so.6> (If unsymbolized, use tools/stack_decode.py):
[2019-01-26 23:15:47.526][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:117] thr<28> #0 0x7ff123114c04 (unknown)
[2019-01-26 23:15:47.526][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:104] thr<28> obj</home/jblatt/.cache/bazel/_bazel_jblatt/b6258e8f3eabbcfebfab41d31f03811a/sandbox/linux-sandbox/4/execroot/__main__/bazel-out/k8-fastbuild/bin/test/integration/mixer_fault_test.runfiles/__main__/test/integration/mixer_fault_test>
[2019-01-26 23:15:47.553][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #1 0x209e113 google::protobuf::internal::GeneratedMessageReflection::HasBit()
[2019-01-26 23:15:47.571][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #2 0x20923d6 google::protobuf::internal::GeneratedMessageReflection::HasField()
[2019-01-26 23:15:47.587][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #3 0x2037f03 google::protobuf::TextFormat::Printer::PrintField()
[2019-01-26 23:15:47.603][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #4 0x2036d68 google::protobuf::TextFormat::Printer::Print()
[2019-01-26 23:15:47.621][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #5 0x20367e1 google::protobuf::TextFormat::Printer::Print()
[2019-01-26 23:15:47.638][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #6 0x20332f1 google::protobuf::TextFormat::Printer::PrintToString()
[2019-01-26 23:15:47.656][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #7 0x2033144 _ZNK6google8protobuf7Message11DebugStringB5cxx11Ev
[2019-01-26 23:15:47.673][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #8 0x50af1b Envoy::Utils::GrpcTransport<>::GrpcTransport()
[2019-01-26 23:15:47.690][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #9 0x50aaff Envoy::Utils::GrpcTransport<>::GetFunc()::{lambda()#1}::operator()()
[2019-01-26 23:15:47.707][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #10 0x50a857 std::_Function_handler<>::_M_invoke()
[2019-01-26 23:15:47.724][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #11 0x582383 std::function<>::operator()()
[2019-01-26 23:15:47.758][28][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:114] thr<28> #12 0x580b9a istio::mixerclient::MixerClientImpl::Check()
```
Issues 1 and 2 were fixed by replacing the large number of objects
dynamically allocated in the MixerClientImpl::Check function and passed
to an async client lambda with two shared pointers that keep all memory
alive for the lifetime of the Envoy filter (which is about as long as
the lifetime as the upstream request). This also had a nice side effect
of removing a few large protobuf copies.   It will also make it *much* easier to retry failed mixer requests because the shared pointers will preserve the memory in retry lambdas.  See https://github.com/istio/istio/issues/9992 which I will work on next.

**Which issue this PR fixes**

Partial fix for https://github.com/istio/istio/issues/8224

Note that the original bug described this as a problem with a slow telemetry upstream. I noticed that the policy check issues were often accompanied by log messages like:

```
libprotobuf ERROR src/istio/mixerclient/report_batch.cc:75] Mixer Report failed with: UNAVAILABLE:upstream connect error or disconnect/reset before headers
```
The policy checks encountered similar issues but didn't log them.   So
in addition to the shared pointer bug fix for the leak and the sigsegv,
I also added additional logging, a simple logging abstraction so errors
in the mixer client could be relayed to envoy's logger, and additional
stats so we could track error rates, cache effectiveness, etc. 

This should handle the policy check side. No errors were found for the
telemetry side, but I'll add additional counters and logging for that
too.

MixerClientImpl is pretty much a rewrite.    I recommend using the 'unified view' to review that one.


duderino:
The counter changes in this PR will break the istio/istio build.  Please merge PR https://github.com/istio/istio/pull/11591 to fix.

src/istio/mixerclient/check_cache.cc
```diff
@@ -147,18 +148,23 @@ Status CheckCache::CacheResponse(const Attributes &attributes,
   }
   utils::HashType signature;
   if (!referenced.Signature(attributes, "", &signature)) {
-    GOOGLE_LOG(ERROR) << "Response referenced mismatchs with request";
-    GOOGLE_LOG(ERROR) << "Request attributes: " << attributes.DebugString();
-    GOOGLE_LOG(ERROR) << "Referenced attributes: " << referenced.DebugString();
+    if (MIXER_ERROR_ENABLED) {
+      MIXER_ERROR(
+          "Response referenced mismatchs with request.  Request attributes: "
```
douglas-reid: s/mismatchs/mismatched

Also, how frequently do we expect this to trigger?  This will push a bunch of stuff to the log.

src/istio/mixerclient/quota_cache.cc
```diff
@@ -218,9 +227,12 @@ void QuotaCache::SetResponse(const Attributes& attributes,
 
   utils::HashType signature;
   if (!referenced.Signature(attributes, quota_name, &signature)) {
-    GOOGLE_LOG(ERROR) << "Quota response referenced mismatchs with request";
-    GOOGLE_LOG(ERROR) << "Request attributes: " << attributes.DebugString();
-    GOOGLE_LOG(ERROR) << "Referenced attributes: " << referenced.DebugString();
+    if (MIXER_WARN_ENABLED) {
+      MIXER_WARN(
+          "Quota response referenced mismatchs with request. Request "
```
douglas-reid: same as above; only I note here that we use WARN instead of ERROR.  Is that intentional?

src/istio/mixerclient/report_batch.cc
```diff
@@ -67,10 +68,13 @@ void ReportBatch::FlushWithLock() {
...
   auto request = batch_compressor_->Finish();
   ReportResponse* response = new ReportResponse;
+
+  // TODO(jblatt) should an async call be made while this lock is held?  Can the
+  // request send block()?
   transport_(request, response, [this, response](const Status& status) {
     delete response;
     if (!status.ok()) {
-      GOOGLE_LOG(ERROR) << "Mixer Report failed with: " << status.ToString();
+      MIXER_WARN("Mixer Report failed with: %s", status.ToString().c_str());
```
douglas-reid: Can we handle timeout logs differently than other failures?  We have a lot of complaints about the logs filling up with these statements... and there isn't anything that can really be done.  I'd like to see some sort of rate limiting or sampling on this to ease the volume -- especially for timeouts.

duderino: How about a (0 == counter++ % 100) guard?   What frequency would you like?

src/istio/mixerclient/check_cache.cc
duderino: I didn't add that one, just converted it to use Envoy log formatting instead of using libprotobufs logger.   So the frequency is the same, but now it can be controlled with envoy's log threshold and format is consistent

duderino: (will fix the typo though)

src/istio/mixerclient/quota_cache.cc
duderino: I think  we should pick one threshold for both this one and the one above.  I'm tempted to say both of these are WARN - they shouldn't have happened, someone needs to look at them, but only the transaction is borked, the proxy itself can carry on happily.  Thoughts?

src/istio/mixerclient/report_batch.cc
douglas-reid: that sounds reasonable to me.  @mandarjog any thoughts here?

src/istio/mixerclient/quota_cache.cc
douglas-reid: WARN sounds good to me.

src/istio/mixerclient/check_cache.cc
douglas-reid: ok. yeah, i guess this should be rare. just curious.  we probably need a general audit of logging across Istio.

src/envoy/utils/stats.h
```diff
@@ -28,12 +28,28 @@ namespace Utils {
...
+  COUNTER(total_quota_cache_hit_denies)  \
+  COUNTER(total_remote_quota_calls)  \
+  COUNTER(total_remote_quota_accepts)  \
+  COUNTER(total_remote_quota_denies)  \
+  COUNTER(total_remote_quota_prefetch_calls)  \
+  COUNTER(total_remote_calls)  \
+  COUNTER(total_remote_call_successes)  \
+  COUNTER(total_remote_call_timeouts)  \
+  COUNTER(total_remote_call_send_errors)  \
+  COUNTER(total_remote_call_other_errors)  \
```
kyessenov: mininit: misaligned end slashes

include/istio/mixerclient/check_response.h
```diff
@@ -22,21 +22,15 @@
...
-  bool is_check_cache_hit{false};
+// The CheckResponseInfo exposes policy and quota check details to the check
+// callbacks.
+class CheckResponseInfo {
+ public:
+  virtual ~CheckResponseInfo(){};
 
-  // Whether this quota response is from cache.
-  bool is_quota_cache_hit{false};
+  virtual const ::google::protobuf::util::Status& status() const = 0;
```
kyessenov: question: do `status` and `routeDirective` need to be virtual?

src/istio/control/http/attributes_builder.h
```diff
@@ -27,7 +27,8 @@ namespace http {
 // The context for each HTTP request.
 class AttributesBuilder {
  public:
-  AttributesBuilder(RequestContext* request) : request_(request) {}
+  AttributesBuilder(istio::mixer::v1::Attributes* attributes)
```
kyessenov: question: can we avoid a raw pointer here? what stops us from using references?

duderino: Yes, I'd love to get rid of the raw pointers.

Initially I tried reworking all this code to use the CheckContext shared pointer, but it was a massive diff.   So I exercised restraint.  And Mixer v2 is coming.

src/envoy/utils/stats.h
duderino: I'll align them but no promises if make check replaces my formatting with something else :)

include/istio/mixerclient/check_response.h
duderino: The check context shared pointer with all the state now implements the check response.   Before there was a copy into a result class which included a protobuf copy.   What's worse, a virtual call or a protobuf copy?   I figured the latter.

kyessenov: Ok. I still think we could avoid it, but that probably requires more refactoring to erase mixerclient/envoy boundary than desired.

src/istio/control/http/attributes_builder.h
kyessenov: Ok, thanks for explaining. Whether or v2 is coming soon, this is production code. I am a bit concerned about the loss of context for the lifetime of the parameter, but if we can validate that the filter never outlives the attributes with integration tests under dynamic xDS reloads, it's fine. dynamic_listener test in istio is supposed to test for that, so if it passes, then it's probably OK.

duderino: Completely agree on the use of raw pointers.  I would love to do this, but I want to minimize this overly large PR and I think the current PR makes things no worse.  I have also verified that none of these raw pointers are used beyond the lifetime of the calling function.

So I'll do it if you guys are willing to review all the changes.   I think a separate PR on top of this would make that easier.   @mandarjog since he's also reviewing this.

include/istio/mixerclient/check_response.h
duderino: I suppose I could create a wrapper object that delegates instead of making a virtual function call, but that would increment shared pointer ref counts.   Still, this feels like premature optimization to me.  For the gazillion inefficiencies of the mixer filter, replacing a virtual function call that I introduced to avoid some deep protobuf copies doesn't seem worth it

duderino:
@qiwzhang Apologies for asking for your review since you've moved on to another project, but could you take a look at this?   I think it'd really benefit from your expertise and experience with this code.

There are essentially three large changes here:

1) A bunch of logging.  That's just boilerplate and should be easy to review by anyone, so you can skip if you're busy.

2) Combining all the state needed by a remote check request into shared pointers to fix a memory leak and make quota check retries possible/safer (actual retry and lots of tests for that are coming in https://github.com/istio/proxy/pull/2113 - but that badly needs a rebase which I will do after this is merged).   This needs a bit more attention, but I think others on the team are qualified to review it.

3) A large rework of the MixerClientImpl and a few much smaller changes to QuotaCache and CheckCache.  That was done to expose the control flow of policy/quota checks so we could introduce a lot more counters.   That also made a large diff on the corresponding test cases. 

3) is the set of changes that most badly needs your review.   I think it's fine if you only have time to do that.  Others have and can review 1 and 2. 

Thanks

src/envoy/utils/utils_test.cc
```diff
@@ -48,8 +49,9 @@ TEST(UtilsTest, ParseMessageWithUnknownField) {
 }
 
 TEST(UtilsTest, CheckResponseInfoToStreamInfo) {
-  ::istio::mixerclient::CheckResponseInfo
-      check_response;  // by default status is unknown
+  auto attributes = std::make_shared<::istio::mixerclient::SharedAttributes>();
+  ::istio::mixerclient::CheckContext check_response(
+      false, attributes);  // by default status is unknown
```
silentdai:
```suggestion
      false /* fail_open */, attributes);  // by default status is unknown
```

src/istio/control/client_context_base.h
```diff
@@ -40,17 +41,19 @@ class ClientContextBase {
...
   // Use mixer client object to make a Check call.
   ::istio::mixerclient::CancelFunc SendCheck(
       ::istio::mixerclient::TransportCheckFunc transport,
-      ::istio::mixerclient::CheckDoneFunc on_done, RequestContext* request);
+      ::istio::mixerclient::CheckDoneFunc on_done,
+      ::istio::mixerclient::CheckContextSharedPtr& check_context);
 
   // Use mixer client object to make a Report call.
-  void SendReport(const RequestContext& request);
+  void SendReport(istio::mixerclient::SharedAttributesPtr& attributes);
```
silentdai: It's extremely uncommon to use non-const shared_ptr. It's only change the raw pointer need a non-const shared_ptr.
Could you confirm?

src/istio/control/http/attributes_builder_test.cc
```diff
@@ -14,6 +14,7 @@
  */
 
 #include "src/istio/control/http/attributes_builder.h"
+#include "google/protobuf/stubs/status.h"
```
silentdai: nit: I believe we have a tool to reorder the header files

src/istio/control/client_context_base.h
duderino: If the consuming code wanted to increment the ref count on attributes, could it do that if attributes were const?

qiwzhang:
@duderino This PR is too big. It is mixing bug fixing with code refractory.  

For crash,  it is not clear to me how the crash happened,  without knowledge, I am not sure the new code will fix it.   Could you use issues to explain the root cause for each crash?  I will review it the issue.

For the code refractory, I like to see it breaks into smaller PR.

src/istio/control/client_context_base.h
silentdai: Yes. `const shared_ptr` restricts you from changing the raw pointer(reset() or swap()) but allows changing control block(use count)

duderino: ACK.  Will update all of these to use const.  Thanks

src/envoy/http/mixer/filter.cc
```diff
@@ -147,15 +147,17 @@ void Filter::setDecoderFilterCallbacks(
 }
 
 void Filter::completeCheck(const CheckResponseInfo& info) {
-  auto status = info.response_status;
+  auto status = info.status();
```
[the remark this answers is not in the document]
duderino: Good catch

duderino:
I'm closing this in favor of three smaller PRs.

1) Logging only:  https://github.com/istio/proxy/pull/2116
2) Replace many raw pointers with 2 shared pointers (this is the big one, sorry):  https://github.com/istio/proxy/pull/2117
3) Additional counters.  Will be small and will be based on 2.

src/istio/control/client_context_base.h
duderino: Was done on https://github.com/istio/proxy/pull/2117

src/envoy/http/mixer/filter.cc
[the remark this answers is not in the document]
duderino: Was done on https://github.com/istio/proxy/pull/2117

src/istio/control/http/attributes_builder.h
duderino: @PiotrSikora came up with a similar criticism when we were discussing a breakout of this PR.  See https://github.com/istio/proxy/pull/2117

I will do this (and I really would love to!) if you agree to review the resulting changes.  I think it would be good to do this in a new PR after 2117 is merged or else it'll become even larger

include/istio/mixerclient/client.h
```diff
@@ -16,11 +16,12 @@
 #ifndef ISTIO_MIXERCLIENT_CLIENT_H
 #define ISTIO_MIXERCLIENT_CLIENT_H
 
+#include <vector>
 #include "environment.h"
 #include "include/istio/quota_config/requirement.h"
 #include "options.h"
-
-#include <vector>
+#include "src/istio/mixerclient/check_context.h"
```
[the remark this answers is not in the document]
duderino: Verified it's needed by this code in that file:

```
class MixerClient {

  virtual CancelFunc Check(istio::mixerclient::CheckContextSharedPtr& context,
                           TransportCheckFunc transport,
                           CheckDoneFunc on_done) = 0;
```

src/envoy/utils/utils_test.cc
duderino: Was done on #2117

src/istio/control/http/attributes_builder_test.cc
duderino: Would love to learn more about this tool.  Until then, I manually fixed this on #2117

PR #2113, opened 2019-02-10
Add retries to policy checks on failed transport error

duderino:

On transport error only, retry a policy/quota check with exponential backoff and jitter

**Which issue this PR fixes** 

https://github.com/istio/istio/issues/9992

This PR takes advantage of the newly introduced CheckContext with its memory scoped by a smart pointer.  This lets us more easily implement retry - each retry lambda just increments the smart pointer.

Note that I also moved the mixer client cancellation lambda into the CheckContext to make sure it too stays valid as long as the CheckContext is referenced.  

A new retry lambda was also introduced and handled similarly to the cancellation lambda.   

The semantic difference between the two:  the cancellation lambda is invoked if the downstream client goes away.  the retry lambda is invoked if a transport error occurs and the mixer filter wants to wait a bit before retrying against another mixer upstream.

1) I recommend reviewing the changes with ignore whitespace because clang-format varies the indentation on the lambda inside mixer_client_impl.cc quite a bit.   See https://github.com/istio/proxy/pull/2113/files?w=1

2) After this is merged I will update https://github.com/istio/istio/pull/11591 to import the new istio/proxy.  That pull request makes the istio/mixerclient unit tests pass against the new mixer filter counters.

### DIFF
--- a/include/istio/control/http/request_handler.h
+++ b/include/istio/control/http/request_handler.h
@@ -35,10 +35,13 @@ class RequestHandler {
   // * extract attributes from the config.
   // * if necessary, forward some attributes to downstream
   // * make a Check call.
-  virtual ::istio::mixerclient::CancelFunc Check(
-      CheckData* check_data, HeaderUpdate* header_update,
-      ::istio::mixerclient::TransportCheckFunc transport,
-      ::istio::mixerclient::CheckDoneFunc on_done) = 0;
+  virtual void Check(CheckData* check_data, HeaderUpdate* header_update,
+                     const ::istio::mixerclient::TransportCheckFunc& transport,
+                     const ::istio::mixerclient::CheckDoneFunc& on_done) = 0;
+
+  virtual void ResetCancel() = 0;
+
+  virtual void CancelCheck() = 0;
 
   // Make a Report call. It will:
   // * check service config to see if Report is required
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -76,7 +76,7 @@ FilterHeadersStatus Filter::decodeHeaders(HeaderMap& headers, bool) {
                        decoder_callbacks_->connection());
   Utils::HeaderUpdate header_update(&headers);
   headers_ = &headers;
-  cancel_check_ = handler_->Check(
+  handler_->Check(
       &check_data, &header_update,
       control_.GetCheckTransport(decoder_callbacks_->activeSpan()),
       [this](const CheckResponseInfo& info) { completeCheck(info); });
@@ -203,14 +203,12 @@ void Filter::completeCheck(const CheckResponseInfo& info) {
 
 void Filter::onDestroy() {
   ENVOY_LOG(debug, "Called Mixer::Filter : {} state: {}", __func__, state_);
-  if (state_ != Calling) {
-    cancel_check_ = nullptr;
+  if (state_ != Calling && handler_) {
+    handler_->ResetCancel();
   }
   state_ = Responded;
-  if (cancel_check_) {
-    ENVOY_LOG(debug, "Cancelling check call");
-    cancel_check_();
-    cancel_check_ = nullptr;
+  if (handler_) {
+    handler_->CancelCheck();
   }
 }
 
--- a/src/envoy/tcp/mixer/filter.cc
+++ b/src/envoy/tcp/mixer/filter.cc
@@ -43,14 +43,12 @@ void Filter::initializeReadFilterCallbacks(
 }
 
 void Filter::cancelCheck() {
-  if (state_ != State::Calling) {
-    cancel_check_ = nullptr;
+  if (state_ != State::Calling && handler_) {
+    handler_->ResetCancel();
   }
   state_ = State::Closed;
-  if (cancel_check_) {
-    ENVOY_LOG(debug, "Cancelling check call");
-    cancel_check_();
-    cancel_check_ = nullptr;
+  if (handler_) {
+    handler_->CancelCheck();
   }
 }
 
@@ -61,7 +59,7 @@ void Filter::callCheck() {
   state_ = State::Calling;
   filter_callbacks_->connection().readDisable(true);
   calling_check_ = true;
-  cancel_check_ = handler_->Check(
+  handler_->Check(
       this, [this](const CheckResponseInfo &info) { completeCheck(info); });
   calling_check_ = false;
 }
@@ -140,7 +138,7 @@ Network::FilterStatus Filter::onNewConnection() {
 void Filter::completeCheck(const CheckResponseInfo &info) {
   const auto &status = info.status();
   ENVOY_LOG(debug, "Called tcp filter completeCheck: {}", status.ToString());
-  cancel_check_ = nullptr;
+  handler_->ResetCancel();
   if (state_ == State::Closed) {
     return;
   }
--- a/src/istio/control/client_context_base.cc
+++ b/src/istio/control/client_context_base.cc
@@ -31,6 +31,7 @@ using ::istio::mixerclient::MixerClientOptions;
 using ::istio::mixerclient::QuotaOptions;
 using ::istio::mixerclient::ReportOptions;
 using ::istio::mixerclient::Statistics;
+using ::istio::mixerclient::TimerCreateFunc;
 using ::istio::mixerclient::TransportCheckFunc;
 using ::istio::utils::CreateLocalAttributes;
 using ::istio::utils::LocalNode;
@@ -38,6 +39,16 @@ using ::istio::utils::LocalNode;
 namespace istio {
 namespace control {
 namespace {
+
+static constexpr uint32_t MaxDurationSec = 24 * 60 * 60;
+
+static uint32_t DurationToMsec(const ::google::protobuf::Duration& duration) {
+  uint32_t msec =
+      1000 * (duration.seconds() > MaxDurationSec ? MaxDurationSec
+                                                  : duration.seconds());
+  msec += duration.nanos() / 1000 / 1000;
+  return msec;
+}
 
 CheckOptions GetJustCheckOptions(const TransportConfig& config) {
   if (config.disable_check_cache()) {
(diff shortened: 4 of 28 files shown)


include/istio/control/http/request_handler.h
```diff
@@ -35,10 +35,13 @@ class RequestHandler {
...
   // * make a Check call.
-  virtual ::istio::mixerclient::CancelFunc Check(
-      CheckData* check_data, HeaderUpdate* header_update,
-      ::istio::mixerclient::TransportCheckFunc transport,
-      ::istio::mixerclient::CheckDoneFunc on_done) = 0;
+  virtual void Check(CheckData* check_data, HeaderUpdate* header_update,
+                     ::istio::mixerclient::TransportCheckFunc transport,
+                     ::istio::mixerclient::CheckDoneFunc on_done) = 0;
+
+  virtual void ResetCancel() = 0;
```
silentdai: nit: We may adopt the macro PURE in envoy/common/pure.h

include/istio/mixerclient/check_response.h
```diff
@@ -22,21 +22,15 @@
...
 
-// The CheckResponseInfo holds response information in detail.
-struct CheckResponseInfo {
-  // Whether this check response is from cache.
-  bool is_check_cache_hit{false};
+// The CheckResponseInfo exposes policy and quota check details to the check
+// callbacks.
+class CheckResponseInfo {
+ public:
+  virtual ~CheckResponseInfo(){};
```
silentdai: nit : 
```
virtual ~CheckResponseInfo() = default;
```

duderino:
Remaining work:

1) add tests for new policy check retry and cancellation counters
2) add test for exponential backoff
3) address code reviewer comments
4) rebase

duderino:
@JimmyCYJ and @silentdai could you review this?

duderino:
I'll fix the tsan and asan errors.

src/istio/mixerclient/client_impl.cc
```diff
@@ -170,82 +191,136 @@ CancelFunc MixerClientImpl::Check(CheckContextSharedPtr &context,
...
+  // This lambda and any lambdas it creates for retry will inc the ref count
+  // on the CheckContext shared pointer.
+  //
+  // The other captures (this/MixerClientImpl, transport func, and on_done func
+  // are assumed to be long-lived objects that exist until shutdown.
+  //
+  CancelFunc cancel_func = transport(
+      context->request(), context->response(),
+      [this, context, transport, on_done](const Status &status) {
+        context->resetCancel();
```
silentdai: I am confused by this cancel stuff.
- When this lambda is invoked, the `resetCancel` invalid future cancel attempt
- this lambda doesn't always succeed. A retry may occurs.

In case there is a retr, does the `retry` retore the `cancel`? If so, where is it? If not, is it a big issue?

src/envoy/tcp/mixer/filter.cc
```diff
@@ -140,7 +138,7 @@ Network::FilterStatus Filter::onNewConnection() {
 void Filter::completeCheck(const CheckResponseInfo &info) {
   const auto &status = info.status();
   ENVOY_LOG(debug, "Called tcp filter completeCheck: {}", status.ToString());
-  cancel_check_ = nullptr;
+  handler_->ResetCancel();
```
silentdai: Is the `completeCheck` always invoked after `callCheck`(so that handler_ is not nullptr)?

silentdai:
@duderino 
WTR asan, 
```
!std::equal(prefix.begin(), prefix.begin() + prefix.size(),
                      counter->name().begin()))
```
Is the asan error happens if the prefix is longer than counter->name() ?
If so, `absl::StartsWith(counter->name(), prefix)` could be a good replacement.

duderino:
The tsan issue may actually be a timeout - really a gross timeout because stuff that should have run in milliseconds was taking minutes.  

We recently upgraded clang to clang-7 on the build images.   I reproduced the timeouts on a stock ubuntu + clang-7.  When I did a 'sudo apt-get install lld-7' the timeouts went away.

Will respond to @silentdai's other finds tonight or tomorrow.  Thanks @silentdai for the weekend review.

src/istio/mixerclient/client_impl.cc
duderino: I'll try to explain....

Transactions are cancelled when the downstream client goes away and when the overall transaction times out.

In the original code there was a cancel lambda that is invoked when the transaction is cancelled.  This cancel lambda tells the grpc library to cancel any outstanding grpc request sent to the mixer server.   If a request is not outstanding, then the cancel lambda will be null and won't be called.

In the new code, it's also possible for requests to be in a retry wait when the transaction is cancelled.   In this case we don't tell the grpc library to cancel the request because no request has been passed to it.  Instead we tell envoy to cancel the retry timer.

In check_context.h you'll find this:

```
  void cancel() {
    if (cancel_func_) {
      MIXER_DEBUG("Cancelling check call");
      cancel_func_();
      cancel_func_ = nullptr;
    }

    if (retry_timer_) {
      MIXER_DEBUG("Cancelling retry");
      retry_timer_->Stop();
      retry_timer_ = nullptr;
    }
  }
```

duderino: And yes, once the retry lambda is invoked, it will restore the cancel lambda assuming it can successfully pass a request to the grpc library

src/envoy/tcp/mixer/filter.cc
duderino: I think callCheck is always invoked before completeCheck.  I'm pretty unfamiliar with this code, but other code in the same function calls handler_ without checking for null.

src/istio/control/client_context_base.cc
```diff
@@ -48,9 +59,25 @@ CheckOptions GetJustCheckOptions(const TransportConfig& config) {
...
-      config.network_fail_policy().policy() == NetworkFailPolicy::FAIL_CLOSE) {
-    options.network_fail_open = false;
+  if (config.has_network_fail_policy()) {
+    if (config.network_fail_policy().policy() ==
+        NetworkFailPolicy::FAIL_CLOSE) {
+      options.network_fail_open = false;
+    }
+
+    if (0 <= config.network_fail_policy().max_retry()) {
+      options.retries = config.network_fail_policy().max_retry();
```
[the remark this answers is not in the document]
duderino: As long as the protobuf defines it as a uint32, then the value returned will always be >= 0.  So currently this if statement is useless.   It could be seen as an extreme form of defensive programming in case the return type ever changed to a signed int... or I could just remove it and unconditionally clobber options.retries.   Thoughts?

src/istio/mixerclient/client_impl.cc
```diff
@@ -170,82 +191,136 @@ CancelFunc MixerClientImpl::Check(CheckContextSharedPtr &context,
...
+        }
+
+        MIXER_DEBUG(
+            "CheckResult transport=%s, policy=%s, quota=%s, attempt=%u",
+            status.ToString().c_str(),
+            result == TransportResult::SUCCESS
+                ? context->policyStatus().ToString().c_str()
+                : "NA",
+            result == TransportResult::SUCCESS && context->quotaCheckRequired()
+                ? context->policyStatus().ToString().c_str()
```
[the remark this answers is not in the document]
duderino: I'm impressed you're reading this so thoroughly.  Thank you for the diligent code review :)

```
  // Return a combination of the error code name and message.
  string ToString() const;
```

So it returns a copy that isn't stored in a variable... then I call c_str() on it.   Will it be destroyed before or after I call c_str()?

Interestingly, because the log threshold is warn in these tests, asan won't actually see this.

@jplevyak do you know what will happen here?

duderino:
I am working on the macos failure.   After creating a macos dev env to debug this, I'm pretty sure the failures are in the test case itself, not the code under test.   The mixer filter periodically syncs stats from the mixer client impls and the envoy stats store.  Something about the timing of that periodic sync is causing the stats to appear to be 0 to the test case when diagnostic log statements say they actually have the expected values.   I'll fix this, but I don't want to delay rc0 because of it.

src/envoy/http/mixer/filter.cc
```diff
@@ -75,7 +75,7 @@ FilterHeadersStatus Filter::decodeHeaders(HeaderMap& headers, bool) {
                        decoder_callbacks_->connection());
   Utils::HeaderUpdate header_update(&headers);
   headers_ = &headers;
-  cancel_check_ = handler_->Check(
+  handler_->Check(
```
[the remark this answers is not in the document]
duderino: See explanation in https://github.com/istio/proxy/pull/2113#discussion_r257526806

src/istio/mixerclient/client_impl.h
```diff
@@ -35,28 +39,39 @@ class MixerClientImpl : public MixerClient {
...
 
   void Report(const SharedAttributesSharedPtr& attributes) override;
 
   void GetStatistics(Statistics* stat) const override;
 
  private:
+  void RemoteCheck(CheckContextSharedPtr context,
+                   const TransportCheckFunc& transport,
+                   const CheckDoneFunc& on_done);
+
```
JimmyCYJ: Could you add a short comment for RetryDelay?

test/integration/mixer_fault_test.cc
```diff
@@ -0,0 +1,1217 @@
...
+  // These three vectors could store Server directly if
+  // Envoy::Stats::IsolatedStoreImpl was made movable.
+  std::vector<ServerPtr> origin_servers_;
+  std::vector<ServerPtr> policy_servers_;
+  std::vector<ServerPtr> telemetry_servers_;
+  Envoy::Network::Address::InstanceConstSharedPtr
+      envoy_address_;  // at most 1 envoy
+};
+
+TEST_F(MixerFaultTest, HappyPath) {
```
JimmyCYJ: Could you please add a short comment describing the test case here and below?

test/integration/mixer_fault_test.cc
```diff
@@ -0,0 +1,1217 @@
...
+  std::vector<Envoy::Network::TcpListenSocketPtr> telemetry_listeners_;
+  // These three vectors could store Server directly if
+  // Envoy::Stats::IsolatedStoreImpl was made movable.
+  std::vector<ServerPtr> origin_servers_;
+  std::vector<ServerPtr> policy_servers_;
+  std::vector<ServerPtr> telemetry_servers_;
+  Envoy::Network::Address::InstanceConstSharedPtr
+      envoy_address_;  // at most 1 envoy
+};
+
```
JimmyCYJ: Could you please add a short comment describing the test case here and below?

duderino:
Will fix the failing macos build and add commentary to the mixer_fault_test.cc in a subsequent PR.  Thanks @JimmyCYJ

PR #2109, opened 2019-02-08
Update Envoy SHA to latest with deterministic hash (release-1.1).

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

PR #2114, opened 2019-02-11
Warn user of using mTLS PERMISSIVE mode and suggest to upgrade to STRICT mode

yangminzhu:
/cc @diemtvu

### DIFF
--- a/src/envoy/http/authn/http_filter.cc
+++ b/src/envoy/http/authn/http_filter.cc
@@ -34,7 +34,26 @@ namespace Istio {
 namespace AuthN {
 
 AuthenticationFilter::AuthenticationFilter(const FilterConfig& filter_config)
-    : filter_config_(filter_config) {}
+    : filter_config_(filter_config) {
+  for (const auto& method : filter_config.policy().peers()) {
+    switch (method.params_case()) {
+      case iaapi::PeerAuthenticationMethod::ParamsCase::kMtls:
+        if (method.mtls().mode() == iaapi::MutualTls_Mode_PERMISSIVE) {
+          ENVOY_LOG(
+              warn,
+              "mTLS PERMISSIVE mode is used, connection can be either "
+              "plaintext or TLS, and client cert can be omitted. "
+              "Please consider to upgrade to mTLS STRICT mode for more secure "
+              "configuration that only allows TLS connection with client cert. "
+              "See https://istio.io/docs/tasks/security/mtls-migration/");
+          return;
+        }
+        break;
+      default:
+        break;
+    }
+  }
+}
 
 AuthenticationFilter::~AuthenticationFilter() {}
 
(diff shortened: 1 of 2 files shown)


JimmyCYJ:
Please update proxy SHA in istio/istio once this PR is merged. Thanks.

yangminzhu:
@wenchenglu Could you merge this? Thank you.

src/envoy/http/authn/peer_authenticator.cc
```diff
@@ -40,6 +40,15 @@ bool PeerAuthenticator::run(Payload* payload) {
   for (const auto& method : policy_.peers()) {
     switch (method.params_case()) {
       case iaapi::PeerAuthenticationMethod::ParamsCase::kMtls:
+        if (method.mtls().mode() == iaapi::MutualTls_Mode_PERMISSIVE) {
```
PiotrSikora: This is logged at runtime, correct? Isn't it going to spam the logs?

yangminzhu: Yes, it's logged at runtime, I think moving it to log when the authN filter is initialized should be better, though it might be ignored very easily as there will only be 1 line of it.

It's best to do this in the validation stage but the last time I checked it doesn't support returning a warning message, but only an error message that would prevent the yaml to be applied.

PiotrSikora:
You need to wait until #2109 is merged, since your PR is now out-of-sync with test infra.

PiotrSikora:
@yangminzhu #2109 is merged now, can you rebase on top of it? Thanks!

wenchenglu:
@duderino 
PTAL if you have any suggestion on the log message to highlight that users should upgrade to strict from permissive mode.

yangminzhu:
@wenchenglu @duderino Can we merge this if we need it for 1.1? I will need to update the proxy sha in Istio to make it into 1.1. Thanks.

src/envoy/http/authn/http_filter.cc
```diff
@@ -34,7 +34,26 @@ namespace Istio {
 namespace AuthN {
 
 AuthenticationFilter::AuthenticationFilter(const FilterConfig& filter_config)
-    : filter_config_(filter_config) {}
+    : filter_config_(filter_config) {
+  for (const auto& method : filter_config.policy().peers()) {
```
duderino: can we wrap this entire loop with something that logs every 1000 calls?

yangminzhu: This is already logs at low frequency. Most users only have 1 method in peers TLS related setting. Also note per @PiotrSikora's comment, I have moved this from the data path to the control path, it should only log when a new authN policy is first initialized and won't trigger for any requests. Is this enough to leave it as-is?

PiotrSikora: Oh, I didn't notice that you moved this to control path. This should be fine, then. Thanks!

duderino: yeah, SGTM to me too

duderino: And apparently it wasn't:  https://github.com/istio/istio/issues/11925

yangminzhu: Yeah, it turns out Envoy actually allocate the filter for every request, so putting it in the constructor doesn't make it on the data path. I moved it to the filter factory `createFilterFactory()` (https://github.com/istio/proxy/pull/2125) and confirmed it's only called when a new config is received by Envoy, not on every request.

PR #2115, opened 2019-02-12
Update to latest istio/api on release-1.1 branch

duderino:
Update istio/api to latest on 1.1 branch


hklai:
Actually please update https://github.com/istio/proxy/blob/release-1.1/istio.deps as well.

duderino:
@JimmyCYJ @rshriram @PiotrSikora PTAL

PR #2116, opened 2019-02-14
Added simple logging abstraction so mixer client logs can be relayed to envoy logs.

duderino:
When running outside of envoy, mixer client logs will go to stderr.

It also reduces log spam from failed reports and remote policy checks by logging one out of every 100 messages per @mandarjog and @douglas-reid request.

This is a breakout PR of https://github.com/istio/proxy/pull/2107. 

**Which issue this PR fixes** 

This will help find future issues related to https://github.com/istio/istio/issues/8224

### DIFF
--- a/src/istio/control/client_context_base.cc
+++ b/src/istio/control/client_context_base.cc
@@ -17,6 +17,7 @@
 #include "include/istio/mixerclient/check_response.h"
 #include "include/istio/utils/attribute_names.h"
 #include "include/istio/utils/attributes_builder.h"
+#include "src/istio/utils/logger.h"
 
 using ::google::protobuf::util::Status;
 using ::istio::mixer::v1::config::client::NetworkFailPolicy;
@@ -98,17 +99,20 @@ CancelFunc ClientContextBase::SendCheck(TransportCheckFunc transport,
     on_done(check_response_info);
   };
 
-  // TODO: add debug message
-  // GOOGLE_LOG(INFO) << "Check attributes: " <<
-  // request->attributes->DebugString();
+  if (MIXER_DEBUG_ENABLED) {
+    MIXER_DEBUG("Check attributes: %s",
+                request->attributes->DebugString().c_str());
+  }
+
   return mixer_client_->Check(*request->attributes, request->quotas, transport,
                               local_on_done);
 }
 
 void ClientContextBase::SendReport(const RequestContext& request) {
-  // TODO: add debug message
-  // GOOGLE_LOG(INFO) << "Report attributes: " <<
-  // request.attributes->DebugString();
+  if (MIXER_DEBUG_ENABLED) {
+    MIXER_DEBUG("Report attributes: %s", request.attributes->DebugString().c_str());
+  }
+
   mixer_client_->Report(*request.attributes);
 }
 
--- a/src/istio/control/client_context_base.cc
+++ b/src/istio/control/client_context_base.cc
@@ -99,19 +99,16 @@ CancelFunc ClientContextBase::SendCheck(TransportCheckFunc transport,
     on_done(check_response_info);
   };
 
-  if (MIXER_DEBUG_ENABLED) {
-    MIXER_DEBUG("Check attributes: %s",
-                request->attributes->DebugString().c_str());
-  }
+  MIXER_DEBUG("Check attributes: %s",
+              request->attributes->DebugString().c_str());
 
   return mixer_client_->Check(*request->attributes, request->quotas, transport,
                               local_on_done);
 }
 
 void ClientContextBase::SendReport(const RequestContext& request) {
-  if (MIXER_DEBUG_ENABLED) {
-    MIXER_DEBUG("Report attributes: %s", request.attributes->DebugString().c_str());
-  }
+  MIXER_DEBUG("Report attributes: %s",
+              request.attributes->DebugString().c_str());
 
   mixer_client_->Report(*request.attributes);
 }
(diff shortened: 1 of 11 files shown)


src/istio/control/client_context_base.cc
```diff
@@ -98,17 +99,20 @@ CancelFunc ClientContextBase::SendCheck(TransportCheckFunc transport,
     on_done(check_response_info);
   };
 
-  // TODO: add debug message
-  // GOOGLE_LOG(INFO) << "Check attributes: " <<
-  // request->attributes->DebugString();
+  if (MIXER_DEBUG_ENABLED) {
```
qiwzhang: Look each MIXER_DEBUG has this "if MIXER_DEBUG_ENABLE" check.   Can they be combined?

duderino: Wonderful idea if I can make it work.  I'll do it if I can ensure that log arguments won't be serialized into nice pretty printed messages, etc when the log threshold wouldn't log them.

duderino: Done. src/istio/utils/logger_test.cc shows that arguments to the logging macros will not be needlessly invoked if the log message is under the logging threshold.

src/istio/utils/logger.h
```diff
@@ -0,0 +1,93 @@
...
+
+extern void setLogger(std::unique_ptr<Logger> logger);
+extern Logger &getLogger();
+
+}  // namespace utils
+}  // namespace istio
+
+#define STRINGLIT2(x) #x
+#define STRINGLIT(x) STRINGLIT2(x)
+#define FILE_LINE "[" __FILE__ ":" STRINGLIT(__LINE__) "] "
```
silentdai: :+1:

src/istio/utils/logger.h
```diff
@@ -0,0 +1,93 @@
...
+                                FILE_LINE FORMAT, ##__VA_ARGS__)
+#define MIXER_WARN_INT(FORMAT, ...)                                 \
+  istio::utils::getLogger().log(istio::utils::Logger::Level::WARN_, \
+                                FILE_LINE FORMAT, ##__VA_ARGS__)
+#define MIXER_ERROR_INT(FORMAT, ...)                                 \
+  istio::utils::getLogger().log(istio::utils::Logger::Level::ERROR_, \
+                                FILE_LINE FORMAT, ##__VA_ARGS__)
+
+#define MIXER_TRACE(FORMAT, ...)            \
+  if (MIXER_TRACE_ENABLED) {                \
```
silentdai: The below else will match the `if (MIXER_TRACE_ENABLED)`.
```
if (condition) 
  MIXER_TRACE
else 
  foo()
```
Consider we have the if w/o parentheses everywhere, I would suggest below
```suggestion
  do { \
    if (MIXER_TRACE_ENABLED) {                \
      MIXER_TRACE_INT(FORMAT, ##__VA_ARGS__); \
    } \
  } while(0);
```

src/istio/utils/logger_test.cc
```diff
@@ -0,0 +1,133 @@
...
+TEST_F(LoggerTest, CallArgsOnlyIfLoggable) {
+  CountingArgument entity;
+  int expected_to_string_calls = 0;
+  int expected_is_loggable_calls = 0;
+  int expected_write_buffer_calls = 0;
+
+  // TRACE and DEBUG shouldn't be logged and shouldn't have any affect on the
+  // arguments to be logged.
+
+  MIXER_TRACE("%s", entity.toString());
```
silentdai: nit: I was wondering why this doesn't need c_str(). Maybe rename this method name to c_str() ?

src/istio/utils/logger.h
qiwzhang: I saw a lot of macro codes like this.  I still don't know why.  But I think it is a right way to do.

duderino: Nice one

src/istio/utils/logger_test.cc
duderino: Yep will do

src/istio/utils/logger.h
silentdai: In my suggestion the last semicolon is not nessesary. Acutally it would better not to append the semicolon.

@qiwzhang `do{}while(0);` and `foo(bar)` are statements while `if ... w/o else` is not. The point is just to make the `MIXER_TRACE` close to a function call.

src/istio/mixerclient/report_batch.cc
```diff
@@ -70,7 +74,12 @@ void ReportBatch::FlushWithLock() {
   transport_(request, response, [this, response](const Status& status) {
     delete response;
     if (!status.ok()) {
-      GOOGLE_LOG(ERROR) << "Mixer Report failed with: " << status.ToString();
+      if (MIXER_WARN_ENABLED &&
+          0 == REPORT_FAIL_LOG_MESSAGES++ % REPORT_FAIL_LOG_MODULUS) {
+        MIXER_WARN("Mixer Report failed with: %s", status.ToString().c_str());
```
douglas-reid: my only question here is whether or not we also want to track (via stats) the number of times this is happening to make monitoring easier.

if so, we can do that in a follow-up PR if it is going to slow down this PR.

duderino: Yep, that's coming in a separate PR.  Already done and tested, just trying to create smaller PRs

duderino:
@qiwzhang could you give me an LGTM here conditional on automated checks passing?

src/istio/utils/logger_test.cc
duderino: done

src/istio/utils/logger.h
duderino: done

duderino:
@JimmyCYJ this has been reviewed pretty thoroughly but needs an LGTM from an OWNER.  Can you help?

PR #2118, opened 2019-02-15
Additional counters for mixer policy check

duderino:
This adds many more counters to the mixer policy check so we can track different error conditions and assess the cache effectiveness of both the policy and quota caches.

This is a breakout PR of #2107.

**Which issue this PR fixes**

This will help find future issues related to [istio/istio#8224](https://github.com/istio/istio/issues/8224)


src/envoy/utils/stats.cc
```diff
@@ -53,47 +53,40 @@ void MixerStatsObject::OnTimer() {
...
-  if (new_stats.total_report_calls > old_stats_.total_report_calls) {
-    stats_.total_report_calls_.add(new_stats.total_report_calls -
-                                   old_stats_.total_report_calls);
-  }
-  if (new_stats.total_remote_report_calls >
-      old_stats_.total_remote_report_calls) {
-    stats_.total_remote_report_calls_.add(new_stats.total_remote_report_calls -
-                                          old_stats_.total_remote_report_calls);
-  }
+  CHECK_AND_UPDATE_STATS(total_check_calls_);
```
JimmyCYJ: Do we need trailing underscore "_". It should be new_stats.total_check_calls instead of new_stats.total_check_calls_, no?

JimmyCYJ: I see, you have updated stats names with "_". Please ignore my comment.

duderino: @JimmyCYJ thank you for jumping on this PR!  appreciate it :)

src/istio/mixerclient/client_impl.cc
```diff
@@ -62,36 +87,63 @@ CancelFunc MixerClientImpl::Check(CheckContextSharedPtr &context,
...
+
     //
     // If policy cache accepts the request and a quota check is not required,
     // immediately accept the request.
     //
-    context->setFinalStatus(context->policyStatus());
-    on_done(*context);
-    return nullptr;
+    ++total_check_cache_hit_accepts_;
+    if (!context->quotaCheckRequired()) {
```
mandarjog: nit: This is a bit hard to follow, there is a repeated code block with only an increment in the middle.

Could this not be the else block of line 117, `if (context->quotaCheckRequired()) `

PR #2123, opened 2019-02-20
clang-6/gcc: compiler barking fix

kyessenov:


src/envoy/http/mixer/control_factory.h
```diff
@@ -82,6 +82,7 @@ class ControlFactory : public Logger::Loggable<Logger::Id::config> {
         case istio::utils::Logger::Level::ERROR_:
           return ENVOY_LOG_CHECK_LEVEL(error);
```
PiotrSikora:
```
default:
  NOT_REACHED_GCOVR_EXCL_LINE;
```
is more in line with Envoy's coding style, and it should fix this as well.

kyessenov: done

duderino:
retroactive /lgtm

PR #2125, opened 2019-02-21
log authn permissive mode only when config is received

yangminzhu:
Signed-off-by: Yangmin Zhu <ymzhu@google.com>

Envoy actually allocates the filter for every request, we need to move the log out of the constructor to avoid spamming the logs.

https://github.com/istio/istio/issues/11925

The next step is to update the Istio proxy sha to pick this change for 1.1.

**Release note**:

```release-note
```


duderino:
Thanks for the quick turnaround @yangminzhu  This will go into 1.1.0-rc.1

Issue #2124, opened 2019-02-21
flaky/failing test: mixer_fault_test


kyessenov:
Clean build with bazel 0.22:
```
//test/integration:mixer_fault_test                                      FAILED in 4.2s
[ RUN      ] MixerFaultTest.HappyPath
test/integration/mixer_fault_test.cc:371: Failure
Expected equality of these values:
  client->class2xxResponses()
    Which is: 0
  requests_to_send
    Which is: 900
test/integration/mixer_fault_test.cc:372: Failure
Expected equality of these values:
  0
  client->class4xxResponses()
    Which is: 900
test/integration/mixer_fault_test.cc:381: Failure
Expected equality of these values:
  origin_callbacks.requestsReceived()
    Which is: 0
  requests_to_send
    Which is: 900
[  FAILED  ] MixerFaultTest.HappyPath (3392 ms)
[ RUN      ] MixerFaultTest.FailClosedAndClosePolicySocketAfterAccept
terminate called after throwing an instance of 'std::bad_function_call'
  what():  bad_function_call
[2019-02-21 00:16:18.524][31][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:81] Caught Aborted, suspect faulting address 0x3ea0000000f
[2019-02-21 00:16:18.524][31][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:67] Backtrace (use tools/stack_decode.py to get line numbers):
[2019-02-21 00:16:18.551][31][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:73] #0: Envoy::SignalAction::sigHandler() [0x5654fde750e2]
[2019-02-21 00:16:18.552][31][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:73] #1: __restore_rt [0x7f06aa52a890]
```

/assign @duderino

kyessenov:
Please triage. I'm not sure what's happening in the test for it to segfault...

duderino:
@kyessenov I've never seen this before.  Can you tell me more about the build env so I can try to reproduce?

kyessenov:
```sh
bazel version
[...]

PR #2126, opened 2019-02-22
Deflake macos MixerFaultTest by broadening assertion ranges.

duderino:

Fix flake in macos tests that was introduced by https://github.com/istio/proxy/pull/2113

### DIFF
--- a/test/integration/mixer_fault_test.cc
+++ b/test/integration/mixer_fault_test.cc
@@ -682,19 +682,19 @@ TEST_F(MixerFaultTest, TolerateTelemetryBlackhole) {
 #ifndef __APPLE__
   // On Linux every connection will be successfully established.
   EXPECT_EQ(client->connectSuccesses(), connections_to_initiate);
-  EXPECT_EQ(0, client->connectFailures());
+  EXPECT_EQ(client->connectFailures(), 0);
   // Client close callback called for every client connection.
   EXPECT_EQ(client->localCloses(), connections_to_initiate);
   // Client response callback is called for every request sent
   EXPECT_EQ(client->responsesReceived(), requests_to_send);
   // Every response was a 2xx class
   EXPECT_EQ(client->class2xxResponses(), requests_to_send);
-  EXPECT_EQ(0, client->class4xxResponses());
-  EXPECT_EQ(0, client->class5xxResponses());
-  EXPECT_EQ(0, client->responseTimeouts());
+  EXPECT_EQ(client->class4xxResponses(), 0);
+  EXPECT_EQ(client->class5xxResponses(), 0);
+  EXPECT_EQ(client->responseTimeouts(), 0);
   // No client sockets are rudely closed by server / no client sockets are
   // reset.
-  EXPECT_EQ(0, client->remoteCloses());
+  EXPECT_EQ(client->remoteCloses(), 0);
 
   // Origin server should see all requests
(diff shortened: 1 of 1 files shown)


kyessenov:
https://github.com/istio/proxy/issues/2124 is happening on linux. Does this help with that issue too?

test/integration/mixer_fault_test.cc
```diff
@@ -545,16 +546,12 @@ TEST_F(MixerFaultTest, FailClosedAndSendPolicyResponseSlowly) {
   // Every response was a 5xx class
   EXPECT_EQ(0, client->class2xxResponses());
   EXPECT_EQ(0, client->class4xxResponses());
-  EXPECT_EQ(requests_to_send, client->class5xxResponses());
+  EXPECT_EQ(client->class5xxResponses(), requests_to_send);
```
JimmyCYJ: I think EXPECT_EQ(requests_to_send, client->class5xxResponses()) is consistent with other EXPECT_EQ(expected result, tested result).

JimmyCYJ:
> #2124 is happening on linux. Does this help with that issue too?

I think this addresses failures on macos.

test/integration/mixer_fault_test.cc
```diff
@@ -563,6 +560,44 @@ TEST_F(MixerFaultTest, FailClosedAndSendPolicyResponseSlowly) {
   // Policy server closes every connection
   EXPECT_EQ(policy_cluster.connectionsAccepted(),
             policy_cluster.localCloses() + policy_cluster.remoteCloses());
+#else
+  // Macos is a bit flakier than Linux, so broaden assetion ranges to reduce
```
PiotrSikora: Nit: it's `macOS`.

test/integration/mixer_fault_test.cc
```diff
@@ -563,6 +560,44 @@ TEST_F(MixerFaultTest, FailClosedAndSendPolicyResponseSlowly) {
...
+  // test flakes.
+
+  // Most connections are successfully established.
+  EXPECT_IN_RANGE(client->connectSuccesses(), 0.8 * connections_to_initiate,
+                  connections_to_initiate);
+  EXPECT_IN_RANGE(client->connectFailures(), 0, 0.2 * connections_to_initiate);
```
PiotrSikora: Could you add this as well?
```
EXPECT_EQ(client->connectSuccesses() + client->connectFailures(), connections_to_initiate);
```
Thanks!

test/integration/mixer_fault_test.cc
PiotrSikora: Yeah, I agree with @JimmyCYJ that the style in all those checks is a bit inconsistent.

But taking into account the range tests for macOS, I think that the tests above should move runtime-values to the left-hand side for readability, i.e.:
```
EXPECT_EQ(client->class2xxResponses(), 0);
EXPECT_EQ(client->class3xxResponses(), 0);
```

duderino: The readability issues I will fix

(they are consistent with yoda conditions but in this case provide no actual value)

duderino: done

test/integration/mixer_fault_test.cc
duderino: done

duderino:
thanks @JimmyCYJ and @PiotrSikora

PR #2128, opened 2019-02-23
Add additional telemetry report counters

duderino:

Adds counters to the mixer telemetry reports so we can see how many succeed vs. fail and get a coarse grained understanding of the cause of the failures.

**Which issue this PR fixes**

Helps diagnose issues related to https://github.com/istio/istio/issues/8224

### DIFF
--- a/src/istio/mixerclient/report_batch.cc
+++ b/src/istio/mixerclient/report_batch.cc
@@ -15,6 +15,7 @@
 
 #include "src/istio/mixerclient/report_batch.h"
 #include "include/istio/utils/protobuf.h"
+#include "src/istio/mixerclient/status_util.h"
 #include "src/istio/utils/logger.h"
 
 using ::google::protobuf::util::Status;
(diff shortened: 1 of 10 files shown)


src/istio/mixerclient/report_batch.cc
```diff
@@ -71,12 +72,34 @@ void ReportBatch::FlushWithLock() {
 
   ++total_remote_report_calls_;
   auto request = batch_compressor_->Finish();
-  ReportResponse* response = new ReportResponse;
+  std::shared_ptr<ReportResponse> response{new ReportResponse()};
```
duderino: note to reviewer, i slipped this in while I was at it.  let me know if you want me to revert

JimmyCYJ: I am fine with this change. Thanks!

JimmyCYJ:
Could you fix the MixerFaultTest.CancelCheck that is failed at macos here or in  a followup PR?

duderino:
@JimmyCYJ yes the PR is waiting for a lgtm:  https://github.com/istio/proxy/pull/2126

PR #2137, opened 2019-03-01
telemetry: add grpc message counters

kyessenov:
Add message counters for gRPC streams:
```
request.grpc.message_count: number of input messages
response.grpc.message_count: number of output messages
```

Counts are reported at the end of the request, identically to a normal HTTP request.
The implementation works by peeking into 5 bytes prefixing every message in h2 data for gRPC.

/assign @PiotrSikora 
/assign @duderino

### DIFF
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -53,19 +53,20 @@ void incrementCounter(Buffer::Instance& data, GrpcMessageCounter* counter) {
       case GrpcMessageCounter::ExpectByte3:
       case GrpcMessageCounter::ExpectByte4:
         data.copyOut(pos, 1, &byte);
-        counter->current_size <<= 8;
+        counter->current_size = counter->current_size << 8;
         counter->current_size = counter->current_size | byte;
         pos += 1;
         counter->state =
             static_cast<GrpcMessageCounter::GrpcReadState>(counter->state + 1);
         break;
       case GrpcMessageCounter::ExpectMessage:
-        if (data.length() >= pos + counter->current_size) {
+        uint32_t available = data.length() - pos;
+        if (counter->current_size <= available) {
           pos += counter->current_size;
           counter->state = GrpcMessageCounter::ExpectByte0;
         } else {
           pos = data.length();
-          counter->current_size -= data.length() - pos;
+          counter->current_size = counter->current_size - available;
         }
         break;
     }
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -32,49 +32,6 @@ namespace Envoy {
 namespace Http {
 namespace Mixer {
 
-namespace {
-
-// detect gRPC message boundaries: each message is prefixed by 5 bytes
-// length-prefix https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
-void incrementCounter(Buffer::Instance& data, GrpcMessageCounter* counter) {
-  uint64_t pos = 0;
-  unsigned byte = 0;
-  while (pos < data.length()) {
-    switch (counter->state) {
-      case GrpcMessageCounter::ExpectByte0:
-        // skip compress flag, increment message count
-        counter->count += 1;
-        counter->current_size = 0;
-        pos += 1;
-        counter->state = GrpcMessageCounter::ExpectByte1;
-        break;
-      case GrpcMessageCounter::ExpectByte1:
-      case GrpcMessageCounter::ExpectByte2:
-      case GrpcMessageCounter::ExpectByte3:
-      case GrpcMessageCounter::ExpectByte4:
-        data.copyOut(pos, 1, &byte);
-        counter->current_size = counter->current_size << 8;
-        counter->current_size = counter->current_size | byte;
-        pos += 1;
-        counter->state =
-            static_cast<GrpcMessageCounter::GrpcReadState>(counter->state + 1);
-        break;
-      case GrpcMessageCounter::ExpectMessage:
-        uint64_t available = data.length() - pos;
-        if (counter->current_size <= available) {
-          pos += counter->current_size;
-          counter->state = GrpcMessageCounter::ExpectByte0;
-        } else {
-          pos = data.length();
-          counter->current_size = counter->current_size - available;
-        }
-        break;
-    }
-  }
-}
-
-}  // namespace
-
 Filter::Filter(Control& control)
     : control_(control),
       state_(NotStarted),
@@ -138,7 +95,7 @@ FilterDataStatus Filter::decodeData(Buffer::Instance& data, bool end_stream) {
   ENVOY_LOG(debug, "Called Mixer::Filter : {} ({}, {})", __func__,
             data.length(), end_stream);
   if (grpc_request_) {
-    incrementCounter(data, &grpc_request_counter_);
+    Utils::IncrementMessageCounter(data, &grpc_request_counter_);
   }
 
   request_total_size_ += data.length();
@@ -192,7 +149,7 @@ FilterHeadersStatus Filter::encodeHeaders(HeaderMap& headers, bool) {
 
 FilterDataStatus Filter::encodeData(Buffer::Instance& data, bool) {
   if (grpc_request_) {
-    incrementCounter(data, &grpc_response_counter_);
+    Utils::IncrementMessageCounter(data, &grpc_response_counter_);
   }
   return FilterDataStatus::Continue;
 }
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -60,7 +60,7 @@ void incrementCounter(Buffer::Instance& data, GrpcMessageCounter* counter) {
             static_cast<GrpcMessageCounter::GrpcReadState>(counter->state + 1);
         break;
       case GrpcMessageCounter::ExpectMessage:
-        uint32_t available = data.length() - pos;
+        uint64_t available = data.length() - pos;
         if (counter->current_size <= available) {
           pos += counter->current_size;
           counter->state = GrpcMessageCounter::ExpectByte0;
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -16,6 +16,7 @@
 #include "src/envoy/http/mixer/filter.h"
 
 #include "common/common/base64.h"
+#include "common/grpc/common.h"
 #include "common/protobuf/utility.h"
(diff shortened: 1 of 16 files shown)


douglas-reid:
are `request` and `response` the right way to describe a bidi stream?  would something like `server.grpc_message_count` and `client.grpc_message_count` make more sense? Are there concerns over possible confusion with unary requests/response counts?

kyessenov:
I added the message by analogy to the existing `response.grpc_status` and
`response.grpc_message`. I am not a fan of `client` and `server`, since it
is hard to distinguish from `request` and `response`, respectively. A
top-level `grpc` namespace could be appropriate, however.

kyessenov:
/assign @qiwzhang 
/assign @JimmyCYJ 
to help review

src/envoy/http/mixer/filter.h
```diff
@@ -33,6 +33,30 @@ struct PerRouteServiceConfig : public Router::RouteSpecificFilterConfig {
   std::string hash;
 };
 
+// The struct to store gRPC message counter state.
+struct GrpcMessageCounter {
+  GrpcMessageCounter() : state(ExpectByte0), current_size(0), count(0){};
+
+  // gRPC uses 5 byte header to encode subsequent message length
```
JimmyCYJ: Could you add the link to gRPC spec https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md, or state here that gRPC message has 1 byte Compressed-Flag+4 byte Message-Length?

src/envoy/http/mixer/filter.cc
```diff
@@ -30,6 +31,49 @@ namespace Envoy {
 namespace Http {
 namespace Mixer {
 
+namespace {
+
+// detect gRPC message boundaries: each message is prefixed by 5 bytes
+// length-prefix https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
+void incrementCounter(Buffer::Instance& data, GrpcMessageCounter* counter) {
+  uint64_t pos = 0;
```
JimmyCYJ: I am not sure if it is correct to add pos=0 here. 
Is Filter::encodeData() called only once in the response pipeline, or called multiple times until all response data are sent out?
If Filter::encodeData could be invoked multiple times, then in one call, Buffer::Instance& data might have only 1 byte Compressed-Flag + 1~3 bytes Message-Length, in the next call, Buffer::Instance& data might have the rest bytes of Message-Length. The worse case could be that Buffer::Instance& data only has one byte in a call.

JimmyCYJ: Counting messages in this way might slow down the gRPC streaming. I am not sure how much difference the performance would become. Maybe we need to conduct some load test for analysis?

JimmyCYJ:
@qiwzhang @lizan Could you take a look?

src/envoy/http/mixer/report_data.h
```diff
@@ -57,15 +57,21 @@ class ReportData : public ::istio::control::http::ReportData,
...
+  uint64_t request_grpc_message_count_;
+  uint64_t response_grpc_message_count_;
 
  public:
   ReportData(const HeaderMap *headers, const HeaderMap *response_trailers,
-             const StreamInfo::StreamInfo &info, uint64_t request_total_size)
+             const StreamInfo::StreamInfo &info, uint64_t request_total_size,
+             uint64_t request_grpc_message_count,
+             uint64_t response_grpc_message_count)
       : headers_(headers),
```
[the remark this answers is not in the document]
kyessenov: Hmm, ReportData is already such a struct used temporarily to pass data between envoy and istio control. Do you think we need another layer of structs? Should we expose the fields instead and not use a constructor?

src/envoy/http/mixer/filter.h
kyessenov: The link is in .cc file. It seems that it's more relevant to the implementation than the header.

src/envoy/http/mixer/filter.cc
kyessenov: I'm not sure what the concern here. The state machine (ExpectByte0, etc) works across data buffer boundaries. The pos variable is the index into the current data buffer bytes. The two work independently.

src/envoy/http/mixer/filter.cc
```diff
@@ -30,6 +31,49 @@ namespace Envoy {
...
+  unsigned byte = 0;
```
lizan: nit `uint8_t`

src/envoy/http/mixer/filter.cc
kyessenov: Regarding the perf impact: yes, it is important to measure that. My feeling is that relative to what the rest of envoy is doing, it is probably not as big as it might seem.

lizan: agreed it shouldn't be big as it only process very few pieces of the data.

src/envoy/http/mixer/filter.cc
```diff
@@ -30,6 +31,49 @@ namespace Envoy {
 namespace Http {
 namespace Mixer {
 
+namespace {
+
+// detect gRPC message boundaries: each message is prefixed by 5 bytes
+// length-prefix https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
+void incrementCounter(Buffer::Instance& data, GrpcMessageCounter* counter) {
```
lizan: definitely need a test for this method

src/envoy/http/mixer/filter.cc
JimmyCYJ: > I'm not sure what the concern here. The state machine (ExpectByte0, etc) works across data buffer boundaries. The pos variable is the index into the current data buffer bytes. The two work independently.

I am thinking about a scenario like this. Assume Filter::encodeData(Buffer::Instance& data, bool) is called twice, in the first call, the passed Buffer::Instance& data has three bytes, 1 byte of Compressed-Flag and 2 bytes of Message-Length. In the second call, the passed Buffer::Instance& data has 2 bytes of Message-Length and Message. I am not sure if this will happen, so I would like to confirm whether we need to handle that case.

kyessenov: What happens is that on the second data buffer, we have state `ExpectByte3` and `pos = 0`. Then after first two bytes in the second buffer, the state is `ExpectMessage` and the rest of the buffer is skipped over up to the consumed length prefix.

src/envoy/http/mixer/filter.cc
kyessenov: OK. If it helps, there is an integration test https://github.com/istio/istio/pull/12195. There is definite lack of any unit testing in this particular part of the code, but we can add something for this specific function.

src/envoy/http/mixer/filter.cc
JimmyCYJ: I see, the state is kept in counter->state.

src/envoy/http/mixer/report_data.h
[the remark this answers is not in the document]
JimmyCYJ: I am fine with this change, maybe we need some refactoring if this parameter list grows in the future, or leave a TODO for that.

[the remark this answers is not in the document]
kyessenov: Added TODO.

src/envoy/http/mixer/filter.cc
kyessenov: Done

src/envoy/http/mixer/filter.cc
kyessenov: Added a unit test.

src/envoy/utils/message_counter_test.cc
```diff
@@ -0,0 +1,95 @@
...
+    // message spans two buffers
+    Buffer::addSeq(buffer1, {1, 0, 0, 0, 2, 0xFF});
+    Buffer::addSeq(buffer2, {0xFF, 0, 0, 0, 0, 2, 0xFF, 0xFF});
+    IncrementMessageCounter(buffer1, &counter);
+    IncrementMessageCounter(buffer2, &counter);
+    EXPECT_EQ(counter.state, GrpcMessageCounter::GrpcReadState::ExpectByte0);
+    EXPECT_EQ(counter.count, 2);
+  }
+
+  {
```
lizan: can you add a case that with message_size is 0? i.e. addRepeated(buffer, 10, 0) should result count 2.

kyessenov: Added.

kyessenov:
@lizan @PiotrSikora hoping to hear if this matches your expectations.

I re-used https://github.com/istio/istio/pull/12195 to run 100 batches of message 1..1000 growing linearly in size over bi-di xDS stream. This should heavily exercise gRPC path in envoy. I got the following pprof data https://gist.github.com/kyessenov/7a41190f3cacd6aa38138297d05acb5d. It's estimated at 4% for 173 samples inside the message counter function. Does that sound too much or about right?

mandarjog:
@kyessenov @PiotrSikora can this be upstreamed into envoy?

`Streaminfo` contains bytesSent / bytesReceived for h2/grpc connections it is reasonable to expect envoy to keep track of this.

qiwzhang:
My project needs message count too.  We can use it if it is in upstream Envoy.

src/envoy/utils/message_counter.cc
```diff
@@ -0,0 +1,59 @@
...
+            static_cast<GrpcMessageCounter::GrpcReadState>(counter->state + 1);
+        break;
+      case GrpcMessageCounter::ExpectMessage:
+        uint64_t available = data.length() - pos;
+        if (counter->current_size <= available) {
+          pos += counter->current_size;
+          counter->state = GrpcMessageCounter::ExpectByte0;
+        } else {
+          pos = data.length();
+          counter->current_size = counter->current_size - available;
```
PiotrSikora: Nit: this code is a bit inconsistent in using `+=`, but not `-=`, `<<=` nor `|=`.

src/envoy/http/mixer/filter.h
```diff
@@ -104,6 +103,11 @@ class Filter : public StreamFilter,
   // trailers.
   uint64_t request_total_size_{0};
 
+  // True for gRPC requests
```
PiotrSikora: Nit: add dot at the end of your comment (here and in few other places in this PR).

kyessenov:
To clarify, my 4% number came from a heavy bi-di use case. I don't have overhead numbers for a basic unary call right now.

duderino:
@kyessenov @PiotrSikora @mandarjog I wonder if we could get this merged by defaulting the feature off?   That seems unsatisfying to me, but then again, even a 4% *upper bound* worries me

PR #2139, opened 2019-03-05
Add the support of bypassing JWT authn for CORS requests

### LINKED ISSUE #651: JWT-AUTH and OPTIONS requests
Hi,

I started to play with the `jwt-auth` and it works well, I could even integrate it with my Keycloak IDP. But now I'm trying to do a request from the browser and it fails on the `OPTIONS` preflight because the authorization header is not present (the browser remove it automatically). Is there an easy way to tell the filter to skip 'OPTIONS' requests (even if it's a bit hacky, it's for demo purposes) ?

Thx

lei-tang:
**What this PR does / why we need it**: Add the support of bypassing JWT authn for CORS requests

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #651 

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/jwt_auth/jwt_authenticator.cc
+++ b/src/envoy/http/jwt_auth/jwt_authenticator.cc
@@ -62,6 +62,17 @@ void JwtAuthenticator::Verify(HeaderMap &headers,
   headers_ = &headers;
   callback_ = callback;
 
+  // Per the spec
+  // http://www.w3.org/TR/cors/#cross-origin-request-with-preflight-0, CORS
+  // pre-flight requests shouldn't include user credentials.
+  if (headers_->Method() &&
+      LowerCaseString(kOptionsHttpMethod) ==
+          LowerCaseString(headers_->Method()->value().c_str())) {
+    ENVOY_LOG(debug, "CORS preflight requests are passed through.");
+    DoneWithStatus(Status::OK);
+    return;
+  }
+
   ENVOY_LOG(debug, "Jwt authentication starts");
   std::vector<std::unique_ptr<JwtTokenExtractor::Token>> tokens;
   store_.token_extractor().Extract(headers, &tokens);
@@ -214,15 +225,6 @@ void JwtAuthenticator::VerifyKey(const PubkeyCacheItem &issuer_item) {
 
 bool JwtAuthenticator::OkToBypass() {
   if (store_.config().allow_missing_or_failed()) {
-    return true;
-  }
-
-  // Per the spec
-  // http://www.w3.org/TR/cors/#cross-origin-request-with-preflight-0, CORS
-  // pre-flight requests shouldn't include user credentials.
-  if (headers_->Method() &&
-      LowerCaseString(kOptionsHttpMethod) ==
-          LowerCaseString(headers_->Method()->value().c_str())) {
     return true;
   }
 
--- a/src/envoy/http/jwt_auth/jwt_authenticator.cc
+++ b/src/envoy/http/jwt_auth/jwt_authenticator.cc
@@ -25,9 +25,12 @@ namespace {
 // The HTTP header to pass verified token payload.
 const LowerCaseString kJwtPayloadKey("sec-istio-auth-userinfo");
 
+// The CORS OPTIONS HTTP method
+const LowerCaseString kOptionsHttpMethod("options");
+
 // Extract host and path from a URI
-void ExtractUriHostPath(const std::string& uri, std::string* host,
-                        std::string* path) {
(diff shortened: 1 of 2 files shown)


src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -214,11 +217,20 @@ bool JwtAuthenticator::OkToBypass() {
     return true;
   }
 
+  // Per the spec
+  // http://www.w3.org/TR/cors/#cross-origin-request-with-preflight-0, CORS
+  // pre-flight requests shouldn't include user credentials.
+  if (headers_->Method() &&
```
qiwzhang: how about bail out earlier, like in line 63?

lei-tang: Done.

src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -62,6 +62,17 @@ void JwtAuthenticator::Verify(HeaderMap &headers,
   headers_ = &headers;
   callback_ = callback;
 
+  // Per the spec
+  // http://www.w3.org/TR/cors/#cross-origin-request-with-preflight-0, CORS
+  // pre-flight requests shouldn't include user credentials.
+  if (headers_->Method() &&
+      LowerCaseString(kOptionsHttpMethod) ==
```
qiwzhang: I think Envoy has fixed string for OPTIONS

https://github.com/envoyproxy/envoy/blob/master/source/common/http/headers.h#L176

qiwzhang: I prefer to do this check in Fitler.cc right inside the decodeHeaders().

qiwzhang:
Could you help to add this for Envoy one too?

It is here

https://github.com/envoyproxy/envoy/blob/master/source/extensions/filters/http/jwt_authn/filter.cc#L23

src/envoy/http/jwt_auth/jwt_authenticator.cc
lei-tang: decodeHeaders() does not have tests. Placing the check in Verify() will enable tests. Plus, placing the check in Verify() is equivalent to placing the check in decodeHeaders().

lei-tang:
Sure, I can add this PR to Envoy repo after this PR is merged.

src/envoy/http/jwt_auth/jwt_authenticator.cc
lei-tang: Done: changed to use the OPTIONS fixed string from Envoy.

src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -66,9 +63,9 @@ void JwtAuthenticator::Verify(HeaderMap &headers,
   // http://www.w3.org/TR/cors/#cross-origin-request-with-preflight-0, CORS
   // pre-flight requests shouldn't include user credentials.
   if (headers_->Method() &&
-      LowerCaseString(kOptionsHttpMethod) ==
+      LowerCaseString(Http::Headers::get().MethodValues.Options) ==
```
qiwzhang: Why do we need to lower case the method?   I thought http methods are pre-defined, they are always upper case.

lei-tang: Done: lowercase method has been removed.

src/envoy/http/jwt_auth/jwt_authenticator.cc
```diff
@@ -49,16 +49,26 @@ void ExtractUriHostPath(const std::string& uri, std::string* host,
...
-                              JwtAuthenticator::Callbacks* callback) {
+void JwtAuthenticator::Verify(HeaderMap &headers,
+                              JwtAuthenticator::Callbacks *callback) {
   headers_ = &headers;
   callback_ = callback;
 
+  // Per the spec
+  // http://www.w3.org/TR/cors/#cross-origin-request-with-preflight-0, CORS
+  // pre-flight requests shouldn't include user credentials.
+  if (headers_->Method() && Http::Headers::get().MethodValues.Options ==
```
lizan: This bypasses too many requests, e.g. WebDAV uses OPTIONS for different purpose and those shouldn't be bypassed.

lei-tang: https://github.com/istio/proxy/pull/2140 is created to add more checks for CORS requests.

PR #2142, opened 2019-03-07
generate another map literal for mixer global dictionary

silentdai:
Global dictionary is need by compressor in mixer. This is supposed to be a const value at compile time. However, currently we only have the forward index as const var but not reverse index. 
This PR avoid the reverse index generation from vector to map. The generation is triggered on each envoy worker thread per listener spawns.
It saves huge cpu and temporary heap usage when a listener is not healthy.

**Which issue this PR fixes** 
Step one for 1.1 istio cpu regresion

an intermediate generated cc file can be verified by running
`bazel build src/istio/mixerclient:global_dictionary_header_gen`

**Release note**:
TBD

### DIFF
--- a/src/istio/mixerclient/attribute_compressor.cc
+++ b/src/istio/mixerclient/attribute_compressor.cc
@@ -159,13 +159,8 @@ class BatchCompressorImpl : public BatchCompressor {
 
 }  // namespace
 
-GlobalDictionary::GlobalDictionary() {
-  const std::vector<std::string>& global_words = GetGlobalWords();
-  for (unsigned int i = 0; i < global_words.size(); i++) {
-    global_dict_[global_words[i]] = i;
-  }
-  top_index_ = global_words.size();
-}
+GlobalDictionary::GlobalDictionary()
+    : global_dict_(GetGlobalDictionary()), top_index_(global_dict_.size()) {}
 
 // Lookup the index, return true if found.
 bool GlobalDictionary::GetIndex(const std::string name, int* index) const {
--- a/src/istio/mixerclient/attribute_compressor.h
+++ b/src/istio/mixerclient/attribute_compressor.h
@@ -38,7 +38,7 @@ class GlobalDictionary {
   int size() const { return top_index_; }
 
  private:
-  std::unordered_map<std::string, int> global_dict_;
+  const std::unordered_map<std::string, int>& global_dict_;
   // the last index of the global dictionary.
   // If mis-matched with server, it will set to base
   int top_index_;
--- a/src/istio/mixerclient/create_global_dictionary.py
+++ b/src/istio/mixerclient/create_global_dictionary.py
@@ -45,25 +45,45 @@ namespace {
  *    bazel-mixerclient/external/mixerapi_git/mixer/v1/global_dictionary.yaml \
  *      > src/global_dictionary.cc
  */
-
-const std::vector<std::string> kGlobalWords{
 """
 
-BOTTOM = r"""};
+INDENT = "    "
 
+FORWORD_INDEX_START = """const std::vector<std::string> kGlobalWords{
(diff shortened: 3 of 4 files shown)


src/istio/mixerclient/attribute_compressor.cc
```diff
@@ -159,7 +159,7 @@ class BatchCompressorImpl : public BatchCompressor {
 
 }  // namespace
 
-GlobalDictionary::GlobalDictionary() {
+GlobalDictionary::GlobalDictionary() : global_dict_(GetGlobalDictionary()) {
   const std::vector<std::string>& global_words = GetGlobalWords();
   for (unsigned int i = 0; i < global_words.size(); i++) {
     global_dict_[global_words[i]] = i;
```
kyessenov: Delete these lines since index is already available?

kyessenov:
Curious: do you have any data on savings? The map is only a few hundred words long, so it's surprising it is that costly (even if it multiplied by num of listeners * workers).

src/istio/mixerclient/attribute_compressor.cc
silentdai: Hmm, my force push only picked up sign off. Will fix!

silentdai:
@kyessenov  Thanks for the early comment! I have a post in istio-networking group addressing the context. My experiment does not reflect the real world work load. So far I cannot provide how much improvement this PR would bring. 

I plan to commit something obviously correct(hopefully) and evaluate the impact at reasonable load.

silentdai:
The failed tsan test is //test/integration:istio_http_integration_test

It's flaky at my devserver. The failing reason is heap use after free. It's not related to my change. I will look into it tomorrow

silentdai:
The flakiness is not high.
`bazel  test  --runs_per_test=20  --config=clang-tsan  -- //test/integration:istio_http_integration_test ` passes

Can any member pull the /retest trigger? Thanks!

mandarjog:
Yuchen, did this piece show up high on the flame graph?

silentdai:
`bazel  test  --runs_per_test=50  --config=clang-tsan  -- //test/integration:istio_http_integration_test`
//test/integration:istio_http_integration_test                           FAILED in 2 out of 50 in 45.9s

Not that bad

silentdai:
Yes it is one of the 3 major blocks.
[Raw flame graph here](https://gist.github.com/silentdai/db042103debb0aa08efa2bc5da0a9ab8)
flame_graph

kyessenov:
@silentdai 
Yes that test flakes but not too bad. 
How do I access this networking mailing list? I thought we all moved to discuss.

silentdai:
@kyessenov Forgive me... I don't know which is the preferred channel. What discuss are you referring to?

kyessenov:
https://discuss.istio.io/ took over WG mailing lists. I'm not sure which mailing list you sent the analysis to.

kyessenov:
Thanks for forwarding. Mixer-dictionary related integration tests are in istio/istio unfortunately. If you want to validate your code, you can run:
```
env ENVOY_PATH=<path-to-bazel-bin/src/envoy/envoy go test -v ./mixer/test/client/...
```

silentdai:
I checkout istio/istio release 1.1 branch and specify newly build envoy 
env ENVOY_PATH=$WORKSPACE/proxy/bazel-out/k8-fastb
uild/genfiles/external/envoy go test -v ./mixer/test/client/.. | grep -v RUN |awk '{print $2}' |sort | uniq -c
``
     50 PASS:
```

duderino:
@silentdai I'd like to keep this out of the 1.1 release unless someone has the bandwidth to do more thorough testing.

It does look like a good thing for release 1.2 since that will get a lot of attention.  Send it there instead?

PR #2144, opened 2019-03-07
Fixes environment-dependent failures in MixerFaultTest

dmitri-d:
Fixes intermittent/environment-dependent failures in MixerFaultTest. Please see https://github.com/istio/proxy/issues/2124.

The failures were due to lambda parameter in ServerCallbackHelper::ServerCallbackHelper going out of scope.

### DIFF
--- a/test/integration/int_server.cc
+++ b/test/integration/int_server.cc
@@ -491,12 +491,9 @@ LocalListenSocket::~LocalListenSocket() {}
 
 ServerCallbackHelper::ServerCallbackHelper(
     ServerRequestCallback request_callback,
-    ServerAcceptCallback accept_callback, ServerCloseCallback close_callback)
-    : accept_callback_(accept_callback),
-      request_callback_(request_callback),
-      close_callback_(close_callback) {
+    ServerAcceptCallback accept_callback, ServerCloseCallback close_callback) {
   if (request_callback) {
-    request_callback_ = [this, &request_callback](
+    request_callback_ = [this, request_callback](
                             ServerConnection &connection, ServerStream &stream,
                             Envoy::Http::HeaderMapPtr request_headers) {
       ++requests_received_;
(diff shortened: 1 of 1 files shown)


dmitri-d:
@duderino: This should fix https://github.com/istio/proxy/issues/2124.

test/integration/int_server.cc
```diff
@@ -496,14 +496,14 @@ ServerCallbackHelper::ServerCallbackHelper(
       request_callback_(request_callback),
       close_callback_(close_callback) {
   if (request_callback) {
-    request_callback_ = [this, &request_callback](
+    instrumented_request_callback_ = [this](
```
duderino: Ah!  Super cool.   I wonder if an equivalent fix would have been to capture the request_callback lambda by value?

duderino:
@dmitri-d thank you so much for fixing my bug.  I really appreciate it.

Technically, I can't merge this until after 1.1 goes out.   https://discuss.istio.io/t/istio-1-1-lockdown-release-status/1275 . Let me talk to the other admins and see if the no merge rule applies to test cases.  Seems like it shouldn't but....

Also, if not merging this for a week or so has any adverse impact for you please let me know.

/lgtm

knrc:
@duderino this shouldn't impact us, we'll apply it locally until it gets merged.

test/integration/int_server.cc
dmitri-d: Didn't think of that, brain == fried. definitely worth a try!

silentdai: Great catch! If you fllow @duderino advice, could you try capture using expression? It might cost less while absolutely never more expensive. Something like
```
request_callback_ = [this, request_callback = std::move(request_callback) ](
```

dmitri-d:
@duderino: updated to pass callbacks to lambdas by value as you suggested.

dmitri-d:
The fix has been merged into master, closing the PR here.

duderino:
It'd be nice to have this in 1.1 too for subsequent 1.1.x releases, so reopening it.  Thanks @dmitri-d

PR #2150, opened 2019-03-12
Building 1.1rc4

duderino:

**Release note**:

```release-note
```

### DIFF
--- a/istio.deps
+++ b/istio.deps
@@ -4,7 +4,7 @@
 		"name": "ISTIO_API",
 		"repoName": "api",
 		"file": "repositories.bzl",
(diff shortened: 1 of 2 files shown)


linsun:
@Joshua Blatt i saw there is two more commits after 0791a…

do you want to pick those up?
they are doc fixes

duderino:
> @joshua Blatt i saw there is two more commits after 0791a…
> 
> do you want to pick those up?
> they are doc fixes

I'd rather not.  @kyessenov  did a selective dep update on istio/istio to pick up only his stable Any hash fixes so the api version i'm pulling into istio/proxy matches the api version in istio/istio from his selective update

PR #2151, opened 2019-03-12
[draft] add context proto

kyessenov:
Introduce context proto for exchanging data between embedded CEL and envoy.

The context is derived from the istio proxy attribute manifest using wrappers for presence detection.

Populate the initial set of values.

/assign @mandarjog 
/assign @PiotrSikora

### DIFF
--- a/src/istio/context/BUILD
+++ b/src/istio/context/BUILD
@@ -17,8 +17,11 @@
 
 package(default_visibility = ["//visibility:public"])
 
+licenses(["notice"])
+
 load(
     "@envoy//bazel:envoy_build_system.bzl",
+    "envoy_cc_library",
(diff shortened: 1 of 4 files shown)


kyessenov:
CI appears to be dead. Any ideas?

mandarjog:
how are we going to use these functions?
These functions should use sandbox API, *not* envoy api directly. 
Sandbox API already has equivalent functions.

kyessenov:
Sure, but is that really a necessity as a first step? All the info objects in envoy are massive, and will need to be mirrored in the API. I think this context should set the right requirements of what needs to be provided by sandbox API. WDYT @PiotrSikora ?

PR #2161, opened 2019-04-03
Update Proxy with newer envoy

mandarjog:
Includes @lizan 's changes from the other PR, but this pins it to a slightly older envoy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,10 @@ CXX := clang++-7
 endif
 PATH := /usr/lib/llvm-7/bin:$(PATH)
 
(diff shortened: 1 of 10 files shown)


PiotrSikora:
This rolls forward proxy in 1.1 by 200 commits, we didn't do that for 1.0. What's the justification here?

mandarjog:
@PiotrSikora we have abandoned it for now.

However we will have to revisit the issue about if we want to be near the tip of envoy master branch when we are required to pickup envoy security fixes.

At some point Istio will start only using `released` versions of envoy.

PR #2162, opened 2019-04-05
Makefile: allow specify BAZEL binary

silentdai:
Also fix the target in make targe `build_envoy`

Usage:
`BAZEL_BIN=~/bazel0.16/bin/bazel make build_envoy`

My current bazel 0.22 cannot build proxy 1.0 any more. I choose to install bazel 0.16 and fix the Makefile to opt the bazel. See the usage above. 
Also enable the `make build_envoy`

If you have a default old bazel, you can use it as long as you specify BAZE_BIN

**Release note**:

```release-note
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 ## limitations under the License.
 
 TOP := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
-
+BAZEL_BIN ?= bazel
 SHELL := /bin/bash
 LOCAL_ARTIFACTS_DIR ?= $(abspath artifacts)
 ARTIFACTS_DIR ?= $(LOCAL_ARTIFACTS_DIR)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,23 +24,23 @@ HUB ?=
 TAG ?=
 
 build:
-	@bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) //...
+	@${BAZEL_BIN} $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) //...
 
 # Build only envoy - fast
 build_envoy:
-	bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) //src/envoy/mixer:envoy
+	${BAZEL_BIN} $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) //src/envoy/tcp/mixer:filter_lib //src/envoy/http/mixer:filter_lib
```
duderino: Why did you narrow the build target here?

silentdai: At the moment I proposed this change the target `//src/envoy/mixer:envoy` doesn't exist anymore. At that moment I assume it's better something than nothing.
I explored and I think the goal is to build an envoy binary. Will fix

silentdai:
Too many conflicts and there is no value to merge into 1.0 as the active are 1.2, 1.1 and master.
Abandon this one and will review.

PR #2158, opened 2019-03-28
update envoy to latest

lizan:
Signed-off-by: Lizan Zhou <lizan@tetrate.io>

**What this PR does / why we need it**:
To pick up TLS logging, fault injection controlled by header.

**Special notes for your reviewer**:

**Release note**:

```release-note
None
```

PR #2166, opened 2019-04-10
release 1.1: Update envoy to enable path normalization

silentdai:
On top of #2158 

And introduce the path normalization.

- Update envoy dependency to 1.10  at 04/10/2019
- Update .bazelrc to adopt the compilation option to enable the path normalization by default
- Update WORKSPACE so that it support envoy build from http github repo and local envoy. The latter is heavily used by debugging.

Q: How to build or test the path normalization?
A: Any make/bazel command should pickup the compilation option automatically. See .bazelrc

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -32,11 +32,15 @@ bind(
     actual = "//external:ssl",
 )
 
+# envoy commit date 04/10/2019
+# bazel version: 0.22.0
+
 # When updating envoy sha manually please update the sha in istio.deps file also
 #
 # Determine SHA256 `wget https://github.com/envoyproxy/envoy/archive/COMMIT.tar.gz && sha256sum COMMIT.tar.gz`
-ENVOY_SHA = "805683f835bd63e4b7b9d89059aa0d3783924a93"
-ENVOY_SHA256 = "75a029fb3904c17f47c7f723e2a04468bfc007bb4cfc74fe21f82cf799d8a904"
+ENVOY_SHA = "ac7aa5ac8a815e5277b4d4659c5c02145fa1d56f"
+ENVOY_SHA256 = "3f13facc893ef0c5063c7391a1ffca8de0f52425c8a7a49ef45e69dbb5e7304b"
+LOCAL_ENVOY_PROJECT = "/PATH/TO/ENVOY"
 
 http_archive(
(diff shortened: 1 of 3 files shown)


WORKSPACE
```diff
@@ -32,11 +32,15 @@ bind(
     actual = "//external:ssl",
 )
 
+# envoy commit date 04/10/2019
+# bazel version: 0.22.0
```
duderino: nice touch

WORKSPACE
```diff
@@ -45,8 +49,15 @@ http_archive(
...
 )
 
-load("@envoy//bazel:repositories.bzl", "GO_VERSION", "envoy_dependencies")
+#local_repository(
+#     name = "envoy",
+#     path = LOCAL_ENVOY_PROJECT
+#)
 
+load("@envoy//bazel:api_repositories.bzl", "envoy_api_dependencies")
+envoy_api_dependencies()
```
duderino: why did you have to add this?

WORKSPACE
```diff
@@ -45,8 +49,15 @@ http_archive(
     sha256 = ENVOY_SHA256,
 )
 
-load("@envoy//bazel:repositories.bzl", "GO_VERSION", "envoy_dependencies")
+#local_repository(
+#     name = "envoy",
+#     path = LOCAL_ENVOY_PROJECT
+#)
```
duderino: did stuff for building against a local envoy slip in here?   I think it's fine to keep it here commented out, but if you do, please add a few comments describing what it's for

WORKSPACE
silentdai: Envoy is splitting /api to external repository. It's WIP and I am seeing error when I introduce envoy using local_repository. 

With this fix, proxy build works perfectly no matter we introduce envoy by http or by local_repository.

WORKSPACE
silentdai: Leave here because I probably pasted these lines 10 times in the last 2 weeks... I haven't figure out if we can use bazel arg to switch between local or http_archive
Let me add the comment.

WORKSPACE
silentdai: It might be another pain if we forgot to update the comment... We will see.

PR #2179, opened 2019-04-18
Update envoy sha and fix bulid break

### LINKED ISSUE #2177: Update istio authn filter to align with changes in Envoy
https://github.com/envoyproxy/envoy/pull/6144 Did some large refactor to the SSL struct in Envoy which will break the Isio authentication filter.

- the `Ssl::Connection` is renamed to `Ssl::ConnectionInfo`
- the `uriSanPeerCertificate()` is changed to return `std::vector<std::string>` from `std::string`

/cc @snowp @lizan A quick question, there is a `TODO(snowp): Remove this in favor of StreamInfo::downstreamSslConnection.` in the original envoy PR, I'm wondering is this going to happen soon? I would prefer to resolve the conflicts in one PR in Istio side.

yangminzhu:
There are a couple of breaking changes in upstream Envoy, they are resolved as following:
- Change to use `Ssl::ConnectionInfo` from `Ssl::Connection` as renamed in https://github.com/envoyproxy/envoy/pull/6144

- Change to return the first item of `uriSanPeerCertificate()` and `uriSanLocalCertificate()` as it now returns a vector, changed in https://github.com/envoyproxy/envoy/pull/6144

- Include `envoy_api_dependencies` in WORKSPACE as introduced in https://github.com/envoyproxy/envoy/pull/6229

- Use `getStringView()` to replace `c_str()` in `HeaderString` as removed in https://github.com/envoyproxy/envoy/pull/6564

- Pass `Envoy::Filesystem::fileSystemForTest()` to `Envoy::Api::Impl` as introduced in https://github.com/envoyproxy/envoy/pull/6072

Fixes #2177

@kyessenov for the mixer changes.


yangminzhu:
The tests on prow failed to build, probably due to the gcc version (5.4.0) is too low (we use gcc-6 in circle-ci).

`
W0418 00:24:06.754] /usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/ext/new_allocator.h:120:23: error: no matching constructor for initialization of 'std::pair<Envoy::Http::LowerCaseString, std::__cxx11::basic_string<char> >'
`

lizan:
LGTM, waiting for prow toolchains update

PR #2180, opened 2019-04-22
fixing a comment that is out of sync with the code

huang195:
Comment is out of sync with the code. Fixing the comment to be not confusing


huang195:
duplicate. Closing

PR #2182, opened 2019-04-24
Retrieve tracing header from request headers during report

bianpengyuan:
Fixes https://github.com/istio/istio/issues/13391

Even though tracing happens before all filters inside envoy (https://github.com/envoyproxy/envoy/blob/master/source/common/http/conn_manager_impl.cc#L735), the tracing header injection actually happens at envoy.router filter (https://github.com/envoyproxy/envoy/blob/master/source/common/router/router.cc#L1218).

The current behavior of mixer filter is that it extracts request headers once at `decodeHeaders` callback during check time. However request header is not yet ready by that time. This PR extracts b3 tracing headers at report time and add them into request.header attribute.

### DIFF
--- a/src/istio/control/http/attributes_builder_test.cc
+++ b/src/istio/control/http/attributes_builder_test.cc
@@ -15,15 +15,15 @@
 
 #include "src/istio/control/http/attributes_builder.h"
 
+#include "gmock/gmock.h"
 #include "google/protobuf/stubs/status.h"
 #include "google/protobuf/text_format.h"
 #include "google/protobuf/util/message_differencer.h"
+#include "gtest/gtest.h"
 #include "include/istio/utils/attribute_names.h"
 #include "include/istio/utils/attributes_builder.h"
 #include "src/istio/control/http/mock_check_data.h"
 #include "src/istio/control/http/mock_report_data.h"
-#include "gmock/gmock.h"
-#include "gtest/gtest.h"
 
 using ::google::protobuf::TextFormat;
 using ::google::protobuf::util::MessageDifferencer;
--- a/src/istio/control/http/attributes_builder_test.cc
+++ b/src/istio/control/http/attributes_builder_test.cc
@@ -15,15 +15,15 @@
 
 #include "src/istio/control/http/attributes_builder.h"
 
-#include "gmock/gmock.h"
 #include "google/protobuf/stubs/status.h"
 #include "google/protobuf/text_format.h"
 #include "google/protobuf/util/message_differencer.h"
-#include "gtest/gtest.h"
 #include "include/istio/utils/attribute_names.h"
 #include "include/istio/utils/attributes_builder.h"
 #include "src/istio/control/http/mock_check_data.h"
 #include "src/istio/control/http/mock_report_data.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
 
 using ::google::protobuf::TextFormat;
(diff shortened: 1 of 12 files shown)


bianpengyuan:
@kyessenov @JimmyCYJ ptal thanks!

src/istio/control/http/attributes_builder_test.cc
```diff
@@ -765,6 +780,11 @@ TEST(AttributesBuilderTest, TestReportAttributes) {
         map["server"] = "my-server";
         return map;
       }));
+  EXPECT_CALL(mock_data, GetTracingHeaders(_))
+      .WillOnce(Invoke([](std::map<std::string, std::string> &m) {
```
mandarjog: This is testing that if getTracingHeaders works correctly, the correct attribute will be built.
Is there an integration test that sets tracing to 100% and does not mock out the rest?

bianpengyuan: No, I don't think such a integration test framework exist in istio-proxy repo right? My understanding is that all istio-proxy related integration test needs to be written in istio/istio repo. So we need to add a test there with SHA update.

mandarjog: Take a look at test/integration/istio_http_integration_test.cc

bianpengyuan:
@duderino Could you help to merge this? This is a bug fix for mixer tracing adapter due to a behavior change in envoy. Mixer adapter relies on b3 headers to instrument spans. Mixer client assumes request headers are all ready at check time so it only processes request header once. However envoy injects those tracing headers at envoy.router filter, which causes mixer filter missing tracing headers, or mistakenly using b3 headers for downstream span to instrument upstream span. This PR adds functionality to extract tracing headers at report time, an integration test has already been added. I will add another test in istio/istio with sha update. Thanks!

bianpengyuan:
@duderino integration test is in a pending PR now: https://github.com/istio/istio/pull/13671

bianpengyuan:
@duderino gentle ping. Let me know if this should be merged to master.

kyessenov:
@bianpengyuan I think we need to manually cherry pick this PR to master and link both PRs to the underlying issue.

douglas-reid:
Can we merge this PR?

silencehe09:
> Can we merge this PR?
When will this PR be released ?

silencehe09:
> Can we merge this PR?

When will this PR be released ?

douglas-reid:
@duderino many thanks!

@silencehe09 hopefully a subsequent 1.1.X series of releases will include this change.

@bianpengyuan can you track the use of this in istio/istio and merge up into `master` ?

bianpengyuan:
@douglas-reid https://github.com/istio/istio/pull/13671 pick this change into istio/istio and https://github.com/istio/proxy/pull/2197 picks the change into istio/proxy master.

zhongfox:
hi @douglas-reid, seems we encounter the same issue described in issue 
https://github.com/istio/istio/issues/13391, we are using istio 1.1.7

when and which istio version may merge this pr?
thanks.

douglas-reid:
@fpesce @duderino can you provide any insights on 1.1.X and 1.2 release plans that would include this PR ?

bianpengyuan:
@douglas-reid @duderino @fpesce  The change is already in 1.2. Question is do we still want to cherry pick this into 1.1 and do a 1.1.8 release or we should just wait for 1.2.

PR #1914, opened 2018-08-14
Update clang to 6.0 and use it for release binaries.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

PR #2203, opened 2019-05-07
fixing issue with bazel shutdown causing circleci hung

huang195:
`bazel shutdown` is causing some clean up problems in circleci that's blocking CI.


huang195:
macos CI check is somewhat different from linux version, which uses a fixed 0.22 version of bazel. macos relies on brew which could install newer version of bazel. At the moment, it install 0.25 version of bazel.

huang195:
for macos, the flag `--incompatible_bzl_disallow_load_after_statement=false` is a wet bandaid that will eventually have to fall off. Its default is already to `true`. In the very near future, the flag will be removed, and all bzl files need to be modified to be compliant with this new check.

huang195:
@PiotrSikora @lizan @qiwzhang can you help to take a look at this PR? I'm looking to revert some of the changes from this PR (https://github.com/istio/proxy/pull/1914) as it's causing CircleCI to consistently hang. I'm at a loss with macos as brew is installing bazel 0.25 and it's introducing quite a bit of incompatibility issues that older version of bazel 0.22 that we are using for linux does not have.

huang195:
FYI, after digging around on github on recently merged PRs onto the master branch, I see that the last time macos CI ran successfully was when brew installed 0.23 version of bazel :(

lizan:
add `--incompatible_bzl_disallow_load_after_statement=false` to mac bazel CLI should fix mac CI

huang195:
@lizan yeah I tried to add the flag, but the build hit other problems, so I removed the flag. See this circleci build that had this flag set to `false`: https://circleci.com/gh/istio/proxy/6151:

```
ERROR: /Users/distiller/.cache/bazel/external/envoy/bazel/BUILD:113:1: error loading package '@com_github_grpc_grpc//src/proto/grpc/health/v1': in /Users/distiller/.cache/bazel/external/com_github_grpc_grpc/bazel/grpc_build_system.bzl: Label '@com_github_grpc_grpc//:bazel/cc_grpc_library.bzl' crosses boundary of subpackage '@com_github_grpc_grpc//bazel' (perhaps you meant to put the colon here: '@com_github_grpc_grpc//bazel:cc_grpc_library.bzl'?) and referenced by '@envoy//bazel:grpc_health_proto'
ERROR: Analysis of target '//src/envoy:envoy' failed; build aborted: Analysis failed
```
Any idea what could be causing this problem?

.bazelrc
```diff
@@ -19,6 +19,7 @@ build:asan --define tcmalloc=disabled
 build:asan --build_tag_filters=-no_asan
 build:asan --test_tag_filters=-no_asan
 build:asan --define signal_trace=disabled
+build:asan --test_env=ASAN_OPTIONS=detect_odr_violation=0
```
lizan: use `detect_odr_violation=1` I don't think asan will fail with 1, but will fail with 2 (which is default)

huang195: asan is still reporting errors with `detect_odr_violation=0`. I'll try to set it to 1 and see if it passes.

huang195:
@lizan I feel like going down a rabbit hole chasing down all the dependencies that bazel 0.25 on macos is causing. I'm going to open another PR that fixes all the other CI cases except macos, and close this one. As it's not one of the "required" CIs, hopefully, it will at least not block all the PRs that would otherwise succeed.

PR #2209, opened 2019-05-10
Resolve CI failing: remove bazel shutdown command

huang195:
@lizan this is the simpler version of the PR here: https://github.com/istio/proxy/pull/2203.
This should at least pass all the `required` CI use cases. /ptal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,30 +33,25 @@ CXX := clang++-7
 endif
 PATH := /usr/lib/llvm-7/bin:$(PATH)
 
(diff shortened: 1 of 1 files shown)


duderino:
@huang195 @crazyxy @lizan as far as I can tell, this PR broke the macos build.

huang195:
@duderino Getting rid of bazel shutdown fixed the timeout problem, but also exposed another problem that's caused by a newer version of bazel (0.25?) that's pulled in by brew. You probably need to pull a subset of changes from this PR to completely fix macos.

PR #2230, opened 2019-05-22
Make proxy reference envoy-wasm

bianpengyuan:
Try out building istio-proxy with envoyproxy/envoy-wasm repo in extension branch.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -37,17 +37,17 @@ bind(
 # Determine SHA256 `wget https://github.com/envoyproxy/envoy/archive/COMMIT.tar.gz && sha256sum COMMIT.tar.gz`
 # envoy commit date  05/16/2019
 # bazel version: 0.25.0
-ENVOY_SHA = "829b905ca0fdc85233c3969247e53a62a52ac627"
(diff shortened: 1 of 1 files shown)


kyessenov:
Should we just use master? It's painful to set-up test infra for each branch.

bianpengyuan:
@kyessenov This is the master of envoy/envoy-wasm repo. Are you referring other repo?

kyessenov:
@bianpengyuan I was referring to the target branch. It looks like you want to merge into istio:extension. This is a new branch that has no test-infra set-up, which mean I have a green merge button available to me without test checks.

bianpengyuan:
@kyessenov I do see circle test running for the branch, though it is not blocking the change. The only concern is envoy-wasm might not be as stable (I don't think there are tech infra setup at all in that repo). But considering the code change made there are all in filters scope, so maybe it is fine. cc @mandarjog

bianpengyuan:
We need to test out how long the build is going run on prow. Since right now we are building from envoy-wasm which builds a v8 engine and is very slow, we might not want to make master slow for changes not related to new extensions.

There is already a wasm branch, we should just use that one that it has already had test-infra set up.

PR #2236, opened 2019-05-25
Merge branch 'master' into wasm

bianpengyuan:


bianpengyuan:
looks like the last batch of merge squashed all commits and make commit log different. It causes conflict when merging master to this branch and will cause conflict when merging wasm back to master.

Seems like there is not change yet made to wasm branch. There is only one PR from @gargnupur open against this branch. So Nupur, I am wondering  are you ok with deleting and recreating wasm branch? The PR will get closed when branch deleted but it could be reopened after recreation.

gargnupur:
@bianpengyuan : Sure, checking if I can delete branch wasm..

bianpengyuan:
Close as we prefer to use master branch.

PR #2246, opened 2019-05-29
remove bazel shutdown command (#2209)

crazyxy:
Cherry-pick #2209


huang195:
@duderino Getting rid of `bazel shutdown` fixed the timeout problem, but also exposed another problem that's caused by a newer version of bazel (0.25?) that's pulled in by brew. You probably need to pull a subset of changes from this PR to completely fix macos.

PR #2249, opened 2019-05-30
Remove strip for release build

crazyxy:
**What this PR does / why we need it**: Improve ability to debug Envoy in prod and compile release Envoy with symbols

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -60,12 +60,13 @@ test --test_env=HEAPCHECK=normal --test_env=PPROF_PATH
 # enable path normalization by default. See https://github.com/envoyproxy/envoy/pull/6519
 test --define path_normalization_by_default=true
 
-# Release builds without debug symbols.
+# Release builds with debug symbols
 build:release -c opt
-build:release --strip=always
+build:release --copt -g
 
 # Release builds with debug symbols
 build:release-symbol -c opt
+build:release-symbol --copt -g
 
 # Add compile option for all C++ files
 build --cxxopt -Wnon-virtual-dtor


crazyxy:
> What are the implications on the proxy performance?

AFAIK, exposing debug symbols won't degrade the proxy performance but it will consume more disk space. @mandarjog WDYT?

.bazelrc
```diff
@@ -60,12 +60,13 @@ test --test_env=HEAPCHECK=normal --test_env=PPROF_PATH
...
 
-# Release builds without debug symbols.
+# Release builds with debug symbols
 build:release -c opt
-build:release --strip=always
+build:release --copt -g
 
 # Release builds with debug symbols
 build:release-symbol -c opt
+build:release-symbol --copt -g
```
PiotrSikora: Doesn't this effectively make `release` and `release-symbol` the same?

We're already building *both* of them as part of release process, see: https://github.com/istio/proxy/blob/ac78dc0193284534bea9e9551f784ee6d33494b2/scripts/release-binary.sh#L62
https://github.com/istio/proxy/blob/ac78dc0193284534bea9e9551f784ee6d33494b2/scripts/release-binary.sh#L75

crazyxy: The bazel build document says that 

>opt means build with optimization enabled and with assert() calls disabled (-O2 -DNDEBUG). Debugging information will not be generated in opt mode unless you also pass --copt -g.

https://docs.bazel.build/versions/master/user-manual.html I think bazel document has some mistake. I tested the release binary and gdb can load symbols from it. I will close this pr.

PiotrSikora: Note: I was not complaining about adding `--copt -g`, but about making `release` and `release-symbols` configs exactly the same (by removing `--strip=always`).

crazyxy: Sorry for misreading. I should remove one of them after removing `--strip=always`. I thought `release` and `release-symbols` could be referred in multiple places so I kept both of them. I would send another pr to cleanup the redundancy. But anyway this pr is no longer needed.

mandarjog:
No implications on data path performance, but the size of the image is considerably larger.

PR #2251, opened 2019-05-31
Import common files into this repo.

geeknoid:


kyessenov:
I think we need to organize common files by language better. There is no need to import all these go tools here, but we also need to export c++ tools to commons.

geeknoid:
"all these go tools" is one config file and two small scripts. I don't see the need for any segregation by language at this point, it's not worth the cost. We can just augment the common-files repo with C++-specific stuff alongside the Go files.

kyessenov:
Are you OK with putting bazel stuff into common files then?

geeknoid:
The only thing in common-files should be things that are shared between
repos. If the bazel stuff is specific to the proxy repo only, then it
doesn't need to be in the common space.

kyessenov:
Fair enough. Let's wait if we need other repos to host C++ extensions, and then pull the trigger. It would be nice to share some bazel definitions.

PR #2257, opened 2019-06-03
Build 1.1.8

duderino:
Switch from envoyproxy/envoy to istio/envoy repo to pick up three commits from jplev related to reducing 503s from connection reuse.

See https://github.com/envoyproxy/envoy/compare/ac7aa5ac8a815e5277b4d4659c5c02145fa1d56f...istio:8193a430e75f69dc2bc6fa1ddca5976f5853c954 for the diff.


duderino:
macos is a known failure

PR #2258, opened 2019-06-03
Add opencensus library

bianpengyuan:
This will be used by stackdriver extension. It is a copy of opencensus-library (https://github.com/census-instrumentation/opencensus-cpp) with modification to make it fit with envoy's silo thread model and wasm API. Test is not included in this PR since code is mostly unchanged and should remain unchanged. Exporter is not included in this PR and will be added in following PR with tests.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -82,3 +82,16 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_depe
 go_rules_dependencies()
 
 go_register_toolchains(go_version = GO_VERSION)
+
+# TODO(bianpengyuan): remove this googleapis dep when upstream imports it in
+# https://github.com/envoyproxy/envoy/pull/5387
+load("//extensions/stackdriver/opencensus:opencensus.bzl", "telemetry_googleapis")
+
+telemetry_googleapis()
+
+load("@telemetry_googleapis//:repository_rules.bzl", "switched_rules_by_language")
+
+switched_rules_by_language(
+    name = "com_google_googleapis_imports",
+    cc = True,
+)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -83,7 +83,7 @@ go_rules_dependencies()
 
 go_register_toolchains(go_version = GO_VERSION)
 
-# TODO(bianpengyuan): remove this googleapis dep when upstream imports it in
+# TODO(bianpengyuan): remove this googleapis dep when upstream envoy-wasm repo merges
 # https://github.com/envoyproxy/envoy/pull/5387
 load("//extensions/stackdriver/opencensus:opencensus.bzl", "telemetry_googleapis")
 
--- a/extensions/stackdriver/opencensus/BUILD
+++ b/extensions/stackdriver/opencensus/BUILD
@@ -1,0 +1,15 @@
+# Copyright 2018, OpenCensus Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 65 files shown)


silentdai:
It's a pain to introduce the code text. You may want to explain
1. what the changes you made to the original code
2. why we cannot import opencensus-cpp project as it support native bazel build
3. How to update opencensus-cpp in the future

silentdai:
Maybe you don't need this PR any more: https://github.com/envoyproxy/envoy/pull/5387

kyessenov:
@bianpengyuan can you delineate the upstream vs your changes in some straightforward fashion?
We don't need to review the upstream code, just the changes you made.

@silentdai The native extension could be used once it is exposed through WASM API. Right now it is not, so we can't access its capabilities from WASM extensions.

silentdai:
@kyessenov @bianpengyuan  I am not arguing that there is no reference to the opencensus code. It's common pattern that introduce first and use in the future :)

I am more interested in the changes @bianpengyuan made to opencensus.

PiotrSikora:
@bianpengyuan are you going to upstream your changes or are we going to maintain this in-tree fork going forward?

I'm not sure how big are your changes, but perhaps you could import this library as Bazel external with patches applied locally?

bianpengyuan:
@PiotrSikora I think we are going to maintain this in-tree fork. Change is not trivial and touch whole bunch of files because opencensus library built to support multi-thread process with singleton objects, so patching is not quite ideally. In fact the whole stackdriver exporter needs to be customized to use WASM API.

The code within this PR should just remain unchanged over time unless we found bug or performance bottleneck in it. The stats part of opencensus library is pretty stable and change very little. The logic there is not very complicated so it is feasible to maintain our own fork. If one day opencensus integrates with envoy's stats system, for sure we switch from this in tree fork to that.

bianpengyuan:
@silentdai @kyessenov Sorry for the delay!

@silentdai I added a bit more text to read me about changes made to the library. As I said in the last comment it should be fine to maintain our own fork. It is like implementing our own adapter like the one in mixer.

@kyessenov  Code change that needs review are
* all build files, which are changed to use envoy rules.
* The three files that import WASM API: https://github.com/istio/proxy/pull/2258/files#diff-410dc608d480b030c986bc084e2f99c2R31, https://github.com/istio/proxy/pull/2258/files#diff-410dc608d480b030c986bc084e2f99c2, and https://github.com/istio/proxy/pull/2258/files#diff-f03a8ec4119feb871dcfdd671540fa09. Those files use absl::clock to get time now, which introduces whole bunch of clock system calls which is not supported by emscripten ABI. So a clock WASM API (`getCurrentTimeNanoseconds`) is used. As a result of that all absl::time objects are modified to uint64_t.
* Exposing a Flush method here: https://github.com/istio/proxy/pull/2258/files#diff-fee52e53cbaa8d888cc55e127608b38aR37. It was originally a private method called by an exporter thread. It is exposed here to make it callable in WASM context.

kyessenov:
I find it hard to review this. I suggest to follow @PiotrSikora and structure it as a bazel fetch plus patch application.

1. You don't really have to use envoy rules to build envoy stuff. It might be worth filing an issue in opencensus to define a macro instead of using bazel rules directly, to accommodate cases when there is a need to customize build environment.

2. This should be another issue to provide macros in opencensus to stub out clock system calls, or provide alternative macro definitions. That's a reasonable request I think.

3. Bazel patch should be fairly simple for changing visibility.

bianpengyuan:
@kyessenov Ack, let me try it out and see if it is feasible.

PR #2214, opened 2019-05-14
Update Envoy SHA to latest.

jplevyak:

PR #2259, opened 2019-06-03
Build 1.1.8 part 2

duderino:
Update istio/envoy sha to pick up https://github.com/envoyproxy/envoy//pull/6994 and https://github.com/istio/envoy/pull/76


nrjpoddar:
@duderino we noticed that the macos build is always failing in release-1.1 branch. I think you need to cheery-pick parts or whole of this PR https://github.com/istio/proxy/pull/2214 to make this build with pinned version of bazel.

PR #2262, opened 2019-06-07
Add a simple setup for testing communication between 2 envoys

gargnupur:
This setup starts one echo app and one client envoy and one server envoy.


test/envoye2e/env/var.go
```diff
@@ -0,0 +1,302 @@
+// Copyright 2019 Istio Authors
```
kyessenov: Remove this. You can inline any references to the functions here.

gargnupur: Removed

PR #2268, opened 2019-06-12
Switch from envoyproxy/envoy to istio/envoy before shipping Istio 1.2

duderino:
We will cherry pick into istio/envoy as necessary and we will do a full refresh of Envoy again for 1.3.

@fpesce I'm just switching to our istio/envoy fork for this one.  I will follow up with a few cherry picks from envoyproxy/envoy that we need for the 1.2 release.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -47,7 +47,7 @@ http_archive(
     name = "envoy",
     sha256 = ENVOY_SHA256,
     strip_prefix = "envoy-" + ENVOY_SHA,
-    url = "https://github.com/envoyproxy/envoy/archive/" + ENVOY_SHA + ".tar.gz",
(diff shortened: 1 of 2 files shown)


fpesce:
@duderino I don't know the state of istio/proxy in terms of branch/master divergence.
Would you mind making the change on master and then cherrypicking to 1.2?

duderino:
@fpesce and I are having a discussion about whether istio/proxy:master should point to envoyproxy/envoy or istio/envoy

fpesce:
I'm merging that, let's have a discussion with Envoy resident dev from Istio about the best way to create a permanent 5-days old envoyproxy/envoy master dependencies on istio/proxy master

PR #2270, opened 2019-06-13
Report StopIteration if connection is closed

crazyxy:
**What this PR does / why we need it**: Fix segment fault.

**Which issue this PR fixes** : fixes https://github.com/istio/istio/issues/14707

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/tcp/mixer/filter.cc
+++ b/src/envoy/tcp/mixer/filter.cc
@@ -168,12 +168,14 @@ void Filter::completeCheck(const CheckResponseInfo &info) {
 // Network::ConnectionCallbacks
 void Filter::onEvent(Network::ConnectionEvent event) {
   if (filter_callbacks_->upstreamHost()) {
-    ENVOY_LOG(debug, "Called tcp filter onEvent: {} upstream {}",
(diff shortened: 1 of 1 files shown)


crazyxy:
> This seems a envoy bug, can you file an issue in envoy and add a TODO saying this is a workaround for that?

I also made the change for upstream Envoy. I am writing test for the change and will send the pr later.

src/envoy/tcp/mixer/filter.cc
```diff
@@ -167,6 +169,7 @@ void Filter::completeCheck(const CheckResponseInfo &info) {
 void Filter::onEvent(Network::ConnectionEvent event) {
   if (filter_callbacks_->upstreamHost()) {
     ENVOY_LOG(debug, "Called tcp filter onEvent: {} upstream {}",
+              filter_callbacks_->connection(),
```
duderino: I think you need to add a third {} to the debug log

crazyxy: Good catch. I think `ENVOY_CONN_LOG` should be used here.

bianpengyuan:
@crazyxy just FYI I am investigating prow build failure.

PR #2281, opened 2019-06-17
Set Istio authn filter to prefer using Envoy jwt filter if found

pitlv2109:
This PR allows Istio authn filter to prefer using Envoy jwt filter over Istio jwt filter if found. This is the first step to use Envoy jwt filter in Istio.

As for the integration test, it is exact the same as one in `istio_http_integration_test.cc`, except this test uses Envoy jwt filter, rather than Istio jwt filter. Later on if we deprecate Istio jwt filter, we can just remove one of these two files. 

Since Pilot is not yet configured to convert jwt policy to Envoy jwt config, this PR should have very little to none risk.


src/envoy/http/authn/filter_context.cc
```diff
@@ -74,24 +74,47 @@ void FilterContext::setPrincipal(const iaapi::PrincipalBinding& binding) {
...
+    ENVOY_LOG(debug, "Now try fetching dynamic_metadata for filter {}",
+             Utils::IstioFilterName::kJwt);
+    filter_it = dynamic_metadata_.filter_metadata().find(Utils::IstioFilterName::kJwt);
+    if (filter_it == dynamic_metadata_.filter_metadata().end()) {
+      ENVOY_LOG(debug, "No dynamic_metadata found for filter {}",
+                Utils::IstioFilterName::kJwt);
+      return false;
+    }
+  } else {
+    isUsingEnvoyJwtFilter = true;
```
diemtvu: It's a bit hard to read. I suggest to have a separate getJwtPayload that deal with only upstream JWT filter, and one with Istio Jwt (i.e this one, but rename). This function then can simply be

```
bool FilterContext::getJwtPayload(const std::string& issuer,
                                  std::string* payload) const {	                                  std::string* payload) const {
  return getJwtPayloadFromUpstreamJwt(issuer, payload) || getJwtPayloadFromIstioJwt(...);
}
```

test/integration/istio_http_integration_test_with_envoy_jwt_filter.cc
```diff
@@ -0,0 +1,563 @@
...
+  return fmt::sprintf(kMixerFilterTemplate, kSourceUID, kTelemetryBackend,
+                      kPolicyBackend);
+}
+
+
+// checkJwtRawClaim finds the serialized jwt payload and check to see if all key-value pairs
+// from the jwt claim is there.
+// Returns false if it cannot find the serialized jwt payload or if any key-value pair does not
+// matched the expectation.
+bool checkJwtRawClaim(const ::google::protobuf::RepeatedPtrField<::std::string>& words) {
```
diemtvu: It's ok for now, but may be good to know why upstream JWT filter change the order of claims in the raw payload.

+ @qiwzhang Do you have any insight?

src/envoy/http/authn/filter_context.cc
```diff
@@ -74,24 +74,47 @@ void FilterContext::setPrincipal(const iaapi::PrincipalBinding& binding) {
 
 bool FilterContext::getJwtPayload(const std::string& issuer,
                                   std::string* payload) const {
-  const auto filter_it =
-      dynamic_metadata_.filter_metadata().find(Utils::IstioFilterName::kJwt);
+  // Try getting the Jwt payload from Envoy jwt_authn filter.
+  // If not found, try getting the Jwt payload from Istio jwt-auth filter.
+  bool isUsingEnvoyJwtFilter = false;
```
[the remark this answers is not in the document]
pitlv2109: Sure. Thanks!

src/envoy/http/authn/filter_context.cc
pitlv2109: Sounds good. Thanks!

test/integration/istio_http_integration_test_with_envoy_jwt_filter.cc
pitlv2109: I emailed Wayne this morning but he's OOO until next week.

test/integration/istio_http_integration_test_with_envoy_jwt_filter.cc
```diff
@@ -0,0 +1,563 @@
...
+        payload_in_metadata: testing@secure.istio.io
+      testing-rbac:
+        issuer: testing-rbac@secure.istio.io
+        local_jwks:
+          inline_string: "%s"
+        payload_in_metadata: testing-rbac@secure.istio.io
+    rules: 
+    - match:
+        prefix: /
+      requires:
```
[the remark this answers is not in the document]
pitlv2109: This looks much cleaner. Thanks. I think before it was a copy-paste issue from my part.

src/envoy/http/authn/filter_context.cc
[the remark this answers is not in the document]
pitlv2109: Done. PTAL.

src/envoy/http/authn/filter_context.cc
pitlv2109: Done. PTAL.

src/envoy/http/authn/filter_context.cc
```diff
@@ -74,7 +74,42 @@ void FilterContext::setPrincipal(const iaapi::PrincipalBinding& binding) {
...
+  if (entry_it == data_struct.fields().end()) {
+    return false;
+  }
+
+  if (entry_it->second.struct_value().fields().empty()) {
+    return false;
+  }
+
+  // Serialize the payload from Envoy jwt filter first before writing it to |payload|.
+  Protobuf::util::MessageToJsonString(entry_it->second.struct_value(), payload);
```
yangminzhu: I think we don't need to convert it back to string once the istio jwt filter is completely removed, which should improve the performance and also simplify the code. Could you put a TODO as a reminder?

diemtvu: It depends whether we want to use a structured proto, i.e payload, or the proto struct directly in the next step. Either way, I agree this can be revised later.

pitlv2109: Didn't see @diemtvu's comment, so I added a TODO anyway. I think we can revise later.

src/envoy/http/authn/filter_context.cc
```diff
@@ -74,7 +74,42 @@ void FilterContext::setPrincipal(const iaapi::PrincipalBinding& binding) {
 
 bool FilterContext::getJwtPayload(const std::string& issuer,
                                   std::string* payload) const {
-  const auto filter_it =
+  return getJwtPayloadFromEnvoyJwtFilter(issuer, payload) ||
```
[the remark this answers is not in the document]
pitlv2109: Not sure what a short-circuit intention is :D but I add a line saying we prefer using payload from Envoy jwt filter.

JimmyCYJ:
Please add tests into istio/mixer/tests/client/ to cover the new filter. Thanks.

yangminzhu:
For https://github.com/istio/istio/issues/9838

PR #2285, opened 2019-06-24
Add TCP Flow test

gargnupur:
Add support for TCP server and a tcp flow test
Also added support to add new filters in http filters

### DIFF
--- a/test/envoye2e/env/tcp_envoy_conf.go
+++ b/test/envoye2e/env/tcp_envoy_conf.go
@@ -26,17 +26,27 @@ static_resources:
   - name: client
     connect_timeout: 5s
     type: STATIC
-    hosts:
-    - socket_address:
-        address: 127.0.0.1
-        port_value: {{.Ports.ClientToServerProxyPort}}
+    load_assignment:
+      cluster_name: client
+      endpoints:
+      - lb_endpoints:
+        - endpoint:
+            address:
+              socket_address:
+                address: 127.0.0.1
+                port_value: {{.Ports.ClientToServerProxyPort}}
   - name: server
     connect_timeout: 5s
     type: STATIC
-    hosts:
-    - socket_address:
-        address: 127.0.0.1
-        port_value: {{.Ports.ProxyToServerProxyPort}}
+    load_assignment:
+      cluster_name: server
+      endpoints:
+      - lb_endpoints:
+        - endpoint:
+            address:
+              socket_address:
+                address: 127.0.0.1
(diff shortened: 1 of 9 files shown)


src/envoy/http/authn/filter_context.cc
```diff
@@ -102,15 +102,17 @@ bool FilterContext::getJwtPayloadFromEnvoyJwtFilter(const std::string& issuer,
     return false;
   }
 
-  // Serialize the payload from Envoy jwt filter first before writing it to |payload|.
-  // TODO (pitlv2109): Return protobuf Struct instead of string, once Istio jwt filter is removed.
-  // Also need to change how Istio authn filter processes the jwt payload.
+  // Serialize the payload from Envoy jwt filter first before writing it to
```
mandarjog: nit: why are all these changes showing up here? are we using a different formatting ?

gargnupur: I ran scripts/run_gofmt.sh that we are using in istio.io/istio... looks like it was added only 7 days back.. I can revert these changes if we don't intend to use this script for formatting?

PiotrSikora: `make check` is broken (see #2287), so the last few commits were committed without formatting being properly enforced.

PiotrSikora: Fixed, please merge with `master`.

gargnupur: Thanks @PiotrSikora

test/envoye2e/env/tcp_envoy_conf.go
```diff
@@ -0,0 +1,135 @@
...
+  address:
+    socket_address:
+      address: 127.0.0.1
+      port_value: {{.Ports.ClientAdminPort}}
+static_resources:
+  clusters:
+  - name: client
+    connect_timeout: 5s
+    type: STATIC
+    hosts:
```
kyessenov: `hosts` is deprecated, please update the config file. I think the new way is similar to EDS load assignment API.

kyessenov:
I can LGTM once the static bootstrap is updated to change the deprecated field.https://github.com/envoyproxy/data-plane-api/blob/master/envoy/api/v2/cds.proto#L191

test/envoye2e/env/tcp_envoy_conf.go
gargnupur: Fixed.. also fixed it in envoy_conf.go

PR #2287, opened 2019-06-24
Fix checks.

PiotrSikora:
Broken in #2251.


Makefile
```diff
@@ -53,9 +53,13 @@ test_asan:
 test_tsan:
 	PATH=$(PATH) CC=$(CC) CXX=$(CXX) bazel $(BAZEL_STARTUP_ARGS) test $(BAZEL_TEST_ARGS) --config=clang-tsan --test_env=TSAN_OPTIONS=suppressions=$(TOP)/tsan.suppressions -- $(BAZEL_TARGETS) $(SANITIZER_EXCLUSIONS)
 
+check:
+	@echo >&2 "Please use \"make lint\" instead."
```
venilnoronha: What's the reason for renaming this?

PiotrSikora: This was renamed in #2251, presumably to align with `make lint` in [istio/istio](https://github.com/istio/istio), and not in this PR.

Here, I'm only adding a guard to make sure that people/scripts using `make check` will see the change, instead of `make check` being silently ignored, like it was for the past few weeks.

PR #2289, opened 2019-06-25
Limit resource usage on Prow.

PiotrSikora:
While there, merge BAZEL_TEST_ARGS into BAZEL_BUILD_ARGS.


PiotrSikora:
@utka this should limit resource usage to 12GB RAM and 8CPUs. Let me know if I should increase/decrease those.

PiotrSikora:
yay! This seems to work :)

PR #2291, opened 2019-06-26
Fix header parsing in JWT filter

yangminzhu:
Fix header parsing in JWT filter

### DIFF
--- a/src/envoy/http/jwt_auth/jwt.h
+++ b/src/envoy/http/jwt_auth/jwt.h
@@ -231,8 +231,6 @@ class Jwt : public WithStatus {
   int64_t Exp();
 
  private:
-  const EVP_MD* md_;
-
   Json::ObjectSharedPtr header_;
   std::string header_str_;
   std::string header_str_base64url_;
--- a/src/envoy/http/jwt_auth/jwt.cc
+++ b/src/envoy/http/jwt_auth/jwt.cc
@@ -273,13 +273,7 @@ Jwt::Jwt(const std::string &jwt) {
     return;
   }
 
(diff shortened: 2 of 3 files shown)


src/envoy/http/jwt_auth/jwt.h
```diff
@@ -231,8 +231,6 @@ class Jwt : public WithStatus {
   int64_t Exp();
 
  private:
-  const EVP_MD* md_;
```
silentdai: I am surprised a pointer can be removed without touching destructor.
Could you elaborate who owns the object `md_` used to reference?

yangminzhu: Does it matter who owns it? I thought it's a const pointer pointing to a static variable.

yangminzhu:
@fpesce @duderino 
The test `proxy-presubmit.sh` seems getting stuck forever, I have tried to retest, will try again to see if it works.

src/envoy/http/jwt_auth/jwt.h
yangminzhu: Talked offline and we think this is not a concern, resolved.

silentdai: Thank you for the explanation! `md_` may not be introduced at the beginning. 
/lgtm

silentdai: Sorry I was heading to a meeting so I didn't complete what I should have pointed out: The `static variable` owns so `Jwt` doesn't need to free it. That's what I wanted to know. Thanks again!

fpesce:
Fixes istio/istio#15084

PR #2298, opened 2019-07-01
Fix jwt crash on invalid RSA public keys

qiwzhang:

If RSA public key is invalid, e.g. bad Base64 encoded in "n" or "e" field,  jwt_auth filter will crash.

**Release note**:
```
None
```

### DIFF
--- a/src/envoy/http/jwt_auth/jwt.cc
+++ b/src/envoy/http/jwt_auth/jwt.cc
@@ -537,7 +537,11 @@ void Pubkeys::ExtractPubkeyFromJwkRSA(Json::ObjectSharedPtr jwk_json) {
 
   EvpPkeyGetter e;
   pubkey->evp_pkey_ = e.EvpPkeyFromJwkRSA(n_str, e_str);
-  keys_.push_back(std::move(pubkey));
+  if (e.GetStatus() == Status::OK) {
+    keys_.push_back(std::move(pubkey));
+  } else {
+    UpdateStatus(e.GetStatus());
+  }
 }
 
 void Pubkeys::ExtractPubkeyFromJwkEC(Json::ObjectSharedPtr jwk_json) {
@@ -567,7 +571,11 @@ void Pubkeys::ExtractPubkeyFromJwkEC(Json::ObjectSharedPtr jwk_json) {
 
   EvpPkeyGetter e;
   pubkey->ec_key_ = e.EcKeyFromJwkEC(x_str, y_str);
-  keys_.push_back(std::move(pubkey));
+  if (e.GetStatus() == Status::OK) {
+    keys_.push_back(std::move(pubkey));
+  } else {
+    UpdateStatus(e.GetStatus());
+  }
 }
 
 std::unique_ptr<Pubkeys> Pubkeys::CreateFrom(const std::string &pkey,
--- a/src/envoy/http/jwt_auth/jwt.cc
+++ b/src/envoy/http/jwt_auth/jwt.cc
@@ -333,8 +333,12 @@ bool Verifier::VerifySignatureRSA(EVP_PKEY *key, const EVP_MD *md,
                                   size_t signed_data_len) {
   bssl::UniquePtr<EVP_MD_CTX> md_ctx(EVP_MD_CTX_create());
 
-  EVP_DigestVerifyInit(md_ctx.get(), nullptr, md, nullptr, key);
-  EVP_DigestVerifyUpdate(md_ctx.get(), signed_data, signed_data_len);
+  if (EVP_DigestVerifyInit(md_ctx.get(), nullptr, md, nullptr, key) != 1) {
+    return false;
+  }
+  if (EVP_DigestVerifyUpdate(md_ctx.get(), signed_data, signed_data_len) != 1) {
+    return false;
+  }
   return (EVP_DigestVerifyFinal(md_ctx.get(), signature, signature_len) == 1);
 }
 
--- a/src/envoy/http/jwt_auth/pubkey_cache.h
+++ b/src/envoy/http/jwt_auth/pubkey_cache.h
@@ -82,6 +82,9 @@ class PubkeyCacheItem : public Logger::Loggable<Logger::Id::filter> {
       if (status != Status::OK) {
         ENVOY_LOG(warn, "Invalid inline jwks for issuer: {}, jwks: {}",
                   jwt_config_.issuer(), inline_jwks);
+        throw EnvoyException(
+            fmt::format("Invalid inline jwks for issuer: {}, jwks: {}",
+                        jwt_config_.issuer(), inline_jwks));
       }
     }
   }
--- a/src/envoy/http/jwt_auth/pubkey_cache.h
+++ b/src/envoy/http/jwt_auth/pubkey_cache.h
@@ -80,11 +80,12 @@ class PubkeyCacheItem : public Logger::Loggable<Logger::Id::filter> {
                              // inline jwks never expires.
                              std::chrono::steady_clock::time_point::max());
       if (status != Status::OK) {
-        ENVOY_LOG(warn, "Invalid inline jwks for issuer: {}, jwks: {}",
-                  jwt_config_.issuer(), inline_jwks);
-        throw EnvoyException(
-            fmt::format("Invalid inline jwks for issuer: {}, jwks: {}",
-                        jwt_config_.issuer(), inline_jwks));
+        ENVOY_LOG(warn,
+                  "Invalid inline jwks for issuer: {}, jwks: {}, error: {}",
+                  jwt_config_.issuer(), inline_jwks, StatusToString(status));
+        throw EnvoyException(fmt::format(
+            "Invalid inline jwks for issuer: {}, jwks: {}, error: {}",
+            jwt_config_.issuer(), inline_jwks, StatusToString(status)));
       }
     }
   }
--- a/src/envoy/http/jwt_auth/auth_store.h
+++ b/src/envoy/http/jwt_auth/auth_store.h
@@ -71,6 +71,7 @@ class JwtAuthStoreFactory : public Logger::Loggable<Logger::Id::config> {
       : config_(std::make_shared<const ::istio::envoy::config::filter::http::
                                      jwt_auth::v2alpha1::JwtAuthentication>(
             config)),
+        dummy_store_(config_),
         tls_(context.threadLocal().allocateSlot()) {
     tls_->set([config = this->config_](Event::Dispatcher&)
                   -> ThreadLocal::ThreadLocalObjectSharedPtr {
@@ -86,6 +87,8 @@ class JwtAuthStoreFactory : public Logger::Loggable<Logger::Id::config> {
  private:
   // The auth config.
   JwtAuthenticationConstSharedPtr config_;
+  // A dummy Auth store to verify config is valid
+  JwtAuthStore dummy_store_;
   // Thread local slot to store per-thread auth store
   ThreadLocal::SlotPtr tls_;
 };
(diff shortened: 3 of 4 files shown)


mandarjog:
/ok-to-test

who removed wayne from the org ?

qiwzhang:
@yangminzhu Done

src/envoy/http/jwt_auth/jwt.cc
```diff
@@ -333,8 +333,12 @@ bool Verifier::VerifySignatureRSA(EVP_PKEY *key, const EVP_MD *md,
                                   size_t signed_data_len) {
   bssl::UniquePtr<EVP_MD_CTX> md_ctx(EVP_MD_CTX_create());
```
fpesce: I think there is a memory leak here.
I don't thee when the allocated value returned by EVP_MD_CTX_create() is free()'d.

src/envoy/http/jwt_auth/jwt.cc
```diff
@@ -333,8 +333,12 @@ bool Verifier::VerifySignatureRSA(EVP_PKEY *key, const EVP_MD *md,
...
 
-  EVP_DigestVerifyInit(md_ctx.get(), nullptr, md, nullptr, key);
-  EVP_DigestVerifyUpdate(md_ctx.get(), signed_data, signed_data_len);
+  if (EVP_DigestVerifyInit(md_ctx.get(), nullptr, md, nullptr, key) != 1) {
+    return false;
```
fpesce: there should be a EVP_MD_CTX_destroy() of whatever was returned by EVP_MD_CTX_create().

src/envoy/http/jwt_auth/jwt.cc
```diff
@@ -333,8 +333,12 @@ bool Verifier::VerifySignatureRSA(EVP_PKEY *key, const EVP_MD *md,
...
+  }
+  if (EVP_DigestVerifyUpdate(md_ctx.get(), signed_data, signed_data_len) != 1) {
+    return false;
```
fpesce: there should be a EVP_MD_CTX_destroy() of whatever was returned by EVP_MD_CTX_create().

src/envoy/http/jwt_auth/jwt.cc
```diff
@@ -333,8 +333,12 @@ bool Verifier::VerifySignatureRSA(EVP_PKEY *key, const EVP_MD *md,
...
+  }
   return (EVP_DigestVerifyFinal(md_ctx.get(), signature, signature_len) == 1);
```
fpesce: It looks like there was a memory leak before, how does EVP_MD_CTX_create() get deallocated ?

src/envoy/http/jwt_auth/jwt.cc
yangminzhu: See https://github.com/google/boringssl/blob/b96d4708652ba962e147eb0733d62afea07b31b0/include/openssl/digest.h#L109

src/envoy/http/jwt_auth/jwt.cc
qiwzhang: It is assigned to a smart_ptr bssl::UniquePtr.  It should free the buffer.

yangminzhu: Should we call `EVP_MD_CTX_free()` explicitly in this case?

fpesce: How does smart_ptr knows which function to call to free it?

src/envoy/http/jwt_auth/jwt.cc
qiwzhang: So far,  EVP_MD_CTX_create() only allocate a piece of buffer.  bssl::UniquePtr will free the buffer. So we are OK.

If EVP_MD_CTX_create() allocate embedded multiple buffers, then we have memory leak.

src/envoy/http/jwt_auth/jwt.cc
qiwzhang: So far, EVP_MD_CTX_create() only allocate a piece of buffer. bssl::UniquePtr will free the buffer. So we are OK.

If EVP_MD_CTX_create() allocate embedded multiple buffers, then we have memory leak.

src/envoy/http/jwt_auth/jwt.cc
qiwzhang: Please see my reply in other comments.

src/envoy/http/jwt_auth/jwt.cc
fpesce: It would work if and only if `EVP_MD_CTX_create()` would not be a structure with other fields allocated.
But from the actual openSSL code, you can see there is a need to cleanup fields for the opaque structure pointed by `EVP_MD_CTX`, noticeably `ctx->md_data`, c.f.:

I'm good at C, not C++, but I can't see how smart_ptr would figure out how to free fields of an opaque structure. (opaque meaning, the structure type name appears in .h, but its fields are only in .c, so the scope prevents you to know the content of it, it's basically like an object with only private elements, not calling its memory-releasing/destruction function will create a memory leak).

```c
void EVP_MD_CTX_destroy(EVP_MD_CTX *ctx)
	{
	EVP_MD_CTX_cleanup(ctx);
	OPENSSL_free(ctx);
	}

/* This call frees resources associated with the context */
int EVP_MD_CTX_cleanup(EVP_MD_CTX *ctx)
	{
	/* Don't assume ctx->md_data was cleaned in EVP_Digest_Final,
	 * because sometimes only copies of the context are ever finalised.
	 */
	if (ctx->digest && ctx->digest->cleanup
	    && !EVP_MD_CTX_test_flags(ctx,EVP_MD_CTX_FLAG_CLEANED))
		ctx->digest->cleanup(ctx);
	if (ctx->digest && ctx->digest->ctx_size && ctx->md_data
	    && !EVP_MD_CTX_test_flags(ctx, EVP_MD_CTX_FLAG_REUSE))
		{
		OPENSSL_cleanse(ctx->md_data,ctx->digest->ctx_size);
		OPENSSL_free(ctx->md_data);
		}
	if (ctx->pctx)
		EVP_PKEY_CTX_free(ctx->pctx);
#ifndef OPENSSL_NO_ENGINE
	if(ctx->engine)
		/* The EVP_MD we used belongs to an ENGINE, release the
		 * functional reference we held for this reason. */
		ENGINE_finish(ctx->engine);
#endif
	memset(ctx,'\0',sizeof *ctx);

	return 1;
	}
```

src/envoy/http/jwt_auth/jwt.cc
fpesce: it does allocate multiple buffer, it has a ->md_data field in it, that could be allocated (c.f.: https://github.com/majek/openssl/blob/master/crypto/evp/digest.c#L206
)

fpesce: Also ->pctx is allocate as soon as the verify init is called:
https://github.com/openssl/openssl/blob/master/crypto/evp/m_sigver.c#L29

qiwzhang: borningssl Smart point is very small.  It knows which destroy function call.

It is defined [here](https://github.com/google/boringssl/blob/6433a91dcb153304da4ef6a3d47a98d5860252a4/include/openssl/base.h#L534)

fpesce:
/lgtm
boringssl has smartptr with destructors correctly called.

fpesce:
@silentdai please give us an approval when you've got time!

src/envoy/http/jwt_auth/pubkey_cache.h
```diff
@@ -82,6 +82,9 @@ class PubkeyCacheItem : public Logger::Loggable<Logger::Id::filter> {
       if (status != Status::OK) {
         ENVOY_LOG(warn, "Invalid inline jwks for issuer: {}, jwks: {}",
                   jwt_config_.issuer(), inline_jwks);
+        throw EnvoyException(
+            fmt::format("Invalid inline jwks for issuer: {}, jwks: {}",
```
yangminzhu: I'm trying to test this but it seems this doesn't catch the bad config.

yangminzhu: Could you add `UpdateStatus(e.GetStatus());` after line 565 in https://github.com/istio/proxy/blob/a975561b980463f08689d3debe33bb9eefc80c3d/src/envoy/http/jwt_auth/jwt.cc#L565

And also add the `StatusToString(status)` to the exception here?

yangminzhu: I think we actually have validation code to deal with the bad public key but the issue is the result of the validation is not propagated to the caller properly (missing the `UpdateStatus(e.GetStatus());`). If we have this, we would have been able to catch this in very early stage instead of in segfault in runtime.

qiwzhang: Yes,  I just read the code.  It is very late in throwing exception.  The is constructed in local thread slot.  It is not done in the filter config update.  We should remove the code.  Let me see how easy to add the code in Filter config update.

yangminzhu: Don't you only just need to call `UpdateStatus(e.GetStatus());` to propagate the status? I thought this would be a easy fix.

qiwzhang: Done.  add a dummy store in config thread.

src/envoy/http/jwt_auth/pubkey_cache.h
```diff
@@ -80,8 +80,12 @@ class PubkeyCacheItem : public Logger::Loggable<Logger::Id::filter> {
                              // inline jwks never expires.
                              std::chrono::steady_clock::time_point::max());
       if (status != Status::OK) {
-        ENVOY_LOG(warn, "Invalid inline jwks for issuer: {}, jwks: {}",
-                  jwt_config_.issuer(), inline_jwks);
+        ENVOY_LOG(warn,
+                  "Invalid inline jwks for issuer: {}, jwks: {}, error: {}",
+                  jwt_config_.issuer(), inline_jwks, StatusToString(status));
+        throw EnvoyException(fmt::format(
```
yangminzhu: This is still going to be throw exception in worker thread as it's called in the constructor right?

silentdai: I have the same feeling since the thread local `JwtAuthStore` which owns `PubkeyCache` which construct PubkeyCacheItem on per worker thread. I didn't see the one catching this exception.
https://github.com/istio/proxy/blob/d70d7779ec3b067a43e9439dd9f925395dcdac20/src/envoy/http/jwt_auth/auth_store.h#L71

silentdai: IMHO line 77 could also throw exception

qiwzhang: I added dummy_store_ in auth_store.h  line 91.   Its purpose is to create a pubkey cache in main thread (config thread).  This store is not used,  it just want to parse the config and validate it.  If anything is wrong, the exception will  be throw in the main thread which will be catch by Envoy LDS subscriber config_update.  The bad LDS will be discarded.

For an invalid LDS config,  it will not be used,  so the exception will NOT be throw in the worker thread.

silentdai: @qiwzhang Oh, I saw it and the comment but I didn't realize it is addressing the thread model issue.   I think it is tricky but it should works.

silentdai:
/lgtm from networking perspective. 
Leave @yangminzhu for final approval WRT to the business logic

src/envoy/http/jwt_auth/pubkey_cache.h
yangminzhu: Thanks for the explanation

silentdai:
I accidently clicked wrong button
/approve

fpesce:
We have a Continuous Integration problem here, it seems that compilation gets stuck on proxy-presubmit-asan test.

fpesce:
@yangminzhu We should cherry pick this to 1.2 right ?

yangminzhu:
@fpesce Yes, I think so.

PR #2308, opened 2019-07-15
Unfold Set-Cookie header on direct response from Mixer

elevran:
This PR allows setting multiple `Set-Cookie` headers from a mixer response.
Per the RFC, Set-Cookie headers should be sent on separate lines. However, the mixer response encodes headers as a map<key,value> which prevents multiple headers from being specified.
This PR's treats the returned value as a comma separated list. While incomplete in handling all possible cookie variants and attributes (e.g., `Expires` attribute may contain a comma in the date), it handles the common case of session based cookies and allows future extension based on customer feedback.

Full discussion is on [#14240](https://github.com/istio/istio/issues/14240)

**Which issue this PR fixes**:
fixes #14240

`Set-Cookie` is not an Envoy inline header so will not get folded by the proxy if multiple headers are added. The current "cookie cutter" is simple and works for a subset of the cookies. This is noted in the code documentation.

**Release note**:
```release-note
FIX: mixer direct response may now include multiple Set-Cookie values, by specifying a comma separated list of cookies to send. The cookies must not include other commas (e.g., in Expire attribute) 
```

### DIFF
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -147,16 +147,18 @@ void Filter::UpdateHeaders(
   }
 }
 
-void Filter::UnfoldSetCookieHeader(HeaderMap& headers, const std::string value) {
+void Filter::UnfoldSetCookieHeader(HeaderMap& headers,
+                                   const std::string value) {
   // Folded cookie syntax can be complicated. See, for example,
   // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Syntax.
   // Unfortunately, the http2 library does not expose/support cookie parsing.
   // Here we handle a much simpler syntax - a comma separated list of cookies.
   // The simplification comes at the cost of preventing the use of 'Expires'
-  // cookie attribute. A more robust parser would require 100-150 LOC and could be
-  // modeled after the following JS or Java code:
+  // cookie attribute. A more robust parser would require 100-150 LOC and could
+  // be modeled after the following JS or Java code:
   // https://github.com/nfriedly/set-cookie-parser/blob/master/lib/set-cookie.js
-  // or https://github.com/google/j2objc/commit/16820fdbc8f76ca0c33472810ce0cb03d20efe25
+  // or
+  // https://github.com/google/j2objc/commit/16820fdbc8f76ca0c33472810ce0cb03d20efe25
   std::istringstream iss(value);
   std::vector<std::string> cookies;
   std::string cookie;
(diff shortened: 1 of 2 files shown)


kyessenov:
Thanks for the fix. I'm thinking -- would it be better to implement this in Mixer instead? The code is easier to test in istio, and I'm less worried about potential problems in golang vs c++.

The place to add it would be somewhere here: https://github.com/istio/istio/blob/0fd6e72e348617041bb622134f9ee16aad7cef8d/mixer/pkg/runtime/dispatcher/session.go#L254

elevran:
@kyessenov while I would prefer to resolve this only inside Mixer and not affect the proxy, not sure about how to go about fixing this in Mixer alone without changing the API between mixer and the filter. The API limits what mixer can instruct the filter to do:
- the current object is a map[string]string so no duplicate keys are allowed
- `Set-Cookie` is unique in that it should not be folded in response to the client
- filter should call setCopy() for headers that need to be set (ie no duplicates) and addReference() for headers that should be duplicated.

It seems that we need to special case the `Set-Cookie` headers in the MIxer, API layer, *and* filter, which would result in a much larger changeset, but I'm open to suggestions...

elevran:
@kyessenov looking at this again, the following may work:
- change mixer to process DirectResponse (e.g., around [here](https://github.com/istio/istio/blob/0fd6e72e348617041bb622134f9ee16aad7cef8d/mixer/pkg/runtime/dispatcher/session.go#L352) as as in other locations like you indicated) to parse the `Set-Cookie` value, if present, and create multiple APPEND directives from it. All other headers should use REPLACE. 
- change filter handling of APPEND to call headers.addReference() always. This would fold *inline* Envoy headers and leave others (including `Set-Cookie`) as multiple header lines

Is that what you had in mind?

elevran:
https://github.com/istio/istio/pull/15581 is an alternative - @kyessenov please advise which is closer to what you had in mind and we can close the other

kyessenov:
SGTM. I can raise this at P&T meeting today just to get some more input on which approach is preferred. I'm not sure I understand why `addCopy` and `addReference` behave differently. I thought it has to do with memory ownership for the underlying strings, not the semantics of the operation.

kyessenov:
Hi, we discussed this a bit, and generally agree on the following:
- if we can fix upstream envoy API to handle `addHeader(`set-cookie`, ...)` correctly, then we should do that, since that affects LUA/WASM/etc as well
- otherwise, we should avoid making changes to istio/proxy and instead work on istio/istio, due to less risk of parsing bugs

elevran:
> Hi, we discussed this a bit, and generally agree on the following:
> 
> * if we can fix upstream envoy API to handle `addHeader(`set-cookie`, ...)` correctly, then we should do that, since that affects LUA/WASM/etc as well
 
Adding the header to Envoy does the right thing, so Lua, WASM etc should work. 
Envoy folds inline headers and leaves others (such as `Set-Cookie`) in separate lines. This assumes that the caller uses multiple calls to addCopy("set-cookie", ...). Each call would be a separate header line.
You're correct regarding memory ownership (I was referring to addCopy vs setCopy...).

> * otherwise, we should avoid making changes to istio/proxy and instead work on istio/istio, due to less risk of parsing bugs

Agree. Abonadoning this PR in favor of concentrating changes in mixer code base.

PR #2311, opened 2019-07-16
Pass istio.io/metadata between peers using HTTP headers

kyessenov:


src/envoy/http/metadata_exchange/config.cc
```diff
@@ -0,0 +1,118 @@
...
+  auto metadata =
+      getMetadataValue(Common::Wasm::MetadataType::Node, NodeMetadataKey);
+  if (metadata.kind_case() == google::protobuf::Value::kStructValue) {
+    std::string metadata_bytes;
+    metadata.struct_value().SerializeToString(&metadata_bytes);
+    metadata_value_ =
+        Base64::encode(metadata_bytes.data(), metadata_bytes.size());
+  }
+
+  logDebug("metadata_value_ id:" + std::to_string(id()) +
```
silentdai: use fmt::format() in `source/common/common/fmt.h`?

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -0,0 +1,118 @@
...
+using Common::Wasm::Null::Plugin::removeResponseHeader;
+using Common::Wasm::Null::Plugin::replaceRequestHeader;
+using Common::Wasm::Null::Plugin::replaceResponseHeader;
+
+void PluginContext::onCreate() {
+  // TODO(kuat): serialize metadata in the root context instead
+  // populate and encode node metadata
+  auto metadata =
+      getMetadataValue(Common::Wasm::MetadataType::Node, NodeMetadataKey);
+  if (metadata.kind_case() == google::protobuf::Value::kStructValue) {
```
silentdai: TIL

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -0,0 +1,118 @@
...
+
+  logDebug("metadata_value_ id:" + std::to_string(id()) +
+           " value:" + metadata_value_);
+}
+
+Http::FilterHeadersStatus PluginContext::onRequestHeaders() {
+  // strip and store downstream peer metadata
+  auto downstream_metadata_value = getRequestHeader(ExchangeMetadataHeader);
+  if (downstream_metadata_value != nullptr &&
+      downstream_metadata_value->view().size() > 0) {
```
silentdai: nit: `!downstream_metadata_value->view().empty()` 
https://github.com/envoyproxy/envoy/pull/7419

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -0,0 +1,118 @@
...
+    removeRequestHeader(ExchangeMetadataHeader);
+    auto downstream_metadata_bytes =
+        Base64::decode(downstream_metadata_value->toString());
```
silentdai: find below to avoid string construction
  static std::string Base64::decodeWithoutPadding(absl::string_view input);

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -0,0 +1,118 @@
...
+    proxy_setMetadataStruct(
+        Common::Wasm::MetadataType::Request, DownstreamMetadataKey,
+        strlen(DownstreamMetadataKey), downstream_metadata_bytes.data(),
```
silentdai: Not sure how good is strlen but you can always use sizeof(DownstreamMetadataKey) -1 to avoid char count by cpu

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -0,0 +1,118 @@
...
+        downstream_metadata_bytes.size());
+  }
+
+  // insert peer metadata struct for upstream
+  if (metadata_value_.size() > 0) {
```
silentdai: if (not metadata_value_.empyt())

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  return Http::FilterHeadersStatus::Continue;
+}
+
+Http::FilterHeadersStatus PluginContext::onResponseHeaders() {
+  // strip and store upstream peer metadata
+  auto upstream_metadata_value = getResponseHeader(ExchangeMetadataHeader);
+  if (upstream_metadata_value != nullptr &&
+      upstream_metadata_value->view().size() > 0) {
```
silentdai: and here

src/envoy/http/metadata_exchange/config.h
```diff
@@ -0,0 +1,83 @@
...
+// DownstreamMetadataKey is the key in the request metadata for downstream peer
+// metadata
+constexpr char DownstreamMetadataKey[] =
+    "envoy.wasm.metadata_exchange.downstream";
+
+// UpstreamMetadataKey is the key in the request metadata for downstream peer
+// metadata
+constexpr char UpstreamMetadataKey[] = "envoy.wasm.metadata_exchange.upstream";
+
+using StringView = absl::string_view;
```
silentdai: In header file, generally it's not good to introduce classes to namespace by keyword `using` . Not sure if there is exception.

src/envoy/http/metadata_exchange/config.cc
lizan: Just declare `DownstreamMetadataKey` as `constexpr absl::string_view`, then you have `.size()` here.

src/envoy/http/metadata_exchange/config.h
kyessenov: Macro expansion below requires it... I agree this is not a good style, but this is a problem in the upstream.

src/envoy/http/metadata_exchange/config.cc
kyessenov: using absl::StrCat instead

src/envoy/http/metadata_exchange/config.cc
kyessenov: Done

src/envoy/http/metadata_exchange/config.cc
kyessenov: Thanks, this is indeed better.

src/envoy/http/metadata_exchange/config.cc
kyessenov: Done, thanks.

kyessenov:
It's committed as an example config, and also to be able to manually test the filter. There is a lot of moving parts, so we want to keep tests flexible for now.

PR #2313, opened 2019-07-18
validate trust domain in authentication filter

yangminzhu:
We want to deny requests from different trust domain by default.

https://github.com/istio/istio/issues/15631

**Release note**:

```release-note
```


yangminzhu:
@myidpt @diemtvu ready for review, PTAL, thanks.

src/envoy/utils/utils.cc
```diff
@@ -34,6 +34,29 @@ const std::string kPerHostMetadataKey("istio");
 // Attribute field for per-host data override
 const std::string kMetadataDestinationUID("uid");
 
+bool getRawPrincipal(const Network::Connection* connection, bool peer,
```
incfly: getCertSAN seems suit the implementation better.

src/envoy/utils/utils.cc
```diff
@@ -120,30 +143,37 @@ bool GetDestinationUID(const envoy::api::v2::core::Metadata& metadata,
...
-        *principal = result.substr(kSPIFFEPrefix.size());
-      } else {
-        *principal = result;
-      }
-      return true;
+  std::string result;
+  if (getRawPrincipal(connection, peer, &result)) {
+    if (hasSPIFFEPrefix(result)) {
+      // Strip out the prefix "spiffe://" in the identity.
+      *principal = result.substr(kSPIFFEPrefix.size());
```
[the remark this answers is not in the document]
yangminzhu: I think it's okay here as this function is used internally and it's the caller's responsibility to make sure `principal` is not null, otherwise they're making a mistake using this function.

src/envoy/http/authn/authenticator_base.cc
```diff
@@ -47,6 +47,34 @@ AuthenticatorBase::AuthenticatorBase(FilterContext* filter_context)
 
 AuthenticatorBase::~AuthenticatorBase() {}
 
+bool AuthenticatorBase::validateTrustDomain(
+    const Network::Connection* connection) const {
+  std::string peer_trust_domain;
+  if (!Utils::GetTrustDomain(connection, true, &peer_trust_domain)) {
+    ENVOY_LOG(error,
```
silentdai: Use ENVOY_CONN_LOG(error, connection, ...)

src/envoy/utils/utils.cc
```diff
@@ -34,6 +34,29 @@ const std::string kPerHostMetadataKey("istio");
 // Attribute field for per-host data override
 const std::string kMetadataDestinationUID("uid");
 
+bool getCertSAN(const Network::Connection* connection, bool peer,
+                std::string* principal) {
+  if (connection) {
+    Ssl::ConnectionInfo* ssl =
+        const_cast<Ssl::ConnectionInfo*>(connection->ssl());
+    if (ssl != nullptr) {
+      const std::vector<std::string> sans =
```
silentdai: const auto& any other type to avoid copy

src/envoy/utils/utils.cc
```diff
@@ -34,6 +34,29 @@ const std::string kPerHostMetadataKey("istio");
 // Attribute field for per-host data override
 const std::string kMetadataDestinationUID("uid");
 
+bool getCertSAN(const Network::Connection* connection, bool peer,
+                std::string* principal) {
+  if (connection) {
```
silentdai: It's severe issue when when the connection is nullptr. Consider add a ENVOY_LOG?

src/envoy/utils/utils.cc
```diff
@@ -34,6 +34,29 @@ const std::string kPerHostMetadataKey("istio");
...
+    Ssl::ConnectionInfo* ssl =
+        const_cast<Ssl::ConnectionInfo*>(connection->ssl());
```
silentdai: Why const_cast? Could you add a comment here?

src/envoy/utils/utils.cc
```diff
@@ -120,30 +143,37 @@ bool GetDestinationUID(const envoy::api::v2::core::Metadata& metadata,
...
 
+bool GetTrustDomain(const Network::Connection* connection, bool peer,
+                    std::string* trust_domain) {
+  std::string cert_san;
+  if (!getCertSAN(connection, peer, &cert_san) || !hasSPIFFEPrefix(cert_san)) {
+    return false;
+  }
+
+  // Strip out the prefix "spiffe://" before getting trust domain.
+  std::size_t slash = cert_san.find('/', kSPIFFEPrefix.size());
```
silentdai: absl::StartsWith

src/envoy/utils/utils.cc
```diff
@@ -34,6 +34,29 @@ const std::string kPerHostMetadataKey("istio");
...
+        return false;
+      }
+      *principal = sans[0];
+      return true;
+    }
+  }
+  return false;
+}
+
+bool hasSPIFFEPrefix(const std::string& san) {
```
silentdai: Probably you are implementing absl::StartsWith

src/envoy/utils/utils.cc
```diff
@@ -34,6 +34,29 @@ const std::string kPerHostMetadataKey("istio");
 // Attribute field for per-host data override
 const std::string kMetadataDestinationUID("uid");
 
+bool getCertSAN(const Network::Connection* connection, bool peer,
+                std::string* principal) {
```
silentdai: You didn't check if principal is nullptr so probably you should declare it as std::string& principal

src/envoy/utils/utils.cc
```diff
@@ -120,30 +143,37 @@ bool GetDestinationUID(const envoy::api::v2::core::Metadata& metadata,
...
+    } else {
+      *principal = cert_san;
     }
+    return true;
   }
   return false;
 }
 
+bool GetTrustDomain(const Network::Connection* connection, bool peer,
+                    std::string* trust_domain) {
```
silentdai: ditto: std::string& trust_domain
Honestly the modern approach to to declare as
optional<std::string> GetTrustDomain(const Network::Connection* connection, bool peer);

src/envoy/utils/utils.cc
incfly: That's what I was wondering as well in previous comments. It seems not just this method is passing raw pointer to the method and assume caller's responsible to ensure the memory safety. other methods around doing similar things.

Wondering what are general C++/Envoy's style guide in these cases? non const reference?

silentdai: >That's what I was wondering as well in previous comments.

:+1: 

> non const reference?

That is what I know

For this particular function as <bool, string> is the output, the absl::optional<std::string> is recommended

yangminzhu: I literally copied the old code, I think this is just okay because it's simply an out param, there are a lot of API having the same assumptions that the caller is going to set up all the out param before calling the function. I can change the newly added functions but will leave the old code as-is in this PR.

silentdai: If this the old convention, you can leave it here and create a issue to address it in the future :)

silentdai: @yangminzhu  Sorry I didn't look into the folded comment!

src/envoy/utils/utils.cc
yangminzhu: Changed to use absl::StartsWith.

src/envoy/utils/utils.cc
yangminzhu: This is a utility function so I would prefer to add the LOG in the caller site (validateX509 in authenticator_base.cc)

src/envoy/utils/utils.cc
yangminzhu: I agree it's better to use optional, but I would prefer to stick with the same convention like other GetXXX functions in utils.h in this PR. Of course could refactor all of them in separate PR.

src/envoy/utils/utils.cc
yangminzhu: Is this comment for hasSPIFFEPrefix() or would you mind to clarify a little bit here? Thanks

src/envoy/utils/utils.cc
yangminzhu: That's the old code I'm not sure the intention, removed the const_cast in this PR, thanks for catching this.

yangminzhu:
@silentdai I addressed the comments so fat, would mind to take a look again? thanks for the review.

src/envoy/utils/utils_test.cc
```diff
@@ -69,4 +121,45 @@ TEST(UtilsTest, CheckResponseInfoToStreamInfo) {
...
+  testGetPrincipal(sans, "spiffe:foo/bar", true);
+}
+
+TEST_P(UtilsTest, GetPrincipalEmpty) {
+  std::vector<std::string> sans;
+  testGetPrincipal(sans, "", false);
+}
+
+TEST_P(UtilsTest, GetTrustDomain) {
+  std::vector<std::string> sans{"spiffe://td/bar", "bad"};
```
silentdai: What do you mean by "bad" here? My understanding ist hat getTrustDomain use only the first element. So the second element is the hint whether the test should successful? If so it is not "bad"

src/envoy/utils/utils.cc
silentdai: Aha, I totally misunderstand this one. I thought you are skipping "spiffe" instead of "spiffe://"
You are good to go. Sorry for the churn

src/envoy/utils/utils_test.cc
yangminzhu: Ah, I just put it here as a second element to make sure the 1st element is used. It could be anything as long as it's different from the 1st element.

PR #2314, opened 2019-07-18
Add Stackdriver plugin config

bianpengyuan:
Add Stackdriver plugin config. The config is parsed by StackdriverRootContext.

### DIFF
--- a/extensions/stackdriver/config/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/stackdriver_plugin_config.proto
@@ -1,0 +1,36 @@
+/* Copyright 2019 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package stackdriver.config;
+
+message PluginConfig {
+  // Indicates whether the module should act as inbound or outbound reporter.
+  // Difference being:
+  //   * inbound and outbound have different metric types to export.
+  //   * inbound reporter writes access logs.
+  enum ReporterKind {
+    INBOUND = 0;
+    OUTBOUND = 1;
+  }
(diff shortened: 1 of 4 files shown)


extensions/stackdriver/config/stackdriver_plugin_config.proto
```diff
@@ -0,0 +1,36 @@
...
+
+syntax = "proto3";
+
+package stackdriver.config;
+
+message PluginConfig {
+  // Indicates whether the module should act as inbound or outbound reporter.
+  // Difference being:
+  //   * inbound and outbound have different metric types to export.
+  //   * inbound reporter writes access logs.
```
kyessenov: Is this still true given there is a boolean `export_access_log` below?

extensions/stackdriver/stackdriver.cc
```diff
@@ -36,14 +36,23 @@ namespace Plugin {
...
 constexpr char kExporterRegistered[] = "registered";
 
 void StackdriverRootContext::onConfigure(
     std::unique_ptr<WasmData> configuration) {
-  // TODO: Add config for Stackdriver plugin, such as reporter kind.
-  UNREFERENCED_PARAMETER(configuration);
+  // Parse configuration JSON string.
+  JsonParseOptions json_options;
+  Status status =
+      JsonStringToMessage(configuration->toString(), &config_, json_options);
```
kyessenov: I think you can use absl::string_view here (`configuration->view()`)

bianpengyuan: Tried but it does not seem to work:
```
extensions/stackdriver/stackdriver.cc:50:7: error: no matching function for call to 'JsonStringToMessage'
      JsonStringToMessage(configuration->view(), &config_, json_options);
      ^~~~~~~~~~~~~~~~~~~
external/com_google_protobuf/src/google/protobuf/util/json_util.h:107:30: note: candidate function not viable: no known conversion from 'StringView' (aka 'absl::string_view') to 'google::protobuf::StringPiece' for 1st argument
PROTOBUF_EXPORT util::Status JsonStringToMessage(
                             ^
external/com_google_protobuf/src/google/protobuf/util/json_util.h:110:21: note: candidate function not viable: requires 2 arguments, but 3 were provided
inline util::Status JsonStringToMessage(StringPiece input,
                    ^
```

extensions/stackdriver/config/stackdriver_plugin_config.proto
bianpengyuan: Yes, I think we just want to export access log at server side. `export_access_log` controls if server exports access log or not.

extensions/stackdriver/stackdriver.cc
kyessenov: Ah, ok. Thanks. I think protobuf hasn't yet pushed StringPiece->string_view update.

PR #2315, opened 2019-07-19
Add NodeInfo and RequestInfo into Stackdriver plugin

bianpengyuan:
NodeInfo is extracted from local node metadata and peer node metadata. RequestInfo is collected from filter callbacks. Both data structure will be used to fill logs and metrics.


extensions/stackdriver/common/constants.h
```diff
@@ -0,0 +1,41 @@
...
+constexpr char kMetadataPodNameKey[] = "name";
+constexpr char kMetadataNamespaceKey[] = "namespace";
+constexpr char kMetadataOwnerKey[] = "owner";
+constexpr char kMetadataWorkloadNameKey[] = "workload_name";
+constexpr char kMetadataContainersKey[] = "ports_to_containers";
+constexpr char kPlatformMetadataKey[] = "platform_metadata";
+constexpr char kGCPClusterLocationKey[] = "gcp_cluster_location";
+constexpr char kGCPClusterNameKey[] = "gcp_cluster_name";
+constexpr char kGCPProjectKey[] = "gcp_project";
+constexpr char kUpstreamMetadataKey[] = "envoy.wasm.metadata_exchange.upstream";
```
mandarjog: These should be move to a common place where mx filter can also use it

extensions/stackdriver/common/context.cc
```diff
@@ -0,0 +1,98 @@
...
+#include "extensions/stackdriver/common/context.h"
+#include "extensions/stackdriver/common/constants.h"
+
+namespace Extensions {
+namespace Stackdriver {
+namespace Common {
+
+// Find value of the given key from the metadata proto map and copy it into the
+// field. If the given key cannot be found, field remains unset.
+void FillNodeMetadataField(
```
mandarjog: Return a bool?

extensions/stackdriver/common/constants.h
kyessenov: I'm actually OK to not introduce a direct code dependency yet. Ultimately, this will be config, so no need to add code dependency right now.

extensions/stackdriver/common/context.h
```diff
@@ -0,0 +1,96 @@
...
+namespace Common {
+
+// NodeInfo represents the information extracted from proxy node metadata, or
+// peer node metadata header. This is used to fill metrics and log labels.
+struct NodeInfo {
+  // Name of the node. e.g. pod name in k8s.
+  std::string name;
+
+  // Namespaace that the node runs in.
+  std::string namespace_name;
```
mandarjog: namespace.

extensions/stackdriver/stackdriver.cc
```diff
@@ -91,6 +99,52 @@ StackdriverOptions StackdriverRootContext::getStackdriverOptions() {
...
+        Common::Wasm::MetadataType::Request, kUpstreamMetadataKey);
+    ExtractNodeMetadata(upstream_metadata, &request_info_.peer_node_info);
+  }
+  request_info_.end_timestamp = proxy_getCurrentTimeNanoseconds();
+  return FilterHeadersStatus::Continue;
+}
+
+FilterDataStatus StackdriverContext::onResponseBody(size_t body_buffer_length,
+                                                    bool) {
+  request_info_.response_size += body_buffer_length;
```
kyessenov: This is already available as `stream_info.bytesReceived()` and `stream_info.bytesSent()`.
This is only OK if is not exposed through API. We need to file an issue to expose it, since it's more efficient to reuse the computation.

extensions/stackdriver/stackdriver.cc
```diff
@@ -91,6 +99,52 @@ StackdriverOptions StackdriverRootContext::getStackdriverOptions() {
...
+}
+
+FilterHeadersStatus StackdriverContext::onResponseHeaders() {
+  if (getRootContext()->reporterKind() ==
+      PluginConfig::ReporterKind::PluginConfig_ReporterKind_OUTBOUND) {
+    auto upstream_metadata = getMetadataStruct(
+        Common::Wasm::MetadataType::Request, kUpstreamMetadataKey);
+    ExtractNodeMetadata(upstream_metadata, &request_info_.peer_node_info);
+  }
+  request_info_.end_timestamp = proxy_getCurrentTimeNanoseconds();
```
kyessenov: Shouldn't this be in Log()?

extensions/stackdriver/stackdriver.cc
```diff
@@ -91,6 +99,52 @@ StackdriverOptions StackdriverRootContext::getStackdriverOptions() {
...
+    auto downstream_metadata = getMetadataStruct(
+        Common::Wasm::MetadataType::Request, kDownstreamMetadataKey);
+    ExtractNodeMetadata(downstream_metadata, &request_info_.peer_node_info);
+  }
+  return FilterHeadersStatus::Continue;
+}
+
+FilterDataStatus StackdriverContext::onRequestBody(size_t body_buffer_length,
+                                                   bool) {
+  request_info_.request_size += body_buffer_length;
```
kyessenov: Already available in stream_info.

extensions/stackdriver/stackdriver.cc
kyessenov: It's also available as `stream_info.requestComplete()`.

extensions/stackdriver/common/context_test.cc
```diff
@@ -0,0 +1,130 @@
...
+namespace Extensions {
+namespace Stackdriver {
+namespace Common {
+
+// Test ExtraceNodeMetadata with all possible metadata field.
+TEST(ContextTest, ExtractNodeMetadata) {
+  google::protobuf::Struct metadata_struct;
+  auto metadata_fields = metadata_struct.mutable_fields();
+  google::protobuf::Value pod_name;
+  *pod_name.mutable_string_value() = "test_pod";
```
kyessenov: You could use TextProto to ease testing here.

extensions/stackdriver/common/context.h
```diff
@@ -0,0 +1,96 @@
...
+  int64_t destination_port = 0;
+
+  // Protocol used the request (HTTP/1.1, gRPC, etc).
+  std::string request_protocol;
+
+  // Response code of the request.
+  int64_t response_code = 0;
+
+  // Host name of destination service.
+  std::string destination_service_host;
```
kyessenov: consider using string_view if Envoy already exposes those interfaces. we can make this struct smaller this way.

extensions/stackdriver/common/context.cc
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+void ExtractNodeMetadata(const google::protobuf::Struct &metadata,
+                         NodeInfo *node_info) {
+  const auto &istio_metadata_fields = metadata.fields();
+  if (istio_metadata_fields.empty()) {
+    return;
+  }
+  FillNodeMetadataField(istio_metadata_fields, kMetadataPodNameKey,
```
kyessenov: this would work better by defining a protobuf and using protobuf to extract values from JSON struct.

mandarjog: Are you suggesting node_info should be a proto ?
There is already a node proto.

kyessenov: Yes, turn `node_info` and then it's a one liner to extract from a struct.

bianpengyuan: I update the function to use JSON struct as a bridge. Indeed code becomes simpler, but I guess performance is worse than the original function?

extensions/stackdriver/stackdriver.cc
```diff
@@ -54,8 +56,23 @@ void StackdriverRootContext::onConfigure(
     return;
   }
 
-  // Only register exporter once in main thread when initiating base WASM
-  // module.
+  // Get node metadata. GetMetadataStruct always returns the whole node metadata
+  // even with a key passed in.
+  // TODO: clean this up after fixing upstream API to respect node metadata key.
+  auto node_metadata = getMetadataStruct(Common::Wasm::MetadataType::Node, "");
+  auto istio_metadata = node_metadata.fields().find(kIstioMetadataKey);
```
mandarjog: you can use
```c++
getMetadataValue(Common::Wasm::MetadataType::Node, "istio.io/metadata");
```

kyessenov: I don't think it respects the key per the comment and my own experience.

extensions/stackdriver/common/context.cc
kyessenov: Thanks. I think we can optimize it later by moving the deserialization into metadata encoder (which already deserializes the struct).

mandarjog: We need to fix this whole thing. Too many conversations here and in the metadata exchange filter. At least here we cache local node in onconfigure.

extensions/stackdriver/stackdriver.cc
mandarjog: Can you check what you are using in metadata exchange ?
Is it trying to get the whole metadata or just our sub key?

extensions/stackdriver/common/context.cc
kyessenov: There is no way to avoid serialization. You have to retrieve dynamic metadata which is done through protobuf JSON serialization in WASM API. Since it's per request, we can't move it to onConfigure.

extensions/stackdriver/stackdriver.cc
kyessenov: I'm using low-lever proxy_get/setMetadataStruct. The problem with C++ version is that it automatically serializes the whole struct, and I didn't want to since I have bytes already.

kyessenov: I think we need a vararg version of this method that can traverse into structs. Basically, something along Find() in https://github.com/envoyproxy/envoy-wasm/issues/42. That way there is no transfer of container objects to get a value.

extensions/stackdriver/stackdriver.cc
mandarjog: Can all this be done in `Log` ?

extensions/stackdriver/common/context.cc
bianpengyuan: The function has been removed and the new code returns JSON parsing status.

extensions/stackdriver/common/context.h
bianpengyuan: This has been renamed to namespace in NodeInfo Protobuf.

extensions/stackdriver/stackdriver.cc
bianpengyuan: Added a todo to track this.

extensions/stackdriver/stackdriver.cc
bianpengyuan: Done and add a todo to switch to requestComplete()

extensions/stackdriver/stackdriver.cc
bianpengyuan: ditto.

extensions/stackdriver/common/context_test.cc
bianpengyuan: Done.

extensions/stackdriver/common/context.cc
bianpengyuan: Done.

extensions/stackdriver/stackdriver.cc
bianpengyuan: Done.

extensions/stackdriver/common/constants.h
bianpengyuan: Ok, I will just keep it as what it is for now.

extensions/stackdriver/common/context.cc
```diff
@@ -0,0 +1,44 @@
...
+#include "google/protobuf/util/json_util.h"
+
+namespace Extensions {
+namespace Stackdriver {
+namespace Common {
+
+using namespace google::protobuf::util;
+using namespace stackdriver::common;
+
+Status ExtractNodeMetadata(const google::protobuf::Struct &metadata,
```
silentdai: naming: `extractNodeMetadata`

silentdai: IIRC envoy replace all `google::protobuf` to `ProtobufWkt`. Not sure why but try if you can follow

extensions/stackdriver/stackdriver.cc
```diff
@@ -54,8 +56,20 @@ void StackdriverRootContext::onConfigure(
...
-  // module.
+  // Get node metadata. GetMetadataStruct always returns the whole node metadata
+  // even with a key passed in.
+  // TODO: change to GetMetadataStruct after fixing upstream API to respect node metadata key.
+  auto node_metadata = getMetadataValue(Common::Wasm::MetadataType::Node, kIstioMetadataKey);
+  status = ExtractNodeMetadata(node_metadata.struct_value(),
+                                    &local_node_info_);
+  if (status != Status::OK) {
+    logWarn("cannot parse local node metadata " +
+            node_metadata.DebugString() + ": " + status.ToString());
```
silentdai: Not sure if you need an early return

extensions/stackdriver/stackdriver.cc
```diff
@@ -91,7 +105,68 @@ StackdriverOptions StackdriverRootContext::getStackdriverOptions() {
...
+FilterDataStatus StackdriverContext::onResponseBody(size_t body_buffer_length,
+                                                    bool) {
+  // TODO: switch to stream_info.bytesSent/bytesReceived to avoid extra compute.
+  request_info_.response_size += body_buffer_length;
+  return FilterDataStatus::Continue;
+}
+
+StackdriverRootContext *StackdriverContext::getRootContext() {
+  RootContext *root = this->root();
+  return static_cast<StackdriverRootContext *>(root);
```
silentdai: Maybe dynamic_cast?

extensions/stackdriver/common/BUILD
```diff
@@ -0,0 +1,67 @@
...
+cc_library(
+    name = "context",
+    srcs = [
+        "context.cc",
+    ],
+    hdrs = [
+        "context.h",
+    ],
+    visibility = [
+        "//extensions/stackdriver:__pkg__",
```
silentdai: TIL

extensions/stackdriver/common/context.h
bianpengyuan: I am not sure how to make it use string_view. WASM apis returns WasmData which owns the underlying string. To avoid copy, we need to change this fields to be WasmData type right? I could do it but it makes code weird, so I prefer to do it later as performance optimization.

kyessenov: Ok. I'm not sure how ownership is passed through WASM either, so let's just copy.

extensions/stackdriver/common/context.cc
kyessenov: Not necessary. This is done to simplify copybara import (protobuf is net::proto2 internally). We are not planning to move this into google3 any time soon.

bianpengyuan: Rename the function to be lower cased. I don't want to introduce dependency on envoy code base because eventually it will be compiled into a wasm module, which ideally should not rely on anything from envoy except the SDK.

extensions/stackdriver/stackdriver.cc
bianpengyuan: Thanks!

extensions/stackdriver/stackdriver.cc
bianpengyuan: I thought about it too and I make it not to early return. Exporter will still be registered and user will get better error message from Opencensus library on how wrong node metadata would affect metric exporting.

extensions/stackdriver/common/context.cc
silentdai: I see. Thank you for the explanation!

kyessenov:
There was a comment about why we're doing this in onHeaders callback instead of Log(). Can you respond to that?

bianpengyuan:
@kyessenov I think I already move it to onLog(): https://github.com/istio/proxy/pull/2315/files#diff-1a85058b98763420131e21f16886e857R168

kyessenov:
Ok, I was referring to everything inside onRequestHeader and onResponseHeaders. We don't need to do any computation on the data path. Everything can be done afterwards for telemetry.

PR #2316, opened 2019-07-20
Update metadata exchange filter

mandarjog:
1. Move Node metadata caching to `RootContext::OnConfigure`
2. Use deterministic serialization
3. Add `id` header.

### DIFF
--- a/src/envoy/http/metadata_exchange/config.cc
+++ b/src/envoy/http/metadata_exchange/config.cc
@@ -51,9 +51,10 @@ void PluginRootContext::onConfigure(
         Base64::encode(metadata_bytes.data(), metadata_bytes.size());
 
     // magic "." to get the whole node.
-    auto node = getMetadataStruct(Common::Wasm::MetadataType::Node, ".");
+    auto node =
+        getMetadataStruct(Common::Wasm::MetadataType::Node, WholeNodeKey);
     for (auto& f : node.fields()) {
-      if (f.first == "id" &&
+      if (f.first == NodeIdKey &&
           f.second.kind_case() == google::protobuf::Value::kStringValue) {
         node_id_ = f.second.string_value();
         break;
@@ -66,7 +67,7 @@ void PluginRootContext::onConfigure(
 }
 
 PluginRootContext* PluginContext::rootContext() {
-  return static_cast<PluginRootContext*>(this->root());
+  return dynamic_cast<PluginRootContext*>(this->root());
 }
 
 std::string PluginContext::node_id() { return rootContext()->node_id(); }
@@ -101,15 +102,15 @@ Http::FilterHeadersStatus PluginContext::onRequestHeaders() {
                       DownstreamMetadataIdKey, downstream_metadata_id->view());
   }
 
-  auto mv = metadata_value();
+  auto metadata = metadata_value();
   // insert peer metadata struct for upstream
-  if (mv.size() > 0) {
-    replaceRequestHeader(ExchangeMetadataHeader, mv);
+  if (metadata.size() > 0) {
+    replaceRequestHeader(ExchangeMetadataHeader, metadata);
   }
 
-  auto nid = node_id();
-  if (nid.size() > 0) {
-    replaceRequestHeader(ExchangeMetadataHeaderId, nid);
+  auto nodeid = node_id();
+  if (nodeid.size() > 0) {
+    replaceRequestHeader(ExchangeMetadataHeaderId, nodeid);
   }
 
   return Http::FilterHeadersStatus::Continue;
@@ -134,15 +135,15 @@ Http::FilterHeadersStatus PluginContext::onResponseHeaders() {
                       UpstreamMetadataIdKey, upstream_metadata_id->view());
   }
 
-  auto mv = metadata_value();
+  auto metadata = metadata_value();
   // insert peer metadata struct for downstream
-  if (mv.size() > 0) {
-    replaceResponseHeader(ExchangeMetadataHeader, mv);
+  if (metadata.size() > 0) {
+    replaceResponseHeader(ExchangeMetadataHeader, metadata);
   }
 
-  auto nid = node_id();
-  if (nid.size() > 0) {
-    replaceResponseHeader(ExchangeMetadataHeaderId, nid);
+  auto nodeid = node_id();
+  if (nodeid.size() > 0) {
+    replaceResponseHeader(ExchangeMetadataHeaderId, nodeid);
   }
 
   return Http::FilterHeadersStatus::Continue;
--- a/src/envoy/http/metadata_exchange/config.cc
+++ b/src/envoy/http/metadata_exchange/config.cc
@@ -15,6 +15,7 @@
 
 #include "src/envoy/http/metadata_exchange/config.h"
 #include "common/common/base64.h"
+#include "config.h"
 
 namespace Envoy {
 namespace Extensions {
@@ -23,6 +24,7 @@ namespace MetadataExchange {
 
 // imports from the low-level API
 using Common::Wasm::Null::NullVmPluginFactory;
+using Common::Wasm::Null::Plugin::getMetadataStruct;
 using Common::Wasm::Null::Plugin::getMetadataValue;
 using Common::Wasm::Null::Plugin::getRequestHeader;
 using Common::Wasm::Null::Plugin::getResponseHeader;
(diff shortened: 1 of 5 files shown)


mandarjog:
output
```
17:20 $ curl -v http://localhost:8090/help
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 8090 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8090 (#0)
> GET /help HTTP/1.1
> Host: localhost:8090
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 200 OK
< content-length: 13
< content-type: text/plain
< location: http://localhost:8090/help
< date: Sat, 20 Jul 2019 00:20:03 GMT
< server: envoy
< x-envoy-upstream-service-time: 2
< x-envoy-peer-metadata-id: test
< x-envoy-peer-metadata: CiIKBmxhYmVscxIYKhYKFAoDYXBwEg0aC3Byb2R1Y3RwYWdlChYKCW5hbWVzcGFjZRIJGgdkZWZhdWx0AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==
<
```

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -58,15 +91,26 @@ Http::FilterHeadersStatus PluginContext::onRequestHeaders() {
...
+    removeRequestHeader(ExchangeMetadataHeaderId);
+    setMetadataStruct(Common::Wasm::MetadataType::Request,
+                      DownstreamMetadataIdKey, downstream_metadata_id->view());
   }
 
+  auto mv = metadata_value();
   // insert peer metadata struct for upstream
-  if (metadata_value_.size() > 0) {
-    replaceRequestHeader(ExchangeMetadataHeader, metadata_value_);
+  if (mv.size() > 0) {
```
kyessenov: C++ style is to use long variable names. `mv` is a golang pattern :D

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -34,22 +35,53 @@ using Common::Wasm::Null::Plugin::removeResponseHeader;
...
+    google::protobuf::io::CodedOutputStream mcs(&md);
+
+    mcs.SetSerializationDeterministic(true);
+    metadata.struct_value().SerializeToCodedStream(&mcs);
+
     metadata_value_ =
         Base64::encode(metadata_bytes.data(), metadata_bytes.size());
+
+    // magic "." to get the whole node.
+    auto node = getMetadataStruct(Common::Wasm::MetadataType::Node, ".");
```
bianpengyuan: It would be nice to make all hardcoded strings (".", "id") const.

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -34,22 +35,53 @@ using Common::Wasm::Null::Plugin::removeResponseHeader;
...
+      }
+    }
   }
 
   logDebug(
       absl::StrCat("metadata_value_ id:", id(), " value:", metadata_value_));
 }
 
+PluginRootContext* PluginContext::rootContext() {
+  return static_cast<PluginRootContext*>(this->root());
```
bianpengyuan: s/static_cast/dynamic_cast (I did the same in the sd plugin based on review from yuchen :))

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -34,22 +35,53 @@ using Common::Wasm::Null::Plugin::removeResponseHeader;
...
-  // TODO(kuat): serialize metadata in the root context instead
-  // populate and encode node metadata
+void PluginRootContext::onConfigure(
+    std::unique_ptr<WasmData> ABSL_ATTRIBUTE_UNUSED configuration) {
   auto metadata =
       getMetadataValue(Common::Wasm::MetadataType::Node, NodeMetadataKey);
   if (metadata.kind_case() == google::protobuf::Value::kStructValue) {
     std::string metadata_bytes;
-    metadata.struct_value().SerializeToString(&metadata_bytes);
+    google::protobuf::io::StringOutputStream md(&metadata_bytes);
```
bianpengyuan: Is this change in order to make serialization deterministic? Why do we need to make it deterministic?

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -23,6 +23,7 @@ namespace MetadataExchange {
 
 // imports from the low-level API
 using Common::Wasm::Null::NullVmPluginFactory;
+using Common::Wasm::Null::Plugin::getMetadataStruct;
```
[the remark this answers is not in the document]
lizan: `using namespace` is prohibited by style guide https://google.github.io/styleguide/cppguide.html#Namespaces

src/envoy/http/metadata_exchange/config.cc
mandarjog: Why? Is it creating a new copy ?

src/envoy/http/metadata_exchange/config.cc
mandarjog: lack of determinism has bitten us many times in the past. It also makes testing easier.

mandarjog:
Updated per comments. PTAL.

src/envoy/http/metadata_exchange/config.cc
bianpengyuan: Ah yeah, indeed it could make test easier when we try to verify the metadata header value. Thanks!

src/envoy/http/metadata_exchange/config.cc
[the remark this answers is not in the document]
bianpengyuan: Ack. Thanks!

src/envoy/http/metadata_exchange/config.cc
bianpengyuan: My understanding was dynamic_cast is safer at downcast by type checking at runtime. But second thought static_cast actually works well here because `this->root()` should definitely return a `PluginRootContext` and could save a bit of runtime. So nvm about this.

src/envoy/http/metadata_exchange/config.cc
kyessenov: Don't know about that. The real issue is depending on bytes as a hash. This is explicitly prohibited by protobuf, and we simply should not depend on the raw bytes.

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -34,22 +35,54 @@ using Common::Wasm::Null::Plugin::removeResponseHeader;
...
       absl::StrCat("metadata_value_ id:", id(), " value:", metadata_value_));
 }
 
+PluginRootContext* PluginContext::rootContext() {
+  return dynamic_cast<PluginRootContext*>(this->root());
+}
+
+std::string PluginContext::node_id() { return rootContext()->node_id(); }
+
+std::string PluginContext::metadata_value() {
```
kyessenov: mayb consider it `inline` to avoid string copying.

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -58,15 +91,26 @@ Http::FilterHeadersStatus PluginContext::onRequestHeaders() {
...
-    proxy_setMetadataStruct(
-        Common::Wasm::MetadataType::Request, DownstreamMetadataKey.data(),
-        DownstreamMetadataKey.size(), downstream_metadata_bytes.data(),
-        downstream_metadata_bytes.size());
+    setMetadataStruct(Common::Wasm::MetadataType::Request,
+                      DownstreamMetadataKey, downstream_metadata_bytes);
+  }
+
+  auto downstream_metadata_id = getRequestHeader(ExchangeMetadataHeaderId);
+  if (downstream_metadata_id != nullptr) {
```
kyessenov: add non-empty condition like the above.

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -34,22 +35,54 @@ using Common::Wasm::Null::Plugin::removeResponseHeader;
...
+      }
+    }
   }
 
   logDebug(
       absl::StrCat("metadata_value_ id:", id(), " value:", metadata_value_));
 }
 
+PluginRootContext* PluginContext::rootContext() {
+  return dynamic_cast<PluginRootContext*>(this->root());
```
kyessenov: Per Google style, dynamic_cast or down_cast is recommended I think. It avoids the issue with potential type conversion.

src/envoy/http/metadata_exchange/config.h
```diff
@@ -55,24 +66,44 @@ class PluginRootContext : public RootContext {
...
+  void onCreate() override{};
   Http::FilterHeadersStatus onRequestHeaders() override;
   Http::FilterHeadersStatus onResponseHeaders() override;
 
  private:
-  std::string metadata_value_;
+  inline PluginRootContext* rootContext() {
+    return dynamic_cast<PluginRootContext*>(this->root());
+  };
+  inline std::string metadata_value() {
```
silentdai: Not sure if metadataValue should be in `class Context`

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -58,15 +73,27 @@ Http::FilterHeadersStatus PluginContext::onRequestHeaders() {
...
+    removeRequestHeader(ExchangeMetadataHeaderId);
+    setMetadataStruct(Common::Wasm::MetadataType::Request,
+                      DownstreamMetadataIdKey, downstream_metadata_id->view());
+  }
+
+  auto metadata = metadata_value();
   // insert peer metadata struct for upstream
-  if (metadata_value_.size() > 0) {
-    replaceRequestHeader(ExchangeMetadataHeader, metadata_value_);
+  if (metadata.size() > 0) {
```
silentdai: prefer `!empty()` over `.size() >  0`

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -23,27 +23,42 @@ namespace MetadataExchange {
...
+    mcs.SetSerializationDeterministic(true);
+    metadata.struct_value().SerializeToCodedStream(&mcs);
+
     metadata_value_ =
         Base64::encode(metadata_bytes.data(), metadata_bytes.size());
+
+    // magic "." to get the whole node.
+    auto node =
+        getMetadataStruct(Common::Wasm::MetadataType::Node, WholeNodeKey);
+    for (auto& f : node.fields()) {
```
silentdai: const?

PR #2317, opened 2019-07-20
Add opencensus registry into sd plugin

bianpengyuan:
Add view, measure and tag registry function into stackdriver plugin.

### DIFF
--- a/extensions/stackdriver/common/constants.h
+++ b/extensions/stackdriver/common/constants.h
@@ -17,6 +17,52 @@ namespace Extensions {
 namespace Stackdriver {
 namespace Common {
 
+// Measure names of metrics.
+constexpr char kServerRequestCountMeasure[] =
+    "istio.io/service/server/request_count_measure";
+constexpr char kServerRequestBytesMeasure[] =
+    "istio.io/service/server/request_bytes_measure";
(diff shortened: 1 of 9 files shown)


bianpengyuan:
Can I get some review on this?

extensions/stackdriver/common/constants.h
```diff
@@ -17,6 +17,52 @@ namespace Extensions {
 namespace Stackdriver {
 namespace Common {
 
+// Measure names of metrics.
+constexpr char kServerRequestCountMeasure[] =
```
kyessenov: `constexpr absl::string_view` is easier to use for constant strings.

bianpengyuan: Why it is easier to use? Just replace the type seems to be not working, for example the following statement won't work, protobuf map cannot use absl::string_view as a key:
```
(*monitored_resource.mutable_labels())[kProjectIDLabel]
```

kyessenov: OK, fair enough. Leave as-is, until protobuf adopts string_view. It's easier because it gives you all the string functions (concat, size, etc).

bianpengyuan:
@kyessenov I think I need an approval label :).

PR #2319, opened 2019-07-23
Fix mx filter broken in the last commit

mandarjog:
1. Fix deterministic serialization by ensuring stream is destroyed before string is used.
2. Split testdata config in 2 parts, server.yaml and client.yaml to separate client and server.
3. Review comments from last review.

### DIFF
--- a/src/envoy/http/metadata_exchange/config.cc
+++ b/src/envoy/http/metadata_exchange/config.cc
@@ -37,21 +37,27 @@ using Common::Wasm::Null::Plugin::replaceRequestHeader;
 using Common::Wasm::Null::Plugin::replaceResponseHeader;
 using Common::Wasm::Null::Plugin::setMetadataStringValue;
 
+bool serializeToStringDeterministic(const google::protobuf::Struct& metadata,
+                                    std::string* metadata_bytes) {
+  google::protobuf::io::StringOutputStream md(metadata_bytes);
+  google::protobuf::io::CodedOutputStream mcs(&md);
+
+  mcs.SetSerializationDeterministic(true);
+  if (!metadata.SerializeToCodedStream(&mcs)) {
+    logWarn(
+        absl::StrCat("unable to serialize Nodemetadata key=", NodeMetadataKey));
+    return false;
+  }
+  return true;
+}
+
 void PluginRootContext::onConfigure(
     std::unique_ptr<WasmData> ABSL_ATTRIBUTE_UNUSED configuration) {
   auto metadata =
       getMetadataValue(Common::Wasm::MetadataType::Node, NodeMetadataKey);
   if (metadata.kind_case() == google::protobuf::Value::kStructValue) {
     std::string metadata_bytes;
-    google::protobuf::io::StringOutputStream md(&metadata_bytes);
-    google::protobuf::io::CodedOutputStream mcs(&md);
-
-    mcs.SetSerializationDeterministic(true);
-    if (!metadata.struct_value().SerializeToCodedStream(&mcs)) {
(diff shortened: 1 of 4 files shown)


src/envoy/http/metadata_exchange/config.cc
```diff
@@ -115,24 +123,36 @@ Http::FilterHeadersStatus PluginContext::onResponseHeaders() {
...
 }
 
+inline void logRequestMetadata(StringView key) {
+  auto val = getMetadataStringValue(Common::Wasm::MetadataType::Request, key);
+  if (!val.empty()) {
+    logDebug(absl::StrCat(key, "=", val));
+  }
+}
+
+void PluginContext::onLog() {
```
kyessenov: Can you remove this? It's only needed for debugging, right?

kyessenov: I think we could use Envoy standard logging and just fix the upstream problems.

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -44,15 +47,19 @@ void PluginRootContext::onConfigure(
     google::protobuf::io::CodedOutputStream mcs(&md);
 
     mcs.SetSerializationDeterministic(true);
-    metadata.struct_value().SerializeToCodedStream(&mcs);
+    if (!metadata.struct_value().SerializeToCodedStream(&mcs)) {
+      logWarn(absl::StrCat("unable to serialize Nodemetadata key=",
+                           NodeMetadataKey));
+    }
+    mcs.Trim();  // trim buffer to the actual size needed.
```
silentdai: Does it imply a flush()? I am not familiar with this code.

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -106,7 +114,7 @@ Http::FilterHeadersStatus PluginContext::onResponseHeaders() {
       !upstream_metadata_value->view().empty()) {
     removeResponseHeader(ExchangeMetadataHeader);
     auto upstream_metadata_bytes =
-        Base64::decode(upstream_metadata_value->toString());
+        Base64::decodeWithoutPadding(upstream_metadata_value->toString());
```
silentdai: I suspect we can use upstream_metadata_value->view()
I remeber there is decode api which accept string view. So you could save another string construction.

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -23,36 +23,49 @@ namespace MetadataExchange {
...
+                                    std::string* metadata_bytes) {
+  google::protobuf::io::StringOutputStream md(metadata_bytes);
+  google::protobuf::io::CodedOutputStream mcs(&md);
+
+  mcs.SetSerializationDeterministic(true);
+  if (!metadata.SerializeToCodedStream(&mcs)) {
+    logWarn(
+        absl::StrCat("unable to serialize Nodemetadata key=", NodeMetadataKey));
+    return false;
+  }
```
mandarjog: I could get rid of `mcs.Trim()` by putting this in a separate function.
Now `CodedOutputStream` is destroyed before the `std::string` is used by the caller, this ensures that flush/Trim is called.

mandarjog:
Updated per review, also removed the obsure `Trim()` call in favor of a separate function.
PTAL.

src/envoy/http/metadata_exchange/config.cc
silentdai: nit: I would suggest return `absl::optional<string>()`
I think we can leave it in future versions as this PR does nothing wrong :)

PR #2320, opened 2019-07-24
authn_wasm: Add skeleton authn filter in WASM.

yangminzhu:
https://github.com/istio/istio/issues/15772

### DIFF
--- a/src/envoy/http/authn_wasm/config.cc
+++ b/src/envoy/http/authn_wasm/config.cc
@@ -1,0 +1,113 @@
+/* Copyright 2019 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/envoy/http/authn_wasm/config.h"
+
+#include "authentication/v1alpha1/policy.pb.h"
+#include "google/protobuf/text_format.h"
+#include "google/protobuf/util/json_util.h"
+
+namespace iaapi = istio::authentication::v1alpha1;
+
+namespace Envoy {
+namespace Extensions {
+namespace Wasm {
+namespace AuthnWasm {
+
+// imports from the low-level API
+using Common::Wasm::Null::NullVmPluginFactory;
+
+// TODO: The WASM library should be improved with:
+//  1) allow logging with a unified prefix.
+//  2) expose macros that evaluates logDebug() only for debug logging.
+using Common::Wasm::Null::Plugin::logDebug;
+using Common::Wasm::Null::Plugin::logError;
+
+using ::google::protobuf::TextFormat;
+using namespace google::protobuf::util;
+
+void PluginRootContext::onConfigure(std::unique_ptr<WasmData> configuration) {
+  logDebug("called PluginRootContext::onConfigure()");
+  JsonParseOptions json_options;
+  Status status = JsonStringToMessage(configuration->toString(),
+                                      &filter_config_, json_options);
+  if (status != Status::OK) {
+    logError("Cannot parse authentication filter config: " +
+             configuration->toString());
+  } else {
+    std::string out_str;
+    TextFormat::PrintToString(filter_config_, &out_str);
+    logDebug("Applied authentication filter config:\n" + out_str);
+  }
+}
+
+void PluginRootContext::onStart() {
+  logDebug("called PluginRootContext::onStart()");
+};
+
+void PluginRootContext::onTick() {
+  logDebug("called PluginRootContext::onTick()");
+};
+
+void PluginContext::onCreate() { logDebug("called PluginContext::onCreate()"); }
+
+Http::FilterHeadersStatus PluginContext::onRequestHeaders() {
+  const auto& config = filter_config();
+  for (const auto& method : config.policy().peers()) {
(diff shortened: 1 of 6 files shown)


src/envoy/http/authn_wasm/config.cc
```diff
@@ -0,0 +1,113 @@
...
+
+using ::google::protobuf::TextFormat;
+using namespace google::protobuf::util;
+
+void PluginRootContext::onConfigure(std::unique_ptr<WasmData> configuration) {
+  logDebug("called PluginRootContext::onConfigure()");
+  JsonParseOptions json_options;
+  Status status = JsonStringToMessage(configuration->toString(),
+                                      &filter_config_, json_options);
+  if (status != Status::OK) {
```
incfly: if the status is not okay, will the filter stuck in a unsuable state? Is this err handling same in compile-in mode filter?

src/envoy/http/authn_wasm/config.cc
```diff
@@ -0,0 +1,113 @@
...
+// imports from the low-level API
+using Common::Wasm::Null::NullVmPluginFactory;
+using Common::Wasm::Null::Plugin::logDebug;
+using Common::Wasm::Null::Plugin::logError;
+
+using ::google::protobuf::TextFormat;
+using namespace google::protobuf::util;
+
+void PluginRootContext::onConfigure(std::unique_ptr<WasmData> configuration) {
+  logDebug("called PluginRootContext::onConfigure()");
```
incfly: different filter will share the same function exported by host? maybe we should then add 'authn' prefix to simulate log scope?

src/envoy/http/authn_wasm/config.cc
```diff
@@ -0,0 +1,113 @@
...
+  return Http::FilterHeadersStatus::Continue;
+}
+
+// Registration glue
+
+Common::Wasm::Null::NullVmPluginRootRegistry* context_registry_{};
+
+class AuthnWasmFactory : public Common::Wasm::Null::NullVmPluginFactory {
+ public:
+  const std::string name() const override { return "envoy.wasm.authn_wasm"; }
```
incfly: authn_wasm seems redudant? wasm is in prefix already.

src/envoy/http/authn_wasm/config.h
```diff
@@ -0,0 +1,82 @@
...
+using istio::envoy::config::filter::http::authn::v2alpha1::FilterConfig;
+using StringView = absl::string_view;
+using Common::Wasm::Null::NullVmPluginRootRegistry;
+using Common::Wasm::Null::Plugin::Context;
+using Common::Wasm::Null::Plugin::ContextFactory;
+using Common::Wasm::Null::Plugin::RootContext;
+using Common::Wasm::Null::Plugin::RootFactory;
+using Common::Wasm::Null::Plugin::WasmData;
+
+// PluginRootContext is the root context for all streams processed by the
```
incfly: is this comment accurate? i saw the `wasm_filter.md` is documented this is a per vm instance.

src/envoy/http/authn_wasm/config.cc
```diff
@@ -0,0 +1,113 @@
...
+void PluginRootContext::onTick() {
+  logDebug("called PluginRootContext::onTick()");
+};
+
+void PluginContext::onCreate() { logDebug("called PluginContext::onCreate()"); }
+
+Http::FilterHeadersStatus PluginContext::onRequestHeaders() {
+  const auto& config = filter_config();
+  std::string out_str;
+  TextFormat::PrintToString(config, &out_str);
```
incfly: serialized computation is invoked for every request header as data path, might not added for now?

if WASM has some feature to load runtime feature (log level) and we can then put that computation there? same thing, i think logDebug is a function not a macro, and the DebugString is always invoked?

src/envoy/http/authn_wasm/config.cc
yangminzhu: I think Envoy will just rejects the config, the filter will not be created.

src/envoy/http/authn_wasm/config.h
yangminzhu: I copied from other files, why is this not accurate? It talks about streams and thread but doesn't mention VMs?

src/envoy/http/authn_wasm/config.cc
yangminzhu: That's a good point but I think it should be solved in the WASM layer, so I will just add a TODO at this moment.

src/envoy/http/authn_wasm/config.cc
yangminzhu: I think this is another issue that should be resolved in the WASM level. I can just remove this for now.

yangminzhu:
@incfly thanks for the review, comments addressed, PTAL.

src/envoy/http/authn_wasm/config.cc
incfly: but how you let envoy know that? The status is not returned or consumed here.

src/envoy/http/authn_wasm/config.h
incfly: Yes.`wasm_filter.md` references the VM not the thread. @mandarjog and @jplevyak in last weekly meeting, we talked about the mapping of VM/worker threads, I can't remember the details though to confirm if it's 1:1 mapping. This seems closed to existing Envoy filter context, not sure how much applied to Wasm one.

> At bootstrap time, a root context is created. The root context
has the same lifetime as the VM/runtime instance and acts as a target
for any interactions which happen at initial setup. It is also used for
interactions that outlive a request.

src/envoy/http/authn_wasm/config.cc
incfly: Have you added TODO here? didn't see the diff yet.

src/envoy/http/authn_wasm/config.cc
incfly: sounds good. might worth file a feature request.

/cc @PiotrSikora @jplevyak

src/envoy/http/authn_wasm/config.cc
incfly: never mind see it above.

PR #2322, opened 2019-07-24
Add function to record SD metrics

bianpengyuan:
A integration test will be added to cover this code after upstream OC library updated.


bianpengyuan:
cc @kyessenov @mandarjog Looks like I can no longer edit reviewers... Please take a look thanks!

bianpengyuan:
@kyessenov @mandarjog Can I have some review on this? I am hoping to do a one-off migration to the common package after this one got in.

extensions/stackdriver/metric/record.cc
```diff
@@ -0,0 +1,89 @@
...
+         {sourceOwnerKey(), request_info.peer_node_info.owner()},
+         {destinationPrincipalKey(), request_info.destination_principal},
+         {destinationWorkloadNameKey(), local_node_info.workload_name()},
+         {destinationWorkloadNamespaceKey(), local_node_info.namespace_()},
+         {destinationOwnerKey(), local_node_info.owner()}});
+  }
+  if (kind == stackdriver::config::PluginConfig::ReporterKind::
+                  PluginConfig_ReporterKind_OUTBOUND) {
+    opencensus::stats::Record(
+        {{clientRequestCountMeasure(), 1},
```
mandarjog: Will all these be converted to direct variable access?

bianpengyuan: Yeah it will return a static variable like this: https://github.com/census-ecosystem/opencensus-cpp-example/blob/master/helloworld/helloworld.cc#L40

extensions/stackdriver/config/stackdriver_plugin_config.proto
```diff
@@ -29,7 +29,7 @@ message PluginConfig {
   ReporterKind kind = 1;
 
   // Controls whether to export access log.
-  bool export_access_log = 2;
+  bool disable_access_logging = 2;
```
silentdai: This break the existing code. Usually it's forbidden but we are at the very early phase. WDYT @mandarjog

extensions/stackdriver/BUILD
```diff
@@ -35,7 +35,7 @@ envoy_cc_library(
     deps = [
         "//extensions/stackdriver/common:context",
         "//extensions/stackdriver/config:stackdriver_plugin_config_cc_proto",
-        "//extensions/stackdriver/metric:registry",
+        "//extensions/stackdriver/metric",
```
silentdai: :metric can be omit? TIL

extensions/stackdriver/config/stackdriver_plugin_config.proto
kyessenov: Recommendation: add v1alpha1 as a last part in the proto package

bianpengyuan: It is fine to update proto definition at this point, but anyway I versioned the proto. PTAL. Thanks!

extensions/stackdriver/BUILD
bianpengyuan: Yeah, if the package name is a dup with the folder name, it gets omitted.

bianpengyuan:
@kyessenov @silentdai Can I get some review on this? Mandar is out for a week and I don't want to block this change for that long :)

PR #2325, opened 2019-07-26
Metadata Exchange Filter

gargnupur:
Work left: 
1) Move writing of metadata to the callback when TLS handshake is complete.
onNewConnection gets called when connection is established but before handshake is complete. Thus couldnt do metadata write here.

Things to note:
1) In Upstream Filter, onWrite happens first, thus we write metadata in "OnWrite".
2) In Downstream Filter, onData happens first, thus we write metadata in "OnData".
3) Both Upstream and Downstream filter consume metadata in OnData.


kyessenov:
FYI The upstream filter facility is now merge into Envoy, so it should be ready to use. We don't have to combine both filters in one PR, upstream can be done separately from downstream.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,162 @@
...
+    envoy::tcp::alpnproxy::header::AlpnProxyHeaderProto proxy_header;
+    if (!proxy_header.ParseFromString(proxy_header_buf)) {
+      config_->stats().header_not_found_.inc();
+      done_on_data_ = true;
+      return Network::FilterStatus::Continue;
+    }
+    data.drain(proxy_header_length);
+    if (proxy_header.has_metadata()) {
+      read_callbacks_->connection().streamInfo().setDynamicMetadata(
+          "filters.network.alpn_proxy.upstream", proxy_header.metadata());
```
kyessenov: Please make this a constant

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,162 @@
...
+  done_on_data_ = true;
+  return Network::FilterStatus::Continue;
+}
+
+Network::FilterStatus AlpnProxyFilter::onNewConnection() {
+  return Network::FilterStatus::Continue;
+}
+
+Network::FilterStatus AlpnProxyFilter::onWrite(Buffer::Instance& data, bool) {
+  config_->stats().metadata_found_.inc();
```
kyessenov: Why is it incremented per each write?

src/envoy/tcp/alpn_proxy/alpn_proxy_test.cc
```diff
@@ -0,0 +1,160 @@
...
+
+using testing::NiceMock;
+using testing::Return;
+using testing::ReturnRef;
+
+namespace Envoy {
+namespace Tcp {
+namespace AlpnProxy {
+namespace {
+bool protoMapEq(const ProtobufWkt::Struct& obj,
```
kyessenov: MessageDifferencer is public now, you can find a use of it in envoy.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Fixed

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Left from testing... removed

src/envoy/tcp/alpn_proxy/alpn_proxy_test.cc
gargnupur: Thanks..Fixed

WORKSPACE
```diff
@@ -72,3 +72,14 @@ envoy_dependencies()
 load("@envoy//bazel:dependency_imports.bzl", "envoy_dependency_imports")
 
 envoy_dependency_imports()
+
+ABSL_SHA = "61c9bf3e3e1c28a4aa6d7f1be4b37fd473bb5529"
+
+ABSL_SHA256 = "7ddf863ddced6fa5bf7304103f9c7aa619c20a2fcf84475512c8d3834b9d14fa"
+
+http_archive(
```
lizan: absl is already loaded by dependencies above, linking different versions of absl is likely causing problems.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,164 @@
...
+          AlpnProxy_FilterDirection_DOWNSTREAM) {
+    config_->stats().downstream_filter_found_.inc();
+    const uint32_t initial_header_length = sizeof(AlpnProxyInitialHeader);
+    if (data.length() < initial_header_length) {
+      config_->stats().initial_header_not_found_.inc();
+      // Not enough data to read. Wait for it to come.
+      return Network::FilterStatus::StopIteration;
+    }
+    std::string initial_header_buf(initial_header_length, 0);
+    data.copyOut(0, initial_header_length, &initial_header_buf[0]);
```
lizan: just linearize and use buffer data, no need copyOut

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,164 @@
...
+      done_on_data_ = true;
+      return Network::FilterStatus::Continue;
+    }
+    const uint64_t proxy_header_length =
+        absl::gntohl(initial_header->data_size);
+    // Drain the initial header length bytes read.
+    data.drain(initial_header_length);
+    if (data.length() < proxy_header_length) {
+      // Not enough data to read. Wait for it to come.
+      return Network::FilterStatus::StopIteration;
```
lizan: You already drained initial_header_length above, in the next iteration of onData when new data come, it will read first `initial_header_length` again but the data no longer contains real magic/length.

WORKSPACE
gargnupur: Thanks.. Removed

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Yeah, I was hoping to refactor this in next PR but just did in this PR only :)

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Thanks!! Updated.

gargnupur:
@kyessenov, @mandarjog , @lizan : Looks like tests are failing because of the following error:
[2019-08-03 16:21:21.167][32983][warning][testing] [external/envoy/test/test_common/environment.cc:175] Testing with IPv6 addresses may not be supported on this machine. If testing fails, set the environment variable ENVOY_IP_TEST_VERSIONS.

and I see this in other PR's too..

mandarjog:
Yes it is failing on PRs since prow nodes were upgraded.

rshriram:
whats the purpose of this filter?

gargnupur:
@rshriram : This filter will be used to exchange node metadata in case of TCP connections.
More info here: https://github.com/envoyproxy/envoy/issues/6532

src/envoy/tcp/alpn_proxy/header/alpn_proxy_header.proto
```diff
@@ -0,0 +1,30 @@
...
+
+option java_outer_classname = "AlpnProxyHeaderProto";
+option java_multiple_files = true;
+option java_package = "io.envoyproxy.envoy.tcp.alpnproxy.header";
+option go_package = "AlpnProxy";
+
+import "google/protobuf/struct.proto";
+
+// [#protodoc-title: AlpnProxyHeaderProto is sent on wire as initial bytes]
+message AlpnProxyHeaderProto {
```
kyessenov: Is this proto needed? Can we use `google.protobuf.Struct` instead of it, or even better, make that proto customizable?

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,234 @@
...
+
+std::unique_ptr<::Envoy::Buffer::OwnedImpl> constructProxyHeaderData(
+    const ::google::protobuf::Struct& metadata) {
+  envoy::tcp::alpnproxy::header::AlpnProxyHeaderProto proxy_header;
+  proxy_header.mutable_metadata()->CopyFrom(metadata);
+
+  AlpnProxyInitialHeader initial_header;
+  initial_header.magic = absl::ghtonl(AlpnProxyInitialHeader::magic_number);
+  initial_header.data_size =
+      absl::ghtonl(proxy_header.SerializeAsString().length());
```
kyessenov: Please serialize to string once and save the string on stack.

src/envoy/tcp/alpn_proxy/alpn_proxy.h
```diff
@@ -0,0 +1,134 @@
...
+  const std::string UpstreamDynamicDataKey =
+      "filters.network.alpn_proxy.upstream";
+  const std::string DownstreamDynamicDataKey =
+      "filters.network.alpn_proxy.downstream";
+
+  enum {
+    WriteMetadata,               // Write node metadata
+    ReadingInitialHeader,        // AlpnProxyInitialHeader is being read
+    ReadingProxyHeader,          // AlpnProxyHeaderProto is being read
+    NeedMoreDataIinitialHeader,  // Need more data to be read
```
kyessenov: typo `Iinitial`

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,234 @@
...
+namespace AlpnProxy {
+namespace {
+
+std::unique_ptr<::Envoy::Buffer::OwnedImpl> constructProxyHeaderData(
+    const ::google::protobuf::Struct& metadata) {
+  envoy::tcp::alpnproxy::header::AlpnProxyHeaderProto proxy_header;
+  proxy_header.mutable_metadata()->CopyFrom(metadata);
+
+  AlpnProxyInitialHeader initial_header;
+  initial_header.magic = absl::ghtonl(AlpnProxyInitialHeader::magic_number);
```
kyessenov: Can you comment why ghtonl is needed? Maybe as part of the header declaration, indicate that most-significant byte first is used.

src/envoy/tcp/alpn_proxy/config/alpn_proxy.proto
```diff
@@ -0,0 +1,46 @@
...
+option java_outer_classname = "AlpnProxyProto";
+option java_multiple_files = true;
+option java_package = "io.envoyproxy.envoy.tcp.alpnproxy.config";
+option go_package = "AlpnProxy";
+
+// [#protodoc-title: AlpnProxy protocol match and data transfer]
+// AlpnProxy protocol match and data transfer
+message AlpnProxy {
+  // Direction of the flow of traffic in which this this AlpnProxy filter
+  // is plcaed.
```
kyessenov: typo: `placed`

src/envoy/tcp/alpn_proxy/config/alpn_proxy.proto
```diff
@@ -0,0 +1,46 @@
...
+  string protocol = 1;
+
+  // The node metadata id whose data will be written to connection data.
+  string node_metadata_id = 2;
+
+  // The human readable prefix to use when emitting
+  string stat_prefix = 3;
+
+  // Defines if this is a downstream or upstream AlpnProxy filter.
+  FilterDirection filter_direction = 4;
```
kyessenov: Is this implicit in the distinction between network and upstream network filters? It'd be better to have less config since the direction should be inferrable from the factory.

src/envoy/tcp/alpn_proxy/config/alpn_proxy.proto
```diff
@@ -0,0 +1,46 @@
...
+  enum FilterDirection {
+    DOWNSTREAM = 0;
+    UPSTREAM = 1;
+  }
+  // Protocol that Alpn should support on the server.
+  // [#comment:TODO(GargNupur): Make it a list.]
+  string protocol = 1;
+
+  // The node metadata id whose data will be written to connection data.
+  string node_metadata_id = 2;
```
kyessenov: I think it's safe to make

src/envoy/tcp/alpn_proxy/config.cc
```diff
@@ -0,0 +1,120 @@
...
+#include "envoy/network/connection.h"
+#include "envoy/registry/registry.h"
+#include "src/envoy/tcp/alpn_proxy/alpn_proxy.h"
+#include "src/envoy/utils/config.h"
+
+namespace Envoy {
+namespace Tcp {
+namespace AlpnProxy {
+
+Network::FilterFactoryCb AlpnProxyConfigFactory::createFilterFactory(
```
kyessenov: This can be safely removed. JSON config is no longer required.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,234 @@
...
+    : scope_(scope),
+      stat_prefix_(stat_prefix),
+      protocol_(protocol),
+      node_metadata_id_(node_metadata_id),
+      filter_direction_(filter_direction),
+      stats_(generateStats(stat_prefix, scope)) {}
+
+Network::FilterStatus AlpnProxyFilter::onData(Buffer::Instance& data, bool) {
+  // No work needed if we have read connection protocol and its not the expected
+  // alpn protocol or if alpn protocol is found but we are still reading data.
```
kyessenov: Do you mean `done reading data`?

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,234 @@
...
+  if (conn_state_ != WriteMetadata) {
+    return;
+  }
+
+  if (local_info_.node().has_metadata()) {
+    auto metadata_fields = local_info_.node().metadata().fields();
+    auto node_metadata = metadata_fields.find(config_->node_metadata_id_);
+    if (node_metadata != metadata_fields.end()) {
+      std::unique_ptr<::Envoy::Buffer::OwnedImpl> buf =
+          constructProxyHeaderData(node_metadata->second.struct_value());
```
kyessenov: Heads up that we are switching to using filter state instead of metadata. Therefore, it might be worth encapsulating all code that has to do with metadata reading and saving into functions.

src/envoy/tcp/alpn_proxy/header/alpn_proxy_header.proto
gargnupur: I feel its good to have it so that later if we want to add a specific field say host name, it can be easily added?
what did you have in mind for making it customizable? using Any?

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Done

src/envoy/tcp/alpn_proxy/alpn_proxy.h
gargnupur: done

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Done here and in header declaration

src/envoy/tcp/alpn_proxy/config/alpn_proxy.proto
gargnupur: fixed

src/envoy/tcp/alpn_proxy/config/alpn_proxy.proto
gargnupur: Thanks.. Fixed

src/envoy/tcp/alpn_proxy/config.cc
gargnupur: Done

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Yup.. changed

src/envoy/tcp/alpn_proxy/header/alpn_proxy_header.proto
kyessenov: Yes, using Any. We could use a type url in the header that defaults to struct.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Thanks for the heads up! Moved into readMetadata and writeMetadata functions

src/envoy/tcp/alpn_proxy/header/alpn_proxy_header.proto
gargnupur: Used Any now.. please check the logic..

This is what am doing=> if Any contains a struct, I take that struct out and pass that as metadata struct value. If not then I construct a new metadata struct value whose key is type_url of any and value is any. 
Kinda torn between using any's value as compared to any itself in metadata struct value's value.. what do you think?

src/envoy/tcp/alpn_proxy/alpn_proxy.h
```diff
@@ -0,0 +1,142 @@
...
+enum FilterDirection { Downstream, Upstream };
+
+/**
+ * Configuration for the AlpnProxy filter.
+ */
+class AlpnProxyConfig {
+ public:
+  AlpnProxyConfig(const std::string& stat_prefix, const std::string& protocol,
+                  const std::string& node_metadata_id,
+                  const FilterDirection& filter_direction, Stats::Scope& scope);
```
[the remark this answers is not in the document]
gargnupur: Thanks Fixed

src/envoy/tcp/alpn_proxy/config/alpn_proxy.proto
```diff
@@ -0,0 +1,37 @@
...
+
+// [#protodoc-title: AlpnProxy protocol match and data transfer]
+// AlpnProxy protocol match and data transfer
+message AlpnProxy {
+  // Protocol that Alpn should support on the server.
+  // [#comment:TODO(GargNupur): Make it a list.]
+  string protocol = 1;
+
+  // The node metadata id whose data will be written to connection data.
+  string node_metadata_id = 2;
```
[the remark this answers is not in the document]
gargnupur: Just saw your PR : https://github.com/istio/proxy/pull/2365.. I had structured it before without the use of bootstrap stuff.. leave it for now and change it in a different PR?

src/envoy/tcp/alpn_proxy/config/alpn_proxy.proto
```diff
@@ -0,0 +1,37 @@
...
+
+  // The human readable prefix to use when emitting
+  string stat_prefix = 3;
```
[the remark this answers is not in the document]
gargnupur: Removed

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,269 @@
...
+      stats_(generateStats(stat_prefix, scope)) {}
+
+Network::FilterStatus AlpnProxyFilter::onData(Buffer::Instance& data, bool) {
+  // No work needed if we have read connection protocol and its not the expected
+  // alpn protocol or if alpn protocol is found but we are done reading data.
+  if ((connection_protocol_read_ && !alpn_protocol_found_) ||
+      (alpn_protocol_found_ && read_done_)) {
+    return Network::FilterStatus::Continue;
+  }
+  if (read_callbacks_->connection().nextProtocol() != config_->protocol_) {
```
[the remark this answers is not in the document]
gargnupur: Thanks for looking.. I think what was needed was !alpn_protocol_found...

Please check logic below:

If ALPN protocol is not found then in first time itself, connection_protocol_read_ will be true and will return from that if and alpn_protocol_found will be false. Then in first if of the function it will return because of (connection_protocol_read_ && !alpn_protocol_found_ ) becoming true and it will not read protocol again and again.
if ALPN protocol found is true.. then first time it will read the protocol, not set connection_protocol_read but will set alpn_protocol_found to true. If we have more data to read, then later times, alpn_protocol_found will be true and it will not enter this if (!alpn_protocol_found_ &&
      read_callbacks_->connection().nextProtocol() != config_->protocol_) and not read protocol again and again.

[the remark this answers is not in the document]
kyessenov: I am not sure. Maybe it would be better to use a state machine instead of the bits and just draw a simple chart for the transition diagram?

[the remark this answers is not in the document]
gargnupur: Done.. extended the state machine I had for reading headers to reading protocol too.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,260 @@
...
+    case ConnProtocolNotRead: {
+      if (read_callbacks_->connection().nextProtocol() != config_->protocol_) {
+        conn_state_ = Invalid;
+        config_->stats().alpn_protocol_not_found_.inc();
+        return Network::FilterStatus::Continue;
+      } else {
+        conn_state_ = WriteMetadata;
+        config_->stats().alpn_protocol_found_.inc();
+      }
+    }
```
kyessenov: I think this has to fall-through to WriteMetadata. Can you make that explicit or add a comment?

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,260 @@
...
+          initial_header_buf.c_str());
+  if (absl::gntohl(initial_header->magic) !=
+      AlpnProxyInitialHeader::magic_number) {
+    config_->stats().initial_header_not_found_.inc();
+    conn_state_ = Invalid;
+    return;
+  }
+  proxy_data_length_ = absl::gntohl(initial_header->data_size);
+  // Drain the initial header length bytes read.
+  data.drain(initial_header_length);
```
kyessenov: Can the data buffer be smaller than the initial buffer size?

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,260 @@
...
+        config_->stats().alpn_protocol_found_.inc();
+      }
+    }
+    case WriteMetadata: {
+      // TODO(gargnupur): Try to move this just after alpn protocol is
+      // determined and first onData is called in Downstream filter.
+      if (config_->filter_direction_ == FilterDirection::Downstream) {
+        writeNodeMetadata();
+      }
+    }
```
kyessenov: Same here, this is intended to fall through I think.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,260 @@
...
+}
+
+void AlpnProxyFilter::tryReadInitialProxyHeader(Buffer::Instance& data) {
+  if (conn_state_ != ReadingInitialHeader &&
+      conn_state_ != NeedMoreDataInitialHeader) {
+    return;
+  }
+  const uint32_t initial_header_length = sizeof(AlpnProxyInitialHeader);
+  if (data.length() < initial_header_length) {
+    config_->stats().initial_header_not_found_.inc();
```
kyessenov: This does not seem right to increment header on buffering. This is mostly random depending how TCP data is chunked, right?

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,260 @@
...
+    case ReadingInitialHeader:
+    case NeedMoreDataInitialHeader: {
+      tryReadInitialProxyHeader(data);
+      if (conn_state_ == NeedMoreDataInitialHeader) {
+        return Network::FilterStatus::StopIteration;
+      }
+      if (conn_state_ == Invalid) {
+        return Network::FilterStatus::Continue;
+      }
+    }
```
kyessenov: Same here. If buffer is large, then we need to drain the proxy header.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,260 @@
...
+  Envoy::ProtobufWkt::Struct struct_metadata =
+      Envoy::MessageUtil::anyConvert<Envoy::ProtobufWkt::Struct>(proxy_data);
+  if (config_->filter_direction_ == FilterDirection::Downstream) {
+    writeMetadata(UpstreamDynamicDataKey, struct_metadata);
+  } else {
+    writeMetadata(DownstreamDynamicDataKey, struct_metadata);
+  }
+}
+
+void AlpnProxyFilter::writeMetadata(const std::string key,
```
kyessenov: nit: naming suggestion setMetadata so as not to confuse with network writes.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,260 @@
...
+  }
+  data.drain(proxy_data_length_);
+
+  Envoy::ProtobufWkt::Struct struct_metadata =
+      Envoy::MessageUtil::anyConvert<Envoy::ProtobufWkt::Struct>(proxy_data);
+  if (config_->filter_direction_ == FilterDirection::Downstream) {
+    writeMetadata(UpstreamDynamicDataKey, struct_metadata);
+  } else {
+    writeMetadata(DownstreamDynamicDataKey, struct_metadata);
+  }
```
kyessenov: set conn_state_ = Done?

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
```diff
@@ -0,0 +1,260 @@
...
+        return Network::FilterStatus::Continue;
+      } else {
+        conn_state_ = WriteMetadata;
+        config_->stats().alpn_protocol_found_.inc();
+      }
+    }
+    case WriteMetadata: {
+      // TODO(gargnupur): Try to move this just after alpn protocol is
+      // determined and first onWrite is called in Upstream filter.
+      if (config_->filter_direction_ == FilterDirection::Upstream) {
```
kyessenov: I'm thinking this guard may not be necessary. This essentially validates the expectation that onData is called first for downstream and onWrite is called first upstream. However, this is not important since one of them has to be called first anyways, and both callbacks do the right thing and transition out of the writing state (so not possible to write twice).

In fact, it might be necessary to remove this guard for server-first protocol. If the server (envoy) writes first on the downstream connection, then there is a bug with this guard. This can happen with MySQL, for example.

kyessenov:
Thank you for changing it to use a state machine. This is more readable IMHO.
@lizan are you still requesting changes? It's preventing the merge.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Done.. removed the else stmt.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Fixed

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Changed.. to be in sync with set.. changed readMetadata to getMetadata too.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: State Machine fall through will take care of it.. in default section?

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: I thought that assumption was needed.. Thanks for the info.. Removed the guard.

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: I thought "return Network::FilterStatus::StopIteration;" is doing that and we don't need to drain "data" and buffer bytes read till now..

src/envoy/tcp/alpn_proxy/alpn_proxy.cc
gargnupur: Please check replied above...
I thought "return Network::FilterStatus::StopIteration;" is doing that and we don't need to drain "data" and buffer bytes read till now..

src/envoy/tcp/alpn_proxy/alpn_proxy_initial_header.h
```diff
@@ -0,0 +1,35 @@
...
+namespace AlpnProxy {
+
+// Used with AlpnProxyHeaderProto to be extensible.
+struct AlpnProxyInitialHeader {
+  uint32_t magic;  // Magic number in network byte order. Most significant byte
+                   // is placed first.
+  static const uint32_t magic_number = 0x23071961;  // decimal 587667809
+  uint32_t data_size;  // Size of the data blob in network byte order. Most
+                       // significant byte is placed first.
+} __attribute__((packed));
```
lizan: use PACKED_STRUCT instead of `__attribute__((packed))`

src/envoy/tcp/alpn_proxy/alpn_proxy_initial_header.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Envoy {
+namespace Tcp {
+namespace AlpnProxy {
+
+// Used with AlpnProxyHeaderProto to be extensible.
+struct AlpnProxyInitialHeader {
+  uint32_t magic;  // Magic number in network byte order. Most significant byte
+                   // is placed first.
+  static const uint32_t magic_number = 0x23071961;  // decimal 587667809
```
lizan: how this number is chosen?

src/envoy/tcp/alpn_proxy/alpn_proxy_initial_header.h
gargnupur: Thanks.. Fixed

src/envoy/tcp/alpn_proxy/alpn_proxy_initial_header.h
gargnupur: It's a random number..

gargnupur:
@lizan: One Pager Implementation doc can be found at: https://docs.google.com/document/d/15jKV_muOZKX4CccnA1w-X73SkqGt8cV3a9TBXxzs0TA/edit?usp=sharing.
Agree with name change suggestion, in a meeting today at Google, metadata exchange was suggested too.. I am thinking to do name change in a separate PR, if that works for you?

lizan:
Thanks for putting the doc!

> I am thinking to do name change in a separate PR, if that works for you?

I'm in favor of doing that in this PR. At very least the registered name should be changed because it appears in config.

mandarjog:
I agree, lets rename it to `metadata_exchange` .
The alpn link is there in the implementation, but one could use the same filter within plaintext tcp, `with prior knowledge`.

Re  `magic_number`, we should check against other common magics, like beginning of an http1, http2 transactions. Is there a systematic way to go about this?

If we can't be certain, we should have the ability override `magic_number` through config.

src/envoy/tcp/alpn_proxy/alpn_proxy_initial_header.h
lizan: Can you double check this doesn't conflict with major protocols? (i.e. couldn't be first 4 bytes of TLS, HTTP/1.1, HTTP/2, MySQL, etc..)

gargnupur:
@lizan, @mandarjog , @kyessenov , @PiotrSikora: looks like newly generated random 0x3D230467 magic number should not conflict with Http, http/2, TLS, Redis, SQL... (googling this number doesn't say it's any protocol's magic number.)

Summarizing the findings from chat here(thanks to @kyessenov,@piotrsikora for the help):
1) HTTP: starts with GET, POST.. binaries of these do not match with our magic number.
2) HTTP/2: starts with a predefined prefix 0x505249202a20485454502f322e300d0a0d0a534d0d0a0d0a.
3) TLS: first 8 bits are TLS major version then next 8 bits TLS minor version.it's {3, 3} for TLS 1.2. First 4 bits of our magic number is 1111, so TLS will take a long time to collide.
4) SQL: https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::Handshake. First byte is protocol version
, whose some info can be found here: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/135d0ebe-5c4c-4a94-99bf-1811eccb9f4a?redirectedfrom=MSDN). Doesn't look like it will conflict.
5) Redis:  looks like data type is the first byte : https://redis.io/topics/protocol
does not collide with the magic number.

src/envoy/tcp/alpn_proxy/alpn_proxy_initial_header.h
gargnupur: Done..

mandarjog:
@lizan ping !

src/envoy/tcp/metadata_exchange/config.cc
```diff
@@ -0,0 +1,116 @@
...
+  return createFilterFactory(
+      dynamic_cast<
+          const envoy::tcp::metadataexchange::config::MetadataExchange&>(
+          config),
+      context);
+}
+
+ProtobufTypes::MessagePtr
+MetadataExchangeUpstreamConfigFactory::createEmptyConfigProto() {
+  return ProtobufTypes::MessagePtr{
```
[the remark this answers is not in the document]
gargnupur: Fixed

src/envoy/tcp/metadata_exchange/metadata_exchange.cc
```diff
@@ -0,0 +1,260 @@
...
+  if (data.length() < initial_header_length) {
+    config_->stats().initial_header_not_found_.inc();
+    // Not enough data to read. Wait for it to come.
+    conn_state_ = NeedMoreDataInitialHeader;
+    return;
+  }
+  std::string initial_header_buf = std::string(
+      static_cast<const char*>(data.linearize(initial_header_length)),
+      initial_header_length);
+  const MetadataExchangeInitialHeader* initial_header =
```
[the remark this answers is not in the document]
gargnupur: But shouldn't marking MetadataExchangeInitialHeader struct as packed attribute help with misalignment errors?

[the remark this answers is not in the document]
lizan: No packed struct doesn't mitigate misalignment errors, because you're doing `reinterpret_cast` from a raw pointer which is not guaranteed to be aligned.

Since you're already doing a copy here, why not just do:
`
MetadataExchangeInitialHeader initial_header;
data.copyOut(0, initial_header_length, &initial_header);
`
then where you copied to is aligned.

PR #2331, opened 2019-07-30
Add prometheus stats filter

mandarjog:
This is to get review started, Will add tests.

1. pluginContext.log delegates to rootContext.report()
2. SimpleStat abstraction counters/historgrams/gauges using lower level metrics API.
3. getStandardMappings functions abstract mapping.
4. Most of the code in common dir comes from stackdriver/common, with some additions.


extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+  inline void record(const Common::RequestInfo& request_info) {
+    recordMetric(metric_id_, valueFn_(request_info));
+  };
+
+ private:
+  uint32_t metric_id_;
+  valueExtractorFn valueFn_;
+};
+
+using SimpleStatSharedPtr = std::shared_ptr<SimpleStat>;
```
jplevyak: A SimpleStat is copy-able so you don't need to use a shared_ptr.   Just treat it as an immediate and don't worry about lifetime.

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+    auto metric_id = metric_->resolveWithFields(vals);
+    logDebug(absl::StrCat(__FUNCTION__, ":", __LINE__, ":", metric_id, "/",
+                          source.name(), dest.name(), requestInfo.mTLS));
+    return std::make_shared<SimpleStat>(metric_id, valueFn_);
+  };
+
+ private:
+  std::string name_;
+  valueExtractorFn valueFn_;
+  MetricSharedPtr metric_;
```
jplevyak: Is this just to make it copy-able?  It looks like this object has only one owner (StatGen), so shared_ptr is not necessary.
Could you just make it immediate?

mandarjog: Metric does not have a default constructor, if I use immediate member, I will have to initialize with something dummy and then replace.
Any ideas ?

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+              (outbound ? unknown : (requestInfo.mTLS ? vMTLS : vNone))),
+      MAPPING_SYM("permissive_response_code",
+                  requestInfo.permissive_response_code, vNone),
+      MAPPING_SYM("permissive_response_policyid",
+                  requestInfo.permissive_response_policyid, vNone)};
+}
+
+// StatGen is dimensioned using standard Istio dimensions.
+// Standard Istio metrics have the following dimensions
+class StatGen {
```
silentdai: It's not clear to me whether `StatGen` is supposed to be copyable or moveable, or shared. I cannot image

Some indicator could be
declare fields as const
explicitly declare copy constructor/move constructor as default.

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+    logDebug(absl::StrCat(__FUNCTION__, ":", __LINE__, ":", metric_id, "/",
+                          source.name(), dest.name(), requestInfo.mTLS));
+    return std::make_shared<SimpleStat>(metric_id, valueFn_);
+  };
+
+ private:
+  std::string name_;
+  valueExtractorFn valueFn_;
+  MetricSharedPtr metric_;
+  Mappings mappings_;
```
silentdai: mappings_ is private and I didn't see it assigned... Did I lose anything?

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+      : name_(name), valueFn_(valueFn) {
+    metric_ =
+        std::make_shared<Metric>(metricType, name, mappings_.metricTags());
+  };
+
+  StatGen() = delete;
+  inline StringView name() const { return name_; };
+
+  // return a SimpleSharedPtr
+  SimpleStatSharedPtr resolve(bool outbound, const common::NodeInfo& source,
```
silentdai: To me it seems `resolve` is a factory method. It make more sense to return unique_ptr.
The caller of the method should feel free to promote to shared_ptr if the caller decides to share

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+    auto vals = mappings_.eval(outbound, source, dest, requestInfo);
+    auto metric_id = metric_->resolveWithFields(vals);
+    logDebug(absl::StrCat(__FUNCTION__, ":", __LINE__, ":", metric_id, "/",
+                          source.name(), dest.name(), requestInfo.mTLS));
+    return std::make_shared<SimpleStat>(metric_id, valueFn_);
+  };
+
+ private:
+  std::string name_;
+  valueExtractorFn valueFn_;
```
silentdai: naming: a type should start with capital letter

extensions/stats/config.h
jplevyak: We can add one and an initialize() method.

jplevyak: We can add one and an initialize() method.

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+
+using google::protobuf::util::JsonParseOptions;
+using google::protobuf::util::Status;
+
+struct Node {
+  common::NodeInfo node_info;
+  // key computed from the
+  std::string key;
+
+  Node(common::NodeInfo nodeInfo) {
```
silentdai: A single param constructor usually declare as `explicit` to avoid auto converting from `common::NodeInfo`

silentdai: also naming: `node_info`

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+  Mapping() = delete;
+
+  std::string name_;
+  mapperFn mapper_;
+};
+std::vector<Mapping> getStandardMappings();
+
+class Mappings {
+ public:
+  Mappings(std::vector<Mapping> mappings) : mappings_(mappings){};
```
silentdai: 1. make it explicit
2. it's rare to pass a constructor a vector copy. If that is the intention, you can do init as ` mappings_(std::move(mappings))`

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+
+  Mappings() : Mappings(getStandardMappings()) {}
```
silentdai: oops. I get the why `mappings_` is not assigned when I read here. Could you make it more explicit? Such as delete this one and create a static method, something like Mappings createFromStandardMappings();

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+
+  // metricTags converts mappings into ordered tags
+  std::vector<MetricTag> metricTags() {
+    std::vector<MetricTag> ret;
+
+    ret.reserve(mappings_.size());
+    for (const auto& mapping : mappings_) {
+      ret.push_back({mapping.name_, MetricTag::TagType::String});
```
silentdai: I lost the type of MetricTag but could you try `ret.emplace_back(mapping.name_, MetricTag::TagType::String)`?

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+        [](bool ABSL_ATTRIBUTE_UNUSED outbound,                          \
+           const common::NodeInfo& ABSL_ATTRIBUTE_UNUSED source,         \
+           const common::NodeInfo& ABSL_ATTRIBUTE_UNUSED dest,           \
+           const Common::RequestInfo& ABSL_ATTRIBUTE_UNUSED requestInfo) \
+            -> std::string {                                             \
+          auto source_labels = source.labels();                          \
+          auto dest_labels = dest.labels();                              \
+          auto val = (expr);                                             \
+          logDebug(absl::StrCat((key), "=", ToString(val)));             \
+          return (SYMIFEMPTY(ToString(val), (sym)));                     \
```
silentdai: kind of complex here but I think ToString execute twice

extensions/stats/config.h
```diff
@@ -0,0 +1,384 @@
...
+      vals.push_back(mapping.mapper_(outbound, source, dest, requestInfo));
+    }
+    return vals;
+  }
+
+ private:
+  std::vector<Mapping> mappings_;
+};
+
+#define MAPPING_SYM(key, expr, sym)                                      \
```
silentdai: I didn't see the lambda capture anything. Why there is lambda for each key? Can't you just create a lambda and use it every where?

extensions/stats/config.h
mandarjog: keeping it as is now, will come back to it later.

extensions/stats/config.h
mandarjog: mappings_ default constructor creates istioStandard metrics mappings.

extensions/stats/config.h
kyessenov: It would be simpler to define a bunch of static functions instead of building lambdas. It is a lot easier to debug functions with names.

kyessenov: Also, the arguments should be collected into a context object here I think, and captured by reference.

extensions/stats/config.cc
```diff
@@ -0,0 +1,161 @@
...
+  NodeInfoSharedPtr node_info = std::make_shared<common::NodeInfo>();
+  // Missed the cache
+  auto metadata = getMetadataStruct(MetadataType::Request, peer_metadata_key);
+  auto status = Common::extractNodeMetadata(metadata, node_info.get());
+  if (status != Status::OK) {
+    logWarn("cannot parse peer node metadata " + metadata.DebugString() + ": " +
+            status.ToString());
+  }
+
+  auto new_nodeinfo = std::make_shared<Node>(node_info);
```
kyessenov: Why shared? The cache owns the node info, right? Shared pointers are generally expensive, and not advised unless required.

extensions/stats/config.h
```diff
@@ -0,0 +1,386 @@
...
+
+// END WASM_PROLOG
+
+namespace Stats {
+
+using StringView = absl::string_view;
+
+constexpr StringView Sep = "#";
+
+// The following need to be std::strings because the receiver expects a string.
```
kyessenov: Seems to be used in StrAppend only... Why not StrCat?

extensions/stats/config.h
```diff
@@ -0,0 +1,386 @@
...
+ private:
+  std::string name_;
+  ValueExtractorFn value_fn_;
+  Mappings mappings_;
+  // TODO convert shared pointer to a direct struct once Metric has a
+  // constructor.
+  MetricSharedPtr metric_;
+};
+
+class RequestsTotal : public StatGen {
```
bianpengyuan: It seems heavy to subclass StatsGen for each stat. Why not simply make each stat an instance of StatGen and pass in a simple lambda as valueFn for initialization?

extensions/common/context.cc
```diff
@@ -0,0 +1,82 @@
...
+
+// WASM_PROLOG
+#ifndef NULL_PLUGIN
+#include "api/wasm/cpp/proxy_wasm_intrinsics.h"
+
+#else  // NULL_PLUGIN
+
+#include "extensions/common/wasm/null/null.h"
+
+namespace Envoy {
```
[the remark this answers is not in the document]
mandarjog: `::WASM::Common` ?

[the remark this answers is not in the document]
bianpengyuan: Yeah, that looks good to me.

extensions/stats/config.cc
silentdai: If the goal is to owned by map but not shard, as we chatted, `Prefer absl::flat_hash_map<Key, std::unique_ptr<Value>> when pointer stability of values is needed.` but we can start with shard_ptr since value as unique_ptr restricted the usage of map api. This could be resolved in future PRs

extensions/stats/config.cc
```diff
@@ -0,0 +1,161 @@
...
+    metadataKey = Common::kUpstreamMetadataKey;
+    reporter = "client";
+  }
+
+  auto peer = node_info_cache_.getPeerById(metadataIdKey, metadataKey);
+
+  // check if this peer has associated metrics
+  // These fields should vary independently of peer properties.
+  // TODO derive this from the mapper
+  auto metric_base_key =
```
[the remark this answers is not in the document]
kyessenov: We can override a hash for absl containers if it helps to avoid concatenating strings

mandarjog:
Please take a look. Thanks.

extensions/common/context_test.cc
```diff
@@ -0,0 +1,151 @@
...
+  Status status = extractNodeMetadata(metadata_struct, &node_info);
+  EXPECT_EQ(status, Status::OK);
+}
+
+}  // namespace Common
+
+// WASM_EPILOG
+#ifdef NULL_PLUGIN
+}  // namespace Wasm
+#endif
```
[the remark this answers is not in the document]
kyessenov: We should run clang-format here...

extensions/stats/plugin.h
```diff
@@ -0,0 +1,405 @@
...
+  X(destination_service_name)       \
+  X(destination_service_namespace)  \
+  X(request_protocol)               \
+  X(response_code)                  \
+  X(response_flags)                 \
+  X(connection_security_policy)
+
+struct IstioDimensions {
+#define X(name) std::string(name);
+  ISTIO_DIMENSIONS
```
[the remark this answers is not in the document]
kyessenov: I'm not sure the purpose of a macro here. Is there anything else going on here besides defining a bunch of string fields in a struct?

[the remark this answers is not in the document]
kyessenov: Instead of X, it might be better for readability to make it a parameter to another macro, e.g. https://github.com/envoyproxy/envoy/blob/master/include/envoy/http/header_map.h#L264

extensions/common/node_info.proto
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package common;
```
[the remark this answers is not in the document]
mandarjog: ack

extensions/stats/config.proto
```diff
@@ -0,0 +1,28 @@
...
+package stats;
+
+message PluginConfig {
+  // Indicates whether the module should act as inbound or outbound reporter.
+  enum Direction {
+    INBOUND = 0;
+    OUTBOUND = 1;
+  }
+  Direction direction = 1;
+  bool debug = 2;
```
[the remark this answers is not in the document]
mandarjog: until this is added thru admin interface, debug here works great.

extensions/stats/plugin.h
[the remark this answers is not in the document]
mandarjog: it gives to_string, equality, list_values *and* the struct itself. I have renamed X to something better.

extensions/stats/plugin.cc
```diff
@@ -0,0 +1,151 @@
...
+      outbound_ ? peer.node_info : local_node_info_;
+
+  ::Wasm::Common::RequestContext ctx{outbound_, source_node_info,
+                                     destination_node_info, request_info};
+
+  // check if this peer has associated metrics
+  // These fields should vary independently of peer properties.
+  // TODO derive this from the mapper
+  auto metric_base_key =
+      absl::StrCat(peer.key, Sep, request_info.request_protocol, Sep,
```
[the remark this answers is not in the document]
mandarjog: principal is the property of workload, so it is redundant wrt other identifiers.

mandarjog:
I have addressed all comments.

extensions/stats/plugin.cc
```diff
@@ -0,0 +1,170 @@
...
+const Node& NodeInfoCache::getPeerById(StringView peer_metadata_id_key,
+                                       StringView peer_metadata_key) {
+  auto peer_id =
+      getMetadataStringValue(MetadataType::Request, peer_metadata_id_key);
+  auto nodeinfo_it = cache_.find(peer_id);
+  if (nodeinfo_it != cache_.end()) {
+    return nodeinfo_it->second;
+  }
+
+  // TODO kick out some elements from cache here if size == MAX
```
kyessenov: Please file an issue for this. Otherwise, this is leaking :)

extensions/stats/plugin.h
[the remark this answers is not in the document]
kyessenov: I was referring to using macro parameters instead of def/undef cycles.

extensions/stats/plugin_test.cc
```diff
@@ -0,0 +1,72 @@
...
+#endif  // NULL_PLUGIN
+
+// END WASM_PROLOG
+
+namespace Stats {
+
+TEST(IstioDimensions, Hash) {
+  IstioDimensions d1;
+  IstioDimensions d2{.request_protocol = "grpc"};
+  IstioDimensions d3{.request_protocol = "grpc", .response_code = "200"};
```
kyessenov: this is strange syntax. is this referring to field with a prefix or is the dot a part of the field name?

mandarjog: This is a literal struct. Is there a better way?

kyessenov: Ok, got it. Looks fine.

extensions/stats/plugin.h
```diff
@@ -0,0 +1,443 @@
...
+    return (
+#define DIMENSION(name) lhs.name == rhs.name&&
+        ISTIO_DIMENSIONS
+#undef DIMENSION
+            lhs.outbound == rhs.outbound);
+  }
+};
+
+// Node holds node_info proto and a computed key
+struct Node {
```
bianpengyuan: The Node is not used any more, no? Did I miss its usage somewhere? If not used, then the key concat is also unnecessary right?

extensions/stats/plugin.h
```diff
@@ -74,70 +74,55 @@ using google::protobuf::util::Status;
...
+#define REPLACE_VALUES(name) \
+  absl::StrReplaceAll(name, HACK_VALUES_REPLACEMENTS),
+    return std::vector<std::string>{STD_ISTIO_DIMENSIONS(REPLACE_VALUES)};
   }
 
   void setFieldsUnknownIfEmpty() {
-#define DIMENSION(name) \
-  if ((name).empty()) { \
-    (name) = unknown;   \
+#define SET_IF_EMPTY(name) \
```
bianpengyuan: Should this be `if not unknown`? For example if for one request a field is known and filled, the in following request the field becomes unknown, its value won't be overrode by `unknown` right?

extensions/stats/plugin.cc
```diff
@@ -87,13 +87,18 @@ void PluginRootContext::report(
       CTXDEBUG("metricKey cache hit ", istio_dimensions_.debug_key(),
                ", stat=", stat.metric_id_, stats_it->first.to_string());
     }
-    incrementMetric(cache_hits_, 1);
+    cache_hits_accumulator_++;
+    if (cache_hits_accumulator_ == 100) {
+      incrementMetric(cache_hits_, cache_hits_accumulator_);
```
bianpengyuan: Why accumulate? Is it expensive?

extensions/stats/plugin.cc
```diff
@@ -137,13 +123,18 @@ const Node& NodeInfoCache::getPeerById(StringView peer_metadata_id_key,
...
     auto it = cache_.begin();
-    for (int i = 0; i < int(max_cache_size_ / 4); ++i, ++it) {
-    }
-    cache_.erase(cache_.begin(), it);
+    cache_.erase(cache_.begin(), std::next(it, max_cache_size_ / 4));
     logInfo(absl::StrCat("cleaned cache, new cache_size:", cache_.size()));
   }
 
-  initializeNode(peer_metadata_key, &(cache_[peer_id]));
+  auto metadata = getMetadataStruct(MetadataType::Request, peer_metadata_key);
```
bianpengyuan: `const auto& metadata`? SD filter is also written in this way, I need to change that as well...

extensions/stats/plugin.cc
mandarjog: Yes, this is adding one metric call into the host for every report. We do not need that level of fidelity. updating every 100 calls is ok.

extensions/stats/plugin.h
bianpengyuan: I think you might need an extra data structure to record which field is filled if dimensions are shared across requests?

extensions/stats/plugin_test.cc
rishigits: This syntax is causing warnings on Gcc 7.3x:

"extensions/stats/plugin_test.cc:44:48: sorry, unimplemented: non-trivial designated initializers not supported
   IstioDimensions d2{.request_protocol = "grpc"};"

Is it possible to do something as follows:

/=============================/
  IstioDimensions d1,d2,d3,d4,d5,d6,d7,d8;
  d2.request_protocol = "grpc";
  d3.request_protocol = "grpc";
  d3.response_code = "200";
  d4.request_protocol = "grpc";
  d4.response_code = "400";
  d5.request_protocol = "grpc";
  d5.source_app = "app_source";
  d6.request_protocol = "grpc";
  d6.source_app = "app_source";
  d6.source_version = "v2";
  d7.outbound = true;d7.request_protocol = "grpc";
  d7.source_app = "app_source";d7.source_version = "v2";
  d8.outbound = true;d8.request_protocol = "grpc";
  d8.source_app = "app_source";d8.source_version = "v2";
/=============================/
Thx.

PR #2339, opened 2019-08-06
Update common files.

geeknoid:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #2340, opened 2019-08-07
Revert "Update common files. (#2339)"

PiotrSikora:
This reverts commit 38a24757efd4687bdf3da1daddcf279070057d66.


PiotrSikora:
@geeknoid could you please stop messing up with those files? Your previous update (#2251) broke  linters and other checks (`make check` became a no-op), this update (#2339) broke tests (`make test` became a no-op). What's the value of having those common files? Our CI is broken as it is, this doesn't help at all.

kyessenov:
Common files bring little value in this repo. We need clang/bazel which is entirely missing from common files.

PR #2312, opened 2019-07-17
Update Envoy-WASM SHA to latest.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

PR #2342, opened 2019-08-07
Fix static linking.

PiotrSikora:
Broken in #2312.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,5 +1,7 @@
-# Copied from https://github.com/envoyproxy/envoy/blob/master/tools/bazel.rc
+# =====================================================================
 # Envoy specific Bazel build/test options.
+# Copied from: https://github.com/envoyproxy/envoy/blob/master/.bazelrc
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -8,11 +10,15 @@
 startup --host_jvm_args=-Xmx512m
 
 build --workspace_status_command=tools/bazel_get_workspace_status
-# enable path normalization by default. See https://github.com/envoyproxy/envoy/pull/6519
-build --define path_normalization_by_default=true
 build --host_force_python=PY2
 
+# Link (statically) against libstdc++.
+build --action_env=BAZEL_LINKLIBS=-l%:libstdc++.a
+build --action_env=BAZEL_LINKOPTS=-lm:-static-libgcc
```
PiotrSikora: I synchronized the other options with upstream while at it, but those lines are the real fix.

PiotrSikora:
Rebased on top of `master` to make sure all tests pass.

PR #2343, opened 2019-08-07
Revert "using sha instraf of git_sha (#2338)"

PiotrSikora:
This reverts commit 41943e54ea595930fdd7c2296067b58556e286f8.


duderino:
@PiotrSikora why was this one reverted?

PiotrSikora:
Because it's was breaking postsubmit (one of the things that was breaking it, not the only one).
`$GIT_SHA` exists, `$SHA` doesn't.

PR #2347, opened 2019-08-07
Properly export PATH, CC and CXX variables to Bazel.

PiotrSikora:


kyessenov:
Eventually, we should pull the trigger and move proxy docker here. I'm about 85% there to tolerate the pain of moving. The whole dance around pushing binaries to some bucket somewhere is entirely unnecessary.

PR #2349, opened 2019-08-08
Migrate SD to use extension common package

bianpengyuan:
This is a pure mechanical change and clean up to migrate sd plugin to use the extension/common package.


extensions/stackdriver/metric/registry.cc
```diff
@@ -34,13 +36,14 @@ MonitoredResource getMonitoredResource(
     const std::string &monitored_resource_type,
     const NodeInfo &local_node_info) {
   google::api::MonitoredResource monitored_resource;
-  monitored_resource.set_type(kContainerMonitoredResource);
+  monitored_resource.set_type(monitored_resource_type);
+  auto platform_metadata = local_node_info.platform_metadata();
```
silentdai: The left side doesn't copy platform_metadata but your PR does the copy. Maybe just because you want to use operator[]?

extensions/stackdriver/metric/registry.cc
```diff
@@ -17,15 +17,17 @@
 #include "extensions/stackdriver/common/constants.h"
 #include "google/api/monitored_resource.pb.h"
 
+#include <typeinfo>
```
silentdai: nits: IIRC bracket include header should be next to the top most "registry.h" 

Feel free to ignore if it's not.

silentdai: Also: where is it used?

extensions/stackdriver/metric/registry.cc
bianpengyuan: Yes, for code simplicity. This code path is only executed once at initialization so a copy is fine.

extensions/stackdriver/metric/registry.cc
bianpengyuan: Oops, yeah, this is left from debugging. Will clean up in following PR. Thanks!

PR #2352, opened 2019-08-09
Test the release scripts.

PiotrSikora:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,8 @@ BAZEL_TARGETS ?= //...
 SANITIZER_EXCLUSIONS ?= -test/integration:mixer_fault_test
 HUB ?=
 TAG ?=
+GCS_BUCKET ?= gs://istio-build/proxy
+
 ifeq "$(origin CC)" "default"
 CC := clang
 endif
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -68,11 +70,17 @@ lint:
...
-	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/push-debian.sh -p $(ARTIFACTS_DIR)
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/push-debian.sh -p "$(ARTIFACTS_GCS_PATH)" -o "$(ARTIFACTS_DIR)"
 
-deb:
-	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) //tools/deb:istio-proxy
+test_release:
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/release-binary.sh -d none -i
+
+push_release:
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/release-binary.sh -d "$(GCS_BUCKET)"
```
duderino: I think the only way of triggering this is with an explicit 'make push_release', right?   It couldn't be called as a side effect of any test, right?

PiotrSikora: Correct.

scripts/release-binary.sh
```diff
@@ -23,30 +23,45 @@ export PATH=/usr/lib/llvm-8/bin:$PATH
...
 
+# Verify that we're building binaries on Ubuntu 16.04 (Xenial).
+CHECK=1
+
 function usage() {
   echo "$0
-    -d  The bucket name to store proxy binary (optional)"
+    -d  The bucket name to store proxy binary (optional).
+        By default, the GCS bucket is set to \"gs://istio-build/proxy\".
+        Use \"none\" to skip the GCS upload.
```
duderino: i wonder if we should make the default 'none'?

PiotrSikora: Yes, but right now `proxy-postsubmit` calls it without arguments, so it would break again. We can do it in subsequent change, though.

PR #2353, opened 2019-08-09
Remove hardcoded defaults from the release scripts.

PiotrSikora:


PiotrSikora:
@kyessenov well, we should be building Docker images here as well, and be done with the release process for the Proxy. I kind of like that we publish binaries and artifacts for others to consume, though.

PR #2354, opened 2019-08-12
Expose telemetry on TCP new Connection callback

nrjpoddar:
Expose Telemetry for BlackHole cluster

Fixes: https://github.com/istio/istio/issues/15867

### DIFF
--- a/include/istio/control/tcp/request_handler.h
+++ b/include/istio/control/tcp/request_handler.h
@@ -29,6 +29,9 @@ class RequestHandler {
  public:
   virtual ~RequestHandler() {}
 
+  // Builds shared attributes required for both Check and Report calls
+  virtual void BuildCheckAttributes(CheckData* check_data) = 0;
+
   // Perform a Check call. It will:
   // * extract downstream tcp connection attributes
   // * check config, make a Check call if necessary.
--- a/src/envoy/tcp/mixer/filter.cc
+++ b/src/envoy/tcp/mixer/filter.cc
@@ -55,8 +55,6 @@ void Filter::cancelCheck() {
 
 // Makes a Check() call to Mixer.
 void Filter::callCheck() {
-  handler_ = control_.controller()->CreateRequestHandler();
-
   state_ = State::Calling;
   filter_callbacks_->connection().readDisable(true);
   calling_check_ = true;
@@ -134,6 +132,8 @@ Network::FilterStatus Filter::onNewConnection() {
                  filter_callbacks_->connection().remoteAddress()->asString(),
                  filter_callbacks_->connection().localAddress()->asString());
 
+  handler_ = control_.controller()->CreateRequestHandler();
+  handler_->BuildCheckAttributes(this);
   // Wait until onData() is invoked.
   return Network::FilterStatus::Continue;
 }
@@ -211,6 +211,9 @@ bool Filter::GetDestinationIpPort(std::string *str_ip, int *port) const {
       filter_callbacks_->upstreamHost()->address()) {
     return Utils::GetIpPort(filter_callbacks_->upstreamHost()->address()->ip(),
                             str_ip, port);
+  } else {
+    return Utils::GetIpPort(
+        filter_callbacks_->connection().localAddress()->ip(), str_ip, port);
   }
   return false;
 }
--- a/src/istio/control/tcp/request_handler_impl.cc
+++ b/src/istio/control/tcp/request_handler_impl.cc
@@ -35,8 +35,7 @@ RequestHandlerImpl::RequestHandlerImpl(
       client_context_(client_context),
       last_report_info_{0ULL, 0ULL, std::chrono::nanoseconds::zero()} {}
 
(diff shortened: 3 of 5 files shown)


nrjpoddar:
@kyessenov PTAL

With these changes along side my Pilot changes that made it is master, I get the following telemetry for BlackHole cluster when TCP proxy is invoked:

```json

        "metric": {
          "__name__": "istio_tcp_connections_closed_total",
          "connection_security_policy": "unknown",
          "destination_app": "unknown",
          "destination_ip": "52.22.188.80",
          "destination_principal": "unknown",
          "destination_service": "BlackHoleCluster",
          "destination_service_name": "BlackHoleCluster",
          "destination_service_namespace": "unknown",
          "destination_version": "unknown",
          "destination_workload": "unknown",
          "destination_workload_namespace": "unknown",
          "instance": "100.96.2.164:42422",
          "job": "istio-mesh",
          "reporter": "source",
          "response_flags": "-",
          "source_app": "sleep",
          "source_principal": "unknown",
          "source_version": "unknown",
          "source_workload": "sleep",
          "source_workload_namespace": "default"
        },
        "value": [
          1565669245.171,
          "1"
        ]
      }
```

nrjpoddar:
Fixing/adding new tests but if I can get your feedback @kyessenov that would reduce churn!

include/istio/control/tcp/request_handler.h
```diff
@@ -29,6 +29,9 @@ class RequestHandler {
  public:
   virtual ~RequestHandler() {}
 
+  // Builds shared attributes required for both Check and Report calls
+  virtual void BuildSharedAttributes(CheckData* check_data) = 0;
```
nrjpoddar: I kept `CheckData` as arg here but open to changing the interface to be completely independent of Check call. I was trying to do the minimal changes needed to get the PR out and get feedback.

kyessenov: You can call it BuildCheckAttributes since this just moves all the build logic for the initial attributes into a separate function.

src/envoy/tcp/mixer/filter.cc
```diff
@@ -211,6 +211,9 @@ bool Filter::GetDestinationIpPort(std::string *str_ip, int *port) const {
       filter_callbacks_->upstreamHost()->address()) {
     return Utils::GetIpPort(filter_callbacks_->upstreamHost()->address()->ip(),
                             str_ip, port);
+  } else {
```
kyessenov: Hmm, how did this work before for server side reports? Is this being overridden somewhere later?
I'm a bit worried since this changes the telemetry for client-side reports for destination IP.

include/istio/control/tcp/request_handler.h
nrjpoddar: Good point, will do that!

src/envoy/tcp/mixer/filter.cc
nrjpoddar: So the change only matters when there's no assigned upstream cluster which gets triggered for BlackHoleCluster or other clusters which are not defined or clusters with no endpoints. For every other case the `if` clause should get invoked and preserve old behavior.

I think this is a net improvement on the existing behavior as it ensures that we always have a DestinationIP where as in the existing implementation there will be scenarios where destination IP and port will be "unknown" or "0".

kyessenov: I think the cluster is not assigned at the moment of the call here. The upstream pool sends a callback after the connection is allowed (or perhaps, this is a race). If that's the case, then there is a change in the behavior.

nrjpoddar: I see. If my understanding is correct then for BlackHole a cluster will never get assigned as without this logic I was never getting Destination IP in my telemetry.

For other cases, you're right this might be a race. I will add some logging and verify if that's the case.

nrjpoddar:
@kyessenov I added this debug code https://github.com/istio/proxy/pull/2354/commits/a215ceb2739091934dad743a052a38185a03e771 and always see either one of the two logs getting hit but never both for any of the use cases I covered.

Please provide an example scenario that you think will/might cause both of these logs to get hit within the context of one request.

src/envoy/tcp/mixer/filter.cc
kyessenov: This is only called in onLog, so the decision from the upstream pool should have already been made by then.

nrjpoddar:
Thanks @kyessenov for the review. How do I resolve/get around the 10 reviewers that were automatically assigned?

sdake:
@PiotrSikora hi - this PR is a candidate for cherrypick to support @nrjpoddar 's use case.  Do you confirm or reject?

Cheers
-steve

nrjpoddar:
> @PiotrSikora hi - this PR is a candidate for cherrypick to support @nrjpoddar 's use case. Do you confirm or reject?
> 
> Cheers
> -steve

@PiotrSikora FYI all the work for istio/istio i.e. Pilot is already in release-1.3 via https://github.com/istio/istio/pull/15512

PR #2358, opened 2019-08-13
release: test upstream

kyessenov:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,7 @@ jobs:
       - run: make lint
       - run: make build
       - run: make test
+      - run: make upstream_test
       - save_cache:
           key: linux_fastbuild-bazel-cache-{{ checksum "WORKSPACE" }}
           paths:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       - run: make lint
       - run: make build
       - run: make test
+      - run: make upstream_test
```
PiotrSikora: Nit: `test_upstream` instead of `upstream_test` to match other targets.

But last time I tried doing this, it wasn't as easy as this PR.

kyessenov: What's hard about it?

kyessenov: 41 tests failing with no artifacts. Looks like fun...

lizan:
Oh, actually, https://github.com/envoyproxy/envoy/pull/8488 should make this PR possible without much change. You just need to sync envoy sha.

kyessenov:
Given that we have envoy/envoy-wasm and istio/envoy, it's not clear to me this is the right place to test upstream.

lizan:
It's more about testing the build configuration etc in istio-proxy is right or not, so probably just run a few of upstream test? e.g. //test/integration/... and //test/server/...

PR #2359, opened 2019-08-14
Record SD metric and add an integration test

bianpengyuan:
Add metric recording logic, also add an integration test with a fake stackdriver stub.


test/envoye2e/stackdriver_plugin/stackdriver_plugin_test.go
```diff
@@ -0,0 +1,159 @@
...
+      vm: "envoy.wasm.vm.null"
+      code:
+        inline_string: "envoy.wasm.null.stackdriver"
+    configuration: >-
+      {
+        "kind": "INBOUND",
+        "monitoringEndpoint": "localhost:12312",
+      }`
+
+const outboundNodeMetadata = `istio.io/metadata:
```
douglas-reid: before we add any more deps on `istio.io/metadata`, we need to resolve the issue of whether or not we are using that structure.

@kyessenov @mandarjog

bianpengyuan: I will wait until we are settled on metadata stuff before pushing this.

bianpengyuan:
@mandarjog @kyessenov Please take a look, thanks!

bianpengyuan:
@mandarjog @kyessenov @douglas-reid Please take a look thanks!

Makefile
```diff
@@ -62,6 +62,7 @@ clean:
 
 test:
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) test $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_DEV) $(BAZEL_TARGETS)
+	GO111MODULE=on go test ./...
```
douglas-reid: just so I understand: we can't use `bazel test` to run the golang tests with `GO111MODULE=on` ?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -18,19 +18,17 @@ syntax = "proto3";
...
-  //   * inbound reporter writes access logs.
-  enum ReporterKind {
-    INBOUND = 0;
-    OUTBOUND = 1;
-  }
-  ReporterKind kind = 1;
-
   // Controls whether to export access log.
-  bool disable_access_logging = 2;
+  bool disable_access_logging = 1;
```
douglas-reid: nit: should we strive to have positive booleans (ex: `enable_access_logging` )?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -18,19 +18,17 @@ syntax = "proto3";
...
 
   // FQDN of destination service that the request routed to.
-  string destination_service_name = 3;
+  string destination_service_name = 2;
```
douglas-reid: can we provide an example here?  and maybe clarify/distinguish between `destination_service_name` and `destination_service_host`?

douglas-reid: Also, is this required?  what is the implication of this information not being populated?

extensions/stackdriver/metric/record.cc
```diff
@@ -24,40 +24,13 @@ constexpr double kNanosecondsPerMillisecond = 1000000.0;
...
-void record(
-    const stackdriver::config::v1alpha1::PluginConfig::ReporterKind &kind,
-    const ::wasm::common::NodeInfo &local_node_info,
-    const ::wasm::common::NodeInfo &peer_node_info,
-    const ::Wasm::Common::RequestInfo &request_info) {
+void record(bool is_outbound, const ::wasm::common::NodeInfo &local_node_info,
+            const ::wasm::common::NodeInfo &peer_node_info,
+            const ::Wasm::Common::RequestInfo &request_info) {
   double latency_ms =
       double(request_info.end_timestamp - request_info.start_timestamp) /
```
douglas-reid: curious: is `start_timestamp` here first (downstream) byte received? and `end_timestamp` last (downstream) byte sent?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -18,19 +18,17 @@ syntax = "proto3";
...
+
+  // The endpoint that plugin targets for metric reporting. If not specified,
+  // the default Stackdriver monitoring endpoint will be used.
+  string monitoring_endpoint = 3;
```
douglas-reid: just curious: how are we doing auth for non-default endpoints?

extensions/stackdriver/stackdriver.h
```diff
@@ -58,8 +58,8 @@ class StackdriverRootContext : public RootContext {
   void onStart(std::unique_ptr<WasmData>) override;
   void onTick() override;
 
-  // Get reporter kind of this filter from plugin config.
-  stackdriver::config::v1alpha1::PluginConfig::ReporterKind reporterKind();
+  // Get direction of traffic relative to this proxy.
+  PluginDirection direction();
```
douglas-reid: is this used other than to check whether or not something is outbound?  could this be `boolean isOutbound()` instead?

test/envoye2e/stackdriver_plugin/fake_stackdriver/fake_stackdriver.go
```diff
@@ -0,0 +1,101 @@
...
+	grpc "google.golang.org/grpc"
+
+	empty "github.com/golang/protobuf/ptypes/empty"
+	"google.golang.org/genproto/googleapis/api/metric"
+	"google.golang.org/genproto/googleapis/api/monitoredres"
+	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
+)
+
+// FakeStackdriverServer is a fake stackdriver server which implements all of monitoring v3 service method.
+type FakeStackdriverServer struct {
```
douglas-reid: could you just have:

```golang
type FakeSDServer struct {
  monitoringpb.MetricsServiceServer
  
  RcvReq chan *monitoringpb.CreateTimeSeriesRequest
}

// CreateTimeSeries implements CreateTimeSeries method.
func (s *FakeStackdriverServer) CreateTimeSeries(ctx context.Context, req *monitoringpb.CreateTimeSeriesRequest) (*empty.Empty, error) {
	s.RcvReq <- req
	return &empty.Empty{}, nil
}

```

instead of having all of the empty interface methods here?

bianpengyuan: It won't work:
```
cannot use fsds (type *FakeStackdriverServer) as type monitoring.MetricServiceServer in argument to monitoring.RegisterMetricServiceServer:
	*FakeStackdriverServer does not implement monitoring.MetricServiceServer (missing CreateMetricDescriptor method)
```

extensions/stackdriver/stackdriver.h
bianpengyuan: Ah, good point. Done.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
bianpengyuan: Opencensus library uses default GCP credential.

Makefile
bianpengyuan: No, it has nothing to do with go mod. I am just too lazy to add bazel rules for the go test part :). If it does not look good to use go native build I could try to add bazel rules for golang code.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
bianpengyuan: proto3 does not allow default value, and we want to enable access log by default with zero config, so `disable_access_logging` might be a better fit here?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
bianpengyuan: No it is not required and it suppose to be filled by control plane. If not provided, request host will be used instead.

extensions/stackdriver/metric/record.cc
bianpengyuan: start_timestamp is the time that the first byte is processed by the sd filter and end_timestamp is the time when sd filter finishes processing the request (when onLog is called). The latency should really not be computed in this way, instead it should be fetched from envoy API: https://github.com/istio/proxy/blob/2011544571a28461331322ed2274c9d7b6693294/extensions/common/context.cc#L81

Makefile
kyessenov: We can add a bazel build eventually. Envoy already comes with all go set-up needed. Just no reason to force it yet.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
kyessenov: we can add an empty embedded message as a control. the message can be populated with other access log settings down the line.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
kyessenov: I'd recommended labelling it as such, e.g. `test_monitoring_endpoint`.

extensions/stackdriver/metric/record.cc
kyessenov: Envoy keeps a monotonic clock to provide duration, which is better than world clock computations that we may be doing here.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
bianpengyuan: Done.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
bianpengyuan: Are you suggesting something like
```
message PluginConfig {
  message AccessLogControl {}
  AccessLogControl access_log_control = 1;
}
```
Then if you want to turn on access log, you still need to specify an empty struct in the config right? "accessLogControl: {}". It does look better to have this for any future access log related settings, though I am not sure what will they be...

kyessenov: Yeah. The value depends whether we expect this accesslogcontrol config to grow. If not, it doesn't really matter.

PR #2364, opened 2019-08-16
Remove Envoy SHA from istio.deps

nrjpoddar:
WORKSPACE is used for Envoy SHA and istio.deps is no longer kept in sync.

### DIFF
--- a/istio.deps
+++ b/istio.deps
@@ -5,12 +5,5 @@
 		"repoName": "api",
 		"file": "repositories.bzl",
 		"lastStableSHA": "36b42252042c0e8b229a0d66059af184c3105f9d"
-	},
-	{
-		"_comment": "",
-		"name": "ENVOY_SHA",
-		"repoName": "envoyproxy/envoy-wasm",
-		"file": "WORKSPACE",
-		"lastStableSHA": "915ed46b694a611f966bed501bcc177162c2df34"
 	}
 ]


nrjpoddar:
@duderino here you go, how do we go about verifying what we might break in istio/istio?

duderino:
@mandarjog @howardjohn @fpesce @sdake @linsun @rshriram @brian-avery @rlenglet  @PiotrSikora do you think it's safe to remove this?  AFAIK nothing uses it and it exists to only confuse poor release managers who should update WORKSPACE instead, but...

PiotrSikora:
AFAIK, some scripts in `istio/istio` look at this file. `istio/proxy` itself doesn't care about it at all.

howardjohn:
Its used in https://github.com/istio/istio/blob/master/bin/init.sh#L208-L217

istio.deps
```diff
@@ -5,12 +5,5 @@
 		"repoName": "api",
 		"file": "repositories.bzl",
 		"lastStableSHA": "36b42252042c0e8b229a0d66059af184c3105f9d"
```
rlenglet: You can remove this one too.
In the `release-1.3` branch it's pointing to a non-existent commit (it's actually a commit in `istio/envoy`, not `istio/api`), and nothing broke.
https://github.com/istio/proxy/blob/release-1.3/istio.deps

So the whole file could be deleted.

howardjohn: You haven't triggered a release of 1.3 yet is why. It is used https://github.com/istio/istio/blob/f8295503296ec6bae1b1047cc1491469d5e72754/release/gcb/generate_manifest.sh#L77-L85

rlenglet: OK, I'll fix that.

nrjpoddar:
@howardjohn thanks for pointing out the references, in the latest releases (release-1.2) `istio.deps` is out of date and only `WORKSPACE` was updated so now I'm even more confused as to how it works?

howardjohn:
@nrjpoddar my understanding is the API sha must be in sync, but the envoysha is just included as metadata into the proxy (https://github.com/istio/istio/blob/92d121c9b8859ca75ca55e8c4caa9bb7d40a514a/pilot/docker/Dockerfile.proxyv2#L6-L8)

> so now I'm even more confused as to how it works?

I feel the same way everytime I learn more about our release process :slightly_smiling_face:

nrjpoddar:
Thanks @howardjohn for the explanation. So the question then is that should we preserve the behavior of setting ENVOY_SHA which is used as a label (it's currently broken) or we don't document this so no one cares about that label?

howardjohn:
It doesn't seem used, I'd be fine with removing. @rshriram @kyessenov any thoughts?

howardjohn:
@nrjpoddar I say we kill it. But we need to do so in istio/istio first. I'll send a PR, once/if that gets merged we can get this merged?

howardjohn:
Opened https://github.com/istio/istio/pull/16679

PR #2368, opened 2019-08-17
Update Envoy-WASM SHA to latest with security fixes. (#2355)

PiotrSikora:

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -35,11 +35,11 @@ bind(
 # When updating envoy sha manually please update the sha in istio.deps file also
 #
 # Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/COMMIT.tar.gz && sha256sum COMMIT.tar.gz`
-# envoy-wasm commit date: 08/08/2019
+# envoy-wasm commit date: 08/13/2019
 # bazel version: 0.28.1
-ENVOY_SHA = "37c612411ce0812466a1aa14dc201300f7f0a6df"
+ENVOY_SHA = "3f90336c39250c472cfd51b157c305582b7a6725"
 
-ENVOY_SHA256 = "281662f0bb5d609aac838e58a838d24e9da3033be98e9bf8d2dd9054ffec04cd"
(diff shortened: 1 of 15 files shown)


PiotrSikora:
> Why are we still pointing to https://github.com/envoyproxy/envoy-wasm/archive/ instead of https://github.com/istio/envoy/archive/???

Because release managers didn't update it yet?

On a side note: if we start pointing to commits in `istio/envoy:release-1.3`, then we cannot cherry-pick changes in [istio/proxy](https://github.com/istio/proxy) anymore, since the changes in `master` branch are going to be pointing at commits that don't exist in `istio/envoy:release-1.3`.

rlenglet:
> Because release managers didn't update it yet?

My understanding is that release managers only manage the `release-1.3` branches.

rlenglet:
I branched `release-1.3` out of `master`.

rlenglet:
> On a side note: if we start pointing to commits in istio/envoy:release-1.3, then we cannot cherry-pick changes in istio/proxy anymore, since the changes in master branch are going to be pointing at commits that don't exist in istio/envoy:release-1.3.

I don't get your point. How can pointing `istio/proxy:release-1.3` to `istio/envoy:release-1.3` and `istio/proxy:master` to whatever your want create any problem?

PiotrSikora:
> I don't get your point. How can pointing `istio/proxy:release-1.3` to `istio/envoy:release-1.3` and `istio/proxy:master` to whatever your want create any problem?

It's not a problem per-se, but problem for cherry-picks, since commit SHAs in `"whatever"` don't exist in `istio/envoy:release-1.3` (not for new commits), which means that commits updating Envoy SHA in `istio/proxy:master` cannot be cherry-picked into `istio/proxy:release-1.3`.

rlenglet:
> It's not a problem per-se, but problem for cherry-picks, since commit SHAs in "whatever" don't exist in istio/envoy:release-1.3 (not for new commits), which means that commits updating Envoy SHA in istio/proxy:master cannot be cherry-picked into istio/proxy:release-1.3.

How are you suggesting to do otherwise, since you're supposed to only backport bug fixes for Istio 1.3? You can't just point to the latest master HEAD from `istio/proxy:release-1.3` every time you need to pick up a fix.

PR #2374, opened 2019-08-20
Update to clang/libc++ 8 and Bazelisk launcher for Bazel.

PiotrSikora:
Updating both at the same time, since we don't have build images
with either:
a) clang/libc++ 7 and Bazelisk.
b) clang/libc++ 8 and old version of Bazel.


PiotrSikora:
cc @istio/release-managers-1-1 @duderino

.bazelversion
```diff
@@ -0,0 +1 @@
+0.22.0
```
duderino: soooooo awesome

PR #2375, opened 2019-08-20
Update to clang/libc++ 8 and Bazelisk launcher for Bazel.

PiotrSikora:
Updating both at the same time, since we don't have build images
with either:
a) clang/libc++ 7 and Bazelisk.
b) clang/libc++ 8 and old version of Bazel.


PiotrSikora:
cc @istio/release-managers-1-2 @duderino

.circleci/Dockerfile
```diff
@@ -10,19 +10,20 @@ FROM circleci/openjdk:latest
...
-    clang-7 clang-format-7 clang-tidy-7 lld-7 libc++-7-dev libc++abi-7-dev \
+    clang-8 clang-format-8 clang-tidy-8 lld-8 libc++-8-dev libc++abi-8-dev \
     rsync ninja-build
 
-# ~100M, depends on g++, zlib1g-dev, bash-completions
-RUN curl -Lo /tmp/bazel.deb https://github.com/bazelbuild/bazel/releases/download/0.25.0/bazel_0.25.0-linux-x86_64.deb && \
-    sudo dpkg -i /tmp/bazel.deb && rm /tmp/bazel.deb
-
+# Use Bazelisk launcher for Bazel (automatically downloads required version of Bazel).
+RUN cd /tmp && \
```
fpesce: being Dockerfile pedantic here, but usually, it is good practice to avoid cd  and use WORKDIR
c.f. https://github.com/hadolint/hadolint/wiki/DL3003

I'm seeing that it is used in the following RUN command (which actually makes a stronger point of placing a WORKDIR /tmp before this 2 RUN), but it can be refactored in the future.

PiotrSikora: Patches welcomed, but that's merged from `master`, not it's not unique to this PR / branch.

PR #2390, opened 2019-08-24
Bug duration: convert duration from nanos to milis

mandarjog:
Mixer duration was in milis.
Converting envoy duration from nanos to milis.


mandarjog:
@PiotrSikora did we just enable asan? How come this never failed before?

PiotrSikora:
> @PiotrSikora did we just enable asan? How come this never failed before?

No, it's been on forever. Maybe something on Prow changed, since we saw it earlier today in other PR as well.

> This fix needs to be first applied on `master` then backported into `release-1.3`. I see no such PR on `master`, and the code in `master` is still the same.
> @mandarjog If `master` doesn't need to be fixed, document why in this PR's description.

Agreed, @mandarjog please fix this on `master` first.

mandarjog:
@rlenglet https://github.com/istio/proxy/pull/2391/files make the equivalent change in master.

PR #2391, opened 2019-08-24
Update sha to pick up Null pointer fix

mandarjog:
1. Update envoy sha
2. Update istio sha
3. reduce log levels for noisy logs in extension. (cherry-picked back from 1.3)
4. convert seconds from nanos to milis


mandarjog:
@PiotrSikora asan is *not* failing here, unless that fix was already made to master.

PR #2400, opened 2019-08-28
Update build image to prowbazel:0.5.10.

PiotrSikora:


PiotrSikora:
Should we use prowbazel images in CircleCI as well? Might not be important, since we'll probably drop it soon, anyway, now that Prow works reliably.

PR #2403, opened 2019-08-29
Use .bazelrc as part of CircleCI's cache key.

PiotrSikora:


PiotrSikora:
Ping.

PR #2405, opened 2019-08-29
Update destination service label

bianpengyuan:
Dashboard expects host label to be `destination_service` instead of `destination_service_host`.


bianpengyuan:
@venilnoronha  @lambdai Could you help to approve this? Thanks!

mandarjog:
There are changes in the example file, and they will be reflected in the bootstrap.

bianpengyuan:
@mandarjog Will make the change there as well

PR #2408, opened 2019-09-04
Use per filter config in route instead of route entry

bianpengyuan:
When blackhole route is matched, routeEntry will be null in `route` and instead the directResponseEntry will be populated. The original logic uses per filter config from routeEntry, which does not work for blackhole case. After this change, the attributes mixer receives are:
```
...
context.reporter.kind         : outbound
context.reporter.uid          : kubernetes://sleep-596b545f8d-sqhdj.default
destination.service.host      : BlackHoleCluster
destination.service.name      : BlackHoleCluster
origin.ip                     : [10 24 0 11]
...
request.host                  : httpbin.org
request.method                : GET
request.path                  : /
request.scheme                : http
request.size                  : 0
request.total_size            : 788
request.url_path              : /
...
```

@douglas-reid @nrjpoddar @mandarjog Should we consider leaving `destination.service.host` unset for `BlackHoleCluster` cluster? So that metric could fall back to `request.host`. It does not seem to be very informative if both service host and name are `BlackHoleCluster`.

fix https://github.com/istio/istio/issues/16629

### DIFF
--- a/src/envoy/http/mixer/filter.cc
+++ b/src/envoy/http/mixer/filter.cc
@@ -48,15 +48,12 @@ Filter::Filter(Control& control)
 }
 
 void Filter::ReadPerRouteConfig(
-    const PerRouteServiceConfig* route_cfg,
+    const PerRouteServiceConfig& route_cfg,
     ::istio::control::http::Controller::PerRouteConfig* config) {
-  if (!route_cfg) {
-    return;
+  if (!control_.controller()->LookupServiceConfig(route_cfg.hash)) {
+    control_.controller()->AddServiceConfig(route_cfg.hash, route_cfg.config);
   }
-  if (!control_.controller()->LookupServiceConfig(route_cfg->hash)) {
-    control_.controller()->AddServiceConfig(route_cfg->hash, route_cfg->config);
-  }
-  config->service_config_id = route_cfg->hash;
+  config->service_config_id = route_cfg.hash;
 }
 
 FilterHeadersStatus Filter::decodeHeaders(HeaderMap& headers, bool) {
@@ -66,8 +63,10 @@ FilterHeadersStatus Filter::decodeHeaders(HeaderMap& headers, bool) {
   ::istio::control::http::Controller::PerRouteConfig config;
   auto route = decoder_callbacks_->route();
   if (route) {
-    ReadPerRouteConfig(
-        route->perFilterConfigTyped<PerRouteServiceConfig>("mixer"), &config);
+    auto route_cfg = route->perFilterConfigTyped<PerRouteServiceConfig>("mixer");
+    if (route_cfg) {
+      ReadPerRouteConfig(*route_cfg, &config);
+    }
   }
   handler_ = control_.controller()->CreateRequestHandler(config);
 
@@ -228,9 +227,10 @@ void Filter::log(const HeaderMap* request_headers,
     // Here Request is rejected by other filters, Mixer filter is not called.
     ::istio::control::http::Controller::PerRouteConfig config;
     auto route_entry = stream_info.routeEntry();
-    ReadPerRouteConfig(
-        route_entry->perFilterConfigTyped<PerRouteServiceConfig>("mixer"),
-        &config);
+    auto route_cfg = route_entry->perFilterConfigTyped<PerRouteServiceConfig>("mixer");
+    if (route_cfg) {
+      ReadPerRouteConfig(*route_cfg, &config);
+    }
     handler_ = control_.controller()->CreateRequestHandler(config);
   }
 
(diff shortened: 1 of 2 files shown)


douglas-reid:
I agree that it seems incorrect to set `destination.service.host` to `BlackholeCluster` or `PassthroughCluster`. 

Should we also update `response.flags` possibly?  Maybe something like:

BlackHole:
```
destination.service.name="BlackholeCluster"
destination.service.host="httpbin.org"
response.code=502
response.flags="NR"
```

Passthrough:
```
destination.service.name="PassthroughCluster"
destination.service.host="httpbin.org"
response.code=200
response.flags="-"
```

?

bianpengyuan:
Ok that requires a change to pilot code as well as metric instances.

The response flag (`context.proxy_error_code      : -`) is not set for black hole route. I think envoy finds a valid route (Blackhole) and does not think it should generate NR error? Also using blackhole as destination service name should already indicate how request is handled well, no?

src/envoy/http/mixer/filter.cc
```diff
@@ -48,22 +48,15 @@ Filter::Filter(Control& control)
 }
 
 void Filter::ReadPerRouteConfig(
-    const Router::RouteEntry* entry,
+    const PerRouteServiceConfig* route_cfg,
```
mandarjog: Const& ?

mandarjog:
Yes set response flag if possible. Though setting it to NR will be correct in spirit but not letter of the law. @kyessenov ?

src/envoy/http/mixer/filter.cc
bianpengyuan: Done.

bianpengyuan:
Then that means we need to override the response flag decision that envoy made based on route name? That seems to be not right. I think the using `blackhole` as destination_service_name is already an indicator that envoy cannot find any valid route for the request. Populating response flag with `NR` is just a duplication of that?

bianpengyuan:
@mandarjog @douglas-reid @kyessenov Any more input on this? Thanks!

mandarjog:
I agree, not setting NR is better, since it *did* find a route.
@kyessenov can you review ?

nrjpoddar:
Hi @bianpengyuan, thanks for this PR. I was on vacation so couldn't reply earlier. Few things to note:

* AFAIK `destination.service.host` is the FQDN of the upstream service and is not at all related to the request host. I made sure that `destination.service.host` & `destination.service.name` are consistent for BlackHole/Passthrough as is the case with other scenarios.
* I agree with you that setting `NR` response flag seems incorrect. This extra piece of information overwrites Envoy set value which doesn't jive well, secondly it's kind of useless as just with that field you cannot really do much. Additionally, for BlackHole clusters which don't have a HTTP service port defined TCP listeners are hit in which case again this flag is useless.

bianpengyuan:
@nrjpoddar thanks! Making destination.service.host and destination.service.name the same makes blackhole / passthrough metric not so informative. I've made a change to make destination_service_host label fallback to request.host if not provided: https://github.com/istio/istio/pull/16892.

nrjpoddar:
Well `destination.service.host` is also used in case for TCP which will now get set to `unknown` after https://github.com/istio/istio/pull/16892/ isn't that correct?

nrjpoddar:
> Well `destination.service.host` is also used in case for TCP which will now get set to `unknown` after [istio/istio#16892](https://github.com/istio/istio/pull/16892) isn't that correct?

@bianpengyuan with regards to TCP ignore the above comment as your changes in Plugin code only apply for HTTP routes. These changes look good.

nrjpoddar:
can this be cherrypicked to 1.3 release branch? @douglas-reid @mandarjog

nrjpoddar:
@bianpengyuan is something blocking its merge?

bianpengyuan:
@nrjpoddar No, just code review. @kyessenov Would you mind to take a look? Thanks!

src/envoy/http/mixer/filter.cc
```diff
@@ -233,7 +227,12 @@ void Filter::log(const HeaderMap* request_headers,
 
     // Here Request is rejected by other filters, Mixer filter is not called.
     ::istio::control::http::Controller::PerRouteConfig config;
-    ReadPerRouteConfig(stream_info.routeEntry(), &config);
+    auto route_entry = stream_info.routeEntry();
```
kyessenov: can you check for null to be defensive here?

kyessenov:
LG except for null check. StreamInfo route entry is populated on best-effort basis so may be null sometimes.

src/envoy/http/mixer/filter.cc
bianpengyuan: Ah, thanks a lot for catching this!

PR #2419, opened 2019-09-12
Remove istio.deps file

nrjpoddar:
`istio.deps` is not used and istio/istio#16679 removed its usage from istio/istio.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -32,8 +32,6 @@ bind(
     actual = "//external:ssl",
 )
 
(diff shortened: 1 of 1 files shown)


nrjpoddar:
@rlenglet you mentioned above that we should monitor daily builds. I don't know what to look for to see if this caused any issues.

rlenglet:
The master daily builds went ok: https://gcsweb.istio.io/gcs/istio-prerelease/daily-build/master-latest-daily/ so this PR was fine!

PR #2421, opened 2019-09-13
wasm: use filter state instead of metadata

kyessenov:
Updates to use the new ABI.


extensions/common/context.cc
```diff
@@ -47,6 +40,37 @@ using Envoy::Extensions::Common::Wasm::Null::Plugin::getResponseTlsVersion;
 namespace Wasm {
 namespace Common {
 
+namespace {
+
+// Return int64 value or 0 if not available
+int64_t getIntValue(std::initializer_list<StringView> parts) {
```
kyessenov: I'll upstream some bits now that it's easy to see how the ABI is used.

kyessenov:
@bianpengyuan @PiotrSikora @jplevyak this is ready for review. This should sync up the tips of envoy-wasm with proxy.

WORKSPACE
```diff
@@ -33,11 +33,11 @@ bind(
 )
 
 # Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
-# envoy-wasm commit date: 08/26/2019
+# envoy-wasm commit date: 09/12/2019
 # bazel version: 0.28.1
-ENVOY_SHA = "ef4da0bda2bee932c575c0c98508c473dd6085c5"
+ENVOY_SHA = "20db8ed2938c9e88dd36605473b0019a23ab2657"
```
PiotrSikora: Since this is pulling a new version of Envoy from upstream, could you also update `bazel version` here, and sync `.bazelversion` and `.bazelrc` with upstream?

kyessenov: It builds fine with 0.28.1. Do we need to update bazel version?

kyessenov: bazelrc diff is very large, and I'm not sure I can fix it correctly.

test/envoye2e/stackdriver_plugin/fake_stackdriver/timeseries.go
```diff
@@ -63,7 +63,7 @@ const ClientRequestCountJSON = `{
 	   "type":"istio.io/service/client/request_count",
 	   "labels":{
 		  "destination_owner":"kubernetes://api/apps/v1/namespaces/default/deployment/ratings-v1",
-		  "destination_port":"20019",
+		  "destination_port":"20017",
```
kyessenov: This seems closer to the truth. Previously, upstream remote port was reported as destination port.

kyessenov: Value 0 requires further investigation.

extensions/stackdriver/stackdriver.cc
```diff
@@ -143,32 +144,44 @@ void StackdriverContext::onLog() {
...
-      logWarn(
-          absl::StrCat("cannot get metadata for: ", kDownstreamMetadataKey));
+    auto metadata_value =
+        getSelectorExpression({"filter_state", kDownstreamMetadataKey});
+    if (!metadata_value) {
+      logWarn(absl::StrCat("cannot get stackdriver metadata for: ",
+                           kDownstreamMetadataKey));
+      return;
+    }
+    google::protobuf::Value downstream_metadata;
```
bianpengyuan: Can this be parsed into google::protobuf::Struct directly instead of google::protobuf::Value?

extensions/stats/plugin.cc
```diff
@@ -170,12 +170,18 @@ const wasm::common::NodeInfo& NodeInfoCache::getPeerById(
...
-  if (getMetadataStruct(MetadataType::Request, peer_metadata_key, &metadata) !=
-      Common::Wasm::WasmResult::Ok) {
+  auto metadata_value =
+      getSelectorExpression({"filter_state", peer_metadata_key});
+  if (!metadata_value.has_value()) {
     LOG_DEBUG(absl::StrCat("cannot get metadata for: ", peer_metadata_key));
     return cache_[""];
   }
+  google::protobuf::Struct metadata;
+  if (!metadata.ParseFromArray(metadata_value.value()->data(),
```
bianpengyuan: It seems this is common enough that we should just extract this into a function and put it in  common context?

extensions/stats/plugin.cc
```diff
@@ -52,12 +52,11 @@ void PluginRootContext::onConfigure(std::unique_ptr<WasmData> configuration) {
     LOG_WARN("cannot parse local node metadata ");
     return;
   }
-  PluginDirection direction;
-  auto dirn_result = getPluginDirection(&direction);
-  if (WasmResult::Ok == dirn_result) {
-    outbound_ = PluginDirection::Outbound == direction;
+  int64_t direction;
+  if (getValue({"traffic_direction"}, &direction)) {
+    outbound_ = envoy::api::v2::core::TrafficDirection::OUTBOUND == direction;
```
bianpengyuan: So PluginDirection is removed in envoy-wasm? This will make envoy proto a requirement for wasm compilation, not sure if it is possible to do it by 1.4...

PiotrSikora:
@kyessenov you need to update `.circleci/config.yml` to use Xcode `11.0.0` to pull recent enough clang to fix V8 build.

WORKSPACE
PiotrSikora: Fine, I can update Bazel in a follow-up, but in general we try to keep Bazel in sync with whatever upstream uses.

src/envoy/http/metadata_exchange/config.cc
```diff
@@ -92,28 +93,18 @@ void PluginRootContext::updateMetadataValue() {
...
-  if (getMetadataStruct(Common::Wasm::MetadataType::Node, WholeNodeKey,
-                        &node) == Common::Wasm::WasmResult::Ok) {
-    for (const auto& f : node.fields()) {
-      if (f.first == NodeIdKey &&
-          f.second.kind_case() == google::protobuf::Value::kStringValue) {
-        node_id_ = f.second.string_value();
-        break;
-      }
-    }
+  auto node_id = getSelectorExpression({"node", "id"});
```
bianpengyuan: why not just getStringValue?

extensions/stackdriver/stackdriver.cc
kyessenov: The ABI function underlying it produces Value instead of Struct since it has to support lists and other values. So the output is the generic Value but then we know that it contains a struct. The overhead is not that big, just a few extra bytes.

extensions/stats/plugin.cc
kyessenov: Right, updated to use `getStructValue`. Messy to juggle multiple repositories.

extensions/stats/plugin.cc
kyessenov: Proto enum is just an int. We can hardcode the value for OUTBOUND if you wish.

src/envoy/http/metadata_exchange/config.cc
kyessenov: Yeah, forgot I added an utility for it.

extensions/stats/plugin.cc
kyessenov: In general, we may want to expose xDS to plugins that need to consume xDS config.
If config is available in xDS, we should not be using additional config on top of that to avoid proliferation of the config.

extensions/stackdriver/stackdriver.h
```diff
@@ -73,7 +73,8 @@ class StackdriverRootContext : public RootContext {
   wasm::common::NodeInfo local_node_info_;
 
   // Indicates the traffic direction relative to this proxy.
-  PluginDirection direction_ = PluginDirection::Unspecified;
+  envoy::api::v2::core::TrafficDirection direction_ =
```
mandarjog: I agree this is unnecessary coupling of extension code with xds api.

extensions/stackdriver/stackdriver.cc
```diff
@@ -141,38 +142,19 @@ void StackdriverContext::onLog() {
...
-      return;
-    }
-
-    auto status = ::Wasm::Common::extractNodeMetadata(downstream_metadata,
-                                                      &peer_node_info_);
-    if (status != Status::OK) {
-      logWarn("cannot parse downstream peer node metadata " +
-              downstream_metadata.DebugString() + ": " + status.ToString());
-    }
+  google::protobuf::Value metadata;
```
mandarjog: This should be wrapper of macro than this often used magic string - “filter_state”

extensions/stats/testdata/client.yaml
```diff
@@ -84,14 +84,14 @@ static_resources:
                 route: 
                   cluster: server
           http_filters:
-          - name: envoy.wasm
```
mandarjog: Perhaps I missed it, but change to server.yaml?

kyessenov: The test config is no longer functional. We can remove it. Filter state is not easily accessible to native filters, which is its downside.

extensions/stackdriver/stackdriver.cc
kyessenov: Are we committed to "filter_state"? I didn't want to introduce API changes before we're sure about the naming. Same applies to `traffic_direction`.

extensions/stackdriver/stackdriver.h
kyessenov: Traffic direction is part of an API, so the question is whether we want to maintain an independent but coupled definition of it in WASM or just use xDS. I'd assume the second is better, since this is bound to evolve (the gateway option still needs to be added as a traffic direction).
Physically, we don't have to mandate protobufs. The client can just treat it as an int, and use a well-known value to compare.

extensions/common/context.cc
```diff
@@ -101,20 +98,20 @@ void populateHTTPRequestInfo(bool outbound, RequestInfo *request_info) {
...
-  getRequestDestinationPort(&request_info->destination_port);
 
+  int64_t destination_port = 0;
   std::string tls_version;
-  bool cert_presented;
 
   if (outbound) {
-    getResponsePeerCertificatePresented(&cert_presented);
-    getResponseTlsVersion(&tls_version);
+    getValue({"upstream", "port"}, &destination_port);
```
bianpengyuan: Hmm, shouldn't outbound proxy use destination port (i.e. the port that client app originally targets to) and inbound proxy use upstream port (i.e. the port that server app accepts the traffic)? This is my mental model:

client --> (destination port) proxy --> proxy (upstream port) --> server

extensions/stackdriver/stackdriver.h
bianpengyuan: Can we use the original plugin direction for now? Maybe add a todo for it? There are several other duplication of envoy proto in WASM sdk such as grpc service (envoyproxy/envoy-wasm#17). For sure we need to remove them, but unless bazel is supported, there is no easy way to get around this duplication I think..

extensions/common/context.cc
kyessenov: Destination port is the local downstream connection port (e.g. the port that the proxy listens on).
Upstream port is the port the upstream connection remote port (e.g. the port the destination listens on). Does that make sense?

extensions/stackdriver/stackdriver.h
kyessenov: The plugin direction enum is gone I think. I can add an enum here to reinterpret the int value.

kyessenov:
@PiotrSikora @jplevyak PTAL. I added an enum for traffic direction to avoid importing xDS protobufs.

WORKSPACE
```diff
@@ -33,11 +33,11 @@ bind(
 )
 
 # Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
-# envoy-wasm commit date: 08/26/2019
+# envoy-wasm commit date: 09/12/2019
```
JimmyCYJ: Could you please update the SHA to the anyone after 09/24/2019, so that it picks the SDS change?

kyessenov: Sure, done.

JimmyCYJ: Thanks a lot!

PR #2424, opened 2019-09-16
jwt: extract claim with whitespce as list

yangminzhu:
If a claim in the JWT token includes spaces, extract it as a list of string delimited with spaces. For example, `"aud": "aud1   aud2"` will be extracted as `"aud": ["aud1", "aud2"]`

Fixes https://github.com/istio/istio/issues/13565

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/authn/authn_utils_test.cc
+++ b/src/envoy/http/authn/authn_utils_test.cc
@@ -37,16 +37,17 @@ const std::string kSecIstioAuthUserinfoHeaderValue =
        "some-other-string-claims": "some-claims-kept"
      }
    )";
-const std::string kSecIstioAuthUserInfoHeaderWithNoAudValue =
+const std::string kSecIstioAuthUserInfoHeaderWithAudValueList =
     R"(
        {
          "iss": "issuer@foo.com",
          "sub": "sub@foo.com",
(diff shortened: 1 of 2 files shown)


src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -116,11 +117,13 @@ TEST(AuthnUtilsTest, GetJwtPayloadFromHeaderTest) {
   EXPECT_TRUE(MessageDifferencer::Equals(expected_payload, payload));
 }
 
-TEST(AuthnUtilsTest, ProcessJwtPayloadWithNoAudTest) {
+TEST(AuthnUtilsTest, ProcessJwtPayloadWithAudListTest) {
```
lei-tang: Why ProcessJwtPayloadWithNoAudTest() is replaced by ProcessJwtPayloadWithAudListTest(), instead of having both of these two tests?

yangminzhu: The ProcessJwtPayloadWithNoAudTest is testing when the aud field is not in the token which is not very useful, we don't need to test a field is not in the token because we always compare it with an expected output. If there anything not in the expected output, the test will fail, so we don't need to specifically have a `NoAudTest`.

PR #2427, opened 2019-09-20
fix(metadata filter): update metadata exchange filter to new EnvoyFilter API

douglas-reid:
Updates the config for the metadata exchange filter from the deprecated EnvoyFilter API bits to the new EnvoyFilter API bits.

I have only done a single spot testing of this in action. I'm hoping the existing tests can further validate that I've done the conversion correctly.

### DIFF
--- a/extensions/stats/testdata/istio/metadata-exchange_filter.yaml
+++ b/extensions/stats/testdata/istio/metadata-exchange_filter.yaml
@@ -3,43 +3,21 @@ kind: EnvoyFilter
 metadata:
   name: metadata-exchange
 spec:
-  filters:
-  - filterConfig:
-      configuration: envoy.wasm.metadata_exchange
-      vm_config:
-        code:
-          inline_string: envoy.wasm.metadata_exchange
-        vm: envoy.wasm.vm.null
-    filterName: envoy.wasm
-    filterType: HTTP
-    insertPosition:
-      index: FIRST
-    listenerMatch:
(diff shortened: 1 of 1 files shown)


bianpengyuan:
@douglas-reid Can we also update the one in istio/istio? https://github.com/istio/istio/blob/master/tests/integration/telemetry/stats/prometheus/testdata/

douglas-reid:
@bianpengyuan see https://github.com/istio/istio/pull/17281.

extensions/stats/testdata/istio/metadata-exchange_filter.yaml
```diff
@@ -3,43 +3,21 @@ kind: EnvoyFilter
...
+  configPatches:
+    - applyTo: HTTP_FILTER
+      match:
+        context: ANY # inbound, outbound, and gateway
+        listener:
+          filterChain:
+            filter:
+              name: "envoy.http_connection_manager"
+      patch:
+        operation: INSERT_BEFORE
```
rshriram: insert before what? missing sub filter match ?

douglas-reid: From the docs: "If no filter is selected, the specified filter will be inserted at the front of the list."

I believe that is what is desired.

douglas-reid:
This went into to istio/istio. Time to push it in here?

PR #2429, opened 2019-09-20
Add Stackdriver access logger

bianpengyuan:
This add sd logger library with wasm grpc API. It does not introduce any usage, nor e2e test. They will be added in later PRs. Also this PR just populate part of the labels that are populated by mixer log adapter. We should make request info and node info to gain parity with mixer attributes after https://github.com/istio/proxy/pull/2421 is in.


extensions/stackdriver/log/exporter.cc
```diff
@@ -0,0 +1,102 @@
...
+
+ExporterImpl::ExporterImpl(RootContext* root_context,
+                           std::string logging_service_endpoint) {
+  context_ = root_context;
+  success_callback_ = [](google::protobuf::Empty&&) {
+    logDebug("successfully sent Stackdriver logging request");
+  };
+
+  failure_callback_ = [](GrpcStatus status, StringView message) {
+    // TODO: add retry.
```
gargnupur: nit:  add your name here?

extensions/stackdriver/log/exporter.cc
```diff
@@ -0,0 +1,102 @@
...
+    logWarn("Stackdriver logging api call error: " +
+            std::to_string(static_cast<int>(status)) + std::string(message));
+  };
+
+  // Construct grpc_service for the Stackdriver gRPC call.
+  GrpcService grpc_service;
+  if (logging_service_endpoint.empty()) {
+    grpc_service.mutable_google_grpc()->set_target_uri(
+        kGoogleStackdriverLoggingAddress);
```
gargnupur: Do we use Google style guide or Envoy style guide in Istio Proxy code? As Envoy suggests naming constants without k.. https://github.com/envoyproxy/envoy/blob/master/STYLE.md

extensions/stackdriver/log/logger.h
```diff
@@ -0,0 +1,79 @@
+/* Copyright 2019 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *constexpr int kDefaultTimeoutMillisecond = 10000;
```
gargnupur: nit: delete this line?

extensions/stackdriver/log/logger.h
```diff
@@ -0,0 +1,79 @@
...
+namespace Log {
+
+#ifdef NULL_PLUGIN
+using Envoy::Extensions::Common::Wasm::Null::Plugin::Extensions::Stackdriver::
+    Log::Exporter;
+#endif
+
+// Size of an entries.write request is 5 Mb:
+// https://cloud.google.com/logging/quotas.
+constexpr int kMaxRotationBytesSize = 4000000;
```
gargnupur: aren't constexpr not suggested to be used in header files?

extensions/stackdriver/log/logger_test.cc
```diff
@@ -0,0 +1,159 @@
...
+  return request_info;
+}
+
+google::logging::v2::WriteLogEntriesRequest expectedRequest(
+    int log_entry_count) {
+  auto request_info = requestInfo();
+  auto peer_node_info = peerNodeInfo();
+  auto node_info = nodeInfo();
+  google::logging::v2::WriteLogEntriesRequest req;
+  req.set_log_name("projects/test_project/logs/server-accesslog-stackdriver");
```
gargnupur: make this a const?

extensions/stackdriver/log/exporter.cc
bianpengyuan: Done.

extensions/stackdriver/log/logger.h
bianpengyuan: Oops, done. Thanks!

extensions/stackdriver/log/logger_test.cc
bianpengyuan: Done.

extensions/stackdriver/log/logger.h
bianpengyuan: Removed.

extensions/stackdriver/log/exporter.cc
bianpengyuan: Hmm, yeah, but stackdriver filter starts with k convention. I think as long as we feel comfortable with it, it is fine to just use the google style?

bianpengyuan:
I am not sure how to fix the macos test.. google logging proto defines `DEBUG` in its log level enum which conflicts with the macro bazel defined when building on macos: https://github.com/bazelbuild/bazel/issues/3513. I tried to add `-UDEBUG` to all the newly added target, it seems like library target could be built, but test target still cannot be built. Not sure how to solve this.. @PiotrSikora do you have any suggestions? is it possible to bypass a test target for a particular platform like macos? Otherwise I think I need to remove the test target.. Thanks!

```
bazel-out/darwin-fastbuild/bin/external/com_google_googleapis/google/logging/type/log_severity.pb.h:65:3: error: expected identifier
  DEBUG = 100,
  ^
<command line>:2:15: note: expanded from here
#define DEBUG 1
```

PiotrSikora:
@bianpengyuan `DEBUG` is used system-wide on macOS (with XCode as well, it's not Bazel issue), and unfortunately it's not something that we can undo.

Could you open an issue against googleapis? It seems to me that it's going to be completely broken on macOS.

bianpengyuan:
@PiotrSikora I created an internal issue about this. I guess there is no way to work around for now except removing the test?

PiotrSikora:
@bianpengyuan well, we might have hard time trying to link against it in the future (i.e. when you'll actually want to use it).

It's a bit weird that the `DEBUG` enum is getting exported in `pb.h` without any prefixes.

extensions/stackdriver/common/utils.cc
```diff
@@ -0,0 +1,56 @@
...
+#include "extensions/stackdriver/common/utils.h"
+#include "extensions/stackdriver/common/constants.h"
+
+namespace Extensions {
+namespace Stackdriver {
+namespace Common {
+
+using google::api::MonitoredResource;
+
+MonitoredResource getMonitoredResource(
```
kyessenov: The common pattern is to pass a pointer/reference to the monitored resource so it's not copied on the method return, e.g. `updatedMonitoredResource(..., MonitoredResource& out)`

extensions/stackdriver/log/exporter.cc
```diff
@@ -0,0 +1,102 @@
...
+constexpr char kGoogleLoggingService[] = "google.logging.v2.LoggingServiceV2";
+constexpr char kGoogleWriteLogEntriesMethod[] = "WriteLogEntries";
+constexpr char kDefaultRootCertFile[] = "/etc/ssl/certs/ca-certificates.crt";
+constexpr int kDefaultTimeoutMillisecond = 10000;
+
+namespace Extensions {
+namespace Stackdriver {
+namespace Log {
+
+ExporterImpl::ExporterImpl(RootContext* root_context,
```
gargnupur: can declare both of these const as they are not getting modified in the function?

extensions/stackdriver/log/logger.h
```diff
@@ -0,0 +1,79 @@
...
+  // buffer. This will be triggered either by a timer or by request size limit.
+  void flush();
+
+  // Current request that the new log entry should be written into.
+  std::unique_ptr<google::logging::v2::WriteLogEntriesRequest>
+      log_entries_request_;
+
+  // Buffer for WriteLogEntriesRequests that are to be exported.
+  std::vector<std::unique_ptr<google::logging::v2::WriteLogEntriesRequest>>
+      request_queue_;
```
gargnupur: just trying to understand.. should access to this queue be thread-safe?
also, google::logging::v2::WriteLogEntriesRequest can become a const?

bianpengyuan: Yes, everything is single threaded. Done.

extensions/stackdriver/log/exporter.cc
bianpengyuan: root_context cannot be const, since `grpcSimpleCall` which is one of its method and used here is not declared as const. But I make the other param const.

extensions/stackdriver/common/utils.cc
bianpengyuan: Done.

bianpengyuan:
@kyessenov @gargnupur I also modify this PR according to the comments from https://github.com/istio/proxy/pull/2439. Please take a look, thanks!

extensions/stackdriver/common/utils.cc
```diff
@@ -0,0 +1,50 @@
...
+namespace Extensions {
+namespace Stackdriver {
+namespace Common {
+
+using google::api::MonitoredResource;
+
+void getMonitoredResource(const std::string &monitored_resource_type,
+                          const ::wasm::common::NodeInfo &local_node_info,
+                          MonitoredResource *monitored_resource) {
+  monitored_resource->set_type(monitored_resource_type);
```
gargnupur: nit: null check that monitored_resource is not null?

extensions/stackdriver/log/exporter.h
```diff
@@ -0,0 +1,86 @@
...
+};
+
+// Exporter writes Stackdriver access log to the backend. It uses WebAssembly
+// gRPC API.
+class ExporterImpl : public Exporter {
+ public:
+  // root_context is the wasm runtime context that this instance runs with.
+  // logging_service_endpoint is an optional param which should be used for test
+  // only.
+  ExporterImpl(RootContext* root_context,
```
gargnupur: nit: can make RootContext also as const?

extensions/stackdriver/log/logger_test.cc
```diff
@@ -0,0 +1,156 @@
...
+  auto exporter_ptr = exporter.get();
+  auto logger = std::make_unique<Logger>(nodeInfo(), std::move(exporter), 900);
+  EXPECT_CALL(*exporter_ptr, exportLogs(::testing::_))
+      .Times(3)
+      .WillRepeatedly(::testing::Invoke(
+          [](const google::logging::v2::WriteLogEntriesRequest& request) {
+            auto expected_request = expectedRequest(3);
+            EXPECT_TRUE(MessageDifferencer::Equals(expected_request, request));
+          }));
+  for (int i = 0; i < 9; i++) {
```
gargnupur: for later: we can improve this test by calling addLogEntry in parallel in multiple threads?

bianpengyuan: I don't think it is necessary. Envoy uses silo thread for filter. We can assume signal thread is always true.

extensions/stackdriver/log/exporter.h
bianpengyuan: root_context cannot be const, since grpcSimpleCall, which is one of its method and used here, is not declared as const.

bianpengyuan:
@gargnupur @kyessenov May I get some review on this?

extensions/stackdriver/log/logger.cc
```diff
@@ -0,0 +1,131 @@
...
+  const auto& project_id =
+      local_node_info.platform_metadata().at(Common::kGCPProjectKey);
+  log_entries_request_->set_log_name("projects/" + project_id + "/logs/" +
+                                     kServerAccessLogName);
+
+  // Set monitored resources derived from local node info.
+  google::api::MonitoredResource monitored_resource;
+  Common::getMonitoredResource(Common::kContainerMonitoredResource,
+                               local_node_info, &monitored_resource);
+  log_entries_request_->mutable_resource()->CopyFrom(monitored_resource);
```
kyessenov: I think you can just pass `mutable_resource` to `getMonitoredResource`.

PR #2431, opened 2019-09-23
feat(stats): add istio_build metric to prom extension

douglas-reid:
This PR adds a Gauge that records the `istio_build` metric when the `stats` extension is loaded in the proxy.

This will allow tracking of proxy version updates/rollouts across a large mesh.

See also: https://github.com/istio/istio/issues/17173, https://github.com/istio/istio/issues/16555

For the example config, here is prometheus output:

```
# TYPE istio_build gauge
istio_build{component="proxy",tag="1.4-dev"} 1
```

And here is the statstd:

```
component.proxy.tag.1.4-dev;._istio_build: 1
```

This metric is meant to match (with available information) the `istio_build` metric defined in [istio/pkg/version](https://github.com/istio/pkg/blob/master/version/stats_linux.go#L35).

Reviewers:
I'm looking for feedback on the use of the Metrics API, as well as if `PluginRootContext::onConfigure` is the appropriate location for this contribution.

### DIFF
--- a/extensions/common/node_info.proto
+++ b/extensions/common/node_info.proto
@@ -34,4 +34,7 @@ message NodeInfo {
   // Platform metadata uses prefixed keys
   // GCP uses gcp_* keys
   map<string, string> platform_metadata = 6 [json_name = "PLATFORM_METADATA"];
+
+  // Version identifier for the proxy.
+  string istio_version = 7 [json_name = "ISTIO_VERSION"];
 }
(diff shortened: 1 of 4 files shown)


douglas-reid:
@ramaraochavali the metric is created in the extension, but the extension uses the underlying Envoy stats pipeline. So this would create a new metric in Envoy stats store with the given tags and push it the configured stats sinks.

ramaraochavali:
@douglas-reid  Ok. Thanks.

mandarjog:
@kyessenov please review.

PR #2433, opened 2019-09-28
Add  metadata exchange filter yaml file with wasm

gargnupur:
It loads wasm for metadata exchange

### DIFF
--- a/extensions/stats/testdata/istio/metadata-exchange_with_wasm_filter.yaml
+++ b/extensions/stats/testdata/istio/metadata-exchange_with_wasm_filter.yaml
@@ -1,0 +1,45 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: metadata-exchange
+spec:
+  filters:
+  - filterConfig:
+      configuration: envoy.wasm.metadata_exchange
+      vm_config:
+        code:
+          filename: /metadata_exchange_1_3_1.wasm
+        vm: envoy.wasm.vm.v8
+    filterName: envoy.wasm
+    filterType: HTTP
+    insertPosition:
+      index: FIRST
+    listenerMatch:
+      listenerProtocol: HTTP
+      listenerType: SIDECAR_INBOUND
+  - filterConfig:
+      configuration: envoy.wasm.metadata_exchange
+      vm_config:
+        code:
+          filename: /metadata_exchange_1_3_1.wasm
+        vm: envoy.wasm.vm.v8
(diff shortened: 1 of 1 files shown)


extensions/stats/testdata/istio/metadata-exchange_with_wasm_filter.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: metadata-exchange
+spec:
+  filters:
+  - filterConfig:
+      configuration: envoy.wasm.metadata_exchange
```
PiotrSikora: I don't think we need `configuration` value here, do we?

gargnupur: As per @mandarjog , looks like its needed

rlenglet:
We'll wait until after post-1.3.2 to merge this.

PiotrSikora:
@rlenglet you're sending mixed signals... is this getting in before 1.3.2 (since it's approved now) or not (because of the `do-not-merge` label)?

cc @lambdai to make sure he rebases on top of this commit, in case it gets merged.

rlenglet:
I approved the PR because it will be OK to merge it at some point.
I added the no-not-merge label because this won't get merged until after 1.3.2 is published.

PiotrSikora:
Could you elaborate why it won't be merged before 1.3.2 release?

Fixes are not prepared for the `release-1.3` branch yet, so merging this now doesn't break anything that's already staged.

Also, this is just an example `.yaml` file, so it literally cannot break anything. This file was missed in the 1.3.1 release, and because of that, the release notes for 1.3.1 point to instructions that don't exist (see: https://github.com/istio/istio.io/pull/5063 and https://github.com/istio/istio.io/pull/5079).

duderino:
@rlenglet looks like this missed 1.3.1 and 1.3.2, but it's a very low risk doc-supporting change.  can we get it into 1.3.3?

rlenglet:
> @rlenglet looks like this missed 1.3.1 and 1.3.2, but it's a very low risk doc-supporting change. can we get it into 1.3.3?

Yes, that is the plan.

rlenglet:
@gargnupur @PiotrSikora Please consider adding an item for this PR into the [1.3.3 release note draft](https://github.com/istio/istio/wiki/Release-Note-1.3.3-Draft) if necessary.

PR #2434, opened 2019-09-30
add upstream info

kyessenov:

Adds:
- `upstream.mtls` a bool indicating whether client-side connection is mTLS
- `upstream.failure_reason` a string recording connection reset log line (usually due to TLS)

/cc @nrjpoddar @douglas-reid


douglas-reid:
can you add detail on how these new attributes impact the calculation of the connection security policy labels, etc.?

nrjpoddar:
Thanks @kyessenov for this PR. Can you add in a comment few examples of how the newly added attributes show up in mixer access logs or telemetry?

include/istio/utils/attribute_names.h
```diff
@@ -81,6 +81,8 @@ struct AttributeName {
   static const char kConnectionId[];
   // Record TCP connection status: open, continue, close
   static const char kConnectionEvent[];
+  static const char kUpstreamMtls[];
```
nrjpoddar: how does work with this attribute `kConnectionMtls`

kyessenov: connection.mtls means downstream mTLS status.
upstream.mtls means upstream mTLS status.
Depending on whether the proxy acts like an outbound or an inbound sidecar, you use one or the other.

kyessenov:
@douglas-reid 

We should modify the mtls label to be `conditional(context.reporter.kind == "client", upstream.mtls | false, connection.mtls | false)`.
Failure reason is more useful in access logs, since we don't have a good catalog of errors yet.

include/istio/utils/attribute_names.h
nrjpoddar: i see, may be this attribute `kConnectionMtls` needs to explicitly say that it's related to downstream. It's can be confusing specially after adding this new attribute.

kyessenov:
@nrjpoddar yes, we'll have to update the attribute vocabulary once this gets in. I think this is better than changing the meaning of the attribute. It was always about downstream, just never properly documented.

Upstream failure reason is coming straight from boringSSL. Yes, it should not be used for metrics, unless we can catalog all the error kinds in the library.

nrjpoddar:
Upstream failure reason is coming straight from boringSSL. Yes, it should not be used for metrics, unless we can catalog all the error kinds in the library. -> I can work with you if needed and we can divide and conquer. Getting this information in metrics, even at low resolution where all SSL related errors are reported as boolean is very useful. An alert via metrics + more information in logs/tracing will be an acceptable short time solution for me.

kyessenov:
@nrjpoddar Sure. I still think we need to propagate full-fidelity info from envoy, in case we get our catalog wrong, or there is a need for more details. As a first pass, we should use a CEL expression (`upstream.failure_reason.startsWith("blah")) as a rough error type metric.

kyessenov:
Hey @PiotrSikora , as a WG lead, can you review or assign someone to review? We need this for 1.4 in case Mixer v2 slips.

src/istio/control/http/attributes_builder.cc
```diff
@@ -280,6 +280,14 @@ void AttributesBuilder::ExtractReportAttributes(
   }
 
   builder.FlattenMapOfStringToStruct(report_data->GetDynamicFilterState());
+
+  builder.AddBool(utils::AttributeName::kUpstreamMtls,
+                  report_data->IsUpstreamMutualTLS());
```
rshriram: just a nit. instead of having a bool for mutual TLs, why not have an enum of sorts? upstream: plaintext, tls, mutualTLS ? We are coming across a lot of setups where the sidecar has to terminate one way TLS on VMs.

kyessenov: Enums are generally painful to deal with in proto. mTLS is special since that's the end state, so it's needed to track the progress on getting to that state. We can populate other TLS properties later (e.g. version, cipher suite ID, etc).

include/istio/utils/attribute_names.h
lizan: +1

src/envoy/utils/utils.h
```diff
@@ -52,9 +52,12 @@ bool GetPrincipal(const Network::Connection* connection, bool peer,
 bool GetTrustDomain(const Network::Connection* connection, bool peer,
                     std::string* trust_domain);
 
-// Returns true if connection is mutual TLS enabled.
+// Returns true if downstream connection is mutual TLS enabled.
```
lizan: rename this to IsDownstreamMutualTLS?

include/istio/utils/attribute_names.h
kyessenov: We can't change the meaning of existing attributes. There are UIs that have expectations. We can change the metric, however, in istio config.

src/envoy/utils/utils.cc
```diff
@@ -178,6 +178,11 @@ bool IsMutualTLS(const Network::Connection* connection) {
          connection->ssl()->peerCertificatePresented();
 }
 
+bool IsUpstreamMutualTLS(const StreamInfo::StreamInfo& stream_info) {
+  const auto ssl = stream_info.upstreamSslConnection();
+  return ssl != nullptr && ssl->peerCertificatePresented();
```
nrjpoddar: I'm a little confused on the usage of `ssl->peerCertificatePresented()`. If this flag is set to true, does this imply client was using TLS for talking to upstream and got a successful certificate from upstream or it should mean that client is using TLS irrespective of how the negotiation went and what upstream server presented?

kyessenov: Ouch, I think this might be an issue. We should be checking whether a local certificate was presented, not peer.

nrjpoddar: that's what I thought. I don't see an Envoy API for that. @lizan may be you know better?

kyessenov: We can extend envoy to report whether a local certificate is presented. That might not mean that the server validated it. Not sure how to get that info.

nrjpoddar:
@kyessenov in general, from a user point of view I only care about if the connection over which this transaction is happening is plaintext, TLS or mTLS. In all of these states, I care about errors if any occur.

I'm struggling to see how users can get benefit from this low level separation of upstream/downstream without getting more confused.

kyessenov:
@nrjpoddar We cannot conflate the two for gateways. We only have set of attributes we report from the gateways, and downstream/upstream connections have very different characteristics.

src/envoy/utils/utils.cc
lizan: we didn't have this in Envoy API because this has been simply reflecting what in the TLS context config, though we're adding more dynamic nature to the config for incremental adoption etc, so it makes sense to add, thanks for posting https://github.com/envoyproxy/envoy/pull/8464.

PR #2435, opened 2019-10-02
Add trust domain and trust domain aliases to mTLS API

pitlv2109:
Add trust domain and trust domain aliases information, so that the authn filter can use this information.

### DIFF
--- a/src/istio/authn/context.proto
+++ b/src/istio/authn/context.proto
@@ -51,6 +51,22 @@ message X509Payload {
   // Identity carries by X509 certification. This is extracted from SAN field
   // for Istio provided certificate.
   string user = 1;
+
+  // The trust domain corresponds to the trust root of a system.
+  // Refer to
+  // [SPIFFE-ID](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#21-trust-domain)
+  string trust_domain = 2;
+
+  // The trust domain aliases represent the aliases of `trust_domain`.
+  // For example, if we have
+  // ```yaml
+  // trustDomain: td1
+  // trustDomainAliases: [“td2”, "td3"]
+  // ```
+  // Any service with the identity `td1/ns/foo/sa/a-service-account`,
+  // `td2/ns/foo/sa/a-service-account`, or `td3/ns/foo/sa/a-service-account`
+  // will be treated the same in the Istio mesh.
+  repeated string trust_domain_aliases = 3;
 };
 
 message Payload {
--- a/src/istio/authn/context.proto
+++ b/src/istio/authn/context.proto
@@ -58,6 +58,8 @@ message X509Payload {
   // The `trust_domains` field contains a list of trust domains. A request from
   // `serviceA` to `serviceB` is accepted at `serviceB` if `serviceA` carries a
   // valid x509 cerficicate with the trust domain from the trust_domains list.
+  // The `trust_domains` field should not be empty. It should contain at least
+  // one trust domain.
   repeated string trust_domains = 2;
 };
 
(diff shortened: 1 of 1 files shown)


src/istio/authn/context.proto
```diff
@@ -51,6 +51,22 @@ message X509Payload {
   // Identity carries by X509 certification. This is extracted from SAN field
   // for Istio provided certificate.
   string user = 1;
+
+  // The trust domain corresponds to the trust root of a system.
+  // Refer to
+  // [SPIFFE-ID](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#21-trust-domain)
+  string trust_domain = 2;
```
diemtvu: Can this ever be empty?

liminw: The user facing config (in MeshConfig) can be empty. If it is empty, we use "cluster.local". We can use the same logic here.  Phillip, can you add a comment specifying the default value?

src/istio/authn/context.proto
```diff
@@ -51,6 +51,22 @@ message X509Payload {
...
+  // The trust domain aliases represent the aliases of `trust_domain`.
+  // For example, if we have
+  // ```yaml
+  // trustDomain: td1
+  // trustDomainAliases: [“td2”, "td3"]
+  // ```
+  // Any service with the identity `td1/ns/foo/sa/a-service-account`,
+  // `td2/ns/foo/sa/a-service-account`, or `td3/ns/foo/sa/a-service-account`
+  // will be treated the same in the Istio mesh.
+  repeated string trust_domain_aliases = 3;
```
PiotrSikora: Why are we using aliases instead of defining multiple trust domains (i.e. `repeated string trust_domains`) above?

liminw: Actually this is a good point. The change currently follows what we have in MeshConfig API. But since authentication filter does not need to distinguish trust_domain vs. trust_domain_alias, they can be collapsed into a single trust_domains field. This also simplifies implementation.

pitlv2109: I think we should keep it as it is to align with MeshConfig. Later on, we need to distinguish between "cluster.local" as the local trust domain -> "cluster.local" as alias, and "some-td" as local trust domain -> "cluster.local" as the local trust domain. Also, this will help later with removing the aliases only when the users are done with the migration.

src/istio/authn/context.proto
pitlv2109: The filter is not using trust domain information as of today so I guess it can be empty. If there is no aliases, I guess we don't need to convert this trust domain to the config either, so I think we can leave it as empti-able?

diemtvu: It's not obvious how this information will be used in each case. Are they used to verify the x509 cert? If it is empty, what does this mean etc. 

As pilot will populate this, so it doesn't matter if it is empti-able or not. But from the implementation point of view, it is clearer (and more independent from pilot/istio logic) if the field is always set explicitly.

src/istio/authn/context.proto
pitlv2109: Per offline discussion, I merged trust domain and trust domain aliases field to one. PTAL. Thanks!

src/istio/authn/context.proto
pitlv2109: Merged two fields to one and added some comment. PTAL. Thanks!

src/istio/authn/context.proto
```diff
@@ -55,18 +55,10 @@ message X509Payload {
...
-  // trustDomainAliases: [“td2”, "td3"]
-  // ```
-  // Any service with the identity `td1/ns/foo/sa/a-service-account`,
-  // `td2/ns/foo/sa/a-service-account`, or `td3/ns/foo/sa/a-service-account`
-  // will be treated the same in the Istio mesh.
-  repeated string trust_domain_aliases = 3;
+  // The `trust_domains` field contains a list of trust domains. A request from
+  // `serviceA` to `serviceB` is accepted at `serviceB` if `serviceA` carries a
+  // valid x509 cerficicate with the trust domain from the trust_domains list.
+  repeated string trust_domains = 2;
```
diemtvu: Can you document if this list can be empty, and what is the expected behavior in that case?

pitlv2109: Added a comment about the usage. PTAL.

PR #2439, opened 2019-10-04
feat(edges): add support for meshtelemetry edge export

douglas-reid:
This PR establishes an initial set of functionality for reporting traffic edges via the Mesh Edges service (at meshtelemetry.googleapis.com). It is needed to maintain equivalent functionality with the Mixer adapter from istio/istio in the `mixer/adapter/stackdriver/contextgraph` package as we transition to the extensibility v2 architecture (aka Mixerless Istio).

This PR does not wire this functionality into any part of the larger system.

A `TODO` list is included to track the remaining work, which includes validation / integration testing.

A copy of the proto for the mesh edges service is included, as the proto is not currently published anywhere else.

Reviewers: Please review this with a detailed eye. I'm only barely c++ literate at this point, so I could use a solid review.

Note: most of this PR is based of the excellent examples provided by @bianpengyuan .

/cc @quentinmit

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -14,6 +14,7 @@
  */
 
 #include "extensions/common/context.h"
+
 #include "google/protobuf/util/json_util.h"
 
 // WASM_PROLOG
@@ -56,8 +57,13 @@ google::protobuf::util::Status extractNodeMetadata(
   }
   google::protobuf::util::JsonParseOptions json_parse_options;
   json_parse_options.ignore_unknown_fields = true;
-  return JsonStringToMessage(metadata_json_struct, node_info,
-                             json_parse_options);
+  status =
+      JsonStringToMessage(metadata_json_struct, node_info, json_parse_options);
+
+  // used to extract a node key at creation time.
+  absl::StrAppend(node_info->mutable_node_key(), node_info->name(), ".",
+                  node_info->namespace_());
+  return status;
 }
 
 google::protobuf::util::Status extractLocalNodeMetadata(
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -14,6 +14,7 @@
  */
 
 #include "extensions/common/context.h"
+
 #include "google/protobuf/util/json_util.h"
 
 // WASM_PROLOG
@@ -56,13 +57,8 @@ google::protobuf::util::Status extractNodeMetadata(
   }
   google::protobuf::util::JsonParseOptions json_parse_options;
   json_parse_options.ignore_unknown_fields = true;
-  status =
-      JsonStringToMessage(metadata_json_struct, node_info, json_parse_options);
-
-  // used to extract a node key at creation time.
-  absl::StrAppend(node_info->mutable_node_key(), node_info->name(), ".",
-                  node_info->namespace_());
-  return status;
+  return JsonStringToMessage(metadata_json_struct, node_info,
+                             json_parse_options);
 }
 
 google::protobuf::util::Status extractLocalNodeMetadata(
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -14,7 +14,6 @@
  */
 
 #include "extensions/common/context.h"
-
 #include "google/protobuf/util/json_util.h"
 
 // WASM_PROLOG
--- a/extensions/stackdriver/edges/TODO
+++ b/extensions/stackdriver/edges/TODO
@@ -1,6 +1,5 @@
 (P0) Figure out mesh_uid data source
 (P0) Integration test for grpc endpoint
-(P1) Caching / don't report the same edge repeatedly
 (P1) Retries
 (P1) Better debugging / monitoring (exported metrics)
 (P1) WASM bits
--- a/extensions/stackdriver/edges/TODO
+++ b/extensions/stackdriver/edges/TODO
@@ -1,0 +1,8 @@
+(P0) Figure out mesh_uid data source
+(P0) Integration test for grpc endpoint
+(P1) Caching / don't report the same edge repeatedly
+(P1) Retries
+(P1) Better debugging / monitoring (exported metrics)
+(P1) WASM bits
+(P2) Support for other platforms / error handling when not on GCP
+(P3) Refactoring / alignment with other Stackdriver extension components
--- a/extensions/stackdriver/edges/TODO
+++ b/extensions/stackdriver/edges/TODO
@@ -1,4 +1,3 @@
-(P0) Figure out mesh_uid data source
 (P0) Integration test for grpc endpoint
 (P1) Destination service shortname (as opposed to host)
 (P1) Retries
--- a/extensions/stackdriver/edges/TODO
+++ b/extensions/stackdriver/edges/TODO
@@ -1,5 +1,6 @@
 (P0) Figure out mesh_uid data source
 (P0) Integration test for grpc endpoint
+(P1) Destination service shortname (as opposed to host)
 (P1) Retries
 (P1) Better debugging / monitoring (exported metrics)
 (P1) WASM bits
--- a/extensions/stackdriver/edges/edge_reporter.cc
+++ b/extensions/stackdriver/edges/edge_reporter.cc
@@ -56,7 +56,7 @@ std::unique_ptr<WorkloadInstance> instanceFromMetadata(
   instance->set_owner_uid(node_info.owner());
   instance->set_workload_name(node_info.workload_name());
   instance->set_workload_namespace(node_info.namespace_());
-  return std::move(instance);
+  return instance;
 };
 
 EdgeReporter::EdgeReporter(
--- a/extensions/stackdriver/edges/edge_reporter.cc
+++ b/extensions/stackdriver/edges/edge_reporter.cc
@@ -61,7 +61,7 @@ void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
 EdgeReporter::EdgeReporter(const ::wasm::common::NodeInfo& local_node_info,
                            std::unique_ptr<MeshEdgesServiceClient> edges_client,
                            TimestampFn now)
-    : now_(now) {
+    : edges_client_(std::move(edges_client)), now_(now) {
   current_request_ = std::make_unique<ReportTrafficAssertionsRequest>();
 
   const auto& project_id =
@@ -80,18 +80,19 @@ EdgeReporter::EdgeReporter(const ::wasm::common::NodeInfo& local_node_info,
                   "/", location, "/meshes/", cluster);
 
   instanceFromMetadata(local_node_info, &node_instance_);
-
-  edges_client_ = std::move(edges_client);
 };
+
+EdgeReporter::~EdgeReporter() { reportEdges(); }
 
 // ONLY inbound
 void EdgeReporter::addEdge(const ::Wasm::Common::RequestInfo& request_info,
+                           const std::string peer_metadata_id_key,
                            const ::wasm::common::NodeInfo& peer_node_info) {
-  auto peer = current_peers_.find(peer_node_info.node_key());
+  auto peer = current_peers_.find(peer_metadata_id_key);
   if (peer != current_peers_.end()) {
     return;
   }
-  current_peers_.emplace(peer_node_info.node_key());
+  current_peers_.emplace(peer_metadata_id_key);
 
   auto* traffic_assertions = current_request_->mutable_traffic_assertions();
   auto* edge = traffic_assertions->Add();
@@ -116,32 +117,26 @@ void EdgeReporter::addEdge(const ::Wasm::Common::RequestInfo& request_info,
 
   if (current_request_->traffic_assertions_size() >
       max_assertions_per_request_) {
-    flush();
+    reportEdges();
   }
 };  // namespace Edges
 
 void EdgeReporter::reportEdges() {
-  flush();
-  for (auto& req : request_queue_) {
-    edges_client_->reportTrafficAssertions(*req);
-  }
-  request_queue_.clear();
-}
-
-void EdgeReporter::flush() {
   if (current_request_->traffic_assertions_size() == 0) {
     return;
   }
 
-  std::unique_ptr<ReportTrafficAssertionsRequest> next =
+  std::unique_ptr<ReportTrafficAssertionsRequest> queued_request =
       std::make_unique<ReportTrafficAssertionsRequest>();
-  next->set_parent(current_request_->parent());
-  next->set_mesh_uid(current_request_->mesh_uid());
+  queued_request->set_parent(current_request_->parent());
+  queued_request->set_mesh_uid(current_request_->mesh_uid());
 
   current_peers_.clear();
-  current_request_.swap(next);
-  *next->mutable_timestamp() = now_();
-  request_queue_.emplace_back(std::move(next));
+  current_request_.swap(queued_request);
+
+  // set the timestamp and then send the queued request
+  *queued_request->mutable_timestamp() = now_();
+  edges_client_->reportTrafficAssertions(*queued_request);
 }
 
 }  // namespace Edges
--- a/extensions/stackdriver/edges/edge_reporter.cc
+++ b/extensions/stackdriver/edges/edge_reporter.cc
@@ -85,7 +85,12 @@ EdgeReporter::EdgeReporter(const ::wasm::common::NodeInfo& local_node_info,
   instanceFromMetadata(local_node_info, &node_instance_);
 };
 
-EdgeReporter::~EdgeReporter() { reportEdges(); }
+EdgeReporter::~EdgeReporter() {
+  // if (current_request_->traffic_assertions_size() == 0 ||
+  // !queued_requests_.empty()) {
+  //   logWarn("EdgeReporter had uncommitted TrafficAssertions when shutdown.");
+  // }
+}
 
 // ONLY inbound
 void EdgeReporter::addEdge(const ::Wasm::Common::RequestInfo& request_info,
@@ -125,6 +130,14 @@ void EdgeReporter::addEdge(const ::Wasm::Common::RequestInfo& request_info,
 };  // namespace Edges
 
 void EdgeReporter::reportEdges() {
+  flush();
+  for (auto& req : queued_requests_) {
+    edges_client_->reportTrafficAssertions(*req.get());
+  }
+  queued_requests_.clear();
+};
+
+void EdgeReporter::flush() {
   if (current_request_->traffic_assertions_size() == 0) {
     return;
   }
@@ -139,7 +152,7 @@ void EdgeReporter::reportEdges() {
 
   // set the timestamp and then send the queued request
   *queued_request->mutable_timestamp() = now_();
-  edges_client_->reportTrafficAssertions(*queued_request);
+  queued_requests_.emplace_back(std::move(queued_request));
 }
 
 }  // namespace Edges
--- a/extensions/stackdriver/edges/edge_reporter.cc
+++ b/extensions/stackdriver/edges/edge_reporter.cc
@@ -17,7 +17,6 @@
 
 #include "extensions/stackdriver/common/constants.h"
 #include "extensions/stackdriver/edges/edges.pb.h"
-#include "google/protobuf/util/time_util.h"
 
 #ifndef NULL_PLUGIN
 #include "api/wasm/cpp/proxy_wasm_intrinsics.h"
@@ -40,7 +39,6 @@ using google::cloud::meshtelemetry::v1alpha1::
 using google::cloud::meshtelemetry::v1alpha1::
     TrafficAssertion_Protocol_PROTOCOL_TCP;
 using google::cloud::meshtelemetry::v1alpha1::WorkloadInstance;
-using google::protobuf::util::TimeUtil;
 
 namespace {
 void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
@@ -57,11 +55,13 @@ void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
   instance->set_workload_name(node_info.workload_name());
   instance->set_workload_namespace(node_info.namespace_());
 };
+
 }  // namespace
 
-EdgeReporter::EdgeReporter(
-    const ::wasm::common::NodeInfo& local_node_info,
-    std::unique_ptr<MeshEdgesServiceClient> edges_client) {
+EdgeReporter::EdgeReporter(const ::wasm::common::NodeInfo& local_node_info,
+                           std::unique_ptr<MeshEdgesServiceClient> edges_client,
+                           TimestampFn now)
+    : now_(now) {
   current_request_ = std::make_unique<ReportTrafficAssertionsRequest>();
 
   const auto& project_id =
@@ -75,7 +75,6 @@ EdgeReporter::EdgeReporter(
       local_node_info.platform_metadata().at(Common::kGCPClusterLocationKey);
   std::string cluster =
       local_node_info.platform_metadata().at(Common::kGCPClusterNameKey);
-  std::string mesh_uid = "unknown";
   absl::StrAppend(current_request_->mutable_mesh_uid(),
                   "//cloudresourcemanager.googleapis.com/projects/", project_id,
                   "/", location, "/meshes/", cluster);
@@ -97,6 +96,8 @@ void EdgeReporter::addEdge(const ::Wasm::Common::RequestInfo& request_info,
   auto* traffic_assertions = current_request_->mutable_traffic_assertions();
   auto* edge = traffic_assertions->Add();
 
+  // TODO(douglas-reid): use the short name for the destination service when
+  // available Right now, this uses destination host instead.
   edge->set_destination_service_name(request_info.destination_service_host);
   edge->set_destination_service_namespace(node_instance_.workload_namespace());
   instanceFromMetadata(peer_node_info, edge->mutable_source());
@@ -122,7 +123,7 @@ void EdgeReporter::addEdge(const ::Wasm::Common::RequestInfo& request_info,
 void EdgeReporter::reportEdges() {
   flush();
   for (auto& req : request_queue_) {
-    edges_client_->reportTrafficAssertions(std::move(req));
+    edges_client_->reportTrafficAssertions(*req);
   }
   request_queue_.clear();
 }
@@ -139,7 +140,7 @@ void EdgeReporter::flush() {
 
   current_peers_.clear();
   current_request_.swap(next);
-  *next->mutable_timestamp() = TimeUtil::GetCurrentTime();
+  *next->mutable_timestamp() = now_();
   request_queue_.emplace_back(std::move(next));
 }
 
--- a/extensions/stackdriver/edges/edge_reporter.cc
+++ b/extensions/stackdriver/edges/edge_reporter.cc
@@ -42,11 +42,12 @@ using google::cloud::meshtelemetry::v1alpha1::
 using google::cloud::meshtelemetry::v1alpha1::WorkloadInstance;
 using google::protobuf::util::TimeUtil;
 
+namespace {
 void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
                           WorkloadInstance* instance) {
   // TODO(douglas-reid): support more than just kubernetes instances
-  instance->set_uid("kubernetes://" + node_info.name() + "." +
-                    node_info.namespace_());
+  absl::StrAppend(instance->mutable_uid(), "kubernetes://", node_info.name(),
+                  ".", node_info.namespace_());
   // TODO(douglas-reid): support more than just GCP ?
   instance->set_location(
       node_info.platform_metadata().at(Common::kGCPClusterLocationKey));
@@ -55,8 +56,8 @@ void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
   instance->set_owner_uid(node_info.owner());
   instance->set_workload_name(node_info.workload_name());
   instance->set_workload_namespace(node_info.namespace_());
-  return;
 };
+}  // namespace
 
 EdgeReporter::EdgeReporter(
     const ::wasm::common::NodeInfo& local_node_info,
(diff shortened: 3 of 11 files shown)


extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,141 @@
...
+using google::cloud::meshtelemetry::v1alpha1::
+    TrafficAssertion_Protocol_PROTOCOL_HTTP;
+using google::cloud::meshtelemetry::v1alpha1::
+    TrafficAssertion_Protocol_PROTOCOL_HTTPS;
+using google::cloud::meshtelemetry::v1alpha1::
+    TrafficAssertion_Protocol_PROTOCOL_TCP;
+using google::cloud::meshtelemetry::v1alpha1::WorkloadInstance;
+using google::protobuf::util::TimeUtil;
+
+std::unique_ptr<WorkloadInstance> instanceFromMetadata(
```
mandarjog: `void instanceFromMetadata (const ::wasm::common::NodeInfo& node_info, WorkloadInstance* instance)` instead

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,141 @@
...
+  node_instance_ = instanceFromMetadata(local_node_info);
+
+  edges_client_ = std::move(edges_client);
+};
+
+// ONLY inbound
+void EdgeReporter::addEdge(const ::Wasm::Common::RequestInfo& request_info,
+                           const ::wasm::common::NodeInfo& peer_node_info) {
+  auto* traffic_assertions = current_request_->mutable_traffic_assertions();
+  auto* edge = traffic_assertions->Add();
```
mandarjog: This using too many allocs and adding too many samples.
Look at https://github.com/istio/proxy/blob/master/extensions/stats/plugin.cc#L118
you can use a map like in the stats plugin, define a key that includes important, or *all* bits you need to use.

extensions/stackdriver/edges/edge_reporter_test.cc
```diff
@@ -0,0 +1,190 @@
...
+  return request_info;
+}
+
+std::unique_ptr<ReportTrafficAssertionsRequest> want() {
+  auto request_info = requestInfo();
+  auto peer_node_info = peerNodeInfo();
+  auto node_info = nodeInfo();
+
+  auto req = std::make_unique<ReportTrafficAssertionsRequest>();
+  req->set_parent("projects/test_project");
```
mandarjog: This is ok, but you can make it easier by using proto parsing. https://github.com/istio/proxy/blob/master/src/istio/control/tcp/request_handler_impl_test.cc#L89

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: done.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: I was avoiding optimizations to start, but have since added a TrafficAssertionsRequest-bound cache that will prevent adding the same edge over and over. there is no cache across reports to the telemetry service yet, but that can be added later.

let me know what you think.

extensions/stackdriver/edges/TODO
```diff
@@ -0,0 +1,7 @@
+(P0) Figure out mesh_uid data source
```
kyessenov: nit: open an issue/epic instead

extensions/stackdriver/edges/edge_reporter.h
```diff
@@ -0,0 +1,86 @@
...
+  void reportEdges();
+
+ private:
+  // flush pushes the current pending request (iff there have been
+  // edges added) into the request buffer, and swaps in a new request
+  // to gather additional edges.
+  void flush();
+
+  // represents the workload instance for the current proxy
+  std::unique_ptr<WorkloadInstance> node_instance_;
```
kyessenov: Unique ptr seems unnecessary since it never escapes outside the class scope. You could just store the proto by value.

extensions/stackdriver/edges/edge_reporter.h
```diff
@@ -0,0 +1,86 @@
...
+  std::vector<std::unique_ptr<ReportTrafficAssertionsRequest>> request_queue_;
+
+  // client used to send requests to the edges service
+  std::unique_ptr<MeshEdgesServiceClient> edges_client_;
+
+  // map of current peers for which edges have been created in the
+  // current_request_;
+  std::map<std::string, bool> current_peers_;
+
+  std::mutex request_mutex_;
```
kyessenov: Why do we need a mutex? Envoy is mostly single-threaded and any mutex on the single thread will hurt performance. Can you add a comment on the reason for a mutex?

extensions/stackdriver/edges/edge_reporter.h
```diff
@@ -0,0 +1,86 @@
...
+
+  // the buffer of pending requests, awaiting transmission to the edges service
+  std::vector<std::unique_ptr<ReportTrafficAssertionsRequest>> request_queue_;
+
+  // client used to send requests to the edges service
+  std::unique_ptr<MeshEdgesServiceClient> edges_client_;
+
+  // map of current peers for which edges have been created in the
+  // current_request_;
+  std::map<std::string, bool> current_peers_;
```
kyessenov: why not use a std::set?

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,156 @@
...
+using google::cloud::meshtelemetry::v1alpha1::
+    TrafficAssertion_Protocol_PROTOCOL_HTTP;
+using google::cloud::meshtelemetry::v1alpha1::
+    TrafficAssertion_Protocol_PROTOCOL_HTTPS;
+using google::cloud::meshtelemetry::v1alpha1::
+    TrafficAssertion_Protocol_PROTOCOL_TCP;
+using google::cloud::meshtelemetry::v1alpha1::WorkloadInstance;
+using google::protobuf::util::TimeUtil;
+
+void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
```
kyessenov: Helper methods are usually hidden using an anonymous namespace: `namespace { void helper() {} }`

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,156 @@
...
+                          WorkloadInstance* instance) {
+  // TODO(douglas-reid): support more than just kubernetes instances
+  instance->set_uid("kubernetes://" + node_info.name() + "." +
```
kyessenov: absl::StrCat should in theory be faster for multiple string concatenations.

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,156 @@
...
+                    node_info.namespace_());
+  // TODO(douglas-reid): support more than just GCP ?
+  instance->set_location(
+      node_info.platform_metadata().at(Common::kGCPClusterLocationKey));
+  instance->set_cluster_name(
+      node_info.platform_metadata().at(Common::kGCPClusterNameKey));
+  instance->set_owner_uid(node_info.owner());
+  instance->set_workload_name(node_info.workload_name());
+  instance->set_workload_namespace(node_info.namespace_());
+  return;
```
kyessenov: nit: unnecessary `return`

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,156 @@
...
+      location + "/meshes/" + cluster);
+
+  node_instance_ = std::make_unique<WorkloadInstance>();
+  instanceFromMetadata(local_node_info, node_instance_.get());
+
+  edges_client_ = std::move(edges_client);
+};
+
+std::string nodeKey(const ::wasm::common::NodeInfo& node_info) {
+  return node_info.name() + "." + node_info.namespace_();
```
kyessenov: This is allocating a string every time. Suggest to save the string by value in the class.

extensions/stackdriver/edges/edge_reporter.h
douglas-reid: Is the use of this class guaranteed to be in a single-threaded context?  If so, it isn't necessary. But if a `flush` can happen, we need someway to make sure that the map isn't checked when it is also being cleared, etc.

extensions/stackdriver/edges/edge_reporter.h
douglas-reid: old golang habits die hard.  Will update.

extensions/stackdriver/edges/edge_reporter.h
kyessenov: WASM is single threaded. So the callbacks should happen on a single execution context I believe. If you want to flush across multiple workers, then that is a deliberate choice that requires creating a shared state and storing the requests buffers in there.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: done.

extensions/stackdriver/edges/edge_reporter.h
douglas-reid: done.

extensions/stackdriver/edges/edge_reporter.h
douglas-reid: ok. removed. added comment to header file about single-threaded assumption.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: updated.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: removed.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: I added this bit to `context.cc`, as NodeInfo is a proto-generated class. I'm not sure if that is better or not.

extensions/stackdriver/edges/edge_reporter_test.cc
douglas-reid: converted where this made sense.

extensions/common/context.cc
```diff
@@ -14,6 +14,7 @@
  */
 
 #include "extensions/common/context.h"
+
```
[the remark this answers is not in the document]
douglas-reid: I've attempted but I think i'm overruled by clang-format.

extensions/stackdriver/edges/mesh_edges_service_client.cc
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "extensions/stackdriver/edges/mesh_edges_service_client.h"
+
```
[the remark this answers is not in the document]
douglas-reid: I've attempted but I think i'm overruled by clang-format.

extensions/stackdriver/edges/mesh_edges_service_client.h
```diff
@@ -0,0 +1,88 @@
...
+// by Stackdriver, using application default credentials.
+class MeshEdgesServiceClientImpl : public MeshEdgesServiceClient {
+ public:
+  // root_context is the wasm runtime context
+  // edges_endpoint is an optional param used to specify alternative service
+  // address.
+  MeshEdgesServiceClientImpl(RootContext* root_context,
+                             std::string edges_endpoint);
+
+  virtual void reportTrafficAssertions(
```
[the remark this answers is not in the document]
douglas-reid: done.

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,148 @@
...
+  current_request_->set_parent("projects/" + project_id);
+
+  // TODO(dougreid): figure out how to get the real mesh uid here.
+  // Using: //.../projects/<project_id>/<location>/meshes/<cluster> as a
+  // placeholder.
+  std::string location =
+      local_node_info.platform_metadata().at(Common::kGCPClusterLocationKey);
+  std::string cluster =
+      local_node_info.platform_metadata().at(Common::kGCPClusterNameKey);
+  std::string mesh_uid = "unknown";
```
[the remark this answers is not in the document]
douglas-reid: nope. removed.

extensions/stackdriver/edges/edge_reporter.h
```diff
@@ -0,0 +1,85 @@
...
+  std::unique_ptr<ReportTrafficAssertionsRequest> current_request_;
+
+  // the buffer of pending requests, awaiting transmission to the edges service
+  std::vector<std::unique_ptr<ReportTrafficAssertionsRequest>> request_queue_;
+
+  // client used to send requests to the edges service
+  std::unique_ptr<MeshEdgesServiceClient> edges_client_;
+
+  // current peers for which edges have been created in current_request_;
+  std::set<std::string> current_peers_;
```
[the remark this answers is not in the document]
douglas-reid: done.

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  std::unique_ptr<ReportTrafficAssertionsRequest> next =
+      std::make_unique<ReportTrafficAssertionsRequest>();
+  next->set_parent(current_request_->parent());
+  next->set_mesh_uid(current_request_->mesh_uid());
+
+  current_peers_.clear();
+  current_request_.swap(next);
+  *next->mutable_timestamp() = TimeUtil::GetCurrentTime();
```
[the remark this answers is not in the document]
douglas-reid: Just switching causes seg faults when used in tests (see below), so I've made a TimestampFn that can be passed in and set a default to use the wasm API.

```
[2019-10-08 17:09:55.505][14][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:81] Caught Segmentation fault, suspect faulting address 0x0
[2019-10-08 17:09:55.505][14][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:69] Backtrace (use tools/stack_decode.py to get line numbers):
[2019-10-08 17:09:55.514][14][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:73] #0: Envoy::SignalAction::sigHandler() [0x5646547c2f32]
[2019-10-08 17:09:55.514][14][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:73] #1: __restore_rt [0x7f6d4be8b3a0]
[2019-10-08 17:09:55.522][14][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:73] #2: Envoy::Extensions::Common::Wasm::Null::Plugin::proxy_getCurrentTimeNanoseconds() [0x56465349f8c8]
[2019-10-08 17:09:55.530][14][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:73] #3: Envoy::Extensions::Common::Wasm::Null::Plugin::getCurrentTimeNanoseconds() [0x5646534a05a4]
```

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,148 @@
...
+  if (current_request_->traffic_assertions_size() >
+      max_assertions_per_request_) {
+    flush();
+  }
+};  // namespace Edges
+
+void EdgeReporter::reportEdges() {
+  flush();
+  for (auto& req : request_queue_) {
+    edges_client_->reportTrafficAssertions(std::move(req));
```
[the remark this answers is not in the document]
douglas-reid: done.

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,148 @@
...
+  auto peer = current_peers_.find(peer_node_info.node_key());
+  if (peer != current_peers_.end()) {
+    return;
+  }
+  current_peers_.emplace(peer_node_info.node_key());
+
+  auto* traffic_assertions = current_request_->mutable_traffic_assertions();
+  auto* edge = traffic_assertions->Add();
+
+  edge->set_destination_service_name(request_info.destination_service_host);
```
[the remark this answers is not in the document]
douglas-reid: I think we want it to be the short name ultimately, but it isn't used on the server side at the moment. I added a TODO to patch when short name is available in metadata, etc.

extensions/stackdriver/edges/edge_reporter.cc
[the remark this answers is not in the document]
bianpengyuan: Ah, it is because the wasm runtime context is empty. Any calls from filter/modules into envoy host needs to attach to a runtime context: https://github.com/envoyproxy/envoy-wasm/blob/724328c7d84bae38d9e4aa7047121a949a6fd511/source/extensions/common/wasm/wasm.cc#L42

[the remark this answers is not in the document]
bianpengyuan: There is no easy way to add unit test for method using WASM APIs, there is no mock provided. Let's switch back to `TimeUtil::GetCurrentTime()` for now and worry about wasm later.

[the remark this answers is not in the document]
bianpengyuan: nvm, I understand your change better now. and yeah it is viable 👍

extensions/common/context.cc
```diff
@@ -56,8 +56,13 @@ google::protobuf::util::Status extractNodeMetadata(
   }
   google::protobuf::util::JsonParseOptions json_parse_options;
   json_parse_options.ignore_unknown_fields = true;
-  return JsonStringToMessage(metadata_json_struct, node_info,
-                             json_parse_options);
+  status =
+      JsonStringToMessage(metadata_json_struct, node_info, json_parse_options);
+
+  // used to extract a node key at creation time.
+  absl::StrAppend(node_info->mutable_node_key(), node_info->name(), ".",
```
kyessenov: why is this needed? this has no side-effects, right?

extensions/common/context.cc
```diff
@@ -56,8 +56,13 @@ google::protobuf::util::Status extractNodeMetadata(
...
+                  node_info->namespace_());
```
kyessenov: I wonder why `namespace_` has an underscore but `name` does not. Is this a typo?

bianpengyuan: No protobuf codegen add that.

kyessenov: Ah, probably because `namespace` is a keyword.

bianpengyuan: yep.

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,149 @@
...
+    TrafficAssertion_Protocol_PROTOCOL_HTTPS;
+using google::cloud::meshtelemetry::v1alpha1::
+    TrafficAssertion_Protocol_PROTOCOL_TCP;
+using google::cloud::meshtelemetry::v1alpha1::WorkloadInstance;
+
+namespace {
+void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
+                          WorkloadInstance* instance) {
+  // TODO(douglas-reid): support more than just kubernetes instances
+  absl::StrAppend(instance->mutable_uid(), "kubernetes://", node_info.name(),
```
kyessenov: nit: use `node_info.node_key()`?

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,149 @@
...
+      local_node_info.platform_metadata().at(Common::kGCPClusterLocationKey);
+  std::string cluster =
+      local_node_info.platform_metadata().at(Common::kGCPClusterNameKey);
+  absl::StrAppend(current_request_->mutable_mesh_uid(),
+                  "//cloudresourcemanager.googleapis.com/projects/", project_id,
+                  "/", location, "/meshes/", cluster);
+
+  instanceFromMetadata(local_node_info, &node_instance_);
+
+  edges_client_ = std::move(edges_client);
```
kyessenov: It's a good idea to keep order of initialization and declaration the same, since the default destructor deletes them in the reverse order of declaration.

kyessenov: In this case, you might just get away with an initializer list above: `edges_client_(edges_client)`. The compiler might insert the move automatically. Make sure the order is the same though in the list and field declaration.

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,149 @@
...
+
+void EdgeReporter::reportEdges() {
+  flush();
+  for (auto& req : request_queue_) {
+    edges_client_->reportTrafficAssertions(*req);
+  }
+  request_queue_.clear();
+}
+
+void EdgeReporter::flush() {
```
kyessenov: I think it would be cleaner to report in the flush. WDYT?

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,149 @@
...
+    edge->set_protocol(TrafficAssertion_Protocol_PROTOCOL_HTTPS);
+  } else if (protocol == "grpc" || protocol == "GRPC") {
+    edge->set_protocol(TrafficAssertion_Protocol_PROTOCOL_GRPC);
+  } else {
+    edge->set_protocol(TrafficAssertion_Protocol_PROTOCOL_TCP);
+  }
+
+  if (current_request_->traffic_assertions_size() >
+      max_assertions_per_request_) {
+    flush();
```
kyessenov: Are we just dropping edges here without reporting? Should we log a warning at least?

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,149 @@
...
+
+  std::unique_ptr<ReportTrafficAssertionsRequest> next =
+      std::make_unique<ReportTrafficAssertionsRequest>();
+  next->set_parent(current_request_->parent());
+  next->set_mesh_uid(current_request_->mesh_uid());
+
+  current_peers_.clear();
+  current_request_.swap(next);
+  *next->mutable_timestamp() = now_();
+  request_queue_.emplace_back(std::move(next));
```
kyessenov: This is confusing. Who owns the current request? Is it the vector of the edge reporter? I think you should make a choice, and then store a pointer in the other. For example, you can keep a vector of unique pointers, and then store a raw pointer as `current_request_`.

extensions/stackdriver/edges/edge_reporter.h
```diff
@@ -0,0 +1,100 @@
...
+// "edges" for a mesh. It should be used **only** to document incoming edges for
+// a proxy. This means that the proxy in which this reporter is running should
+// be the destination workload instance for all reported traffic.
+// This should only be used in a single-threaded context. No support for
+// threading is currently provided.
+class EdgeReporter {
+  typedef std::function<google::protobuf::Timestamp()> TimestampFn;
+
+ public:
+  EdgeReporter(const ::wasm::common::NodeInfo &local_node_info,
```
kyessenov: I think you need a destructor that flushes/reports the edges. You might need to be careful with the lifetime of the client as the reporter is being destructed (e.g. callbacks might become stale).

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: No. We shouldn't be dropping edges. I don't believe there is a hard upper-limit on the server side, and the limit here is just to keep the requests somewhat limited in size.

extensions/common/context.cc
bianpengyuan: Instead of constructing a new key, how about using the node metadata id? It is also available in filter state: https://github.com/istio/proxy/blob/d5701c068d4e04971ba4c640ae27db70b730b632/extensions/stats/plugin.cc#L161

Then in addEdge call, we could add a new param `metadata_id`, which could be used as cache key here.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: I don't have a good feel for the overall behavioral model here. Is it better to separate the batching of requests from the sending, and keep the `addEdge()` flow from potentially directly causing gRPC calls?

I adopted this from @bianpengyuan 's logging PR. @bianpengyuan what are your thoughts here?

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: EdgeReporter owns the current request. But I'll take a stab at simplifying.

extensions/stackdriver/edges/edge_reporter.cc
bianpengyuan: I think the plan is to trigger report by timer but not flush, which could happen in the middle of a request. If export is triggered in flush, will that increase request tail latency?

extensions/stackdriver/edges/edge_reporter.h
bianpengyuan: good point, I need to add this to logging part as well..

extensions/stackdriver/edges/edge_reporter.cc
bianpengyuan: Actually this should all happen in onLog filter callback, which should not affect request tail latency, so yeah I think it is fine to just report the batch whenever flush is triggered.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: Removed `node_key`.

extensions/stackdriver/edges/edge_reporter.h
douglas-reid: Added a simple destructor which should call `reportEdges()`. That should be OK re: lifetimes I believe, as the client should be destructed after completion of the destructor, correct?

extensions/common/context.cc
douglas-reid: Done.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: Entirely removed.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: Simplified down to a single method.

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: I've tried to address ordering. PTAL.

extensions/stackdriver/edges/TODO
douglas-reid: Updated code to grab this from nodeInfo.

extensions/stackdriver/edges/edge_reporter.h
bianpengyuan: Yes it should be fine, it need special care when we add retry in the future.

extensions/stackdriver/edges/edge_reporter.h
```diff
@@ -0,0 +1,95 @@
...
+using Envoy::Extensions::Common::Wasm::Null::Plugin::Extensions::Stackdriver::
+    Edges::MeshEdgesServiceClient;
+#endif
+
+using Envoy::Extensions::Common::Wasm::Null::Plugin::getCurrentTimeNanoseconds;
+using google::cloud::meshtelemetry::v1alpha1::ReportTrafficAssertionsRequest;
+using google::cloud::meshtelemetry::v1alpha1::WorkloadInstance;
+using google::protobuf::util::TimeUtil;
+
+namespace {
```
bianpengyuan: I think it is not recommended to use unnamed namespace in header file: https://google.github.io/styleguide/cppguide.html#Unnamed_Namespaces_and_Static_Variables. The way I can think of is to make it a lambda:
```
class EdgeReporter {
 private:
  TimestampFn now_ = []() {
       return TimeUtil::NanosecondsToTimestamp(getCurrentTimeNanoseconds());
   }
}
```
then just pass in nullptr in actual logic and pass in proto current time function in test. like:
```
EdgeReporter(node_info, client, nullptr)
```

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,141 @@
...
+                  "/meshes/", mesh_id);
+
+  instanceFromMetadata(local_node_info, &node_instance_);
+};
+
+EdgeReporter::~EdgeReporter() { reportEdges(); }
+
+// ONLY inbound
+void EdgeReporter::addEdge(const ::Wasm::Common::RequestInfo& request_info,
+                           const std::string peer_metadata_id_key,
```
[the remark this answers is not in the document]
douglas-reid: done.

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,141 @@
...
+                           const ::wasm::common::NodeInfo& peer_node_info) {
+  auto peer = current_peers_.find(peer_metadata_id_key);
```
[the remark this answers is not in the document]
douglas-reid: cool. done.

extensions/stackdriver/edges/edge_reporter.h
douglas-reid: I added another ctor and pass the lambda in that way. let me know if that's ok.

bianpengyuan:
Ah, I remember the reason of queueing the request. WASM runtime assumes all calls from the WASM module are triggered inside host calls (e.g. onHeaders, onTimers, etc), otherwise host won't have context about the calls from module, even for a grpc call which should always attach to root context.

bianpengyuan:
/hold

Can we switch back to what we have?

douglas-reid:
@bianpengyuan by switch back, you mean returned to a flush + request queue?

bianpengyuan:
@douglas-reid Yes. And actually even the final export call in deconstructor will not work either.. @jplevyak @PiotrSikora Do you have any suggestions on how to deal with this? Is it possible to update the host code to allow calls initiated by the module?

douglas-reid:
@bianpengyuan ok. i've reverted to prior `reportEdges()` + `flush()` logic for now. I've also made the deconstructor do nothing pending further instruction. 

Given the release crunch, etc., if this looks OK, I'd prefer getting this into the codebase and iterating on improvements in follow-ons (as this isn't connected to anything yet).

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -0,0 +1,160 @@
...
+  if (current_request_->traffic_assertions_size() >
+      max_assertions_per_request_) {
+    reportEdges();
+  }
+};  // namespace Edges
+
+void EdgeReporter::reportEdges() {
+  flush();
+  for (auto& req : queued_requests_) {
+    edges_client_->reportTrafficAssertions(*req.get());
```
bianpengyuan: nit: do you need to do `.get()`? Could it just be `*req`?

PR #2441, opened 2019-10-08
Update common files

geeknoid:


geeknoid:
@PiotrSikora I'm trying to modernize this repo to use our common files as all other repos. I can't figure out why all the CI jobs are failing.

In theory, the only difference with this new setup is that we have double makes. The top-level makefile ends up just launching Makefile.core.mk, and Makefile.core.mk contains the same thing as the old Makefile.

Can you tell what is getting confused?

PiotrSikora:
@geeknoid it looks that the LLVM path isn't included in `PATH` in the migrated `Makefile`, when using `make -e` (in which case it gets `PATH` from the environment).

You can either:
1) Remove `-e` from the `make` invocation (assuming it's not needed elsewhere).
2) Use a different variable and override `PATH` in the make targets themselves, e.g.
```
-PATH := /usr/lib/llvm-8/bin:$(PATH)
+LLVMPATH := /usr/lib/llvm-8/bin
```
```
 build:
-       export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_DEV) $(BAZEL_TARGETS)
+       export PATH=$(LLVMPATH):$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_DEV) $(BAZEL_TARGETS)
```

geeknoid:
@PiotrSikora Thanks!

geeknoid:
@PiotrSikora This is finally passing the tests. Can you approve please?

Makefile
```diff
@@ -1,102 +1,102 @@
...
+	--mount type=bind,source="$(PWD)",destination="/work" \
+	--mount type=volume,source=go,destination="/go" \
+	--mount type=volume,source=gocache,destination="/gocache" \
+	-w /work $(IMG)
 
-test_tsan:
-	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_TSAN) //src/envoy:envoy
-	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) test $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_TSAN) --test_env=TSAN_OPTIONS=suppressions=$(TOP)/tsan.suppressions -- $(BAZEL_TARGETS) $(SANITIZER_EXCLUSIONS)
-	GO111MODULE=on go test ./...
+MAKE = $(RUN) make --no-print-directory -e -f Makefile.core.mk
```
PiotrSikora: Note: this still breaks `PATH` in the `BUILD_WITH_CONTAINER=1 make build` builds that fail immediately.

PiotrSikora: Why do we care about all host environment variables to be passed to docker? Perhaps we could explicitly pass those that are needed?

Alternatively, could you replace the `PATH` in `Makefile.core.mk` with something LLVM-specific, and prepend it in make targets, like I suggested earlier?

geeknoid: I'm not trying to get BUILD_WITH_CONTAINER=1 to work at all with this PR. I'm just trying to get the common files on par with all the other repos. After that, I/we can start working on container-based builds.

PiotrSikora: Sounds good.

PR #2443, opened 2019-10-08
Add support for BUILD_WITH_CONTAINER=1.

PiotrSikora:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,11 @@ BAZEL_STARTUP_ARGS := --client_debug $(BAZEL_STARTUP_ARGS)
 BAZEL_BUILD_ARGS := -s $(BAZEL_BUILD_ARGS)
 endif
 
+BUILD_WITH_CONTAINER ?= 0
+ifeq "$(BUILD_WITH_CONTAINER)" "1"
+BAZEL_BUILD_ARGS := $(BAZEL_BUILD_ARGS) --config=docker-clang-libc++
+endif
+
 UNAME := $(shell uname)
 ifeq ($(UNAME),Linux)
 BAZEL_CONFIG_DEV  = --config=libc++


PiotrSikora:
@mandarjog @geeknoid note that this still requires Bazelisk to be available in your system, but everything else is build inside Docker, using `gcr.io/istio-testing/prowbazel:0.5.10` container.

Makefile
```diff
@@ -37,6 +37,11 @@ BAZEL_STARTUP_ARGS := --client_debug $(BAZEL_STARTUP_ARGS)
 BAZEL_BUILD_ARGS := -s $(BAZEL_BUILD_ARGS)
 endif
 
+BUILD_WITH_CONTAINER ?= 0
+ifeq "$(BUILD_WITH_CONTAINER)" "1"
+BAZEL_BUILD_ARGS := $(BAZEL_BUILD_ARGS) --config=docker-clang-libc++
```
kyessenov: Where is the magic that defines this config?

PiotrSikora: `.bazelrc`

kyessenov: Hehe, but that just makes the executor spawn docker containers to run jobs. Bazel itself is not under docker, right?

PiotrSikora: Correct. Everything else is, though.

geeknoid:
My goal is to abandon prowbazel and instead having a container that builds on top of the common build-tools container we use in all other repos. This extended build-tools would include bazel and so the whole build process would run within docker. The only dependency on the environment will be to have 'make' and 'docker', everything else will come from the container.

So I'm not sure if this PR is useful towards that final direction or not. WDYT?

Makefile
```diff
@@ -37,6 +37,11 @@ BAZEL_STARTUP_ARGS := --client_debug $(BAZEL_STARTUP_ARGS)
 BAZEL_BUILD_ARGS := -s $(BAZEL_BUILD_ARGS)
 endif
 
+BUILD_WITH_CONTAINER ?= 0
```
mandarjog: does this mean I do not need anything other than bazel installed locally?

kyessenov: It's not different than just using bazel. Regular bazel uses sandboxes to run jobs, it's just not a docker sandbox. The problem would be that the generated build artifacts cannot be run locally since they link against container libraries. That would make any debugging or symbol resolution harder.

kyessenov: Do keep in mind this warning: ```Note: Builds take noticeably more time to complete when the Docker sandbox feature is enabled. This is normal.``` 

We should probably enable RBE to utilize an entire cluster to run bazel, instead of using docker on a single machine.

PiotrSikora: @mandarjog correct, but ideally Bazelisk instead of Bazel (to avoid updating it by hand).

@kyessenov it's different, since "sandboxed" builds still use the toolchain and build tools from the host environment, whereas docker build uses whatever is installed there.

I agree that we should use RBE (me & John are already sharing one build farm), but that's orthogonal to this change, which tries to remove local build dependencies.

Makefile
mandarjog: What I am looking for is `make build_envoy BULD_WITH_CONTAINER=1` will build me an image without having to install anything locally. Building inside the docker container when you could also build locally is adding an orthogonal feature.

Makefile
kyessenov: Well, there is a limited value to docker-sandbox. We already run bazel inside a docker container, so we don't need nested/peer containers. It would be nice to change Makefiles to use the build container to run everything.

PiotrSikora:
@geeknoid with this PR, you still need to have Bazel(isk) installed on the host, since it's managing the execution inside the container. The difference between `BUILD_WITH_CONTAINER=1` and `BUILD_WITH_CONTAINER=0` is that you also need to have the LLVM toolchain and build tools installed with the latter.

Makefile
PiotrSikora: > We already run bazel inside a docker container, so we don't need nested/peer containers.

Yes, in the CI. The value of this is that developers don't need to have all the other build dependencies installed locally.

Makefile
PiotrSikora: @mandarjog yes, this will build release images for you, assuming that you have Bazel(isk) installed already.

PiotrSikora:
Could we get this in, since it's an improvement over the existing builds that always require toolchains to be installed on the host?

PiotrSikora:
Actually, this isn't pulling prowbazel image, but the one from RBE. Let me look into that...

/hold

Makefile
venilnoronha: It would be good to document the flag and its working here via comments.

PiotrSikora:
This was superseded by a different implementation in common-files.

PR #2445, opened 2019-10-09
update envoy-wasm

yxue:
Latest commit: https://github.com/envoyproxy/envoy-wasm/commit/ec03328688895d99c9f5ae4fd7f8459ef3e95212

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -33,7 +33,7 @@ bind(
 )
 
 # Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
-# envoy-wasm commit date: 09/24/2019
+# envoy-wasm commit date: 10/09/2019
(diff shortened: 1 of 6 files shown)


WORKSPACE
```diff
@@ -35,9 +35,9 @@ bind(
 # Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # envoy-wasm commit date: 09/24/2019
```
bianpengyuan: Update the date?

yxue: Good catch! Updated.

PiotrSikora:
Can we wait until https://github.com/envoyproxy/envoy-wasm/pull/229 is merged? Otherwise we're going to update this and istio/istio twice.

cc @jplevyak

yxue:
@PiotrSikora @bianpengyuan could you help to take a look at the PR?

PR #2446, opened 2019-10-09
[1.2] mixer: calculate header size if cache invalidated

lambdai:
Enforce header byte size calculation when header size cache is in validated.
It should not very expensive: at most 2 (mixer filter + report data) * 2 (req + response) * 2(header + trailer) per http transaction

Fix https://github.com/istio/istio/issues/17735


lambdai:
Please hold on
I see ASAN use after free in 1.1 sibling PR: https://github.com/istio/proxy/pull/2447 
Not sure if it is regression.

lambdai:
OK to merge and run the istio/istio test referring to this SHA, but don't release the next 1.2 until I root cause it

lambdai:
It's not a regression.  The test was failing. I am not going to fix in this PR

PR #2447, opened 2019-10-09
[1.1] mixer: calculate header size if cache invalidated

lambdai:
Enforce header byte size calculation when header size cache is in validated.
It should not very expensive: at most 2 (mixer filter + report data) * 2 (req + response) * 2(header + trailer) per http transaction

Fix istio/istio#17735


lambdai:
It's not a regression. The test was failing. Wont fix now. 
Can reproduce by `bazel  test --config=libc++ --config=clang-asan --runs_per_test=100 //test/integration:istio_http_integration_test`, although the failing test might vary

fpesce:
@duderino I think you want to merge that ASAP in order to build the last 1.1

PR #2449, opened 2019-10-09
Sync .bazelrc with upstream.

PiotrSikora:
This file is often skipped when updating Envoy SHA in WORKSPACE.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,8 @@ UNAME := $(shell uname)
 ifeq ($(UNAME),Linux)
 BAZEL_CONFIG_DEV  = --config=libc++
 BAZEL_CONFIG_REL  = --config=libc++ --config=release
-BAZEL_CONFIG_ASAN = --config=clang-asan --config=libc++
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -41,8 +41,8 @@ UNAME := $(shell uname)
 ifeq ($(UNAME),Linux)
 BAZEL_CONFIG_DEV  = --config=libc++
 BAZEL_CONFIG_REL  = --config=libc++ --config=release
-BAZEL_CONFIG_ASAN = --config=clang-asan --config=libc++
-BAZEL_CONFIG_TSAN = --config=clang-tsan --config=libc++
+BAZEL_CONFIG_ASAN = --config=libc++ --config=clang-asan
+BAZEL_CONFIG_TSAN = --config=libc++ --config=clang-tsan
```
PiotrSikora: Previously, this order was needed because we had to overwrite some options that `clang-asan/tsan` config was setting with options that `libc++` was setting. This is no longer necessary.

PiotrSikora:
Rebased due to conflicts in `WORKSPACE`.

PR #2450, opened 2019-10-10
add alpn http filter

yxue:
Add ALPN http filter to overide TLS ALPN for upstream connection.

### DIFF
--- a/src/envoy/http/alpn/alpn_filter.cc
+++ b/src/envoy/http/alpn/alpn_filter.cc
@@ -1,0 +1,48 @@
+/* Copyright 2019 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/envoy/http/alpn/alpn_filter.h"
+
+#include "common/network/application_protocol.h"
+
+namespace Envoy {
+namespace Http {
+namespace Alpn {
+
+AlpnFilterConfig::AlpnFilterConfig(
+    const istio::envoy::config::filter::http::alpn::v2alpha1::FilterConfig
+        &proto_config) {
+  for (const auto &alpn : proto_config.alpn_override()) {
+    alpn_override_.emplace_back(alpn);
+  }
+}
+
+Http::FilterHeadersStatus AlpnFilter::decodeHeaders(Http::HeaderMap &, bool) {
+  const auto &alpn_override = config_->getAlpnOverride();
+  if (!alpn_override.empty()) {
+    ENVOY_LOG(debug, "override with {} ALPNs", alpn_override.size());
+    decoder_callbacks_->streamInfo().filterState().setData(
+        Network::ApplicationProtocols::key(),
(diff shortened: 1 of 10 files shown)


kyessenov:
Can you use the WASM ABI please? This is a trivial filter that _should_ be possible to express in WASM. The problem I foresee is the filter state is statically type and the WASM layer is not aware of it. We use a generic protobuf value to hold all filter state. It is worth discussing that further in Envoy.

yxue:
We are going to extend the trivial filter to select different alpn override according to downstream and upstream protocol. I am not familiar with WASM ABI and not sure if it can be achieved by WASM

src/envoy/http/alpn/alpn_filter.cc
```diff
@@ -0,0 +1,48 @@
...
+namespace Envoy {
+namespace Http {
+namespace Alpn {
+
+AlpnFilterConfig::AlpnFilterConfig(
+    const istio::envoy::config::filter::http::alpn::v2alpha1::FilterConfig
+        &proto_config) {
+  for (const auto &alpn : proto_config.alpn_override()) {
+    alpn_override_.emplace_back(alpn);
+  }
```
lambdai: initialize member outside ctor body
: alpn_override_(proto_config.alpn_override().begin(), alpn_override.alpn_override().end()) {

src/envoy/http/alpn/alpn_filter.h
```diff
@@ -0,0 +1,56 @@
...
+ public:
+  AlpnFilterConfig() = default;
+  explicit AlpnFilterConfig(
+      const istio::envoy::config::filter::http::alpn::v2alpha1::FilterConfig
+          &proto_config);
+
+  const std::vector<std::string> &getAlpnOverride() { return alpn_override_; }
+
+ private:
+  std::vector<std::string> alpn_override_;
```
lambdai: mark vector const: see my comment in constructor

src/envoy/http/alpn/alpn_filter.h
```diff
@@ -0,0 +1,56 @@
...
+namespace Alpn {
+
+class AlpnFilterConfig {
+ public:
+  AlpnFilterConfig() = default;
+  explicit AlpnFilterConfig(
+      const istio::envoy::config::filter::http::alpn::v2alpha1::FilterConfig
+          &proto_config);
+
+  const std::vector<std::string> &getAlpnOverride() { return alpn_override_; }
```
lambdai: const method

src/envoy/http/alpn/alpn_filter.cc
```diff
@@ -0,0 +1,48 @@
...
+
+Http::FilterHeadersStatus AlpnFilter::decodeHeaders(Http::HeaderMap &, bool) {
+  const auto &alpn_override = config_->getAlpnOverride();
+  if (!alpn_override.empty()) {
+    ENVOY_LOG(debug, "override with {} ALPNs", alpn_override.size());
+    decoder_callbacks_->streamInfo().filterState().setData(
+        Network::ApplicationProtocols::key(),
+        std::make_unique<Network::ApplicationProtocols>(alpn_override),
+        Envoy::StreamInfo::FilterState::StateType::ReadOnly);
+  } else {
```
lambdai: IMHO empty vector is also a valid override.
WDYT @PiotrSikora

src/envoy/http/alpn/alpn_filter.h
yxue: Done

src/envoy/http/alpn/alpn_filter.h
yxue: Done

src/envoy/http/alpn/alpn_filter.cc
yxue: Done

src/envoy/http/alpn/alpn_filter.cc
yxue: I cannot find a scenario in which empty override is used.

PR #2451, opened 2019-10-10
Publish release binary with symbols and AddressSanitizer.

PiotrSikora:
Note: libc++ is dynamically linked in this build.

### DIFF
--- a/scripts/release-binary.sh
+++ b/scripts/release-binary.sh
@@ -28,6 +28,12 @@ if [[ "$(uname)" != "Darwin" && "${BAZEL_BUILD_ARGS}" != *"--config=libc++"* ]];
   BAZEL_BUILD_ARGS="${BAZEL_BUILD_ARGS} --config=libc++"
 fi
 
+if [[ "$(uname)" == "Darwin" ]]; then
+  BAZEL_CONFIG_ASAN="--config=macos-asan"
+else
+  BAZEL_CONFIG_ASAN="--config=clang-asan"
+fi
+
 # The bucket name to store proxy binaries.
 DST=""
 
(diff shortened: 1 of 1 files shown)


scripts/release-binary.sh
```diff
@@ -125,6 +131,21 @@ if [ -n "${DST}" ]; then
   gsutil cp "${BINARY_NAME}" "${SHA256_NAME}" "${DST}/"
 fi
 
+# Build the release binary with symbols and AddressSanitizer (ASan).
+# NOTE: libc++ is dynamically linked in this build.
+BINARY_NAME="${HOME}/envoy-asan-${SHA}.tar.gz"
```
kyessenov: nit: don't override bash variables. it's hard enough without aliases.

PiotrSikora: Hm? What do you mean?

kyessenov: The previous block uses the same names for variables. So if you forget to update one of them here, you'll be using a wrong value. This should have been a function or a block to avoid that.

kyessenov: Which is why it's a nit :) We should try to leave the code in better shape than before us.
I think this is not that important. We will need to switch to publishing docker images, which we'll probably use Bazel for everything.

PiotrSikora: Agreed, I looked into publishing Docker in istio/proxy a while ago. The issue is that proxy images include pilot-agent binary, so this repo is not self-contained, and I don't look forward to adding a circular dependency between istio/proxy and istio/istio.

kyessenov: We don't have to do everything at once. Istio/istio can use FROM istio/proxy-docker-image ADD pilot-agent. Eventually, pilot-agent would be a separate thing so it can be moved to be built here.

kyessenov: If you dig into git history far enough, you can find how to package docker image here. It's pretty trivial.

PR #2452, opened 2019-10-10
Use FilterState instead  of  DynamicMetadata  in  Metadata Exchange Filter

gargnupur:
Use FilterState instead  of  DynamicMetadata  in  Metadata Exchange Filter

Also made  the filter  in sync with existing HTTP Metadata Exchange Wasm Plugin,  so that stats(prometheus)  plugin  can use this easily.

### DIFF
--- a/src/envoy/tcp/metadata_exchange/metadata_exchange.cc
+++ b/src/envoy/tcp/metadata_exchange/metadata_exchange.cc
@@ -17,11 +17,14 @@
 #include <string>
 
 #include "absl/base/internal/endian.h"
+#include "absl/strings/str_split.h"
 #include "absl/strings/string_view.h"
 #include "common/buffer/buffer_impl.h"
 #include "common/protobuf/utility.h"
 #include "envoy/network/connection.h"
 #include "envoy/stats/scope.h"
+#include "extensions/common/context.h"
+#include "extensions/common/wasm/wasm_state.h"
 #include "src/envoy/tcp/metadata_exchange/metadata_exchange.h"
 #include "src/envoy/tcp/metadata_exchange/metadata_exchange_initial_header.h"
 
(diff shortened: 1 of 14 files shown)


src/envoy/tcp/metadata_exchange/metadata_exchange.cc
```diff
@@ -224,31 +238,70 @@ void MetadataExchangeFilter::tryReadProxyData(Buffer::Instance& data) {
...
+    // select keys from the metadata using the keys
+    const std::set<std::string> keys =
+        absl::StrSplit(keys_value.string_value(), ',', absl::SkipWhitespace());
+    for (auto key : keys) {
+      const auto entry_it = node_metadata.fields().find(key);
+      if (entry_it == node_metadata.fields().end()) {
+        continue;
+      }
+      (*metadata->mutable_fields())[key] = entry_it->second;
     }
```
gargnupur: @kyessenov : a lot of this code is copied from your http metadata exchange filter.. should we add it later to a common util?

kyessenov: Please do. `extensions/common` seems like the appropriate place for this function.

src/envoy/tcp/metadata_exchange/metadata_exchange.h
```diff
@@ -88,6 +86,16 @@ class MetadataExchangeConfig {
 
 using MetadataExchangeConfigSharedPtr = std::shared_ptr<MetadataExchangeConfig>;
 
+// A simple wrapper around generic values
+class MetadataExchangeState : public StreamInfo::FilterState::Object {
```
kyessenov: I think you need to use WasmState for this to be accessible from within Wasm. The generic resolver is not aware of all filter states, but it knows how to unpack a WasmState.

gargnupur: Thanks.. just created PR to expose WasmState: https://github.com/envoyproxy/envoy-wasm/pull/241

src/envoy/tcp/metadata_exchange/metadata_exchange.cc
gargnupur: Done!

gargnupur:
@kyessenov : Can you please review this?

PR #2454, opened 2019-10-10
Move node_info_cache to common

bianpengyuan:
So that the code could be shared by all filters.


bianpengyuan:
@mandarjog We'd better get this PR in as well, otherwise performance of sd filter would just suffer.

kyessenov:
I'm not fully understanding the execution model of WASM nowadays. Is this cache shared between requests and is that safe to do so?

bianpengyuan:
@kyessenov great point! This cache is shared between requests, but it could only be accessed by single thread and is non-blocking to access. So I think it should be safe. @jplevyak for confirmation, it should be safe to share state in root context between stream contexts, in both null and real wasm environment right?

bianpengyuan:
@jplevyak can I get some advise on the question above? thanks!

bianpengyuan:
@mandarjog @kyessenov Can I get your review on this? Need this to make sd filter performant..

extensions/stackdriver/stackdriver.cc
```diff
@@ -152,31 +163,19 @@ FilterDataStatus StackdriverContext::onResponseBody(size_t body_buffer_length,
...
-  return dynamic_cast<StackdriverRootContext *>(root);
+StackdriverRootContext* StackdriverContext::getRootContext() {
+  RootContext* root = this->root();
+  return dynamic_cast<StackdriverRootContext*>(root);
 }
 
 void StackdriverContext::onLog() {
-  bool isOutbound = getRootContext()->isOutbound();
+  auto* root = getRootContext();
+  bool isOutbound = root->isOutbound();
```
mandarjog: You can follow the stats filter here and move most of this login into the RootContext code.

mandarjog:
@kyessenov  I don't think there any danger of cache being shared inadvertently here.
It is siloed per thread anyway. We really need to get this in. The header processing takes 30% CPU.  https://github.com/istio/istio/issues/18261
cc: @PiotrSikora

bianpengyuan:
@kyessenov @mandarjog node info cache could be turned off now, let's get this in so that we could measure the planned performance optimization w or w/o cache?

extensions/common/node_info_cache.h
```diff
@@ -0,0 +1,60 @@
...
+
+#endif  // NULL_PLUGIN
+
+namespace Wasm {
+namespace Common {
+
+const size_t DefaultNodeCacheMaxSize = 500;
+const wasm::common::NodeInfo EmptyNodeInfo;
+
+typedef std::shared_ptr<wasm::common::NodeInfo> NodeInfoPtr;
```
kyessenov: nit: const NodeInfo

bianpengyuan: Done

extensions/common/node_info_cache.cc
```diff
@@ -0,0 +1,91 @@
...
+  }
+  auto nodeinfo_it = cache_.find(peer_id);
+  if (nodeinfo_it != cache_.end()) {
+    return nodeinfo_it->second;
+  }
+
+  // Do not let the cache grow beyond max_cache_size_.
+  if (int32_t(cache_.size()) > max_cache_size_) {
+    auto it = cache_.begin();
+    cache_.erase(cache_.begin(), std::next(it, max_cache_size_ / 4));
```
kyessenov: Just to confirm my understanding, the cache revocation strategy is to remove 1/4th of the elements at random?

bianpengyuan: Yes, there cache is not sophisticated enough. We should move to lru. @mandarjog wdyt?

mandarjog: Sure lru works. But this is a trivial implementation...

kyessenov: Yeah, let's not over-complicate until we can measure the effects first.
Can you add a test that exercises cache eviction? I don't see it in this PR.

bianpengyuan: Yeah, definitely not in this PR or 1.4 release, just something we could think about.

bianpengyuan:
@kyessenov Any concerns on this change?

PR #2456, opened 2019-10-11
mixer: add hash policy for maglev

kyessenov:


src/envoy/utils/grpc_transport.cc
```diff
@@ -38,12 +38,20 @@ GrpcTransport<RequestType, ResponseType>::GrpcTransport(
...
-          descriptor(), request, *this, parent_span,
-          absl::optional<std::chrono::milliseconds>(kGrpcRequestTimeoutMs))) {
+      on_done_(on_done) {
   ENVOY_LOG(debug, "Sending {} request: {}", descriptor().name(),
             request.DebugString());
+  Envoy::Http::AsyncClient::RequestOptions options;
+  options.setTimeout(kGrpcRequestTimeoutMs);
+  Protobuf::RepeatedPtrField<envoy::api::v2::route::RouteAction::HashPolicy>
+      hash_policy;
+  hash_policy.Add()->mutable_header()->set_header_name(
```
mandarjog: what is this trying to do? 
adding HostHeader as a hash policy? why 2 calls?

kyessenov:
Test failure: 
```
[ RUN      ] IpVersions/JwtVerificationFilterIntegrationTestWithJwks.RSASuccess1/0
[2019-10-11 05:27:27.894][3346][critical][assert] [external/envoy/test/test_common/environment.cc:269] assert failure: false. Details: failed to open: /home/circleci/.cache/bazel/_bazel_circleci/f16e36219ef33c22efc2ad20f3e3775c/sandbox/processwrapper-sandbox/2814/execroot/__main__/bazel-out/k8-fastbuild/bin/src/envoy/http/jwt_auth/http_filter_integration_test.runfiles/envoy/src/envoy/http/jwt_auth/integration_test/envoy.conf.jwk
[2019-10-11 05:27:27.894][3346][critical][backtrace] [bazel-out/k8-fastbuild/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:83] Caught Aborted, suspect faulting address 0xd6a00000d12
```

@lizan Did something change in the test framework?

src/envoy/utils/grpc_transport.cc
kyessenov: This sets up a hash based on combination of x-istio-attributes and :authority as a fallback. Istio attributes carry source uid. Unfortunately, source IP hash policy won't work since downstream address is not set for async client.

kyessenov:
Nevermind, the issue can be fixed by copying a bunch of code from upstream. Added a cache-busting workspace name. Sorry :)

src/envoy/http/mixer/filter.cc
```diff
@@ -58,7 +58,7 @@ void Filter::ReadPerRouteConfig(
 
 FilterHeadersStatus Filter::decodeHeaders(HeaderMap& headers, bool) {
   ENVOY_LOG(debug, "Called Mixer::Filter : {}", __func__);
-  request_total_size_ += headers.byteSize();
+  request_total_size_ += headers.byteSize().value_or(0);
```
PiotrSikora: I think this should be `headers.refreshByteSize()` (see: #2447). cc @lambdai

WORKSPACE
```diff
@@ -35,10 +36,10 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelrc and .bazelversion files.
 #
-# envoy-wasm commit date: 10/09/2019
-ENVOY_SHA = "ec03328688895d99c9f5ae4fd7f8459ef3e95212"
+# envoy-wasm commit date: 10/10/2019
+ENVOY_SHA = "10546ebcef86aa8fc25ff08dcd11196988344afc"
```
PiotrSikora: Could you pull the latest version? It should't require any other changes.

kyessenov: Done

src/envoy/http/mixer/filter.cc
kyessenov: Done.

WORKSPACE
```diff
@@ -36,10 +36,10 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelrc and .bazelversion files.
 #
-# envoy-wasm commit date: 10/10/2019
-ENVOY_SHA = "10546ebcef86aa8fc25ff08dcd11196988344afc"
+# envoy-wasm commit date: 10/11/2019
+ENVOY_SHA = "c152ba0e19f6536bd604da782994536ad73a8164"
```
PiotrSikora: Sorry, could you update again to pull https://github.com/envoyproxy/envoy-wasm/pull/246 (bugfix)?

src/envoy/http/mixer/report_data.h
```diff
@@ -72,10 +72,10 @@ class ReportData : public ::istio::control::http::ReportData,
         response_total_size_(info.bytesSent()),
         request_total_size_(request_total_size) {
     if (response_headers != nullptr) {
-      response_total_size_ += response_headers->byteSize();
+      response_total_size_ += response_headers->refreshByteSize();
```
PiotrSikora: https://github.com/istio/proxy/pull/2447 uses this instead:
```
response_total_size_ += (response_headers->byteSize().has_value()
                             ? response_headers->byteSize().value()
                             : response_headers->byteSizeInternal());
```
not sure what's the reason for using `byteSizeInternal()`, to be honest.

cc @lambdai @asraa

WORKSPACE
```diff
@@ -14,6 +14,7 @@
 #
 ################################################################################
 #
+workspace(name = "proxy")
```
PiotrSikora: Nit: `io_istio_proxy` (Bazel tries to use the reverse domain name notation).

WORKSPACE
kyessenov: Hey, this is not the last proxy change.

kyessenov: Done

WORKSPACE
kyessenov: Done

src/envoy/http/mixer/report_data.h
kyessenov: Ok. I think this is some const mess.

WORKSPACE
PiotrSikora: I know, but this change would pull the bug that I introduced yesterday, and fixed in the aforementioned PR. Thanks!

src/envoy/http/mixer/report_data.h
asraa: Yeah, this is a WIP to clear this mess (see https://github.com/envoyproxy/envoy/issues/8567)
But generally the pattern would be to refresh the byte size before you report data / log data, and then use `byteSize()` inside these functions since they typically take const pointers / refs.

lambdai: yep. refreshByteSize is not allowed on const HeaderMap. How could this compiled?

WORKSPACE
```diff
@@ -14,6 +14,7 @@
 #
 ################################################################################
 #
+workspace(name = "io_istio_proxy")
```
lambdai: Educate me: what is the purpose?

src/envoy/utils/grpc_transport.cc
```diff
@@ -38,12 +38,20 @@ GrpcTransport<RequestType, ResponseType>::GrpcTransport(
...
+      kIstioAttributeHeader.get());
+  hash_policy.Add()->mutable_header()->set_header_name(
+      Envoy::Http::Headers::get().Host.get());
+  options.setHashPolicy(hash_policy);
```
lambdai: options.setHashPolicy(std::move(hash_policy))? Although I am not fully convinced if optimization indeed happens

kyessenov: It's stack allocated and then copied into options. I'd think escape analysis decides to delete it right after this line.

WORKSPACE
kyessenov: Every workspace must be named in bazel. Not sure why this was missing.

PiotrSikora: @lambdai it's to force importers of this workspace to use the same common name, to avoid duplicates.

lambdai: get it!

src/envoy/utils/grpc_transport.cc
lambdai: I agree with the former part and the escape detection. But I don't think compiler would do automatic move even not one is using it.
I will be astonished when compiler make this decision when it's a lock_guard-ish, so as to this.

kyessenov: The argument is taken by reference. The only difference with a temporary is that deletion happens immediately rather than later, right? Either way, not clear if there's a measurable impact on performance.

WORKSPACE
```diff
@@ -35,10 +36,10 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelrc and .bazelversion files.
 #
-# envoy-wasm commit date: 10/09/2019
-ENVOY_SHA = "ec03328688895d99c9f5ae4fd7f8459ef3e95212"
+# envoy-wasm commit date: 10/11/2019
+ENVOY_SHA = "29b71643c999af4e31ba86d41740edbef71c73fe"
```
rshriram: can you use the one from Greg's PR? or is this more recent?

GregHanson: it's more recent

src/envoy/utils/grpc_transport.cc
lambdai: Aha, setHashPolicy is not a protobuf generated method. It doesn't have optimization for HashPolicy&& arg

PR #2458, opened 2019-10-11
Add usage of sd logger and an e2e test

bianpengyuan:


extensions/stackdriver/stackdriver.cc
```diff
@@ -94,27 +103,33 @@ bool StackdriverRootContext::onConfigure(
...
+    logger_->addLogEntry(request_info, peer_node_info);
+  }
 }
 
 inline bool StackdriverRootContext::isOutbound() {
   return direction_ == ::Wasm::Common::TrafficDirection::Outbound;
 }
 
+// TODO(bianpengyuan) Add final export once root context supports onDone.
+// https://github.com/envoyproxy/envoy-wasm/issues/240
```
mandarjog: Can this be added to contextdestructor ?

bianpengyuan: No, wasm runtime assumes callbacks from modules into hosts must be triggered inside a host call into module (e.g. onDone), otherwise it does not have context information about the callback.

extensions/stackdriver/log/exporter.cc
```diff
@@ -56,6 +56,7 @@ ExporterImpl::ExporterImpl(RootContext* root_context,
 
   // Construct grpc_service for the Stackdriver gRPC call.
   GrpcService grpc_service;
+  grpc_service.mutable_google_grpc()->set_stat_prefix("stackdriver_logging");
```
mandarjog: Up on line 52 please add a failure counter one each for metrics, logging etc.

mandarjog: Why are success callbacks and failure call backs lambdas. They are not capturing anything .

extensions/stackdriver/log/exporter.cc
```diff
@@ -56,6 +56,7 @@ ExporterImpl::ExporterImpl(RootContext* root_context,
...
   if (logging_service_endpoint.empty()) {
     grpc_service.mutable_google_grpc()->set_target_uri(
         kGoogleStackdriverLoggingAddress);
```
mandarjog: Can you verify that setting Gooogle_application_credentials env var is honored here?
That is how we will use it on a VM.

bianpengyuan: Yes it does. grpc-cpp library gets it from that env variable. That is how gke gets credentials as well.

extensions/stackdriver/log/exporter.cc
bianpengyuan: I use lambda to save a function definition, and now they are capturing thing now for the metric.

bianpengyuan:
@kyessenov need approval on this one. Thanks!

PR #2460, opened 2019-10-11
initial hermetic docker build rules

kyessenov:
This allegedly works on istio/istio using the same base.
I'm getting local errors though: `/usr/local/bin/envoy: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.27' not found (required by /usr/local/bin/envoy)` which probably is due to my local machine using different glibc.


WORKSPACE
```diff
@@ -71,3 +72,32 @@ envoy_dependencies()
...
+load(
+    "@io_bazel_rules_docker//container:container.bzl",
+    "container_pull",
+)
+
+container_pull(
+    name = "distroless_cc",
+    registry = "gcr.io",
+    repository = "distroless/cc",
+    digest = "sha256:86f16733f25964c40dcd34edf14339ddbb2287af2f7c9dfad88f0366723c00d7"
```
kyessenov: This is latest as of today.

scripts/check-repository.sh
```diff
@@ -25,16 +25,3 @@ if grep -nr "commit =\|remote =" --include=WORKSPACE --include=*.bzl .; then
   echo "To ensure that all dependencies can be stored offline in distdir, only HTTP repositories are allowed."
   exit 1
 fi
-
-# Check whether number of defined `url =` and `sha256 =` kwargs in repository
-# definitions is equal.
-urls_count=$(grep -nr "url =" --include=WORKSPACE --include=*.bzl | wc -l)
```
kyessenov: This check is ridiculously fragile. Removing it.

kyessenov:
@PiotrSikora can you help picking a better base image? I am still not able to create a functional docker image.

src/envoy/BUILD
```diff
@@ -49,3 +50,10 @@ pkg_tar(
     package_dir = "/usr/local/bin/",
     tags = ["manual"],
 )
+
+container_image(
+    name = "istio-proxy",
+    base = "@distroless_cc//image",
+    tags = ["manual"],
+    tars = [":envoy_tar"],
+)
```
PiotrSikora: The build target for `.deb` package is `//tools/deb:istio-proxy`, perhaps we could move this to `//tools/docker:istio-proxy` to be a bit more consistent?

WORKSPACE
```diff
@@ -71,3 +72,32 @@ envoy_dependencies()
...
+container_deps()
+
+load(
+    "@io_bazel_rules_docker//container:container.bzl",
+    "container_pull",
+)
+
+container_pull(
+    name = "distroless_cc",
+    digest = "sha256:86f16733f25964c40dcd34edf14339ddbb2287af2f7c9dfad88f0366723c00d7",
```
PiotrSikora: What's the source for this digest? Is there any commit and/or date that you can add here?

repositories.bzl
```diff
@@ -351,3 +351,12 @@ py_proto_library(
 def mixerapi_dependencies():
     go_x_tools_imports_repositories()
     mixerapi_repositories()
+
+def docker_dependencies():
+    # Download the rules_docker repository at release v0.12.0
```
PiotrSikora: Nit: This comment is a little redundant with the next few lines.

src/envoy/BUILD
kyessenov: Made it //tools/docker:envoy since it's not full istio-proxy image yet (needs the agent).

WORKSPACE
kyessenov: It's the latest docker image. I don't see any release notes, so it's the same behavior as pulling latest image, just reproducible due to the digest.

repositories.bzl
kyessenov: Removed it.

kyessenov:
Thanks, I confirmed that the image works with the docker ASAN build. I'm not sure how the release pipeline works (do we have credentials to push images and where should they be pushed?). Can we fix the publishing step in a follow-up PR?

scripts/release-binary.sh
```diff
@@ -140,6 +143,7 @@ BAZEL_TARGET="${BAZEL_OUT}/src/envoy/envoy_tar.tar.gz"
 cp -f "${BAZEL_TARGET}" "${BINARY_NAME}"
 sha256sum "${BINARY_NAME}" > "${SHA256_NAME}"
 
+
```
PiotrSikora: Nit: extra empty line.

scripts/release-binary.sh
```diff
@@ -117,6 +117,9 @@ if [ -n "${DST}" ]; then
   gsutil cp "${BINARY_NAME}" "${SHA256_NAME}" "${DST}/"
 fi
 
+# Build the docker image with the release binary
+bazel run ${BAZEL_BUILD_ARGS} --config=release //tools/docker:envoy
```
PiotrSikora: If you don't want to publish it in this PR, then maybe comment it out for the time being, since it's going to be built for each merged change, and it seems wasteful if we're not going to publish the resulting artifact.

src/envoy/BUILD
```diff
@@ -48,4 +48,5 @@ pkg_tar(
     mode = "0755",
     package_dir = "/usr/local/bin/",
     tags = ["manual"],
+    visibility = ["//visibility:public"],
```
PiotrSikora: Hm, I was wondering why this wasn't needed for `//tools/deb:istio-proxy`, but it looks that there is a custom `pkg_tar` target there that packages Envoy binary in the same way.

Could you update `//tools/deb:debian-data` to use this instead and remove the old target?

scripts/release-binary.sh
kyessenov: Ok, added a TODO.

src/envoy/BUILD
kyessenov: Sure. Done.

scripts/release-binary.sh
PiotrSikora: Erm, I meant that you should comment out the `bazel run` step, until those images are published.

WORKSPACE
```diff
@@ -72,3 +73,33 @@ envoy_dependencies()
...
+
+container_deps()
+
+load(
+    "@io_bazel_rules_docker//container:container.bzl",
+    "container_pull",
+)
+
+container_pull(
+    name = "distroless_cc",
```
PiotrSikora: Nit: could this be added to `tools/docker/BUILD` instead? or does it have to be in the `WORKSPACE`?

kyessenov: I think this is a workspace level object. Not 100% sure.

PR #2462, opened 2019-10-15
feat(forwarded attributes): add support for ignore_forwarded_attributes

douglas-reid:
Adds code to the http `request_handler_impl` to ignore forwarded attributes when the http client configuration specifies to ignore the attributes.

See: istio/istio#15188, istio/istio#12063

### DIFF
--- a/src/istio/control/http/client_context.h
+++ b/src/istio/control/http/client_context.h
@@ -55,10 +55,6 @@ class ClientContext : public ClientContextBase {
   // Get the service config cache size
   int service_config_cache_size() const { return service_config_cache_size_; }
 
-  const bool IgnoreForwardedAttributes() const {
-    return config_.ignore_forwarded_attributes();
-  }
-
  private:
   // The http client config.
   const ::istio::mixer::v1::config::client::HttpClientConfig& config_;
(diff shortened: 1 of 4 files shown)


src/istio/control/http/client_context.h
```diff
@@ -55,6 +55,10 @@ class ClientContext : public ClientContextBase {
   // Get the service config cache size
   int service_config_cache_size() const { return service_config_cache_size_; }
 
+  const bool IgnoreForwardedAttributes() const {
```
kyessenov: `const bool` is redundant since bool is always const. just use `bool`.

kyessenov: style: use snake case to make it consistent with the accessor above.

kyessenov: TBH. you don't even need this function since config is accessible via `config()`.

src/istio/control/http/service_context.h
```diff
@@ -53,6 +53,10 @@ class ServiceContext {
     return service_config_ && !service_config_->disable_report_calls();
   }
 
+  bool ignore_forwarded_attributes() const {
+    return client_context_->IgnoreForwardedAttributes();
```
kyessenov: Is it sufficient to scope the setting per listener? Do we have a case when one listener servers both gateway and sidecar destinations, so needs to elect to disable forwarding by service config?

PR #2466, opened 2019-10-16
Fix SD filter root context and update e2e test

bianpengyuan:
VM id still necessary as `local_wasms` is only keyed by vm_id:
https://github.com/envoyproxy/envoy-wasm/blob/16a5cdbf450e4d3bbeead962af5c29bc002ce8b7/source/extensions/common/wasm/wasm.cc#L2829

If vm id is not specified, then the first initialized wasm filter (as well aslistener context) win the race. It should be keyed by vm_id + root_id IMHO..

Also update e2e test to make an envoy instance run with two listeners with different directions. I verified that the test failed without this change.


bianpengyuan:
@jplevyak Please take a look at this PR, Thank you!

jplevyak:
I talked to Piotr and we are going to change the default so that if you do not specify a vm_id then you will get a unique VM per plugin.   In order to share VMs you will have to specify a vm_id which match for the two plugins/filters.

bianpengyuan:
@jplevyak Ok, thanks for the heads up!

PR #2467, opened 2019-10-17
Values stored in context is incorrect under high load

LSChyi:
Sidecar is not sending check request under hight traffic load.
This is just a code change that prove/as workaround that the values stored in context is incorrect, need more investigation to fix it.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #17967

**Release note**:

```release-note
```


LSChyi:
@googlebot I signed it!

PR #2468, opened 2019-10-17
feat(edges): enable edge reporting in stackdriver extension

douglas-reid:
This PR hooks in the edges reporting library into the full stackdriver extension. This will be disabled by default, but config params are added to support testing and transition to on-by-default.

This is a follow-on to https://github.com/istio/proxy/pull/2439.

### DIFF
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -59,7 +59,6 @@ constexpr char kStackdriverExporter[] = "stackdriver_exporter";
 constexpr char kExporterRegistered[] = "registered";
 constexpr int kDefaultLogExportMilliseconds = 10000;                      // 10s
 constexpr long int kDefaultEdgeReportDurationNanoseconds = 600000000000;  // 10m
-constexpr int kNanosPerMillis = 1000000;
 
 bool StackdriverRootContext::onConfigure(
     std::unique_ptr<WasmData> configuration) {
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -13,14 +13,17 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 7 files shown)


extensions/stackdriver/stackdriver.cc
```diff
@@ -120,6 +134,18 @@ void StackdriverRootContext::record(const RequestInfo &request_info,
   if (enableServerAccessLog()) {
     logger_->addLogEntry(request_info, peer_node_info);
   }
+  if (enableEdgeReporting()) {
```
bianpengyuan: Is reporting every 10s too frequent for edge? Will this cause too much computation for the backend?

If we want to make it less frequent, there is only one timer per WASM vm right now, so we need to keep an edge reporting counter `n` to trigger edge reporting every `10 * n` seconds.

douglas-reid: every 10s is too much. took a stab at a configurable duration.

extensions/stackdriver/stackdriver.cc
```diff
@@ -51,7 +57,9 @@ using ::Wasm::Common::RequestInfo;
 
 constexpr char kStackdriverExporter[] = "stackdriver_exporter";
 constexpr char kExporterRegistered[] = "registered";
-constexpr int kDefaultLogExportMilliseconds = 10000;  // 10s
+constexpr int kDefaultLogExportMilliseconds = 10000;                      // 10s
+constexpr long int kDefaultEdgeReportDurationNanoseconds = 600000000000;  // 10m
+constexpr int kNanosPerMillis = 1000000;
```
bianpengyuan: Looks like this line not needed.

bianpengyuan:
@douglas-reid Please pull in master to fix the macos test, thansk!

douglas-reid:
rebased and forced push.

PR #2469, opened 2019-10-17
Run go tests with sanitizers.

PiotrSikora:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,14 +61,19 @@ clean:
 	@bazel clean
 
 test:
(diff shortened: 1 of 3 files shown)


bianpengyuan:
Let's just revert the stackdriver test to what it was before https://github.com/istio/proxy/commit/08cec447dfc349b314e226dc66132dbdb5d08fc9 as null vm does not really create a VM based on vm_id, which causes race between filters if they have different configs.

kyessenov:
I can fix it up in a separate PR as well.

PiotrSikora:
@bianpengyuan @kyessenov yeah, I think fixing it in a separate PR is fine.

PR #2473, opened 2019-10-18
remove config time build metric

mandarjog:
Removing build_metric until it can be properly added back from the sandbox.

fixes https://github.com/envoyproxy/envoy-wasm/issues/255#issuecomment-543237444

Sandbox does not support metric recording during config change.

I have tested this change and it *does not* crash.


bianpengyuan:
Please merge master to fix the macos test.

douglas-reid:
Is there a bug tracking the re-add of this feature?

douglas-reid:
Can we roll this back now (and expose the metric)?

There were several fixes to the wasm branch related to the root cause?

bianpengyuan:
Yes, let's add it back. The underlying issue should have been fixed now and we have test that covers config change.

PR #2474, opened 2019-10-18
Fix building extensions as WebAssembly modules.

PiotrSikora:

### DIFF
--- a/extensions/stats/plugin.cc
+++ b/extensions/stats/plugin.cc
@@ -36,6 +36,13 @@ namespace Plugin {
 
 namespace Stats {
 
+// TODO(PiotrSikora): move this back to Proxy-WASM SDK.
+enum TrafficDirection : int {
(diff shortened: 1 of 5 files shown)


PiotrSikora:
Ping.

extensions/stats/plugin.cc
```diff
@@ -36,6 +36,13 @@ namespace Plugin {
 
 namespace Stats {
 
+// TODO(PiotrSikora): move this back to Proxy-WASM SDK.
```
kyessenov: This got moved already. Care to update it to use Wasm::Common::TrafficDirection?

PiotrSikora: Done.

PiotrSikora:
Yeah, and we need a presubmit as well, to catch build breaks... but I'm not sure if we can do docker-inside-docker for the `./build_wasm.sh` calls.

PR #2475, opened 2019-10-18
Retire Linux jobs for istio/proxy on CircleCI.

PiotrSikora:
Part of istio/istio#16030.


bianpengyuan:
Do we have coverage of `linux_release` in prow?

PiotrSikora:
@bianpengyuan we do... in postsubmits. I've added tests for `test_release` in presubmits when Prow was completely broken, but I don't think that we need it anymore, and it takes forever to run.

PR #2478, opened 2019-10-18
Update Clang and libc++ to version 9.0.

PiotrSikora:


PiotrSikora:
Could we get this in? Everything else is now broken, since Prow images were already updated...

PR #2479, opened 2019-10-21
use different alpn override for different upstream protocol

yxue:

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/alpn/alpn_filter.h
+++ b/src/envoy/http/alpn/alpn_filter.h
@@ -22,19 +22,29 @@ namespace Envoy {
 namespace Http {
 namespace Alpn {
 
+using AlpnOverride =
+    absl::flat_hash_map<Http::Protocol, std::vector<std::string>>;
+
 class AlpnFilterConfig {
  public:
-  AlpnFilterConfig() = default;
-  explicit AlpnFilterConfig(
+  AlpnFilterConfig(
       const istio::envoy::config::filter::http::alpn::v2alpha1::FilterConfig
-          &proto_config);
+          &proto_config,
+      Upstream::ClusterManager &cluster_manager);
 
-  const std::vector<std::string> &getAlpnOverride() const {
-    return alpn_override_;
+  Upstream::ClusterManager &clusterManager() { return cluster_manager_; }
+
+  const std::vector<std::string> &alpnOverride(const Http::Protocol &protocol) {
+    return alpn_override_[protocol];
   }
 
  private:
-  const std::vector<std::string> alpn_override_;
+  Http::Protocol getHttpProtocol(
+      const istio::envoy::config::filter::http::alpn::v2alpha1::FilterConfig::
+          Protocol &protocol);
+
+  AlpnOverride alpn_override_;
+  Upstream::ClusterManager &cluster_manager_;
 };
 
 using AlpnFilterConfigSharedPtr = std::shared_ptr<AlpnFilterConfig>;
--- a/src/envoy/http/alpn/alpn_filter.h
+++ b/src/envoy/http/alpn/alpn_filter.h
@@ -22,7 +22,7 @@ namespace Envoy {
 namespace Http {
 namespace Alpn {
 
-using AlpnOverride =
+using AlpnOverrides =
     absl::flat_hash_map<Http::Protocol, std::vector<std::string>>;
 
 class AlpnFilterConfig {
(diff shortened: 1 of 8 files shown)


src/envoy/http/alpn/alpn_filter.h
```diff
@@ -22,19 +22,29 @@ namespace Envoy {
...
-          &proto_config);
+          &proto_config,
+      Upstream::ClusterManager &cluster_manager);
+
+  Upstream::ClusterManager &clusterManager() { return cluster_manager_; }
 
-  const std::vector<std::string> &getAlpnOverride() const {
-    return alpn_override_;
+  const std::vector<std::string> &alpnOverride(const Http::Protocol &protocol) {
+    return alpn_override_[protocol];
```
lambdai: This is an obvious read only api
Please don't change `alpn_override_`. 
Please 
1. Mark alpnOverride() const
2. return `const std::vector<std::string> * const` since we cannot return reference (alternatively return a default empty string vector else where)

src/envoy/http/alpn/alpn_filter.h
```diff
@@ -22,19 +22,29 @@ namespace Envoy {
...
   }
 
  private:
-  const std::vector<std::string> alpn_override_;
+  Http::Protocol getHttpProtocol(
+      const istio::envoy::config::filter::http::alpn::v2alpha1::FilterConfig::
+          Protocol &protocol);
+
+  AlpnOverride alpn_override_;
```
lambdai: AlpnOverrides alpn_overrides_;

src/envoy/http/alpn/alpn_test.cc
```diff
@@ -31,55 +34,139 @@ namespace {
...
+
+    auto config =
+        std::make_shared<AlpnFilterConfig>(proto_config, cluster_manager_);
     auto filter = std::make_unique<AlpnFilter>(config);
     filter->setDecoderFilterCallbacks(callbacks_);
     return filter;
   }
 
  protected:
+  FilterConfig::Protocol getProtocol(Http::Protocol protocol) {
```
lambdai: Could you elaborate? I don't really understand why you don't use Http::Protocol? It's 1:1 mapping and I see the conversion back and forth.

src/envoy/http/alpn/alpn_test.cc
```diff
@@ -31,55 +34,139 @@ namespace {
...
 
-TEST_F(AlpnFilterTest, NoAlpnOverride) {
+TEST_F(AlpnFilterTest, OverrideAlpnUseDownstreamProtocol) {
   NiceMock<StreamInfo::MockStreamInfo> stream_info;
   ON_CALL(callbacks_, streamInfo()).WillByDefault(ReturnRef(stream_info));
-  auto filter = makeDefaultFilter();
-  EXPECT_CALL(stream_info, filterState()).Times(0);
-  EXPECT_EQ(filter->decodeHeaders(headers_, false),
-            Http::FilterHeadersStatus::Continue);
+  AlpnOverride alpn = {{Http::Protocol::Http10, {"foo", "bar"}},
```
lambdai: nits since it's in tests: const

src/envoy/http/alpn/alpn_test.cc
```diff
@@ -31,55 +34,139 @@ namespace {
...
+    EXPECT_EQ(filter->decodeHeaders(headers_, false),
+              Http::FilterHeadersStatus::Continue);
+    EXPECT_TRUE(filter_state.hasData<Network::ApplicationProtocols>(
+        Network::ApplicationProtocols::key()));
+    auto alpn_override = filter_state
+                             .getDataReadOnly<Network::ApplicationProtocols>(
+                                 Network::ApplicationProtocols::key())
+                             .value();
+
+    EXPECT_EQ(alpn_override, alpn[p]);
```
lambdai: let's split it into two expectation:
 hasValue() and value EQ

src/envoy/http/alpn/alpn_test.cc
```diff
@@ -31,55 +34,139 @@ namespace {
...
+    EXPECT_EQ(filter->decodeHeaders(headers_, false),
+              Http::FilterHeadersStatus::Continue);
+    EXPECT_TRUE(filter_state.hasData<Network::ApplicationProtocols>(
+        Network::ApplicationProtocols::key()));
+    auto alpn_override = filter_state
+                             .getDataReadOnly<Network::ApplicationProtocols>(
+                                 Network::ApplicationProtocols::key())
+                             .value();
+
+    EXPECT_EQ(alpn_override, alpn[Http::Protocol::Http2]);
```
lambdai: ditto

src/envoy/http/alpn/alpn_test.cc
yxue: I defined and used a set of protocols in the proto file.

src/envoy/http/alpn/alpn_test.cc
yxue: I checked `hasData` above `getDataReadOnly`

PR #2480, opened 2019-10-22
ci: add docker push postsubmit

kyessenov:
Enable docker-less push for all bazel release configs.
Bazel does not need docker to pull/push images so this should work in theory assuming credentials are set-up correctly.


tools/docker/BUILD
```diff
@@ -15,10 +15,23 @@
...
+)
 
 container_image(
     name = "envoy",
     base = "@distroless_cc//image",
     tags = ["manual"],
     tars = ["//src/envoy:envoy_tar"],
 )
+
+container_push(
```
mandarjog: We also need to build ubutu image along with distroless.

kyessenov: What's wrong with distroless? We already statically link as much as we can.

mandarjog: Nothing wrong with distroless, but we do publish ubuntu images.
I would like to publish images from here, so that istio/istio will only add statically linked binaries like "pilot-agent" and some other config.

We should add a test of running the binary from within the container, so as to not publish unrunnable images.

howardjohn: If we want to build istio/istio image `FROM` this one it should be based on docker.io/istio/base to be consistent with other images

I am worried doing this may be pretty complicated though, in terms of making sure to use the right base image version, updating the dependency, etc. a lot of our tests rely on the `envoy` binary as well for example. What is the benefit of building the images here, and have we considered what istio/istio will look like if we do this?

kyessenov: distroless is basically debian, so it's very close to ubuntu without all the garbage.
Agree about the tests, but that would require using docker to run images, which is something we want to avoid.

@howardjohn The goal is to avoid relying on istio to get working proxy images.

howardjohn: I think having docker images built here is a good idea, just not 100% sure it's a good idea for istio/istio to consume them. Maybe it is, just needs some extra thought.

If we don't plan to directly build `FROM` these images them distroless seems fine to me?

kyessenov: @mandarjog Added bionic-based build. This is very fast actually, so we can add as many as we want.

kyessenov: @howardjohn Proxy should be the source of truth for all proxy build artifacts. Istio should stack any additions (pilot-agent, iptables) on top of it since Istio can produce statically linked go binaries with minimal system requirements. Gradually, as the agent becomes smaller, the proxy can become the complete source of truth for proxy images.

prow/proxy-postsubmit.sh
```diff
@@ -45,3 +45,4 @@ export BAZEL_BUILD_ARGS="--local_ram_resources=12288 --local_cpu_resources=8 --v
 echo 'Create and push artifacts'
 make push_release RELEASE_GCS_PATH="gs://istio-build/proxy"
 make artifacts ARTIFACTS_GCS_PATH="gs://istio-artifacts/proxy/${GIT_SHA}/artifacts/debs"
+./scripts/release-docker.sh
```
PiotrSikora: Could you pull this back into `release-binary.sh`? Building it separately is going to trash Bazel's cache.

Also, GKE uses "make push_release" target to publish binaries, and since we expose all well-defined targets in the `Makefile`, scripts shouldn't be accessed directly. This alone could be fixed by adding this to `Makefile`, but cache trashing is the bigger concern...

tools/docker/BUILD
```diff
@@ -15,10 +15,39 @@
...
+    repository = "istio-testing/envoy",
+    tag = "{BUILD_CONFIG}-{BUILD_SCM_REVISION}",
+)
+
+container_push(
+    name = "push_envoy_bionic",
+    format = "Docker",
+    image = ":envoy_bionic",
+    registry = "gcr.io",
+    repository = "istio-testing/envoy_bionic",
```
PiotrSikora: Is there a way to not hardcode target repository? We need to make this configurable for GKE builds.

kyessenov: Yes, we can add build variables. Let me see how to generalize this.

prow/proxy-postsubmit.sh
kyessenov: Refactored the script since it's becoming unmaintainable. Please take a look at it now.

tools/docker/BUILD
mandarjog: I think we do need to build proxyv2 image `FROM` this. It's proxy repos responsibility to ensure that the container and the built image are compatible with each other.

proxyv2 build can add pilot-agent and whatever else on top of this.

scripts/release-binary.sh
```diff
@@ -89,97 +85,73 @@ if [ -n "${DST}" ]; then
...
+    echo "Copying ${BINARY_NAME} ${SHA256_NAME} to ${DST}/"
+    gsutil cp "${BINARY_NAME}" "${SHA256_NAME}" "${DST}/"
+  fi
+
+  echo "Building ${config} docker image"
+  bazel run ${BAZEL_BUILD_ARGS} ${CONFIG_PARAMS} \
+    //tools/docker:push_envoy \
+    //tools/docker:push_envoy_bionic
+
+  if [ -n "${PACKAGE_BASE_NAME}"]; then
```
PiotrSikora: Could you add a log for this case as well?

WORKSPACE
```diff
@@ -102,4 +102,13 @@ container_pull(
     repository = "distroless/cc",
 )
 
+container_pull(
+    name = "bionic",
```
PiotrSikora: Could we avoid referring to a specific version in the container name? This should be "ubuntu", IMHO.

scripts/release-binary.sh
```diff
@@ -89,97 +85,73 @@ if [ -n "${DST}" ]; then
...
+  if [ -n "${DST}" ]; then
+    # Copy it to the bucket.
+    echo "Copying ${BINARY_NAME} ${SHA256_NAME} to ${DST}/"
+    gsutil cp "${BINARY_NAME}" "${SHA256_NAME}" "${DST}/"
+  fi
+
+  echo "Building ${config} docker image"
+  bazel run ${BAZEL_BUILD_ARGS} ${CONFIG_PARAMS} \
+    //tools/docker:push_envoy \
+    //tools/docker:push_envoy_bionic
```
PiotrSikora: ASan builds are dynamically linked against libc++-9, so it needs to be installed in the container.

WORKSPACE
kyessenov: Ok.

scripts/release-binary.sh
kyessenov: Done.

scripts/release-binary.sh
kyessenov: Skipping ASan docker image. Where do we find a good base with libc++-9?

kyessenov: Found your https://github.com/GoogleContainerTools/distroless/issues/394

kyessenov:
@PiotrSikora let's give this a try? It's impossible to tell 100% this works until we test run on the post-submit machines.

PiotrSikora:
Except that we can, see: istio/test-infra#1993.

scripts/release-binary.sh
```diff
@@ -89,93 +85,78 @@ if [ -n "${DST}" ]; then
...
+  cp -f "${BAZEL_TARGET}" "${BINARY_NAME}"
+  sha256sum "${BINARY_NAME}" > "${SHA256_NAME}"
+
+  if [ -n "${DST}" ]; then
+    # Copy it to the bucket.
+    echo "Copying ${BINARY_NAME} ${SHA256_NAME} to ${DST}/"
+    gsutil cp "${BINARY_NAME}" "${SHA256_NAME}" "${DST}/"
+  fi
+
+  if [ -n "${PUSH_DOCKER_IMAGE}"]; then
```
PiotrSikora: This should have separate build & push steps, otherwise it's breaking existing `make test_release` target.

kyessenov: There is no need to have a separate test release beyond regular bazel build here. I can guard with DST since that seems to be the pattern.

PiotrSikora: Well, we want to verify that you can build the Docker image in `test_release`.

kyessenov: it's just tar on top a fetched layer? docker is just a collection of tar layers.

PiotrSikora: It is, but stuff breaks all the time, and there is no harm in testing it before merging and potentially breaking release.

kyessenov: Ok, added build command. We can't really test docker images without running docker, which we don't.

scripts/release-binary.sh
```diff
@@ -89,93 +85,86 @@ if [ -n "${DST}" ]; then
...
+  cp -f "${BAZEL_TARGET}" "${BINARY_NAME}"
+  sha256sum "${BINARY_NAME}" > "${SHA256_NAME}"
+
+  if [ -n "${DST}" ]; then
+    # Copy it to the bucket.
+    echo "Copying ${BINARY_NAME} ${SHA256_NAME} to ${DST}/"
+    gsutil cp "${BINARY_NAME}" "${SHA256_NAME}" "${DST}/"
+  fi
+
+
```
PiotrSikora: Nit: extra empty line.

scripts/release-binary.sh
```diff
@@ -89,93 +85,86 @@ if [ -n "${DST}" ]; then
...
+  echo "Building ${config} docker image"
+  bazel build ${BAZEL_BUILD_ARGS} ${CONFIG_PARAMS} \
+    //tools/docker:envoy_distroless \
+    //tools/docker:envoy_ubuntu
+
+  if [ -n "${DST}" ]; then
+    if [ -n "${PUSH_DOCKER_IMAGE}"]; then
```
PiotrSikora: Nit: `if [ -n "${DST}" -a -n "${PUSH_DOCKER_IMAGE}"]; then`.

scripts/release-binary.sh
```diff
@@ -89,93 +85,86 @@ if [ -n "${DST}" ]; then
...
+      bazel run ${BAZEL_BUILD_ARGS} ${CONFIG_PARAMS} \
```
PiotrSikora: Nit: please add `echo "Pushing ${config} docker image"

PiotrSikora:
You need to merge `master` to pick-up #2491.

kyessenov:
/test proxy-presubmit-release

I believe tide auto rebases.

tools/docker/BUILD
```diff
@@ -15,10 +15,39 @@
...
+    tars = ["//src/envoy:envoy_tar"],
+)
+
+container_push(
+    name = "push_envoy_distroless",
+    format = "Docker",
+    image = ":envoy_distroless",
+    registry = "gcr.io",
+    repository = "{DOCKER_REPOSITORY}",
+    tag = "{BUILD_CONFIG}-{BUILD_SCM_REVISION}",
```
PiotrSikora: Nit: don't we want this to be "dockerless-{BUILD_CONFIG}-{BUILD_SCM_REVISION}"?

WORKSPACE
PiotrSikora: I thought this was renamed already? Or is the name enforced in the upstream repo, somehow?

tools/docker/BUILD
kyessenov: distroless seems to be the future default, so I thought I'd save on naming.

WORKSPACE
kyessenov: It's not, but this is just an internal name. It's better than "ubuntu" since we can pull other ubuntus if necessary.

tools/docker/BUILD
PiotrSikora: K

PR #2486, opened 2019-10-24
fix(edges reporting): fix error in reporting logic

douglas-reid:

Fixes the faulty reporting logic (initialize nanos field) and adds an integration test.


bianpengyuan:
I removed endpoints from the config and use bootstrap to pass in endpoint in test:
https://github.com/istio/proxy/blob/de7a821bcb3f22e8628c994c8ea09fb25d807cb3/test/envoye2e/stackdriver_plugin/stackdriver_plugin_test.go#L100-L101

Context graph endpoint could be passed in similarly:
https://github.com/istio/proxy/blob/de7a821bcb3f22e8628c994c8ea09fb25d807cb3/extensions/stackdriver/common/constants.h#L75

bianpengyuan:
Looks like a real test failure.

douglas-reid:
@bianpengyuan yeah, it looks like #2494 added a mesh id. should be fixed now.

PR #2487, opened 2019-10-24
Make onconfigure work with multiple calls

bianpengyuan:
Also removes endpoint from filter config. Make test instead fetch it from bootstrap config.


kyessenov:
LG. Does this deal with root context deletion with pending gRPC calls?

bianpengyuan:
No it does not. I tried to add some clean up call in root context deconstructor before, but it does not work because that host lacks context information to act on the call from module. I think it assumes that it should always be host that initiates the call. We need to have something like onLameduck to make it work.

bianpengyuan:
@jplevyak @mandarjog Ptal, thanks!

extensions/stackdriver/stackdriver.cc
```diff
@@ -85,15 +122,22 @@ bool StackdriverRootContext::onConfigure(
     logWarn("Unable to get plugin direction");
   }
 
-  auto exporter =
-      std::make_unique<ExporterImpl>(this, config_.test_logging_endpoint());
-  // logger takes ownership of exporter.
-  logger_ = std::make_unique<Logger>(local_node_info_, std::move(exporter));
+  if (!logger_) {
```
mandarjog: Since we removed all variability from xds config and shoved it into bootstrap, this will work.
But we will have filters that need xds config and in that case this code needs to ensure that no only are things initialized but initialized with the same params. I think the fix is ok right now.

bianpengyuan: `logger` is expensive and unnecessary to update, so we remove the variability here. But say if `logger` needs to be updated based on config, I think should we make callbacks safe by avoiding referencing the object that are up to destruction due to config update.

bianpengyuan: Using `weak_ptr` is also a solution to this.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -20,7 +20,7 @@ package stackdriver.config.v1alpha1;
 import "google/protobuf/duration.proto";
 
 message PluginConfig {
-  // next id: 8
+  // next id: 5
```
yxue: Do we need to reserve these deprecated fields?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -20,7 +20,7 @@ package stackdriver.config.v1alpha1;
 import "google/protobuf/duration.proto";
 
 message PluginConfig {
-  // next id: 8
```
yxue: Do we need to reserve the deprecated fields?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
bianpengyuan: It's fine, this is no user that actually starts to use this proto yet.

PiotrSikora:
You need to merge `master` to pick-up #2491.

bianpengyuan:
Yeah, I guess prow will do that automatically. The job is running now.

PR #2488, opened 2019-10-24
Update common files.

geeknoid:


Makefile
```diff
@@ -67,7 +68,20 @@ $(info Building with the build container: $(IMG).)
 # the path of the file.
 TIMEZONE=`readlink $(READLINK_FLAGS) /etc/localtime | sed -e 's/^.*zoneinfo\///'`
 
-RUN = $(CONTAINER_CLI) run -t -i --sig-proxy=true -u $(UID):docker --rm \
+# Determine the docker.push credential bind mounts.
```
kyessenov: We don't need docker to do docker push. See #2480

geeknoid: This is the generic common files used across all repos. This particular path is not exercised in this repo.

kyessenov: It might be worth looking to generalize this. I was surprised how much faster it was to build and push images with bazel. The ridiculous docker "one-daemon-fits-all" model is terrible in CI.
Alternatively, use anything but docker, e.g. https://github.com/uber/makisu

PR #2492, opened 2019-10-26
Avoid capturing logger pointer in callback lambda

bianpengyuan:
Also add mesh_uid into metric and log labels.


extensions/stackdriver/log/exporter.cc
```diff
@@ -42,19 +42,23 @@ namespace Stackdriver {
...
-  success_callback_ = [this](google::protobuf::Empty&&) {
-    export_call_.increment(1, "logging", true);
+  Metric export_call(MetricType::Counter, "stackdriver_filter",
+                     {MetricTag{"type", MetricTag::TagType::String},
+                      MetricTag{"success", MetricTag::TagType::Bool}});
+  auto success_counter = export_call.resolve("logging", true);
+  auto failure_counter = export_call.resolve("logging", false);
+  success_callback_ = [success_counter](google::protobuf::Empty&&) {
+    // TODO(bianpengyuan): replace this with envoy's generic gRPC counter.
+    incrementMetric(success_counter, 1);
```
mandarjog: @jplevyak this was our solution to simplify the callback by not capturing references.
This lambda just captures the returned `int` and will use the C API to call into the host.
What is the behaviour if `incrementMetric` is called with an invalid id ?

extensions/stackdriver/log/logger_test.cc
```diff
@@ -98,6 +99,7 @@ google::logging::v2::WriteLogEntriesRequest expectedRequest(
   (*top_label_map)["destination_name"] = node_info.name();
   (*top_label_map)["destination_workload"] = node_info.workload_name();
   (*top_label_map)["destination_namespace"] = node_info.namespace_();
+  (*top_label_map)["mesh_uid"] = node_info.mesh_id();
```
kyessenov: This MESH_ID change should be a separate (small) PR.

extensions/stackdriver/log/exporter.cc
```diff
@@ -42,19 +42,23 @@ namespace Stackdriver {
...
     logDebug("successfully sent Stackdriver logging request");
   };
 
-  failure_callback_ = [this](GrpcStatus status, StringView message) {
+  failure_callback_ = [failure_counter](GrpcStatus status, StringView message) {
     // TODO(bianpengyuan): add retry.
-    export_call_.increment(1, "logging", false);
```
kyessenov: (cc @jplevyak) I think we need a regression test for this, but I don't see a simple way to simulate this in envoy test framework. Let's file an issue to create a test case for reproducing the failure to confirm this fix works.

extensions/stackdriver/log/exporter.cc
jplevyak: The wasm container stores a map from int -> metric and does not free any metrics during the lifetime of the VM, so a valid int for a metric will remain so as long as the VM lives.

extensions/stackdriver/log/exporter.cc
jplevyak: What are you trying to simulate?  This failure was caused by two onConfigure calls coming in.  That can be simulated with the existing test framworks although it is only meaningful in the context of a particular plugin which many or may not handle it correctly.

bianpengyuan: Done. https://github.com/istio/istio/issues/18396

extensions/stackdriver/log/logger_test.cc
bianpengyuan: Done. https://github.com/istio/proxy/pull/2494

mandarjog:
@kyessenov PTAL.

@bianpengyuan asan build is failing.

bianpengyuan:
@mandarjog Hmm, it does not look related to my change. Let me try to rerun.

PR #2496, opened 2019-10-28
Make envoy archive configurable from environment.

clarketm:

Introduces a new `envoy_repository_rule` bazel rule. It is a `repository_rule` with attributes that can be configured from the environment (e.g. `ENVOY_REPOSITORY`, `ENVOY_SHA`, `ENVOY_SHA256`, `ENVOY_PREFIX`). If the aforementioned env variables are **not** specified then the locally defined values are used (same as before this change).

### DIFF
--- a/envoy_repository_rule.bzl
+++ b/envoy_repository_rule.bzl
@@ -1,0 +1,42 @@
+# Copyright 2019 Istio Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+def _envoy_repository_rule_impl(ctx):
+    """Core implementation of envoy_repository_rule."""
+
+    # Set repo, sha, sha256, and prefix from environment (if defined) or supplied attribute.
+    _repo = ctx.os.environ.get("ENVOY_REPOSITORY", default = ctx.attr.repository)
+    _sha = ctx.os.environ.get("ENVOY_SHA", default = ctx.attr.sha)
+    _sha256 = ctx.os.environ.get("ENVOY_SHA256", default = ctx.attr.sha256)
+    _prefix = ctx.os.environ.get("ENVOY_PREFIX", default = ctx.attr.prefix)
+
+    # Download and extract archive.
+    ctx.download_and_extract(
+        sha256 = _sha256,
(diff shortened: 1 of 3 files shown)


kyessenov:
My change was to add build variables for stamping. This is something different I think.

clarketm:
@PiotrSikora / @kyessenov 
The idea here is to allow the `sha` and `url` components of `envoy` repo to be configurable from the environment falling back to hard-coding values. So we can reuse this for private builds without needing to diverge commits. This would allow CI to, for example, use `--repo_env` or `--action_env` to override these values. I am not sure if this is the ideal approach. If it is not, then I am open to suggestions to make this more robust.

kyessenov:
I think this is fine. I prefer it over sed-ding bazel files to replace SHAs.

envoy_repository_rule.bzl
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+def _envoy_repository_rule_impl(ctx):
+    """Core implementation of envoy_repository_rule."""
+
+    # Set repo, sha, sha256, and prefix from environment (if defined) or supplied attribute.
+    _repo = ctx.os.environ.get("ENVOY_REPOSITORY", default = ctx.attr.repository)
```
PiotrSikora: Don't you need to add those variables to `.bazelrc` as well? I believe that we restrict values that we pass from the environment.

clarketm: The variables can be provided either in the `.bazelrc`, at `bazel` command execution or presumably using `BAZEL_BUILD_ARGS`. I do not have a concise set of pros/cons for each approach ATM.

PiotrSikora: I guess my point was that if we add them to `.bazelrc` then those variables will be automatically picked from the environment.

Otherwise, yes, we can explicitly pass them to Bazel using `--action_env=`.

clarketm: I like the idea of utilizing `.bazelrc`. It appears to be the most straightforward of the various approaches. Let me add the update.

lizan:
Ugh this breaks things because the repository_rule introduced here is much less than http_archive does. Especially it doesn’t play well with private repos and/or under GH rate limit because this one doesn’t respect netrc.

lizan:
I don’t think we should do more or less than vanilla http_archive here, keeping the repository rule sync to bazel adds another piece to manage even it looks very simple. Anything beyond that should be done via overridde_repository.

clarketm:
@lizan 
The goal here was not to reduce functionality but rather to enhance configurability so that the attributes could be specified in an external build system. Is there a method to leverage the default `http_archive` implementation (or an alternative with feature parity) while making it configurable; I am not familiar with `--override_repository` is it set for a specific rule?   

In the meantime, should I revert this change?

lizan:
@clarketm Yes I understand your intention and thanks to tackle on it. You can load http_archive from @bazel_tools//tools/build_defs/repo:http.bzl but I'm not sure whether it will work because the impl is not public.

`--override_repository` is a command line option to override any workspace repository to a local directory, so basically you can do anything there (such as checkout, patch, etc) before invoking bazel.

PR #2497, opened 2019-10-28
wasm: use raw bytes for filter state, add custom struct parser

kyessenov:
Benchmark result:
```
2019-10-28 16:06:41
Running bazel-bin/extensions/common/context_speed_test
Run on (32 X 2200 MHz CPU s)
CPU Caches:
  L1 Data 32K (x16)
  L1 Instruction 32K (x16)
  L2 Unified 256K (x16)
  L3 Unified 56320K (x1)
Load Average: 0.05, 0.29, 2.10
-----------------------------------------------------------------
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
BM_GenericStructParser      37498 ns        37495 ns        18653
BM_CustomStructParser        5075 ns         5075 ns       139243
BM_MessageParser             1623 ns         1623 ns       430605
```

Seems like the custom struct parser is preferred, since it's not far off from protobuf parser (that is keep struct on the wire and ABI, and parse as message inside plugins).

### DIFF
--- a/extensions/common/BUILD
+++ b/extensions/common/BUILD
@@ -19,6 +19,7 @@ licenses(["notice"])
 
 load(
     "@envoy//bazel:envoy_build_system.bzl",
+    "envoy_cc_binary",
     "envoy_cc_library",
     "envoy_cc_test",
     "envoy_package",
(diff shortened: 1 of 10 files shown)


extensions/common/BUILD
```diff
@@ -64,3 +65,16 @@ envoy_cc_test(
         "@envoy//source/extensions/common/wasm:wasm_lib",
     ],
 )
+
+envoy_cc_binary(
+    name = "context_speed_test",
+    srcs = ["context_speed_test.cc"],
+    external_deps = [
```
PiotrSikora: Nit: this should have `testonly = 1,`

PR #2502, opened 2019-10-31
Use destination service name in stackdriver metric label

bianpengyuan:

**Release note**:

```release-note
```


extensions/common/context.cc
```diff
@@ -163,6 +163,7 @@ void populateHTTPRequestInfo(bool outbound, RequestInfo* request_info) {
     request_info->destination_service_host =
         getHeaderMapValue(HeaderMapType::RequestHeaders, kAuthorityHeaderKey)
             ->toString();
+    request_info->destination_service_name = cluster_name;
```
bianpengyuan: not sure if this is a good fallback or not..

kyessenov: Does cluster_name mean xDS cluster name? I think parsing host FQDN is preferred over that for consistency (we can still of course extract FQDN from cluster name as an option).

bianpengyuan: Yeah, I was thinking about passthrough and blackhole cluster case, where cluster name does not follow Istio convention and is used for service name. Seems like we need more logic here, I removed this line and added a todo for it. Let's get the bug fix in first.

mandarjog: Yes let’s just keep this fix small, just enough to unlock UI.

extensions/stackdriver/metric/record.cc
```diff
@@ -66,7 +66,7 @@ void record(bool is_outbound, const ::wasm::common::NodeInfo &local_node_info,
        {requestProtocolKey(), request_info.request_protocol},
        {serviceAuthenticationPolicyKey(),
         request_info.mTLS ? kMutualTLS : kNone},
-       {destinationServiceNameKey(), request_info.destination_service_host},
+       {destinationServiceNameKey(), request_info.destination_service_name},
```
mandarjog: If the request was sent to an IP, we will still parse the 1st element of it.
Can you verify? No need to block this PR though

bianpengyuan: The 1st element will be parsed when the cluster name follows istio convention. We can add a test for ingressgateway stats in istio/istio repo where it uses an IP.

bianpengyuan:
@kyessenov @PiotrSikora Can we merge this? Thanks!

extensions/stackdriver/metric/record.cc
kyessenov: Are you asking whether we should parsing "10.0.0.123", for example? It would be weird to have destination_service_name "10".

bianpengyuan: I think Mandar does not want that and current logic would not cause an IP to be parsed. We can verify it in istio/istio test.

PR #2503, opened 2019-10-31
test: add TLS SD validation

kyessenov:


kyessenov:
Note: I move all the templates to separate files, since it's easier to edit them this way

PR #2504, opened 2019-11-01
Update Envoy-WASM SHA to latest based on Envoy v1.12.0.

PiotrSikora:


PiotrSikora:
@jwendell OK, let me pull a new version of wee8 to get that fix in the follow-up PR, but this can be merged now, since the Python issue is unrelated to this PR.

PR #2508, opened 2019-11-01
Update Envoy-WASM SHA to latest with wee8 v7.9.317.14.

PiotrSikora:


PiotrSikora:
@jwendell this should fix the Python 3.7 build issue.

PR #2509, opened 2019-11-01
Use RBE in Prow jobs.

PiotrSikora:
Fixes istio/test-infra#2013.

### DIFF
--- a/prow/proxy-common.inc
+++ b/prow/proxy-common.inc
@@ -31,3 +31,17 @@ export BAZEL_BUILD_ARGS="--local_ram_resources=12288 --local_cpu_resources=8 --v
 
 # e2e tests under //test/envoye2e/... use Bazel artifacts.
 export BAZEL_OUT="$(bazel info output_path)/k8-fastbuild/bin"
+
+# Use GCP service account when available.
+if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
+  echo "Detected GOOGLE_APPLICATION_CREDENTIALS, activating..." >&2
+  gcloud auth activate-service-account --key-file="${GOOGLE_APPLICATION_CREDENTIALS}"
+
+  # Use RBE when logged in.
+  BAZEL_BUILD_RBE_INSTANCE="${BAZEL_BUILD_RBE_INSTANCE:-projects/istio-testing/instances/default_instance}"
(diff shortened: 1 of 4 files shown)


prow/proxy-common.inc
```diff
@@ -31,3 +31,17 @@ export BAZEL_BUILD_ARGS="--local_ram_resources=12288 --local_cpu_resources=8 --v
 
 # e2e tests under //test/envoye2e/... use Bazel artifacts.
 export BAZEL_OUT="$(bazel info output_path)/k8-fastbuild/bin"
+
+# Use GCP service account when available.
+if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
+  echo "Detected GOOGLE_APPLICATION_CREDENTIALS, activating..." >&2
+  gcloud auth activate-service-account --key-file="${GOOGLE_APPLICATION_CREDENTIALS}"
+
+  # Use RBE when logged in.
```
lizan: Is prow running under GCE instances (or GKE)? If so I don't think you even need this, bazel will use metadata server to fetch credentials automatically.

PiotrSikora: I think it's GKE, but I'm not sure... @fejta, do you?

PiotrSikora: In any case, I don't think Bazel fetches credentials on it's own, I think you'd at least need to pass `--google_default_credentials=true`. This is more explicit, IMHO.

lizan: It does because I don't pass anything on my GCE dev machine :) But no hurt to add that.

fejta: This runs in containers not a VM so the metadata server is not the right thing to depend on. You can use GKE workload identity to authenticate though. 

I do believe you'll need at least the default credentials flag either way. 

https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

PiotrSikora: This works fine as-is, so I'm just going to leave it.

fejta:
In case it's helpful this is the config kubernetes uses: https://github.com/kubernetes/test-infra/blob/master/.bazelrc

PiotrSikora:
@fejta thanks, but everything works fine already... I was just waiting for other PR to get merged first.

fejta:
Awesome!

PR #2510, opened 2019-11-02
Stop building Debian packages and Docker images in tests.

PiotrSikora:


PiotrSikora:
> Maybe make those targets manual instead?

But we want to build them with `bazel build //...`

PR #2513, opened 2019-11-04
metadata: skip sending metadata for explicit inbound/outbound

kyessenov:
I've kept the root context a singleton so as not to break EnvoyFilter. 
This adds an extra call to ABI to read direction per request.

### DIFF
--- a/extensions/metadata_exchange/plugin.h
+++ b/extensions/metadata_exchange/plugin.h
@@ -26,6 +26,7 @@ static const std::string EMPTY_STRING;
 
 #else
 
+#include "extensions/common/context.h"
 #include "extensions/common/wasm/null/null_plugin.h"
 
 namespace Envoy {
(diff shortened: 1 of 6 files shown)


bianpengyuan:
`not to break EnvoyFilter` means keep the EnvoyFilter config directionless? I suspect it won't work because of the listener initialization race. With the current wasm implementation, the first listener that triggers VM initialization will decide the direction of plugins on all listeners.

kyessenov:
@bianpengyuan To prevent this, I don't store the direction in the root context, only in the per-request context. I think that should work (at least tests indicate so).

bianpengyuan:
Do we have test running a proxy with both inbound and outbound listeners? I think all tests config proxy with single direction, no?

kyessenov:
@bianpengyuan Ah, right. Good idea to test that. Still, do you see a problem with per-request context direction?

kyessenov:
I'll update the xDS test to use a single proxy in some tests. That should help with identifying any issues with multi-context.

extensions/metadata_exchange/plugin.h
```diff
@@ -71,7 +72,9 @@ class PluginRootContext : public RootContext {
 // Per-stream context.
 class PluginContext : public Context {
  public:
-  explicit PluginContext(uint32_t id, RootContext* root) : Context(id, root) {}
+  explicit PluginContext(uint32_t id, RootContext* root) : Context(id, root) {
+    direction_ = ::Wasm::Common::getTrafficDirection();
```
mandarjog: Is this just as efficient as getting during config time?
This call will be made for every request? Why not make the filter direction aware just like everything else?

kyessenov: It feels expensive to run two VMs just to pick one based on the listener direction. I'm not sure how much more expensive adding this extra ABI call is.

PR #2514, opened 2019-11-05
Tolerate missing field in node metadata platform metadata

bianpengyuan:
fix https://github.com/istio/istio/issues/18614

### DIFF
--- a/extensions/stackdriver/edges/edge_reporter.cc
+++ b/extensions/stackdriver/edges/edge_reporter.cc
@@ -48,13 +48,13 @@ void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
                   ".", node_info.namespace_());
   // TODO(douglas-reid): support more than just GCP ?
   const auto& platform_metadata = node_info.platform_metadata();
-  auto iter = platform_metadata.find(Common::kGCPLocationKey);
-  if (iter != platform_metadata.end()) {
-    instance->set_location(iter->second);
+  const auto location_iter = platform_metadata.find(Common::kGCPLocationKey);
+  if (location_iter != platform_metadata.end()) {
+    instance->set_location(location_iter->second);
   }
-  iter = platform_metadata.find(Common::kGCPClusterNameKey);
-  if (iter != platform_metadata.end()) {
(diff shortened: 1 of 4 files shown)


extensions/stackdriver/log/logger.cc
```diff
@@ -44,15 +44,18 @@ Logger::Logger(const ::wasm::common::NodeInfo& local_node_info,
...
   // Set log names.
-  const auto& project_id =
-      local_node_info.platform_metadata().at(Common::kGCPProjectKey);
+  const auto& platform_metadata = local_node_info.platform_metadata();
+  auto iter = platform_metadata.find(Common::kGCPProjectKey);
+  std::string project_id = "";
+  if (iter != platform_metadata.end()) {
+    project_id = iter->second;
+  }
   log_entries_request_->set_log_name("projects/" + project_id + "/logs/" +
```
kyessenov: maybe we should skip setting log name if project ID is not found? or is there a better default?

bianpengyuan: There is no good fallback, we should just use the validation API to reject the listener config. I added a TODO for it, will do it in another PR.

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -47,10 +47,15 @@ void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
   absl::StrAppend(instance->mutable_uid(), "kubernetes://", node_info.name(),
                   ".", node_info.namespace_());
   // TODO(douglas-reid): support more than just GCP ?
-  instance->set_location(
-      node_info.platform_metadata().at(Common::kGCPLocationKey));
-  instance->set_cluster_name(
-      node_info.platform_metadata().at(Common::kGCPClusterNameKey));
+  const auto& platform_metadata = node_info.platform_metadata();
+  auto iter = platform_metadata.find(Common::kGCPLocationKey);
```
kyessenov: nit: const to force const iterator

bianpengyuan: Done.

PR #2520, opened 2019-11-05
Enable TCP Metrics in Stats Plugin

gargnupur:
Enable TCP Metrics in Stats Plugin

This currently works only on destination side and not on source side because 
we are not able to access filter state from upstream filters in downstream filters.
Will add issue# for it soon.

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -126,6 +126,56 @@ void getDestinationService(const std::string& dest_namespace,
   extractServiceName(*dest_svc_host, dest_namespace, dest_svc_name);
 }
 
+void populateRequestInfo(bool outbound, bool use_host_header_fallback,
+                         RequestInfo* request_info,
+                         const std::string& destination_namespace) {
+  // Fill in request info.
+  // Get destination service name and host based on cluster name and host
+  // header.
+  getDestinationService(destination_namespace, use_host_header_fallback,
+                        &request_info->destination_service_host,
+                        &request_info->destination_service_name);
+
+  // Get rbac labels from dynamic metadata.
+  getStringValue({"metadata", kRbacFilterName, kRbacPermissivePolicyIDField},
+                 &request_info->rbac_permissive_policy_id);
+  getStringValue(
+      {"metadata", kRbacFilterName, kRbacPermissiveEngineResultField},
+      &request_info->rbac_permissive_engine_result);
+
+  getStringValue({"request", "url_path"}, &request_info->request_url_path);
+
+  if (outbound) {
+    uint64_t destination_port = 0;
+    getValue({"upstream", "port"}, &destination_port);
+    request_info->destination_port = destination_port;
+    getStringValue({"upstream", "uri_san_peer_certificate"},
+                   &request_info->destination_principal);
+    getStringValue({"upstream", "uri_san_local_certificate"},
+                   &request_info->source_principal);
+  } else {
+    bool mtls = false;
+    if (getValue({"connection", "mtls"}, &mtls)) {
+      request_info->service_auth_policy =
+          mtls ? ::Wasm::Common::ServiceAuthenticationPolicy::MutualTLS
+               : ::Wasm::Common::ServiceAuthenticationPolicy::None;
+    }
+    getStringValue({"connection", "uri_san_local_certificate"},
+                   &request_info->destination_principal);
+    getStringValue({"connection", "uri_san_peer_certificate"},
+                   &request_info->source_principal);
+  }
+
+  uint64_t response_flags = 0;
+  getValue({"response", "flags"}, &response_flags);
+  request_info->response_flag = parseResponseFlag(response_flags);
+
+  getValue({"request", "time"}, &request_info->start_time);
+  getValue({"request", "duration"}, &request_info->duration);
+  getValue({"request", "total_size"}, &request_info->request_size);
+  getValue({"response", "total_size"}, &request_info->response_size);
+}
+
 }  // namespace
 
 StringView AuthenticationPolicyString(ServiceAuthenticationPolicy policy) {
@@ -217,7 +267,9 @@ google::protobuf::util::Status extractLocalNodeMetadata(
 void populateHTTPRequestInfo(bool outbound, bool use_host_header_fallback,
                              RequestInfo* request_info,
                              const std::string& destination_namespace) {
-  // Fill in request info.
+  populateRequestInfo(outbound, use_host_header_fallback, request_info,
+                      destination_namespace);
+
   int64_t response_code = 0;
   if (getValue({"response", "code"}, &response_code)) {
     request_info->response_code = response_code;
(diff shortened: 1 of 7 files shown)


WORKSPACE
```diff
@@ -38,9 +38,9 @@ bind(
 # 2. Update .bazelrc and .bazelversion files.
 #
 # envoy-wasm commit date: 11/01/2019
```
PiotrSikora: Nit: update the date.

extensions/common/context.cc
```diff
@@ -227,6 +227,56 @@ void populateHTTPRequestInfo(bool outbound, RequestInfo* request_info) {
...
+  int64_t destination_port = 0;
+
+  if (outbound) {
+    getValue({"upstream", "port"}, &destination_port);
+    getStringValue({"upstream", "uri_san_peer_certificate"},
+                   &request_info->destination_principal);
+    getStringValue({"upstream", "uri_san_local_certificate"},
+                   &request_info->source_principal);
+  } else {
+    // getValue({"destination", "port"}, &destination_port);
```
PiotrSikora: Why is this commented out?

extensions/common/context.cc
```diff
@@ -227,6 +227,56 @@ void populateHTTPRequestInfo(bool outbound, RequestInfo* request_info) {
...
+      request_info->service_auth_policy =
+          mtls ? ::Wasm::Common::ServiceAuthenticationPolicy::MutualTLS
+               : ::Wasm::Common::ServiceAuthenticationPolicy::None;
+    }
+    getStringValue({"connection", "uri_san_local_certificate"},
+                   &request_info->destination_principal);
+    getStringValue({"connection", "uri_san_peer_certificate"},
+                   &request_info->source_principal);
+  }
+  request_info->destination_port = destination_port;
```
PiotrSikora: This will be unset in the inbound case.

WORKSPACE
gargnupur: Thanks.. Fixed

extensions/common/context.cc
gargnupur: This was giving segmentation fault for TCP, hence commented out for now.

gargnupur: Actually, just looked downstream address is filled in tcp proxy => https://cs.corp.google.com/piper///depot/google3/third_party/envoy/src/source/common/tcp_proxy/tcp_proxy.cc?sq=package:piper+file://depot/google3+-file:google3/(experimental%7Cobsolete)&type=cs&rcl=279545696&g=0&l=208

Stats filter currently, runs before tcp proxy... should it actually run after?

extensions/common/context.cc
gargnupur: Will change based on discussion above

extensions/common/context.cc
gargnupur: Actually, just looked downstream address is filled in tcp proxy and stats filter currently, runs before tcp proxy... should it actually run after?

PiotrSikora: Are filter after TCP proxy even executed? TCP proxy writes directly to the connection object, so I'm not sure if you'd be getting correct callbacks if you install it after TCP proxy.

gargnupur: Yeah, just tried.. filters are not allowed after TCP Proxy... debugging inside wasm to see why I am getting segfault here..

gargnupur: Looks like local destination address is null.(downstreamLocalAddress).. StreamInfo says this can never be null though...

kyessenov: Would it be possible to share the code between this and the one above? There's a lot of duplication.

extensions/common/context.h
```diff
@@ -76,6 +76,14 @@ struct RequestInfo {
   // Response total size in bytes, include header, body, and trailer.
   int64_t response_size = 0;
 
+  // Total Request total size in bytes for TCP, as seen on the upstream
+  // connection.
+  int64_t tcp_request_size = 0;
```
kyessenov: Maybe just the same field and call "Stream request" where stream is either an http1.1/h2 request body or a TCP connection.

extensions/stats/plugin.h
```diff
@@ -399,6 +403,27 @@ class PluginContext : public Context {
...
+      auto rootCtx = rootContext();
+      ::Wasm::Common::populateTCPRequestInfo(rootCtx->outbound(),
+                                             &request_info_);
+      rootCtx->report(request_info_);
+    }
+  }
+  void onUpstreamConnectionClose(PeerType) override {
+    upstream_closed_ = true;
+    if (downstream_closed_ && !tcp_reported_) {
+      tcp_reported_ = true;
```
kyessenov: Dedupe the code between the two callbacks.

extensions/stats/plugin.cc
```diff
@@ -103,6 +103,19 @@ bool PluginRootContext::onConfigure(size_t) {
...
             return request_info.response_size;
           },
+          field_separator, value_separator),
+      StatGen(
+          absl::StrCat(stat_prefix, "tcp_request_bytes"), MetricType::Histogram,
+          [](const ::Wasm::Common::RequestInfo& request_info) -> uint64_t {
+            return request_info.tcp_request_size;
+          },
+          field_separator, value_separator),
+      StatGen(
```
kyessenov: Doesn't this produce a label `tcp_request_size` for HTTP requests, too? Should this be guarded/conditioned by a protocol?

extensions/common/context.h
gargnupur: shouldn't we have rather total_request_size => 
http => body + header + trailer
tcp => connection data size
?

extensions/stats/plugin.h
gargnupur: Done..

extensions/stats/plugin.cc
gargnupur: Will change based on the name we decide above...

extensions/common/context.cc
gargnupur: Done..

extensions/common/context.h
gargnupur: Done.. just reusing the fields as using the values set in streaminfo for tcp for now...

extensions/stats/plugin.cc
gargnupur: Removed tcp specific stuff

gargnupur:
ping... can I please get a review on this PR

extensions/common/context.cc
```diff
@@ -68,6 +68,57 @@ void extractServiceName(const std::string& fqdn, std::string* service_name) {
...
+
+  // Fill in request info.
+  // Try to get fqdn of destination service from cluster name. If not found, use
+  // host header instead.
+  std::string cluster_name = "";
+  getStringValue({"cluster_name"}, &cluster_name);
+  extractFqdn(cluster_name, &request_info->destination_service_host);
+  if (!request_info->destination_service_host.empty()) {
+    // cluster name follows Istio convention, so extract out service name.
+    extractServiceName(request_info->destination_service_host,
```
kyessenov: Can you merge master? I feel like I just saw a variant of this by @bianpengyuan . Can we combine the two?

gargnupur: Looks like the pr is not merged yet... https://github.com/istio/proxy/pull/2548...

gargnupur: Merged master and updated PR

gargnupur:
@kyessenov : merged with master and is ready for review again.. ptal..

extensions/stats/plugin.h
```diff
@@ -393,14 +393,41 @@ class PluginRootContextInbound : public PluginRootContext {
...
+
+  void logTCP() {
+    tcp_reported_ = true;
+    rootContext()->report(true);
+  }
+
+  bool upstream_closed_;
+  bool downstream_closed_;
+  bool tcp_reported_;
+  ::Wasm::Common::RequestInfo request_info_;
```
kyessenov: This looks like it's unused?

src/envoy/tcp/metadata_exchange/metadata_exchange.cc
```diff
@@ -262,7 +262,7 @@ void MetadataExchangeFilter::tryReadProxyData(Buffer::Instance& data) {
   const auto key_metadata_id_it =
       value_struct.fields().find(ExchangeMetadataHeaderId);
   if (key_metadata_id_it != value_struct.fields().end()) {
-    Envoy::ProtobufWkt::Value val = key_metadata_it->second;
+    Envoy::ProtobufWkt::Value val = key_metadata_id_it->second;
```
kyessenov: oops

kyessenov:
Please remove the do-not-merge label once you address the nit.

extensions/stats/plugin.h
gargnupur: Thanks.. removed

PR #2521, opened 2019-11-05
fix(edges): use service name (instead of host) in edge reporting

douglas-reid:
Instead of using `request_info.destination_service_host` for the `destination_service_name` field, this PR uses `request_info.destination_service_name`.

It also updates the kube URLs in the repo to be the proper plural forms.


bianpengyuan:
@douglas-reid Do you want to do a manual cherry-pick for this?

douglas-reid:
@bianpengyuan yes. i'll try in a few.

PR #2523, opened 2019-11-05
fix(edges): add debug log for generated requests

douglas-reid:
Adding a debug log to view outgoing requests to mesh telemetry endpoint.

### DIFF
--- a/extensions/stackdriver/edges/mesh_edges_service_client.cc
+++ b/extensions/stackdriver/edges/mesh_edges_service_client.cc
@@ -87,8 +87,8 @@ MeshEdgesServiceClientImpl::MeshEdgesServiceClientImpl(
 
 void MeshEdgesServiceClientImpl::reportTrafficAssertions(
     const ReportTrafficAssertionsRequest& request) const {
-
-  LOG_TRACE("mesh edge services client: sending request '" + request.DebugString() + "'");
+  LOG_TRACE("mesh edge services client: sending request '" +
+            request.DebugString() + "'");
 
   context_->grpcSimpleCall(
       grpc_service_, kMeshEdgesService, kReportTrafficAssertions, request,
--- a/extensions/stackdriver/edges/mesh_edges_service_client.cc
+++ b/extensions/stackdriver/edges/mesh_edges_service_client.cc
@@ -87,6 +87,9 @@ MeshEdgesServiceClientImpl::MeshEdgesServiceClientImpl(
 
 void MeshEdgesServiceClientImpl::reportTrafficAssertions(
     const ReportTrafficAssertionsRequest& request) const {
+
+  logDebug("mesh edge services client: sending request '" + request.DebugString() + "'");
+
   context_->grpcSimpleCall(
       grpc_service_, kMeshEdgesService, kReportTrafficAssertions, request,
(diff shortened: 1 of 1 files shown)


extensions/stackdriver/edges/mesh_edges_service_client.cc
```diff
@@ -87,6 +87,9 @@ MeshEdgesServiceClientImpl::MeshEdgesServiceClientImpl(
 
 void MeshEdgesServiceClientImpl::reportTrafficAssertions(
     const ReportTrafficAssertionsRequest& request) const {
+
+  logDebug("mesh edge services client: sending request '" + request.DebugString() + "'");
```
bianpengyuan: This is actually quite expensive when we switch to v8. Log level information is not available inside the VM, so this function needs to copy the log message to the envoy host and let it make the decision on whether to log it. So could we hard code something in node metadata to switch it on/off? Also could you switch to use `LOG_DEBUG` like:
https://github.com/istio/proxy/blob/d9139fa71feae49a7d17b220e91cc1648f11229e/extensions/stats/plugin.cc#L122

douglas-reid: I switched to `LOG_TRACE`.

I'm not clear how to proceed on making this less expensive.

PiotrSikora: We used to have macro for in-VM evaluated logging, which log level was evaluated inside VM, before calling host, but I "accidentally" removed it during one of the upgrades (see: [#2395](https://github.com/istio/proxy/pull/2395/files#diff-70887b3a407a875b71e9cb77b627856e)).

Perhaps revisiting that concept would be a good idea? Ideally, we would measure the cost of this call, since in-VM evaluated logging is configured outside of Envoy's log level control (i.e. we cannot change log level using Envoy's admin endpoint), so there is a trade-off to consider.

douglas-reid: @PiotrSikora I'd love to revisit that evaluation. This level of debugging was essential for speeding up the debug work on this integration. Having to maintain a separate build for troubleshooting is not ideal. Hopefully, it will not be used much going forward, but it would be great if it was available (and efficiently).

mandarjog: I did not realize that the in vm was no longer there.
What ever we do has to be type
If (level) { construct strings } otherwise we will construct string *and* car them over to the host on every call.

Request.debugString is an expensive call, since it happens here once every 10 mins it is ok. Having this type of logTrace in other places is not good.
I will see if it has already affected the stats filter.

PiotrSikora: Agreed, I'm all for re-adding it back, ideally as part of the C++ SDK and not just here.

Long-term, we should figure out how to integrate it with Envoy's log levels, since changing them at runtime has been really helpful during debugging.

PiotrSikora: In the meantime, can you guys decide whether this should block this PR or not?

bianpengyuan: I think this is fine for the edge part. In prod the code path should only be trigger every several minutes. Let's get it in.

PR #2527, opened 2019-11-06
Fill response flag for stats plugin

bianpengyuan:
This needs https://github.com/istio/envoy/pull/116 to be in to work with 1.4 and envoyproxy/envoy-wasm to be updated to work with master.

### DIFF
--- a/extensions/common/BUILD
+++ b/extensions/common/BUILD
@@ -31,9 +31,11 @@ envoy_cc_library(
     name = "context",
     srcs = [
         "context.cc",
+        "util.cc",
     ],
     hdrs = [
         "context.h",
+        "util.h",
     ],
     repository = "@envoy",
     visibility = ["//visibility:public"],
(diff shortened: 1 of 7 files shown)


bianpengyuan:
Can I get some review on this? Thanks!

bianpengyuan:
@kyessenov Does this look like the a flake to you? It seems to be different from client timeout we saw before. There is no crashes and does not return any mismatches.. so likely the prometheus endpoint does not return metrics. Could this also happen because of staled listeners?

```
--- FAIL: TestStatsParallel (48.68s)
    stats_xds_test.go:233: failed to match all stats
FAIL
FAIL	istio.io/proxy/test/envoye2e/stats	66.504s
ok  	istio.io/proxy/test/envoye2e/stats_plugin	23.067s
ok  	istio.io/proxy/test/envoye2e/tcp_metadata_exchange	5.045s
Makefile.core.mk:79: recipe for target 'test_tsan' failed
make: *** [test_tsan] Error 1
```

kyessenov:
@bianpengyuan Yeah this is not a timeout. Either prometheus reports incorrect stats, or we failed to report one. Does this happen consistently? We can disable the test for now to get this in.

bianpengyuan:
@kyessenov haven't retried it. Let me retest it.

kyessenov:
I noticed that it was polling stats endpoint for ~20s. I think it was just starved of CPU resources since TSAN builds are slow.

PR #2528, opened 2019-11-06
Fix building extensions as WebAssembly modules (again).

PiotrSikora:
Those WebAssembly modules are not built as part of the CI yet,
because we cannot run Docker-in-Docker.


bianpengyuan:
If it is not expensive, how about adding emscripten etc into proxy builder image? This PR will need update after https://github.com/istio/proxy/pull/2527

PiotrSikora:
Yeah, I think we have to install wasmsdk inside the build images to make this work, but that's separate from this PR.

bianpengyuan:
Can you please incorporate change from #2527

PiotrSikora:
> Can you please incorporate change from #2527

I believe that PR is already included in the latest merge? Or did you have anything else in mind?

bianpengyuan:
It adds two new files `extensions/common/util.{h.cc}` which I believe should also be added into stats plugin's source?

PiotrSikora:
@bianpengyuan good catch, it looks that `build_wasm.sh` scripts were not exiting on failures, so this error was masked by metadata plugin building successfully. Fixed now.

PiotrSikora:
Also, note that this removes `extensions/stats/plugin.wasm`, which presumably was committed by accident?

/hold

PR #2532, opened 2019-11-07
Update envoy SHA for istio/envoy#118.

lambdai:

pick up istio/envoy#118 and fix the cpu burn issue istio/istio#18229

`bazel test  //test/...` is passing


lambdai:
Confirm istio/istio#18229 is fixed

PR #2535, opened 2019-11-07
fix(edges): use unknown for missing metadata instance values

douglas-reid:
If metadata values are missing, we should report Workload Instance information with "unknown" values instead of leaving them empty. This PR sets defaults via changes to `instanceFromMetadata()`.

**Which issue this PR fixes**

fixes https://github.com/istio/istio/issues/18711

### DIFF
--- a/extensions/stackdriver/edges/edge_reporter.cc
+++ b/extensions/stackdriver/edges/edge_reporter.cc
@@ -40,25 +40,39 @@ using google::cloud::meshtelemetry::v1alpha1::
     TrafficAssertion_Protocol_PROTOCOL_TCP;
 using google::cloud::meshtelemetry::v1alpha1::WorkloadInstance;
 
+constexpr char kUnknown[] = "unknown";
+
+std::string valueOrUnknown(std::string value) {
+  if (value.length() == 0) {
+    return kUnknown;
+  }
+  return value;
+}
+
 namespace {
 void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
                           WorkloadInstance* instance) {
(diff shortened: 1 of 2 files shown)


kyessenov:
@douglas-reid fix your DCO to use company email :)

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -40,25 +40,39 @@ using google::cloud::meshtelemetry::v1alpha1::
     TrafficAssertion_Protocol_PROTOCOL_TCP;
 using google::cloud::meshtelemetry::v1alpha1::WorkloadInstance;
 
+constexpr char kUnknown[] = "unknown";
+
+std::string valueOrUnknown(std::string value) {
```
kyessenov: avoid copy of string: maybe use string_view as return type and receive a `const std::string&`?

douglas-reid: updated to use `const std::string&`. switching to string_view will require more code munging as they aren't accepted in the proto setters.

kyessenov: still copies to return. but it's one less copy.

douglas-reid: shouldn't modern c++ compilers be able to avoid copies in this scenario? i find the standard impossible to parse, but this feels like a situation in which copy eliding could apply.

kyessenov: there are too many compilers out there to say definitively. probably the very latest ones do, but not sure about older clang, gcc, and emscripten.

PR #2539, opened 2019-11-07
Update metadata exchange filter config

bianpengyuan:

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stats/testdata/istio/metadata-exchange_filter.yaml
+++ b/extensions/stats/testdata/istio/metadata-exchange_filter.yaml
@@ -6,7 +6,7 @@ spec:
   configPatches:
     - applyTo: HTTP_FILTER
       match:
-        context: ANY # inbound, outbound, and gateway
+        context: SIDECAR_INBOUND
         listener:
           filterChain:
             filter:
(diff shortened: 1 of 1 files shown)


kyessenov:
Do we need to separate VMs? Is this because context is shared?

bianpengyuan:
Yes, I think after mx filter became direction aware, @douglas-reid recently runs into the issue that metadata is all destination side. Adding vm id fixes the issue for him. @douglas-reid Could you confirm this?

kyessenov:
Hmm, I wonder why we couldn't reproduce it in https://github.com/istio/proxy/pull/2515

bianpengyuan:
yeah there is actually a race. If there is no vm id, the first initialized listener decides the direction of mx plugin on all listeners. vm id is more like a hack which acts an annotation of listener directions...

kyessenov:
@jplevyak what happens when you don't specify `vm_id`? I'm seeing two per-requests contexts created sharing the root context:
```
[2019-11-07 16:46:57.662][26074][warning][wasm] [external/envoy/source/extensions/common/wasm/wasm.cc:1677] wasm log : **** onRequestHeaders drection: 2 id: 2 root id: 1
[2019-11-07 16:46:57.663][26074][warning][wasm] [external/envoy/source/extensions/common/wasm/wasm.cc:1677] wasm log : **** onRequestHeaders drection: 1 id: 3 root id: 1
[2019-11-07 16:46:57.665][26074][warning][wasm] [external/envoy/source/extensions/common/wasm/wasm.cc:1677] wasm log : **** onResponseHeaders drection: 1 id: 3 root id: 1
[2019-11-07 16:46:57.666][26074][warning][wasm] [external/envoy/source/extensions/common/wasm/wasm.cc:1677] wasm log : **** onResponseHeaders drection: 2 id: 2 root id: 1
```

bianpengyuan:
Looks like we switch back to the old behavior where if `vm_id` is not specified, then an vm will be created for each plugin? I updated https://github.com/istio/istio/pull/18768 to remove all `vm_id` in test. It used to fail all the time if vm id is not provided. Let's see if it passes there.

douglas-reid:
FWIW, config of this nature was causing crashes in one of the versions of the proxy i built, so I switched back to the `ANY` config.

PR #2541, opened 2019-11-08
Revert "Update metadata exchange filter config"

bianpengyuan:
Reverts istio/proxy#2539

This is not needed since per stream context actually has corresponding listener config. Mx filter gets direction in each stream, so it does not need vm id to distinguish.

### DIFF
--- a/extensions/stats/testdata/istio/metadata-exchange_filter.yaml
+++ b/extensions/stats/testdata/istio/metadata-exchange_filter.yaml
@@ -6,7 +6,7 @@ spec:
   configPatches:
     - applyTo: HTTP_FILTER
       match:
-        context: SIDECAR_INBOUND
+        context: ANY # inbound, outbound, and gateway
         listener:
           filterChain:
             filter:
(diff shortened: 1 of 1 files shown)


extensions/stats/testdata/istio/metadata-exchange_filter.yaml
```diff
@@ -19,46 +19,6 @@ spec:
             config:
               configuration: envoy.wasm.metadata_exchange
```
kyessenov: can you remove this configuration or make it empty? it's not used.

bianpengyuan: Do we have a test verify metadata exchange filter with empty configuration? I am just worrying the wasm layer might bypass onConfigure call when it is not provided.

kyessenov: I tried it without in one of the tests, it worked. However, I'm not sure it is not bypassing, so let's leave it as-is then.

PR #2546, opened 2019-11-10
v2: Add ability to control host header fallback

mandarjog:
1. Add config to control ability to disable host header fallback
2. Update stats and stackdriver filter to use it.

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -156,7 +156,9 @@ google::protobuf::util::Status extractLocalNodeMetadata(
   return extractNodeMetadata(node, node_info);
 }
 
-void populateHTTPRequestInfo(bool outbound, RequestInfo* request_info) {
+// Host header is used if use_host_header_fallback==true.
+// Normally it is ok to use host header within the mesh, but not at ingress.
+void populateHTTPRequestInfo(bool outbound, bool use_host_header_fallback, RequestInfo* request_info) {
   // TODO: switch to stream_info.requestComplete() to avoid extra compute.
   request_info->end_timestamp = getCurrentTimeNanoseconds();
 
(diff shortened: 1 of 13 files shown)


mandarjog:
Removing cherry-pick label, since automatic cherry-pick is not possible.

extensions/common/context.cc
```diff
@@ -156,7 +156,10 @@ google::protobuf::util::Status extractLocalNodeMetadata(
   return extractNodeMetadata(node, node_info);
 }
 
-void populateHTTPRequestInfo(bool outbound, RequestInfo* request_info) {
+// Host header is used if use_host_header_fallback==true.
+// Normally it is ok to use host header within the mesh, but not at ingress.
```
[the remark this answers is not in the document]
mandarjog: you can configure it however you like. This is an explanation.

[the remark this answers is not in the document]
mandarjog: Some people will will trust host header from all traffic and some people will not trust it at all.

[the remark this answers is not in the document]
howardjohn: Makes sense, I probably should have read more than the first line

extensions/common/context.cc
```diff
@@ -181,16 +184,16 @@ void populateHTTPRequestInfo(bool outbound, RequestInfo* request_info) {
   std::string cluster_name = "";
   getStringValue({"cluster_name"}, &cluster_name);
   extractFqdn(cluster_name, &request_info->destination_service_host);
-  if (request_info->destination_service_host.empty()) {
-    // fallback to host header.
+  if (!request_info->destination_service_host.empty()) {
+    // cluster name follows Istio convention, so extract out service name.
+    extractServiceName(request_info->destination_service_host,
+                       &request_info->destination_service_name);
+  } else if (use_host_header_fallback) {
```
bianpengyuan: Can we make this if (user_host_header_fallback && !cluster_name.empty())?

bianpengyuan: Actually how about just `if (!cluster_name.empty())` ?

mandarjog: what is the else clause ? Do you want to just set it to cluster name then ?
For non-pilot control planes it will not be meaningful.

extensions/common/context.cc
```diff
@@ -156,7 +156,10 @@ google::protobuf::util::Status extractLocalNodeMetadata(
   return extractNodeMetadata(node, node_info);
 }
 
-void populateHTTPRequestInfo(bool outbound, RequestInfo* request_info) {
+// Host header is used if use_host_header_fallback==true.
```
PiotrSikora: Host header will be used only when the destination service is missing, so this comment is misleading.

extensions/common/context.cc
bianpengyuan: No, the else clause still falls back to request header, but only if there is a known route. My understanding is that if there is a known route, request host must be some constrained value, instead of random, no?

extensions/common/context.cc
mandarjog: Updated comment.

extensions/common/context.cc
mandarjog: known route or unknown route is orthogonal to using host header.

You can have a known route with a wildcard match, and you still need to decide if it ok to use the host header.

bianpengyuan: I see. Thanks!

extensions/common/context.cc
mandarjog: oop,s @PiotrSikora I updated comment only in the other PR.

mandarjog:
@bianpengyuan https://github.com/istio/proxy/pull/2547 .  please review this PR

PR #2548, opened 2019-11-11
Extract service name from request host and cluster name

bianpengyuan:
To avoid introducing throw away code to get node information from root context, this change also moves request_info out stream context, which removes computation of request duration and size and use property abi to fetch them from host directly. These changes should be trivial and safe, we could do manual testing to de-risk cherry-pick of this pr to 1.4.


extensions/common/context.cc
```diff
@@ -62,11 +62,40 @@ void extractFqdn(const std::string& cluster_name, std::string* fqdn) {
   }
 }
 
-// Extract service name from service fqdn.
-void extractServiceName(const std::string& fqdn, std::string* service_name) {
-  const std::vector<std::string>& parts = absl::StrSplit(fqdn, '.');
-  if (parts.size() > 0) {
-    *service_name = parts[0];
+// Extract service name from service host.
+void extractServiceName(const std::string& host, std::string* service_name,
```
rshriram: how is this logic portable for non k8s platforms or for external services that are added via service entries?

bianpengyuan: Mixer v1 uses host as service name for external service, so the logic here indeed breaks parity. Is there a good indicator of external service cluster?

extensions/stackdriver/stackdriver.cc
```diff
@@ -184,10 +184,17 @@ void StackdriverRootContext::onTick() {
...
   const auto peer_node_info_ptr = getPeerNode();
   const NodeInfo& peer_node_info =
       peer_node_info_ptr ? *peer_node_info_ptr : ::Wasm::Common::EmptyNodeInfo;
+  const auto& destination_node_info =
+      isOutbound() ? peer_node_info : local_node_info_;
+
+  ::Wasm::Common::RequestInfo request_info;
+  ::Wasm::Common::populateHTTPRequestInfo(isOutbound(), useHostHeaderFallback(),
+                                          &request_info,
+                                          destination_node_info.namespace_());
```
lambdai: I couldn't understand how could namepace_ function is defined. Could you share the knowledge?

extensions/common/context.cc
```diff
@@ -62,11 +62,45 @@ void extractFqdn(const std::string& cluster_name, std::string* fqdn) {
...
+  if (host[name_pos] == ':') {
+    // host is `short_service:port`, return short_service name.
+    *service_name = host.substr(0, name_pos);
+    return;
+  }
+
+  auto namespace_pos = host.find_first_of(".:", name_pos + 1);
+  std::string service_namespace = "";
+  if (namespace_pos == std::string::npos) {
+    service_namespace = host.substr(namespace_pos + 1);
```
lambdai: better not to rely on the impl of `std::string::npos` 
how about just `service_namespace = host`?

extensions/common/context.cc
mandarjog: Yes there is. We already know the destination namespace, if we do not know the destination namespace,, ie. metadata exchange was not sent by the other side, then it is an external service.

So we should use the full service host. It does bring into question how we choose to use service short name. alternatively, there *is no* short name for external services.

@kyessenov @douglas-reid what do you think about not having service name at all for external services?

extensions/stackdriver/stackdriver.cc
bianpengyuan: It is generated from proto: https://github.com/istio/proxy/blob/90b1d9f97e105d20698ece0ebd6fe2f75b44c7ea/extensions/common/node_info.proto#L27. Underscore is used to avoid reserved word `namespace`.

lambdai: `Underscore is used to avoid reserved word namespace.`
TIL

extensions/common/context.cc
bianpengyuan: This is actually a bug, `namespace_pos` should be `name_pos`. Not sure I follow the `npos` concern.

lambdai: Got it. There you go.
A typical value of std::string::npos is `-1`. Prior to fix I thought your intention is `service_namespace = host.substr(-1 +1) /* = host.substr(0),  = host */` :)

extensions/common/context.cc
rshriram: Sorry resolved by mistake. What happens in the egress gateway? Are we going to have different hosts for same gateway service ?

douglas-reid: @mandarjog I think we need a before / after of the telemetry for an external service -- and we should probably run that by the WG.

kyessenov:
It appears upstream SSL info is not filled in onLog. I'll need to look further into this.

kyessenov:
The failure seems to be caused by an uninitialized field service_auth_policy in request info struct.

extensions/common/context.cc
bianpengyuan: I've added more comments how the external service name is decided. It should have achieved parity with the Mixer v1. Ptal. Thanks!

extensions/common/context.cc
```diff
@@ -233,6 +278,11 @@ void populateHTTPRequestInfo(bool outbound, bool use_host_header_fallback,
   uint64_t response_flags = 0;
   getValue({"response", "flags"}, &response_flags);
   request_info->response_flag = parseResponseFlag(response_flags);
+
+  getValue({"request", "time"}, &request_info->start_time);
+  getValue({"request", "duration"}, &request_info->duration);
```
kyessenov: I don't think this works with protobuf types. I believe these are bit-serialized as absl::Time and absl::Duration: https://github.com/envoyproxy/envoy-wasm/blob/master/source/extensions/common/wasm/context.cc#L302. Change the types of struct fields to match what the value should be.

bianpengyuan: I see, it would be good to make the doc here more clear: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/rbac_filter.

kyessenov: Yeah. We didn't want protobuf on the ABI, so the types changed to bit-representations whenever possible.

bianpengyuan: To me it seems makes more sense to use primitive in abi (epoch, nanoseconds), but anyway.

kyessenov: Good point. It might cause a bit of conversion overhead, but it's not set in stone.

bianpengyuan:
Can I get some review on this PR? It would be nice to get this into 1.4 as well, so that we could gain more parity with mixer.

extensions/common/context.cc
```diff
@@ -47,25 +47,83 @@ namespace Common {
...
+//   use the last part as destination host; Otherwise, use host header as
+//   destination host. To get destination service name from host: if destination
+//   host is already a short name, use that as destination service; otherwise if
+//   the second part of destination host is destination namespace, use first
+//   part as destination service name. Otherwise, fallback to use destination
+//   host for destination service name.
+void getDestinationService(const std::string& dest_namespace,
+                           bool use_host_header, std::string* dest_svc_host,
+                           std::string* dest_svc_name) {
+  std::string cluster_name = "";
```
kyessenov: nit: tautological assignment. strings are empty by default.

extensions/common/context.cc
```diff
@@ -47,25 +47,83 @@ namespace Common {
...
+
+  if (cluster_name == kBlackHoleCluster ||
+      cluster_name == kPassThroughCluster ||
+      cluster_name == kInboundPassthroughClusterIpv4 ||
+      cluster_name == kInboundPassthroughClusterIpv6) {
+    *dest_svc_name = cluster_name;
+    return;
+  }
+
+  const std::vector<std::string>& parts = absl::StrSplit(cluster_name, '|');
```
kyessenov: nit: wasteful string allocations. prefer `vector<absl::string_view>`

bianpengyuan: Done.

extensions/common/context.cc
bianpengyuan: Done.

PR #2549, opened 2019-11-11
Update Envoy-WASM SHA to latest with HTTP inspector fixes.

PiotrSikora:
Pulling in istio/envoy#120.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -37,10 +37,10 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/istio/envoy/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelrc and .bazelversion files.
 #
-# envoy commit date: 11/06/2019
-ENVOY_SHA = "bc4e866c4236e4ae7df959a4cb9a3bbd4923f73a"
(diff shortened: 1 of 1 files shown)


rshriram:
Why is this getting an exception when the ALS one is not? Shouldn't we wait until we decide on what to do about the proxy version (basing off envoy stable or master) ?

PiotrSikora:
@rshriram this is a security fix.

PiotrSikora:
@rshriram can you remove the `do-not-merge` label, since it's blocking merge and the release?

howardjohn:
I think @rshriram 's discussion is reasonable, but we should not block the release on this

rshriram:
My point is that we have a very buggy ALS system in 1.4 release today. And fixes for those are not allowed because of deviation from the baseline release. So why is this ? We could turn off sniffing (we already do) in inbound. So, this might as well wait for an entire release ?

PiotrSikora:
@rshriram the whole point of stable releases of Envoy and using them in Istio Proxy is to support bugfixes such as the ALS one. I don't recall anybody not allowing it to be cherry-picked...

lizan:
@PiotrSikora @rshriram yeah sure let's cherry-pick, PTAL istio/envoy#121

PR #2551, opened 2019-11-12
update common file

bianpengyuan:
Meanwhile test the new proxy builder image with prow.


bianpengyuan:
@venilnoronha Yeah, I run `make update-common` to generate this change.

bianpengyuan:
Hmm... that's strange, why tide is not here..

bianpengyuan:
@clarketm Any idea?

clarketm:
@bianpengyuan 
> @clarketm Any idea?

I see an issue with time in the `tide` logs. I think this job is to blame: https://github.com/istio/test-infra/blob/c33be824a61027045427c0d5f61c681a5c6263ec/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml#L42. If we omit the `context` here it should default to job name and not conflict.

```console
 jsonPayload: {
  base-sha: "90b1d9f97e105d20698ece0ebd6fe2f75b44c7ea"   
  branch: "master"   
  component: "tide"   
  controller: "sync"   
  error: "error setting up context checker for pr 2553: contexts prow/proxy-presubmit.sh are defined as required and optional"   
  file: "prow/tide/tide.go:449"   
  func: "k8s.io/test-infra/prow/tide.(*Controller).filterSubpools.func1"   
  level: "error"   
  msg: "Error initializing subpool."   
  org: "istio"   
  repo: "proxy"   
 }
```

bianpengyuan:
I see. Should this be fixed or `context` should be the only key for job? But anyway let me remove the job. New builder test is already done.

howardjohn:
@bianpengyuan we should probably remove the context field from proxy entirely, as we have done with other repos. Basically context is what shows up on the github list. It is confusing since in some places it shows up as the context (like `prow/proxy-presubmit.sh`) and other places the name (like `prow/proxy-presubmit` I think). If you remove context everything uses `name`

clarketm:
@bianpengyuan 
The core problem was there were *different* jobs with the *same* `context` field. As @howardjohn stated a good way around this is to simply **omit** the `context` field and let Prow *default* it to the job `name` (of course job names should be unique as well).

howardjohn:
the job names are also not unique for proxy. We should probably just make
it over to use the config generation, I don't think there are any blockers
anymore?

clarketm:
> the job names are also not unique for proxy. We should probably just make it over to use the config generation, I don't think there are any blockers anymore?

Personally, I do not know the history for why these were not generated. The less attractive alternative is to update the job names.

PR #2553, opened 2019-11-12
using request path when it is grpc request

bianpengyuan:

**Release note**:

```release-note
```

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -207,6 +207,8 @@ void populateHTTPRequestInfo(bool outbound, bool use_host_header_fallback,
       getHeaderMapValue(HeaderMapType::RequestHeaders, kMethodHeaderKey)
           ->toString();
 
+  getStringValue({"request", "url_path"}, &request_info->request_url_path);
+
   int64_t destination_port = 0;
 
   if (outbound) {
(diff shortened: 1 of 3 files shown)


duderino:
@bianpengyuan @mandarjog it would be good to confirm this path has been normalized before using it

bianpengyuan:
@duderino Thanks for reminding! Yes, this is going to be controlled by config. User could control whether to use traffic data as stats label.

kyessenov:
@duderino the path is coming straight from header_map in Envoy sans the query string. So it is whatever envoy is feeding us which I assume is not normalized. gRPC has to be able to parse it though, and it's not clear whether gRPC accepts dots and double slashes.

bianpengyuan:
Oh, by normalization it means path normalization.. This is for stats only, do we need to do normalization? Isn't it just an opaque string for the system?

kyessenov:
There is no risk of authz violation here. The only risk is cardinality explosion, but we already discussed that.

duderino:
how do you mitigate cardinality explosion?  I recall a prior incident where unbounded strings brought down a user's prometheus cluster.   They considered that a severe Istio bug.

bianpengyuan:
We are making it configurable, so that user could disable using unbounded string like this if they don't trust the source peer.

duderino:
Nice job guys

PR #2556, opened 2019-11-13
Allow ENVOY_SHA to be the same as one defined in WORKSPACE

clarketm:

This is an extension (and follow-up) of #2555. This change will allow overriding of the `ENVOY_REPOSITORY` and `ENVOY_PREFIX` externally while retaining the same versioned `sha` for `ENVOY_SHA` - by extracting the value from `WORKSPACE`. For a broader discussion on the need, check out: https://github.com/istio/test-infra/pull/2091#discussion_r345572359

### DIFF
--- a/prow/proxy-common.inc
+++ b/prow/proxy-common.inc
@@ -15,6 +15,7 @@
 WD=$(dirname $0)
 WD=$(cd $WD; pwd)
 ROOT=$(dirname $WD)
+WORKSPACE="${ROOT}/WORKSPACE"
 
 # Exit immediately for non zero status
 set -e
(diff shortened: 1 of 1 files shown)


prow/proxy-common.inc
```diff
@@ -47,9 +48,10 @@ if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
 fi
 
 # Override envoy.
-if [[ "${ENVOY_REPOSITORY:-}" && "${ENVOY_SHA:-}" && "${ENVOY_PREFIX:-}" ]]; then
+if [[ "${ENVOY_REPOSITORY:-}" && "${ENVOY_PREFIX:-}" ]]; then
   TMP_DIR=$(mktemp -d -t envoy-XXXXXXXXXX)
   trap 'rm -rf ${TMP_DIR:?}' EXIT
+  ENVOY_SHA="${ENVOY_SHA:-$(grep -Pom1 "^ENVOY_SHA = \"\K[a-zA-Z0-9]+" "$WORKSPACE")}"
```
howardjohn: not familiar with bazel at all - is there a native way to read the variable instead of parsing it out?

If not lgtm, thanks!

clarketm: I am not a strong bazeler either. I can research a bit more into this.

kyessenov: It's better to use a well-structured bazel file that is shared. Something like https://github.com/envoyproxy/envoy/blob/master/bazel/repository_locations.bzl

lizan: that one doesn't have git sha in structured way either (only as a part of URL). 

btw What's the purpose here? download same tar from different server?

clarketm: > It's better to use a well-structured bazel file that is shared. Something like...

@kyessenov a native bazel solution would be great. I tried that initially and failed to provide a robust solution #2496. I plan to iterate on this as I develop more bazel knowledge.

> btw What's the purpose here? download same tar from different server?

@lizan yes, the idea here is to be able to pull same SHA from different fork/mirror https://github.com/istio/test-infra/pull/2091#discussion_r345572359. More specifically, to swap out the `ENVOY_REPOSITORY` with a *private* fork while retaining the configured SHA.

clarketm:
> @clarketm: #2556 failed to apply on top of branch "release-1.4":
> 
> ```
> error: Failed to merge in the changes.
> Using index info to reconstruct a base tree...
> M	prow/proxy-common.inc
> Falling back to patching base and 3-way merge...
> Auto-merging prow/proxy-common.inc
> CONFLICT (content): Merge conflict in prow/proxy-common.inc
> Patch failed at 0001 Allow ENVOY_SHA to be the same as one defined in WORKSPACE
> ```

After #2562 should work

PR #2559, opened 2019-11-13
Update envoy-wasm sha

lizan:

```release-note
None
```


lizan:
@jplevyak PTAL this is mostly fixing changes by envoyproxy/envoy-wasm#274

lizan:
@PiotrSikora ASAN is RBE timeout, what's the RBE cluster setting we're using?

PiotrSikora:
I don't think there are any settings on the RBE side other than number of instances and their type?

lizan:
@PiotrSikora I'm talking about the machine type, envoy-ci is on n2-standard-2 FYI

.bazelrc
```diff
@@ -8,7 +8,7 @@ import %workspace%/envoy.bazelrc
 
 # Overrides workspace_status_command
 build --workspace_status_command=tools/bazel_get_workspace_status
-
+build:remote --remote_timeout=7200
```
yxue: What's this for?

lizan: Without this RBE will timeout for wee8 build, because istio-testing is using different type of machine than upstream.

PR #2569, opened 2019-11-15
update xeniel to bionic

bianpengyuan:
Fixes istio/istio#18977

**Release note**:

```release-note
```

### DIFF
--- a/scripts/release-binary.sh
+++ b/scripts/release-binary.sh
@@ -37,13 +37,13 @@ fi
 # The bucket name to store proxy binaries.
 DST=""
 
-# Verify that we're building binaries on Ubuntu 16.04 (Xenial).
+# Verify that we're building binaries on Ubuntu 18.04 (Bionic).
 CHECK=1
 
 function usage() {
   echo "$0
     -d  The bucket name to store proxy binary (optional).
-    -i  Skip Ubuntu Xenial check. DO NOT USE THIS FOR RELEASED BINARIES.
+    -i  Skip Ubuntu Bionic check. DO NOT USE THIS FOR RELEASED BINARIES.
         Cannot be used together with -d option."
   exit 1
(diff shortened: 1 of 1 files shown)


bianpengyuan:
Fixes https://github.com/istio/istio/issues/18977

bianpengyuan:
Close since binary built with bionic will break VM use case and distroless image.

kyessenov:
@PiotrSikora what are our expectations for the release image? Do we preserve the minimal requirements from Envoy (GCC 7 or Clang/LLVM 7 on Ubuntu 16LTS). The recent moves to clang-9 and libc++ make me think that no longer works.

lizan:
@kyessenov the move of toolchain and standard library doesn't affect where the binary can be run. istio/proxy is already on libc++ since last release.

kyessenov:
I recall that ASAN build requires dynamic linking, so the base does matter.
It's fine if we xenial, but someone needs to make sure all dynamically linked libraries are there.

lizan:
ASAN build is not dynamically linked because it is not intended to ship (run out side or build container). The only dynamic linking library is libc for the release binary.

PiotrSikora:
@kyessenov building on Ubuntu 16.04 is purely about minimizing required glibc version. Otherwise those binaries won't run on CentOS, etc.

@lizan that's not entirely true. ASan is dynamically linked right now (see: https://github.com/envoyproxy/envoy/pull/8555), and we definitely intend to publish ASan Docker images in the foreseeable future.

lizan:
@PiotrSikora let's build another statically linked asan+ubsan image/deb with statically linked libc++, especially for ubsan, there's ubsan-minimal runtime, which is the one for shipping.

PiotrSikora:
@lizan SGTM.

PR #2571, opened 2019-11-16
ci: enable RBE in postfix again

lizan:

**Release note**:

```release-note
None
```

### DIFF
--- a/scripts/release-binary.sh
+++ b/scripts/release-binary.sh
@@ -64,8 +64,10 @@ fi
 
 # Make sure the release binaries are built on x86_64 Ubuntu 16.04 (Xenial)
 if [ "${CHECK}" -eq 1 ]; then
-  UBUNTU_RELEASE=${UBUNTU_RELEASE:-$(lsb_release -c -s)}
-  [[ "${UBUNTU_RELEASE}" == 'xenial' ]] || { echo 'Must run on Ubuntu 16.04 (Xenial).'; exit 1; }
+  if [[ "${BAZEL_BUILD_ARGS}" != *"--config=remote-"* ]]; then
+    UBUNTU_RELEASE=${UBUNTU_RELEASE:-$(lsb_release -c -s)}
+    [[ "${UBUNTU_RELEASE}" == 'xenial' ]] || { echo 'Must run on Ubuntu 16.04 (Xenial).'; exit 1; }
+  fi
   [[ "$(uname -m)" == 'x86_64' ]] || { echo 'Must run on x86_64.'; exit 1; }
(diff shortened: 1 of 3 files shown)


lizan:
I guess this will expose same issue in https://prow.istio.io/view/gcs/istio-prow/logs/proxy-postsubmit/818 in presubmit, so I can debug from there.

lizan:
erm, I guess recent change of testing image solved this @PiotrSikora, let's try RBE in postsubmit again? Bazel shouldn't use remote executor to pull/push image, those are happening locally (in  https://prow.istio.io/view/gcs/istio-prow/logs/proxy-postsubmit/818 it shows path `/home/bootstrap` means it is local, remote is `/b/f/w/...`)

scripts/release-binary.sh
```diff
@@ -64,8 +64,10 @@ fi
 
 # Make sure the release binaries are built on x86_64 Ubuntu 16.04 (Xenial)
 if [ "${CHECK}" -eq 1 ]; then
-  UBUNTU_RELEASE=${UBUNTU_RELEASE:-$(lsb_release -c -s)}
-  [[ "${UBUNTU_RELEASE}" == 'xenial' ]] || { echo 'Must run on Ubuntu 16.04 (Xenial).'; exit 1; }
+  if [[ "${BAZEL_BUILD_ARGS}" != *"--config=remote-"* ]]; then
```
PiotrSikora: Sorry, but this is wrong. Why did we need this change? Per discussion in istio/proxy#2569, we need to keep building binaries on Ubuntu 16.04, and RBE shouldn't be excluded from that, since those are the binaries that we publish.

bianpengyuan: It is prow that runs this script right? I think if rbe is configured, local builder image os version should not matter.

lizan: @PiotrSikora binary linked at RBE is targeted at what glibc image for the RBE image (in this case Xenial) so it doesn't matter what local image is. #2574  is the real fix for RBE postsubmit though.

PiotrSikora: Do we have a way to guard and verify this for RBE images? I really don't want binary requirements to be accidentally "upgraded", as it happens from time to time.

lizan: @PiotrSikora I agree that we should do that. However the point of avoiding this check is that if we're doing RBE build, the binary requirement has nothing to do with the local builder image.

PiotrSikora: @lizan agreed, my point is that instead of checking RBE and not local builder image, we completely disabled this check in RBE builds.

PR #2572, opened 2019-11-16
Update stackdriver config with edges

mandarjog:
Master config has diverged, so this is directly sent to 1.4.

### DIFF
--- a/extensions/stackdriver/testdata/stackdriver_filter.yaml
+++ b/extensions/stackdriver/testdata/stackdriver_filter.yaml
@@ -21,7 +21,7 @@ spec:
             config:
               root_id: stackdriver_outbound
               configuration: |
-                {}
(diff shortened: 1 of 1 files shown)


extensions/stackdriver/testdata/stackdriver_filter.yaml
```diff
@@ -21,7 +21,7 @@ spec:
             config:
               root_id: stackdriver_outbound
               configuration: |
-                {}
+                {"enable_mesh_edges_reporting": false, "meshEdgesReportingDuration": "600s"}
```
bianpengyuan: The default is 600s. No need to explicitly config it.

mandarjog: That way people know which arg to change if needed.
Same reason why we have enable_mesh_edges: false

bianpengyuan: Ah, I see.

PR #2575, opened 2019-11-18
feat(metric): add istio_build metric to prom stats at config time

douglas-reid:
This adds the `istio_build` metric back to the proxy.

This is a revert of #2473, which was a revert of #2431.


douglas-reid:
@bianpengyuan PTAL. I added a test.

bianpengyuan:
Looks like go format fails: https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_proxy/2575/test_proxy/249/build-log.txt

douglas-reid:
@bianpengyuan unfortunately, it doesn't not fail when I run `BUILD_WITH_CONTAINER=1 make lint` so i'm a bit of a loss on how to debug.

bianpengyuan:
Retest passed format checking. Looks like go-fmt is flaky.

PR #2576, opened 2019-11-20
Add option to skip pushing docker image in release-binary

bianpengyuan:

**Release note**:

```release-note
```

### DIFF
--- a/scripts/release-binary.sh
+++ b/scripts/release-binary.sh
@@ -44,14 +44,16 @@ function usage() {
   echo "$0
     -d  The bucket name to store proxy binary (optional).
     -i  Skip Ubuntu Xenial check. DO NOT USE THIS FOR RELEASED BINARIES.
-        Cannot be used together with -d option."
+        Cannot be used together with -d option.
+    -s  Skip pushing envoy docker image."
   exit 1
 }
 
-while getopts d:i arg ; do
+while getopts d:i:s arg ; do
   case "${arg}" in
     d) DST="${OPTARG}";;
     i) CHECK=0;;
+    s) PUSH_DOCKER_IMAGE="";;
     *) usage;;
   esac
 done
@@ -90,7 +92,7 @@ fi
 # k8-dbg is the output directory for -c dbg builds.
 for config in release release-symbol debug
 do
-  PUSH_DOCKER_IMAGE="true"
+  PUSH_DOCKER_IMAGE="${PUSH_DOCKER_IMAGE:-true}"
   case $config in
     "release" )
       CONFIG_PARAMS="--config=release"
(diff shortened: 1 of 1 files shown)


bianpengyuan:
@PiotrSikora Could you help to merge upstream into envoy-wasm? We need https://github.com/envoyproxy/envoy/commit/648cfc3b6c1410d8b27aa33b6fc0392ab230d601 to build..

bianpengyuan:
Then binary won't be pushed to gcs as well. In gke internal build we don't need to push the envoy docker image.

bianpengyuan:
What's the plan of this image? Are we going to build istio-proxy image from it or is it only for testing? I am getting some permission issue when building istio/proxy 1.4 with gke release pipeline because of the image push. If this is no plan to actually use this image, I think we could just add an option to bypass pushing it. If this image could potentially be used in official release, then I probably should fix the internal pipeline.

PiotrSikora:
Ah, OK. It makes sense separating them, then.

You need to sync envoy-wasm first to fix the CI, but please do this in a separate PR. Thanks!

scripts/release-binary.sh
```diff
@@ -44,14 +44,16 @@ function usage() {
...
+    -s  Skip pushing envoy docker image."
   exit 1
 }
 
-while getopts d:i arg ; do
+while getopts d:i:s arg ; do
   case "${arg}" in
     d) DST="${OPTARG}";;
     i) CHECK=0;;
+    s) PUSH_DOCKER_IMAGE="";;
```
PiotrSikora: You might want to switch this to `0`/`1` to match `CHECK`.

bianpengyuan: Done

scripts/release-binary.sh
```diff
@@ -143,7 +150,7 @@ do
     //tools/docker:envoy_distroless \
     //tools/docker:envoy_ubuntu
 
-  if [ -n "${DST}" -a -n "${PUSH_DOCKER_IMAGE}" ]; then
+  if [ -n "${DST}" -a "${PUSH_DOCKER_IMAGE}" -eq 1 ]; then
```
PiotrSikora: This is now leftover from original code, you can drop `-n "${DST}"` since it's completely unrelated to pushing Docker images.

scripts/release-binary.sh
```diff
@@ -40,18 +40,25 @@ DST=""
...
+PUSH_DOCKER_IMAGE=1
+
 function usage() {
   echo "$0
     -d  The bucket name to store proxy binary (optional).
+        If not provided, both envoy binary push and docker image push are skipped.
     -i  Skip Ubuntu Xenial check. DO NOT USE THIS FOR RELEASED BINARIES.
-        Cannot be used together with -d option."
+        Cannot be used together with -d option.
+    -s  Skip pushing envoy docker image. This should only be useful if -d is provided,
```
PiotrSikora: Since pushing to GCS is optional, this should be as well, for consistency.

Could you change this to `-p` (push) and default to false?

PR #2578, opened 2019-11-21
fix bazel calls when using local envoy and --override_repository

clarketm:
For performing **build/test** operations with a **local** envoy source; one configured via the `--override_repository` flag.

Otherwise, bazel will emit errors like:
```console
ERROR: An error occurred during the fetch of repository 'envoy'
```
and 
```console
ERROR: no such package @envoy//...
``` 

on `bazel info` calls (since the *override* flag is unknown to the command) which is problematic for some/most `make` targets that need to be aware of the envoy binary or bazel output locations.

cc @howardjohn @yxue

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -53,7 +53,7 @@ BAZEL_CONFIG_ASAN = --config=macos-asan
 BAZEL_CONFIG_TSAN = # no working config
 endif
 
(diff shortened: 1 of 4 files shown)


clarketm:
Test failures are related to https://github.com/envoyproxy/envoy/pull/9072

PR #2583, opened 2019-11-26
wip

douglas-reid:
This PR only exists to have github build a proxy image that I can use in an istio/istio Docker image, as `BUILD_WITH_CONTAINERS=1` no longer works on my machine.

### DIFF
--- a/extensions/stackdriver/metric/record.cc
+++ b/extensions/stackdriver/metric/record.cc
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+#include "absl/strings/match.h"
+
 #include "extensions/stackdriver/metric/record.h"
 
 #include "extensions/stackdriver/common/constants.h"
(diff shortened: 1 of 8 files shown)


extensions/stackdriver/metric/record.cc
```diff
@@ -25,6 +27,10 @@ namespace Metric {
 void record(bool is_outbound, const ::wasm::common::NodeInfo &local_node_info,
             const ::wasm::common::NodeInfo &peer_node_info,
             const ::Wasm::Common::RequestInfo &request_info) {
+
+  if (absl::StartsWith(request_info.user_agent, "kube-probe") || absl::StartsWith(request_info.user_agent, "Prometheus")) {
```
douglas-reid: @mandarjog @bianpengyuan where should we put this kinda logic in a non-hack branch?

PR #2585, opened 2019-11-27
WIP Initial implementation of authnv2 filter.

incfly:
https://github.com/istio/istio/issues/19077

- Unit test added, no integration yet.
- No config needed for the filter.
- Many clean up compared with old authn filter, consolidate the functionality in single cc library.


incfly:
@yangminzhu @diemtvu 

FYI, I may still working on some polishment, but would like to get some early high level feedback, thanks.

kyessenov:
High-level feedback: why is this not written against Wasm C interface?
You can still compile it in with minimal performance degradation (due to memcopies everywhere).

src/envoy/http/authnv2/filter_test.cc
```diff
@@ -0,0 +1,205 @@
...
+      : request_headers_{{":method", "GET"}, {":path", "/"}},
+        ssl_(std::make_shared<NiceMock<Ssl::MockConnectionInfo>>()),
+        test_time_(),
+        stream_info_(Http::Protocol::Http2, test_time_.timeSystem()) {}
+
+  void initialize(const std::vector<std::string>& jwts) {
+    // Jwt metadata filter input setup.
+    const std::string jwt_name =
+        Extensions::HttpFilters::HttpFilterNames::get().JwtAuthn;
+    auto& metadata = stream_info_.dynamicMetadata();
```
kyessenov: Please stop using dynamic metadata for any performance-critical work (e.g. this seems to be used 100% of time).
It has huge performance overhead. Use filter state instead.

incfly: The purpose of the filter is to feed data into rbac filter.

My understanding is that we'll need to use RBAC filter to `metadata matcher` to enforce the policy, https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/rbac/v2/rbac.proto#envoy-api-msg-config-rbac-v2-rbac

I don't know if the filter state can be shared and supported by rbac filter.

Follow-up question, why the filter state is more performant than metadata?  and what are the rough number?

kyessenov: Got it, I think we'll need to extend RBAC matcher to support filter state. We found that protobuf struct would consume up to 10% of CPU time just to pass a value between two places. It's just very inefficient.

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+  if (connection->ssl() == nullptr ||
+      !connection->ssl()->peerCertificatePresented()) {
+    return false;
+  }
+  Utils::GetPrincipal(connection, true, &peer_principle);
+  if (peer_principle != "") {
+    setKeyValue(authn_data, istio::utils::AttributeName::kSourcePrincipal,
+                peer_principle);
+  }
+  return true;
```
yangminzhu: Probably should return false when `peer_principle == ""`.

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+  // sorted.
+  for (const auto& entry : jwt_metadata.fields()) {
+    const std::string& issuer = entry.first;
+    if (issuer_selected == "" || issuer_selected.compare(issuer) > 0) {
+      issuer_selected = issuer;
+    }
+  }
+  if (issuer_selected != "") {
+    const auto& jwt_entry = jwt_metadata.fields().find(issuer_selected);
+    Protobuf::util::MessageToJsonString(jwt_entry->second.struct_value(),
```
yangminzhu: You convert the struct to string here and later convert the string back to struct in `ProcessJwt`. I think we could use the struct directly in the 2 functions.

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// Helper function to set a key/value pair into Struct.
+static void setKeyValue(::google::protobuf::Struct& data,
+                        const std::string& key, const std::string& value) {
+  (*data.mutable_fields())[key].set_string_value(value);
+}
+
+// Returns true if the attribute populated to authn filter succeeds.
+bool ProcessJwt(const std::string& jwt, ProtobufWkt::Struct& authn_data) {
```
yangminzhu: This function and also the `ExtractStringList` duplicates a lot of the existing logic, can you refactor to reuse the same code?

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+FilterHeadersStatus AuthnV2Filter::decodeHeaders(HeaderMap&, bool) {
+  ENVOY_LOG(debug, "AuthnV2Filter::decodeHeaders start\n");
+  auto& metadata = decoder_callbacks_->streamInfo().dynamicMetadata();
+  auto& authn_data =
+      (*metadata.mutable_filter_metadata())[Utils::IstioFilterName::kAuthnV2];
+
+  // Always try to get principal and set to output if available.
+  ProcessMtls(decoder_callbacks_->connection(), authn_data);
+  std::string jwt_payload = "";
+  std::string issuer = extractJwtFromMetadata(metadata, &jwt_payload);
```
diemtvu: `issuer` is used for logging purpose only. It may be cleaner to move the logging into the function and return `void`.

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+  Envoy::Json::ObjectSharedPtr json_obj;
+  try {
+    json_obj = Json::Factory::loadFromString(jwt);
+    // ENVOY_LOG(debug, "{}: json object is {}", __FUNCTION__,
+    //           json_obj->asJsonString());
+  } catch (...) {
+    return false;
+  }
+
+  ProtobufWkt::Struct claim_structs;
```
diemtvu: We currently have a hard-coded logic to handle [RcToken-type JWT ](https://github.com/istio/proxy/blob/master/src/envoy/http/authn/authenticator_base.cc#L129). Basically, if the token come from `ingress-authorization` header, we extract claims differently. I don't like that logic, and want to see how to do it better in here (we don't have the location information for this filter neither, at least at the moment).

May be leave a TODO / issue and support it in the next PR. Alternatively, check if the payload contain [`original_claims`](https://github.com/istio/proxy/blob/master/src/envoy/http/authn/authn_utils.cc#L36) to switch to the RcToken mode.

src/envoy/http/authnv2/filter.cc
incfly: Duplicate with Authn V1 filter implementation? Discussed with @diemtvu , by the time we release the new authn policy, the old filter code, including this function defined over there will be gone

The time window of co-existence of two filter, should be very short, if I understand plan correctly.

src/envoy/utils/filter_names.cc
```diff
@@ -22,6 +22,7 @@ namespace Utils {
 const char IstioFilterName::kJwt[] = "jwt-auth";
 const char IstioFilterName::kAuthentication[] = "istio_authn";
 const char IstioFilterName::kAlpn[] = "istio.alpn";
+const char IstioFilterName::kAuthnV2[] = "istio.authn_v2";
```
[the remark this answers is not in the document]
incfly: sure. istio.auth is fine to me.

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+  }
+  Utils::GetPrincipal(connection, true, &peer_principle);
+  if (peer_principle != "") {
+    setKeyValue(authn_data, istio::utils::AttributeName::kSourcePrincipal,
+                peer_principle);
+  }
+  return true;
+}
+
+std::string AuthnV2Filter::extractJwtFromMetadata(
```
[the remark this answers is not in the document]
incfly: you mean return value? Issuer is not a field of the class, rather it's a sub string piece of the field from the jwt payload, how can we return the const str ref?

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+  if (claims->find("azp") != claims->end()) {
+    setKeyValue(authn_data, istio::utils::AttributeName::kRequestAuthPresenter,
+                (*claims)["azp"].list_value().values().Get(0).string_value());
+  }
+
+  // request.auth.claims
+  (*(*authn_data
+          .mutable_fields())[istio::utils::AttributeName::kRequestAuthClaims]
+        .mutable_struct_value())
+      .MergeFrom(claim_structs);
```
lambdai: Is there any validation on `claim_structs`? It is originated from json string. I don't know if it is sanitized already

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+  // request.auth.raw_claims
+  setKeyValue(authn_data, istio::utils::AttributeName::kRequestAuthRawClaims,
+              jwt);
+
+  return true;
+}
+
+// Returns true if the peer identity can be extracted from the underying
+// certificate used for mTLS.
+bool ProcessMtls(const Network::Connection* connection,
```
lambdai: const Network::Connection &

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+  if (filter_it == metadata.filter_metadata().end()) {
+    return "";
+  }
+  const ::google::protobuf::Struct& jwt_metadata = filter_it->second;
+  // Iterate over the envoy.jwt metadata, which is indexed by the issuer.
+  // For multiple JWT, we only select on of them, the first one lexically
+  // sorted.
+  for (const auto& entry : jwt_metadata.fields()) {
+    const std::string& issuer = entry.first;
+    if (issuer_selected == "" || issuer_selected.compare(issuer) > 0) {
```
lambdai: Why only `> 0` matters? Could you explain why `< 0` is different?

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+  if (issuer_selected != "") {
+    const auto& jwt_entry = jwt_metadata.fields().find(issuer_selected);
+    Protobuf::util::MessageToJsonString(jwt_entry->second.struct_value(),
+                                        jwt_payload);
+  }
+  return issuer_selected;
+}
+
+FilterHeadersStatus AuthnV2Filter::decodeHeaders(HeaderMap&, bool) {
+  ENVOY_LOG(debug, "AuthnV2Filter::decodeHeaders start\n");
```
lambdai: I think the context is stream. Maybe ENVOY_STREAM_LOG?

src/envoy/http/authnv2/filter.cc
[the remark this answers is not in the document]
incfly: for argument, we do modify, so, using pointer.

incfly:
@kyessenov No particular reason other than this is faster to iterate and ship. For null sandbox, my understanding is that it offers a way out to not having istio/proxy, but  which could takes quite some time AFAIK. I'm fine with that personally, may take some extra time to wire things up though.

@istio/wg-networking-maintainers-data-plane let me know if we feel strong of using WASM API + null sandbox is a **must** to merge at the moment, otherwise I'll assume we can proceed this filter as ALPN rewriting filter.

src/envoy/http/authnv2/filter.h
```diff
@@ -0,0 +1,91 @@
...
+// What's the usage of the original payload?
+namespace Envoy {
+namespace Http {
+namespace Istio {
+namespace AuthN {
+
+class AuthnV2Filter : public StreamDecoderFilter,
+                      public Logger::Loggable<Logger::Id::filter> {
+ public:
+  AuthnV2Filter() {}
```
lambdai: = default

src/envoy/http/authnv2/filter.cc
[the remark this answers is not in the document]
diemtvu: I mean for return. Probably use `absl::string_view` then. Also, per other comment, may be better don't have return at all (void)

src/envoy/http/authnv2/filter_factory.cc
```diff
@@ -0,0 +1,69 @@
...
+  Http::FilterFactoryCb createFilterFactoryFromProto(const Protobuf::Message&,
+                                                     const std::string&,
+                                                     FactoryContext&) override {
+    return createFilterFactory();
+  }
+
+  ProtobufTypes::MessagePtr createEmptyConfigProto() override {
+    ENVOY_LOG(debug, "Called AuthnFilterConfig : {}", __func__);
+    return std::make_unique<Envoy::ProtobufWkt::Empty>();
+  }
```
lambdai: The best practice is to create the the message defined by your Config

src/envoy/http/authnv2/filter.cc
yangminzhu: It doesn't matter how long is the window, reusing the code is to make sure we have the same behavior before and after. See src/envoy/http/authn/authn_utils.cc

src/envoy/http/authnv2/filter.cc
incfly: This is how it's implemented now. 

1. [`getJwtFromEnvoyFilter`](https://github.com/istio/proxy/blob/master/src/envoy/http/authn/filter_context.cc#L84) calling proto to string serialization
1. passing as proto to validate, in [`validateJwt`[https://github.com/istio/proxy/blob/master/src/envoy/http/authn/authenticator_base.cc#L124)

Plus the extract string does have some string split of the keys, https://github.com/istio/proxy/blob/master/src/envoy/http/authn/authn_utils.cc#L48

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+    }
+  } catch (Json::Exception& e) {
+    // Not convertable to string array
+  }
+}
+
+// Helper function to set a key/value pair into Struct.
+static void setKeyValue(::google::protobuf::Struct& data,
+                        const std::string& key, const std::string& value) {
+  (*data.mutable_fields())[key].set_string_value(value);
```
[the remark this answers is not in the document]
incfly: TIL stands for "today I learned"? Or something else? Anything need to address here?

src/envoy/http/authnv2/filter.cc
```diff
@@ -0,0 +1,236 @@
...
+// The key name for the original claims in an exchanged token
+static const std::string kExchangedTokenOriginalPayload = "original_claims";
+
+// Extract JWT claim as a string list.
+// This function only extracts string and string list claims.
+// A string claim is extracted as a string list of 1 item.
+// A string claim with whitespace is extracted as a string list with each
+// sub-string delimited with the whitespace.
+void ExtractStringList(const std::string& key, const Envoy::Json::Object& obj,
+                       std::vector<std::string>* list) {
```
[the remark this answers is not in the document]
incfly: okay, this seems a diff of envoy style compared with google style guide.

src/envoy/http/authnv2/filter.cc
incfly: take another closer look, and can't understand the reason why old impl convert it back and forth...
i'll take a look see if we can use the struct directly without converting back and forth.

src/envoy/http/authnv2/filter.cc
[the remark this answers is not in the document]
lambdai: Yes it is. Thank you!
My major concern is in the main comment: It seems the message has limited field names. Why do you prefer dynamic keys instead of protobuf message with schema?

src/envoy/http/authnv2/filter.cc
yangminzhu: It's a tech debt that should be cleaned from the old impl as well.

incfly:
Per offline discussion, we'll abandon this PR, instead using the old authn filter to express the new Istio level API. Reasons

Authorization configuration on control plane(s) are hardcoded to read the same metadata for `request.xxx` attributes. New filter with new name requires changes on control plane to configure this filter name differently, and the decision on this is based on whether the new authn policy exists, which might not be handy in __some__ case for the control plane.

The old authn filter while having many limitations, not well implemented, but are expressive enough for new authn api.

Thanks everyone for the review.

PR #2591, opened 2019-12-04
expose service port in stats

yxue:


extensions/stats/plugin.h
```diff
@@ -135,6 +136,7 @@ struct IstioDimensions {
   // destination_version="v1",
   // destination_workload="svc01-0-8",
   // destination_workload_namespace="service-graph01",
+  // destination_port = "80",
```
lambdai: I realize the comment is not correct from the regex. Maybe you want to remove the SPACE character

yxue: done

kyessenov:
@duderino @mandarjog @douglas-reid 
Changes like this need to go through P&T review. 
It's not OK to introduce new standard stat dimensions without thinking through the effect on performance of the monitoring pipeline.

lambdai:
@kyessenov Please advice: It's me that approved. When I looked back still see no hazard. 
Could you provide more details in terms of the risk in this PR so that both @yxue and me could do different in the follow up?

kyessenov:
The set of standard metrics is well-documented https://istio.io/docs/reference/config/policy-and-telemetry/metrics/. It's our goal to keep the set minimal and consistent with Mixer, and any changes to the set need to through working group review.

I'll start the revert process.

douglas-reid:
Is there a way that we could establish a set of owners for the extensions to match what we had in istio/istio?

nrjpoddar:
Do we know why "dstination_port" was added?

lambdai:
It is required by an [analyze tool built in istioctl](https://github.com/istio/istio/pull/19375)

nrjpoddar:
> It is required by an [analyze tool built in istioctl](https://github.com/istio/istio/pull/19375)

Do you need service or container port?

lambdai:
> > It is required by an [analyze tool built in istioctl](https://github.com/istio/istio/pull/19375)
> 
> Do you need service or container port?

Sorry I didn't look into the source of truth...

kyessenov:
I still don't think it's OK to change the default metrics in istio to support an experimental UX tool.
@esnible @istio/wg-user-experience-maintainers we would like to revert the change and table it for 1.6 discussion.

douglas-reid:
A proposal for the feature would be a good start. We maybe could use a separate metric for this purpose, too. But in the near term, we should avoid adding new dimensions without a full community process. Metric cardinality is a sensitive topic for Istio users.

PR #2592, opened 2019-12-04
fix(edges): remove unknown defaults

douglas-reid:
Removes the handling of missing metadata from the edges reporter.

Support on the server-side was added to handle missing values. This is no longer needed, and is actually harmful (as it can lead to invalid locations, etc.).

/cc @quentinmit

### DIFF
--- a/extensions/stackdriver/edges/edge_reporter.cc
+++ b/extensions/stackdriver/edges/edge_reporter.cc
@@ -40,39 +40,29 @@ using google::cloud::meshtelemetry::v1alpha1::
     TrafficAssertion_Protocol_PROTOCOL_TCP;
 using google::cloud::meshtelemetry::v1alpha1::WorkloadInstance;
 
-constexpr char kUnknown[] = "unknown";
-
-std::string valueOrUnknown(const std::string& value) {
-  if (value.length() == 0) {
-    return kUnknown;
-  }
-  return value;
-}
-
 namespace {
 void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
                           WorkloadInstance* instance) {
   // TODO(douglas-reid): support more than just kubernetes instances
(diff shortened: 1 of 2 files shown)


extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -40,39 +40,29 @@ using google::cloud::meshtelemetry::v1alpha1::
...
-}
-
 namespace {
 void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
                           WorkloadInstance* instance) {
   // TODO(douglas-reid): support more than just kubernetes instances
-  absl::StrAppend(instance->mutable_uid(), "kubernetes://",
-                  valueOrUnknown(node_info.name()), ".",
-                  valueOrUnknown(node_info.namespace_()));
+  if ((node_info.name().length() > 0) &&
```
kyessenov: nit: `!_.empty()`

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -40,39 +40,29 @@ using google::cloud::meshtelemetry::v1alpha1::
...
+      (node_info.namespace_().length() > 0)) {
+    absl::StrAppend(instance->mutable_uid(), "kubernetes://", node_info.name(),
```
kyessenov: nit: seems like `kubernetes://` should be a constant

kyessenov:
/hold
in case Doug wants to fix nits.

douglas-reid:
@kyessenov looks like i won't get a chance to address them in `master`. I'll try to make the updates for the manual CP and then follow-up in `master`.

kyessenov:
Bot doesn't listen to me because it thinks I don't exist :(

douglas-reid:
On the plus side, maybe you'll be spared in the inevitable bot uprising.

PR #2602, opened 2019-12-12
Add support for Access Log Policy Filter

gargnupur:
Add support for Access Log Policy Filter

This just adds basic support for sampling Istio's access logs based on client-ip and principal values.
It is not doing anything fancy for now and samples based on following rules:
1) If fail -> log
2) If success, check if current client_ip and principal's request have been logged in the configured log window => If yes, then don't log, otherwise log.

Design Doc: https://docs.google.com/document/d/1Vr6ZgkdhV5avm3TmxSF3rjjjN7a_-btnf4ymXqorKCc/edit#

### DIFF
--- a/extensions/access_log_policy/plugin.cc
+++ b/extensions/access_log_policy/plugin.cc
@@ -117,7 +117,8 @@ void PluginContext::onLog() {
   getStringValue({kConnection, kUriSanPeerCertificate}, &source_principal);
   istio_dimensions_.set_downstream_ip(downstream_ip);
   istio_dimensions_.set_source_principal(source_principal);
-
+  GOOGLE_LOG(INFO)<<"downstream_ip: "<<downstream_ip;
+  GOOGLE_LOG(INFO)<<"source_principal: "<<source_principal;
   absl::Time last_log_time_nanos = lastLogTimeNanos();
   auto cur = absl::Now();
   if ((cur - last_log_time_nanos) > logTimeDurationNanos()) {
--- a/extensions/access_log_policy/plugin.cc
+++ b/extensions/access_log_policy/plugin.cc
@@ -103,6 +103,7 @@ void PluginContext::onLog() {
   getValue({"response", "code"}, &response_code);
   // If request is a failure, log it.
   if (response_code != 200) {
+    GOOGLE_LOG(INFO)<<"Reached  here setting filterstate true, response code 200";
     setFilterStateValue(true);
     return;
   }
@@ -120,12 +121,14 @@ void PluginContext::onLog() {
   absl::Time last_log_time_nanos = lastLogTimeNanos();
   auto cur = absl::Now();
   if ((cur - last_log_time_nanos) > logTimeDurationNanos()) {
+    GOOGLE_LOG(INFO)<<"Reached  here setting filterstate true, time to log";
     if (setFilterStateValue(true)) {
       updateLastLogTimeNanos(cur);
     }
     return;
   }
 
+  GOOGLE_LOG(INFO)<<"Reached  here setting filterstate false";
   setFilterStateValue(false);
 }
 
--- a/extensions/access_log_policy/plugin.cc
+++ b/extensions/access_log_policy/plugin.cc
@@ -103,8 +103,6 @@ void PluginContext::onLog() {
   getValue({"response", "code"}, &response_code);
   // If request is a failure, log it.
   if (response_code != 200) {
-    GOOGLE_LOG(INFO)
-        << "Reached  here setting filterstate true, response code 200";
     setFilterStateValue(true);
     return;
   }
@@ -118,19 +116,15 @@ void PluginContext::onLog() {
   getStringValue({kConnection, kUriSanPeerCertificate}, &source_principal);
   istio_dimensions_.set_downstream_ip(downstream_ip);
   istio_dimensions_.set_source_principal(source_principal);
-  GOOGLE_LOG(INFO) << "downstream_ip: " << downstream_ip;
-  GOOGLE_LOG(INFO) << "source_principal: " << source_principal;
   absl::Time last_log_time_nanos = lastLogTimeNanos();
   auto cur = absl::Now();
   if ((cur - last_log_time_nanos) > logTimeDurationNanos()) {
-    GOOGLE_LOG(INFO) << "Reached  here setting filterstate true, time to log";
     if (setFilterStateValue(true)) {
       updateLastLogTimeNanos(cur);
     }
     return;
   }
 
-  GOOGLE_LOG(INFO) << "Reached  here setting filterstate false";
   setFilterStateValue(false);
 }
 
--- a/extensions/access_log_policy/plugin.cc
+++ b/extensions/access_log_policy/plugin.cc
@@ -1,0 +1,138 @@
+/* Copyright 2019 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "extensions/access_log_policy/plugin.h"
+
+#include <google/protobuf/io/zero_copy_stream_impl_lite.h>
+
+#include "absl/strings/str_cat.h"
+#include "absl/strings/str_split.h"
+#include "extensions/common/istio_dimensions.h"
+#include "extensions/common/node_info.pb.h"
+#include "google/protobuf/util/json_util.h"
+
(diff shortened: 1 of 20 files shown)


douglas-reid:
Can we add a short description of how this works (updates filter state, etc.) and why we chose to make this a separate filter instead of configuration inside of the existing stackdriver extension to the PR description?

I think that information will be useful to others looking to extend Istio, etc.

extensions/access_log_policy/plugin.cc
```diff
@@ -0,0 +1,143 @@
...
+    logWarn(toString(r));
+    return false;
+  }
+  return true;
+}
+
+}  // namespace
+
+constexpr long int kDefaultLogWindowDurationNanoseconds =
+    43200000000000;  // 12h
```
kyessenov: Can you use constexpr with absl::Time or chrono instead of hard-coding a constant? It's easier to read that way.

extensions/access_log_policy/plugin.cc
```diff
@@ -0,0 +1,143 @@
...
+  if (config_.has_log_window_duration()) {
+    log_time_duration_nanos_ =
+        ::google::protobuf::util::TimeUtil::DurationToNanoseconds(
+            config_.log_window_duration());
+  } else {
+    log_time_duration_nanos_ = kDefaultLogWindowDurationNanoseconds;
+  }
+
+  // We currently cache the key based on client_ip and principal only.
+  IstioDimensions::InitIstioDimensions({kClientIp, kPrincipal});
```
kyessenov: I can't find this symbol `InitIstioDimensions`. Is this some macro? Is this a singleton? Doesn't it interfere with stats extension?

extensions/access_log_policy/plugin.cc
```diff
@@ -0,0 +1,143 @@
...
+  // and principal combination.
+  std::string downstream_ip = "";
+  getStringValue({kDestination, kAddress}, &downstream_ip);
+  std::string source_principal = "";
+  getStringValue({kConnection, kUriSanPeerCertificate}, &source_principal);
+  IstioDimensions istio_dimension;
+  istio_dimension.map(kClientIp, downstream_ip);
+  istio_dimension.map(kPrincipal, source_principal);
+
+  long int last_log_time_nanos = lastLogTimeNanos(istio_dimension);
```
kyessenov: Is there a reason why you use architecture dependent type? Is this intentional to be sensitive to 32/64-bits?

extensions/common/istio_dimensions.h
```diff
@@ -0,0 +1,101 @@
...
+// This map maintains the name of Istio dimension and their key to
+// IstioDimensions::dimensions map.
+absl::flat_hash_map<const absl::string_view, unsigned int> dimensions_name_map;
+}  // namespace
+
+// A structure that can hold multiple Istio dimensions(metadata variables).
+// This could be use to key caches based on Istio dimensions for various
+// filters.
+// TODO: Add support for evaluating dynamic Istio dimensions.
+struct IstioDimensions {
```
kyessenov: Can this be merged with the one in stats?

extensions/common/istio_dimensions.h
```diff
@@ -0,0 +1,101 @@
...
+    dimensions_name_map.clear();
+    unsigned int i = 0;
+    for (const auto& elem : dimensions_name_set) {
+      dimensions_name_map[elem] = i++;
+    }
+  }
+
+  // maps property with it's value.
+  void map(const absl::string_view name, std::string value) {
+    dimensions[dimensions_name_map[name]] = value;
```
kyessenov: Why not make use of std::tuple? It's going to be much faster than double map lookup.

gargnupur:
@douglas-reid : thanks for the feedback... added design doc to description...

extensions/access_log_policy/plugin.cc
```diff
@@ -0,0 +1,143 @@
...
+constexpr StringView kAddress = "address";
+constexpr StringView kConnection = "connection";
+constexpr StringView kUriSanPeerCertificate = "uri_san_peer_certificate";
+
+static RegisterContextFactory register_AccessLogPolicy(
+    CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContext));
+
+bool PluginRootContext::onConfigure(size_t) {
+  if (::Wasm::Common::TrafficDirection::Outbound ==
+      ::Wasm::Common::getTrafficDirection()) {
```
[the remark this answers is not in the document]
gargnupur: done..

extensions/access_log_policy/plugin.cc
gargnupur: changing everything to absl::Time

extensions/access_log_policy/plugin.cc
gargnupur: It's a static function within istiodimensions.h

extensions/access_log_policy/plugin.cc
gargnupur: not really.. used what we are using in stackdriver extension for edge reporting... changing everything to be absl::time and duration

extensions/common/istio_dimensions.h
gargnupur: This is derived from the stats one... to make it more generics...

extensions/common/istio_dimensions.h
gargnupur: I did double map lookup to reduce memory footprint... am hoping dimensions name map would be really small.. something like 1-25 entries, which will basically contain name of variable as a string, hence making lookup faster.

And IstioDimensions which would actually be a key to the cache just has a numeric name along with string value as compared to a pair of string name and string value making it less in size...

kyessenov: I think the real question is whether all the keys are known at compile time or not. If they are, then a struct should be faster (maps probably take more memory unless they have compiler optimizations to specialize them to structs). If we don't know all the keys, then map does make sense.

gargnupur: Agreed, one of the cases I had in my mind was supporting dynamic expressions and keying based on that, so this cache can be used to sample say on path or source...

kyessenov: Thoughts @mandarjog @douglas-reid @bianpengyuan ?

mandarjog: I think we can start with a struct. Even for dynamic expressions, we will use one that evaluates to a string right? I think it would still be cheaper.

gargnupur: How about we change the dimensions map to a vector, doesn't really need it to be a map.. we can index into it based on dimensions_name_map value.

Reason, am thinking of this is that say if we share this cache with stats filter, then we would end up declaring 20 variables, and even when mostly empty for logs filter, they would still occupy extra 18 bytes in the key of the logs filter cache. The size of maintaining dimensions_name_map in this case would start paying back after around 25-30 entries..

gargnupur: Talked to @mandarjog offline and changes to struct, the way it's in stats plugin

gargnupur:
@kyessenov : this is ready for review again

extensions/access_log_policy/plugin.cc
```diff
@@ -0,0 +1,137 @@
...
+constexpr StringView kConnection = "connection";
+constexpr StringView kUriSanPeerCertificate = "uri_san_peer_certificate";
+
+static RegisterContextFactory register_AccessLogPolicy(
+    CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContext));
+
+bool PluginRootContext::onConfigure(size_t) {
+  if (::Wasm::Common::TrafficDirection::Inbound !=
+      ::Wasm::Common::getTrafficDirection()) {
+    throw EnvoyException("ASM Acess Logging Policy is an inbound filter only.");
```
kyessenov: This won't work in Wasm, since the compiler does not support exceptions. You need to return false instead.

extensions/access_log_policy/plugin.cc
```diff
@@ -0,0 +1,137 @@
...
+    logWarn(toString(r));
+    return false;
+  }
+  return true;
+}
+
+}  // namespace
+
+constexpr absl::Duration kDefaultLogWindowDurationNanoseconds =
+    absl::Hours(12);  // 12h
```
kyessenov: @bianpengyuan mentioned that absl time does not work well in Wasm since it makes system calls.

gargnupur: @kyessenov ., @bianpengyuan : what is the suggested time measurement to use instead?
std::chrono? Previously I was using long int the way stackdriver plugin does for edge reporting : https://github.com/istio/proxy/blob/master/extensions/stackdriver/stackdriver.h#L108:12...

extensions/access_log_policy/plugin.cc
gargnupur: Thanks.. Will fix in a new pr

extensions/access_log_policy/plugin.cc
kyessenov: Both would work. Just don't depend on absl/time to make it run in Wasm.

bianpengyuan: Yes long int would be ideal if you want to compile the filter to wasm.

gargnupur: Thanks.. will change back to long int..
Looks like we are missing some test coverage wherein we should have an ability to be able to run filters within wasm easily..

PR #2605, opened 2019-12-14
Use SimpleAtoi instead of std::atoi to convert interval

bianpengyuan:

**Release note**:

```release-note
```


kyessenov:
We don't have a mock ABI. Probably a good idea to make one.

bianpengyuan:
Yeah, only e2e test could be added for anything involving ABI. I could add an e2e test that does nothing but just send a request and wait for 60s to see if the server receives anything.

PR #2614, opened 2019-12-27
Update authn filter to use first spiffe san

prabushyam:
Updates authN filter to loop over the cert sans and use the first one **with the prefix** `spiffe://` as the `source.principal`

**Which issue this PR fixes** : fixes https://github.com/istio/istio/issues/19767

**Release note**:

```release-note
AuthN filter looks for the first san with `spiffe://` prefix
```

### DIFF
--- a/src/envoy/utils/utils_test.cc
+++ b/src/envoy/utils/utils_test.cc
@@ -128,6 +128,16 @@ TEST_P(UtilsTest, GetPrincipalNoSpiffePrefix) {
   testGetPrincipal(sans, "spiffe:foo/bar", true);
 }
 
+TEST_P(UtilsTest, GetPrincipalFromSpiffePrefixSAN) {
+  std::vector<std::string> sans{"bad", "spiffe://foo/bar"};
+  testGetPrincipal(sans, "foo/bar", true);
+}
+
+TEST_P(UtilsTest, GetPrincipalFromNonSpiffePrefixSAN) {
+  std::vector<std::string> sans{"foobar", "xyz"};
+  testGetPrincipal(sans, "foobar", true);
+}
+
 TEST_P(UtilsTest, GetPrincipalEmpty) {
   std::vector<std::string> sans;
   testGetPrincipal(sans, "", false);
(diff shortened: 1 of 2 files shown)


src/envoy/utils/utils_test.cc
```diff
@@ -128,6 +128,11 @@ TEST_P(UtilsTest, GetPrincipalNoSpiffePrefix) {
   testGetPrincipal(sans, "spiffe:foo/bar", true);
 }
 
+TEST_P(UtilsTest, GetPrincipalFromSpiffePrefixSAN) {
```
diemtvu: Can you also add a test when both sans are non-spiffe.

src/envoy/utils/utils_test.cc
```diff
@@ -128,6 +128,11 @@ TEST_P(UtilsTest, GetPrincipalNoSpiffePrefix) {
...
+  std::vector<std::string> sans{"bad", "spiffe://foo/bar"};
+  testGetPrincipal(sans, "foo/bar", true);
+}
+
```
diemtvu: For completeness,  can you also add similar test cases for GetTrustDomain below.

src/envoy/utils/utils_test.cc
prabushyam: Thanks, added

src/envoy/utils/utils_test.cc
prabushyam: Added tests for GetTrustDomain as well, thank you.

prabushyam:
@diemtvu @yangminzhu  If this is good to go, could we get this PR merged?

bianpengyuan:
Looks like macos job is stuck. Would you mind to close/re-open or recreate the pr? It happened before and what's weird is that it happens to some particular github id consistently. So it still does not work after reopen/recreation, I could create the same pr and merge that if you are ok with it.

prabushyam:
@bianpengyuan I thought the macos build required some other approval labels. Please feel free to close this PR and recreate it. Thank you

PR #2615, opened 2019-12-30
Extract request info to a class

bianpengyuan:
Extract out request info to a separate class, which lazily load request attributes. It uses a request info proto to cache the attributes that are already fetched from host. This is to avoid unnecessary copy if some filters want to use a subset of attributes. This change also added some attributes that are useful for access logging. The request info protocol does not schematize any thing. It is just an internal proto to keep track of fetched attributes.

To minimize the diff, this code change has some duplicated code from context.cc. They will be cleaned up when real usage is added. Usage of this class is like https://github.com/istio/proxy/compare/master...bianpengyuan:n

### DIFF
--- a/extensions/common/request_info.cc
+++ b/extensions/common/request_info.cc
@@ -1,0 +1,459 @@
+/* Copyright 2019 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License") {}
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "extensions/common/request_info.h"
+
+#include "extensions/common/util.h"
+
+// WASM_PROLOG
+#ifndef NULL_PLUGIN
+#include "proxy_wasm_intrinsics.h"
+
+#else  // NULL_PLUGIN
+
+#include "extensions/common/wasm/null/null_plugin.h"
+
+using Envoy::Extensions::Common::Wasm::HeaderMapType;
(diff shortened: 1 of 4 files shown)


bianpengyuan:
@mandarjog @kyessenov May I get some feedback on this? This is more like an intermediate state before CEL is added. Without this, log will be useless since not enough information is fetched from host.

bianpengyuan:
https://github.com/istio/istio/issues/19705

extensions/common/request_info.cc
```diff
@@ -0,0 +1,459 @@
...
+const char kInboundPassthroughClusterIpv4[] = "InboundPassthroughClusterIpv4";
+const char kInboundPassthroughClusterIpv6[] = "InboundPassthroughClusterIpv6";
+const char B3TraceID[] = "x-b3-traceid";
+const char B3SpanID[] = "x-b3-spanid";
+const char B3TraceSampled[] = "x-b3-sampled";
+
+namespace {
+
+// Extract service name from service host.
+void extractServiceName(const std::string& host,
```
mandarjog: Is this new code or only moved around?

extensions/common/request_info.cc
```diff
@@ -0,0 +1,459 @@
...
+const std::string& RequestInfoImpl::userAgent() {
+  if (!request_info_.has_user_agent()) {
+    getStringValue({"request", "user_agent"},
+                   request_info_.mutable_user_agent()->mutable_value());
+  }
+  return request_info_.user_agent().value();
+}
+
+const std::string& RequestInfoImpl::urlPath() {
+  if (!request_info_.has_url_path()) {
```
mandarjog: What is this call fails?

extensions/common/request_info.cc
```diff
@@ -0,0 +1,459 @@
...
+    *service_name = host;
+    return;
+  }
+  if (host[name_pos] == ':') {
+    // host is `short_service:port`, return short_service name.
+    *service_name = host.substr(0, name_pos);
+    return;
+  }
+
+  auto namespace_pos = host.find_first_of(".:", name_pos + 1);
```
mandarjog: Can you add low level unit tests for request_info.cc?

Which part specifically fix the SD log correlation issues?

bianpengyuan:
Superseded by https://github.com/istio/proxy/pull/2625

PR #2617, opened 2019-12-31
Update authn filter to use first spiffe san

bianpengyuan:
redo https://github.com/istio/proxy/pull/2614.

cc @prabushyam

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/utils/utils.cc
+++ b/src/envoy/utils/utils.cc
@@ -36,6 +36,10 @@ const std::string kPerHostMetadataKey("istio");
 // Attribute field for per-host data override
 const std::string kMetadataDestinationUID("uid");
 
+bool hasSPIFFEPrefix(const std::string& san) {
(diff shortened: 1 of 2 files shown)


prabushyam:
thanks @bianpengyuan 👍

src/envoy/utils/utils.cc
```diff
@@ -47,16 +51,21 @@ bool getCertSAN(const Network::Connection* connection, bool peer,
...
+      // return the first san if no sans have the spiffe:// prefix
       *principal = sans[0];
       return true;
     }
   }
   return false;
 }
 
-bool hasSPIFFEPrefix(const std::string& san) {
-  return absl::StartsWith(san, kSPIFFEPrefix);
```
yxue: Is this the same as the change from line 39-42?

bianpengyuan: Yeah it's the same. I think it will cause build error without moving it above.

PR #2621, opened 2020-01-08
Check generated wasm file at presubmit

bianpengyuan:
Add an script to build and push wasmsdk container, and compile wasm filter with that. This also adds a test for the generated wasm file. Right now only mx filter wasm file is being generated. This will be added to prow presubmit with https://github.com/istio/test-infra/pull/2260 as a non-required job.


bianpengyuan:
cc @jplevyak fyi

bianpengyuan:
This is ready for review. Please take a look. After this is merged, I will create another PR to push wasm files to gcs in post submit.

bianpengyuan:
@PiotrSikora @jplevyak Could you please help reviewing this PR? Thank you!

extensions/access_log_policy/Makefile
```diff
@@ -1,17 +0,0 @@
-DOCKER_SDK=/sdk
```
PiotrSikora: Why is this removed?

bianpengyuan: This filter uses flat hash map which cannot be directly compiled to wasm. I've confirmed with @gargnupur and @mandarjog, it is not a priority to compile this filter into wasm, so remove it from wasm compilation target for now.

test/envoye2e/stats/stats_xds_test.go
```diff
@@ -233,3 +239,44 @@ func TestStatsParallel(t *testing.T) {
...
+	params := &driver.Params{
+		Vars: map[string]string{
+			"ClientPort":                 fmt.Sprintf("%d", ports.AppToClientProxyPort),
+			"BackendPort":                fmt.Sprintf("%d", ports.BackendPort),
+			"ClientAdmin":                fmt.Sprintf("%d", ports.ClientAdminPort),
+			"ServerAdmin":                fmt.Sprintf("%d", ports.ServerAdminPort),
+			"ServerPort":                 fmt.Sprintf("%d", ports.ClientToServerProxyPort),
+			"RequestCount":               "10",
+			"MetadataExchangeFilterCode": "filename: extensions/metadata_exchange/plugin.wasm",
+			"StatsFilterCode":            "filename: extensions/stats/plugin.wasm",
```
kyessenov: nit: could use go sub-tests here.

bianpengyuan: Will clean up in the following PR.

PR #2624, opened 2020-01-14
feat(stats): support grpc status codes in metrics

douglas-reid:
Companion PR to https://github.com/istio/istio/pull/20036.

> This PR implements the changes approved in https://tinyurl.com/yy2cwmv5 to report gRPC status codes in metrics.

TODO:
- [x] update envoywasm to provide `response`.`grpc_status` (https://sourcegraph.com/github.com/envoyproxy/envoy-wasm@542558a722e23225cb1f6e2c8194c093f993f9c8/-/blob/source/extensions/filters/common/expr/context.cc#L124-144)
- [ ] update logs to match other PR
- [x] add grpc-echo client/server for testing
- [ ] update tags in istio/istio to match


PiotrSikora:
@douglas-reid do you want to redo it now on top of #2631 in order to get it into 1.5?

douglas-reid:
yeah, i'll need to rebase, etc.

PiotrSikora:
@douglas-reid it's merged now.

douglas-reid:
@kyessenov @mandarjog please take a look at this PR now.

There are a few important parts, but mostly, a decision is needed regarding the default behavior of `grpc_response_status` label for non-gRPC requests.

The original proposal design doc was under-specified in this regard.

In this PR, if the protocol is not gRPC, then the fallback will be the `UNKNOWN` grpc status code (`2`). This matches the behavior for translating `HTTP OK` -> gRPC codes (doc).

From [http-grpc-status-mapping](https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md):

> 200 is UNKNOWN because there should be a grpc-status in case of truly OK response.

The [spec](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md) for `UNKNOWN`, includes:

> Unknown error. For example, this error may be returned when a Status value received from another address space belongs to an error space that is not known in this address space. Also errors raised by APIs that do not return enough error information may be converted to this error.

as well as

> Error parsing returned status | UNKNOWN | Client

We had previously discussed the following alternatives:
1. Returning `1` (gRPC `OK`)
1. Returning an empty label.

I rejected returning `gRPC OK`, as that conflicts with the mapping.

Returning an empty label is slightly clunky, as the code gets much more complicated, as do the regexes for generating prometheus labels and the weirdness of having non-populated labels for non-gRPC metrics.

Thoughts?

Note: this PR is blocked on a cherry-pick of @kyessenov's hardening PR of upstream Envoy.

douglas-reid:
Update: After thinking about this again, I switched to `grpc_response_status: ""` (empty string) for non-gRPC protocol cases. Please let me know what you think.

src/envoy/http/alpn/alpn_test.cc
```diff
@@ -25,6 +25,7 @@ using istio::envoy::config::filter::http::alpn::v2alpha1::
     FilterConfig_AlpnOverride;
 using testing::NiceMock;
 using testing::Return;
+using testing::ReturnPointee;
```
kyessenov: nit: not used

douglas-reid:
@kyessenov agreed on the templates. deprecation will hit us hard.

PR #2625, opened 2020-01-15
Add extra labels to sd logging

bianpengyuan:
istio/istio#19705


extensions/common/context.cc
```diff
@@ -294,6 +294,27 @@ void populateHTTPRequestInfo(bool outbound, bool use_host_header_fallback,
...
 
+void populateExtendedHTTPRequestInfo(RequestInfo* request_info) {
+  getValue({"source", "address"}, &request_info->source_address);
+  getValue({"destination", "address"}, &request_info->destination_address);
+
+  getValue({"request", "referer"}, &request_info->referer);
+  getValue({"request", "user_agent"}, &request_info->user_agent);
+  getValue({"request", "id"}, &request_info->request_id);
+  std::string trace_sampled;
+  getValue({"request", "headers", "x-b3-sampled"}, &trace_sampled);
```
mandarjog: Does getValue return a bool or similar ?

testdata/stackdriver/gateway_access_log.yaml.tmpl
```diff
@@ -1,20 +1,28 @@
...
+- http_request:
+    request_method: "GET"
+    request_url: "http://127.0.0.1:{{ .Vars.ClientPort }}/"
+    server_ip: "127.0.0.1:{{ .Vars.ServerPort }}"
+    protocol: "http"
+    status: 200
+  labels:
     destination_principal: ""
     destination_service_host: server.default.svc.cluster.local
-    protocol: http
```
mandarjog: Protocol changed from label to http_request, is that correct?

extensions/stackdriver/log/logger.cc
```diff
@@ -72,6 +71,16 @@ Logger::Logger(const ::wasm::common::NodeInfo& local_node_info,
...
   (*label_map)["mesh_uid"] = local_node_info.mesh_id();
+  // Add destination app and version label if exist.
+  const auto& local_labels = local_node_info.labels();
+  auto version_iter = local_labels.find("version");
+  if (version_iter != local_labels.end()) {
+    (*label_map)["destination_version"] = version_iter->second;
+  }
+  auto app_iter = local_labels.find("app");
+  if (app_iter != local_labels.end()) {
+    (*label_map)["destination_app"] = app_iter->second;
```
mandarjog: Note somewhere that this is needed by stack driver to correlate with platform logs.

mandarjog:
@gargnupur please review

extensions/stackdriver/log/logger.cc
bianpengyuan: Done. add a log about it.

testdata/stackdriver/gateway_access_log.yaml.tmpl
bianpengyuan: Yeah, I moved it to http request. Since it already exists there and I think it makes sense to just use that field.

extensions/common/context.cc
bianpengyuan: Yes, good point. Added it to the condition.

extensions/stackdriver/log/logger.cc
```diff
@@ -87,20 +97,55 @@ void Logger::addLogEntry(const ::Wasm::Common::RequestInfo& request_info,
...
+  auto app_iter = peer_labels.find("app");
+  if (app_iter != peer_labels.end()) {
+    (*label_map)["source_app"] = app_iter->second;
+  }
 
-  (*label_map)["request_operation"] = request_info.request_operation;
   (*label_map)["destination_service_host"] =
       request_info.destination_service_host;
   (*label_map)["response_flag"] = request_info.response_flag;
-  (*label_map)["protocol"] = request_info.request_protocol;
```
gargnupur: I think we should keep this?

extensions/stackdriver/stackdriver.cc
```diff
@@ -211,6 +211,7 @@ void StackdriverRootContext::record() {
   ::Extensions::Stackdriver::Metric::record(isOutbound(), local_node_info_,
                                             peer_node_info, request_info);
   if (enableServerAccessLog() && shouldLogThisRequest()) {
+    ::Wasm::Common::populateExtendedHTTPRequestInfo(&request_info);
```
gargnupur: Can this be called for TCP?

extensions/stackdriver/log/logger.cc
bianpengyuan: I moved it to http request. Since it already exists there and I think it makes sense to just use that field.

extensions/stackdriver/stackdriver.cc
bianpengyuan: Most information populated in this function are http based, except source/dst addresses. If it is desired, we could add those two fields into populateTCPRequestInfo.

bianpengyuan:
@lambdai @yxue Could you help me review this? Thanks!

extensions/common/context.cc
```diff
@@ -294,6 +294,27 @@ void populateHTTPRequestInfo(bool outbound, bool use_host_header_fallback,
...
+void populateExtendedHTTPRequestInfo(RequestInfo* request_info) {
+  getValue({"source", "address"}, &request_info->source_address);
+  getValue({"destination", "address"}, &request_info->destination_address);
+
+  getValue({"request", "referer"}, &request_info->referer);
+  getValue({"request", "user_agent"}, &request_info->user_agent);
+  getValue({"request", "id"}, &request_info->request_id);
+  std::string trace_sampled;
+  if (getValue({"request", "headers", "x-b3-sampled"}, &trace_sampled) &&
+      !trace_sampled.empty() && trace_sampled != "0") {
```
yxue: `trace_sampled != "0"` seems weird to me

mandarjog: It is a string, and you can disable sampling by setting it to “0”

mandarjog: Though a nit: we could just check for ==“1” and not worry about empty.

bianpengyuan: https://github.com/istio/proxy/pull/2627

PR #2638, opened 2020-01-28
Add sts to sd filter

bianpengyuan:

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stackdriver/common/utils.cc
+++ b/extensions/stackdriver/common/utils.cc
@@ -71,7 +71,7 @@ void setSTSCallCredentialOptions(
   if (!sts_service) {
     return;
   }
-  sts_service->set_token_exchange_service_uri("http://127.0.0.1:" + sts_port +
+  sts_service->set_token_exchange_service_uri("http://localhost:" + sts_port +
                                               "/token");
   sts_service->set_subject_token_path(kSTSSubjectTokenPath);
   sts_service->set_subject_token_type(kSTSSubjectTokenType);
@@ -82,7 +82,7 @@ void setSTSCallCredentialOptions(
     ::grpc::experimental::StsCredentialsOptions *sts_options,
     const std::string &sts_port) {
   sts_options->token_exchange_service_uri =
-      "http://127.0.0.1:" + sts_port + "/token";
+      "http://localhost:" + sts_port + "/token";
   sts_options->subject_token_path = kSTSSubjectTokenPath;
   sts_options->subject_token_type = kSTSSubjectTokenType;
   sts_options->scope = kSTSScope;
--- a/extensions/stackdriver/common/utils.cc
+++ b/extensions/stackdriver/common/utils.cc
@@ -81,6 +81,9 @@ void setSTSCallCredentialOptions(
 void setSTSCallCredentialOptions(
     ::grpc::experimental::StsCredentialsOptions *sts_options,
     const std::string &sts_port) {
+  if (!sts_options) {
+    return;
+  }
   sts_options->token_exchange_service_uri =
       "http://localhost:" + sts_port + "/token";
   sts_options->subject_token_path = kSTSSubjectTokenPath;
--- a/extensions/stackdriver/common/utils.cc
+++ b/extensions/stackdriver/common/utils.cc
@@ -63,6 +63,20 @@ void getMonitoredResource(const std::string &monitored_resource_type,
   }
 }
 
+void setSTSService(
+    ::envoy::config::core::v3::GrpcService_GoogleGrpc_CallCredentials_StsService
+        *sts_service,
+    const std::string &sts_port) {
+  if (!sts_service) {
+    return;
+  }
+  sts_service->set_token_exchange_service_uri("http://127.0.0.1:" + sts_port +
+                                              "/token");
+  sts_service->set_subject_token_path("/var/run/secrets/tokens/istio-token");
+  sts_service->set_subject_token_type("urn:ietf:params:oauth:token-type:jwt");
+  sts_service->set_scope("https://www.googleapis.com/auth/cloud-platform");
+}
+
 }  // namespace Common
 }  // namespace Stackdriver
 }  // namespace Extensions
--- a/extensions/stackdriver/common/utils.cc
+++ b/extensions/stackdriver/common/utils.cc
@@ -16,6 +16,7 @@
 #include "extensions/stackdriver/common/utils.h"
 
 #include "extensions/stackdriver/common/constants.h"
+#include "grpcpp/grpcpp.h"
 
 namespace Extensions {
 namespace Stackdriver {
@@ -63,7 +64,7 @@ void getMonitoredResource(const std::string &monitored_resource_type,
   }
 }
 
-void setSTSService(
+void setSTSCallCredentialOptions(
     ::envoy::config::core::v3::GrpcService_GoogleGrpc_CallCredentials_StsService
         *sts_service,
     const std::string &sts_port) {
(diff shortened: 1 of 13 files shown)


bianpengyuan:
@JimmyCYJ ptal. This needs to work with https://github.com/istio/istio/pull/20595.

@mandarjog @kyessenov fyi

extensions/stackdriver/log/exporter.cc
```diff
@@ -70,9 +71,22 @@ ExporterImpl::ExporterImpl(RootContext* root_context,
...
+      grpc_service.mutable_google_grpc()
+          ->add_call_credentials()
+          ->mutable_google_compute_engine();
+    } else {
+      auto* sts_service = grpc_service.mutable_google_grpc()
+                              ->add_call_credentials()
+                              ->mutable_sts_service();
+      sts_service->set_token_exchange_service_uri(
+          "http://127.0.0.1:" + sts_port + "/token");
+      sts_service->set_subject_token_path("/tmp/sts-envoy-token.jwt");
```
JimmyCYJ: The token path is /var/run/secrets/tokens/istio-token https://github.com/istio/istio/blob/b28b5a0b2749c2ad8a50c82826860a3183653d9c/pilot/pkg/security/model/authentication.go#L46

extensions/stackdriver/metric/registry.cc
```diff
@@ -42,6 +44,20 @@ StackdriverOptions getStackdriverOptions(
                                        grpc::InsecureChannelCredentials());
     options.metric_service_stub =
         google::monitoring::v3::MetricService::NewStub(channel);
+  } else if (!sts_port.empty()) {
+    ::grpc::experimental::StsCredentialsOptions sts_options;
+    sts_options.token_exchange_service_uri =
+        "http://127.0.0.1:" + sts_port + "/token";
+    sts_options.subject_token_path = "/tmp/sts-envoy-token.jwt";
```
JimmyCYJ: ditto

extensions/stackdriver/stackdriver.cc
```diff
@@ -107,6 +107,16 @@ int getExportInterval() {
   return 60;
 }
 
+// Get port of security token exchange server from node metadata, if not
+// provided, emtpy will be returned.
+std::string getSTSPort() {
+  std::string sts_port = "";
+  if (getValue({"node", "metadata", kSTSPortKey}, &sts_port)) {
```
JimmyCYJ: My PR which adds STS flag is here. We Pilot-agent needs to check this flag and add metadata into bootstrap config.

bianpengyuan: ptal at istio/istio#20595

extensions/stackdriver/log/exporter.cc
bianpengyuan: Done.

extensions/stackdriver/metric/registry.cc
bianpengyuan: Done.

bianpengyuan:
@JimmyCYJ @mandarjog @kyessenov this is ready for review. ptal. thanks!

extensions/stackdriver/log/exporter.cc
```diff
@@ -70,9 +71,23 @@ ExporterImpl::ExporterImpl(RootContext* root_context,
   if (logging_service_endpoint.empty()) {
     grpc_service.mutable_google_grpc()->set_target_uri(
         kGoogleStackdriverLoggingAddress);
-    grpc_service.mutable_google_grpc()
-        ->add_call_credentials()
-        ->mutable_google_compute_engine();
+    if (sts_port.empty()) {
```
mandarjog: Nit: just extract this out as “configure_auth” and use it in 2 places.

mandarjog: The PR has no tests, can we add something here?

bianpengyuan: Extracted the logic to common util. I am not sure how can I test the stub effectively.

extensions/stackdriver/common/utils.cc
```diff
@@ -63,6 +63,20 @@ void getMonitoredResource(const std::string &monitored_resource_type,
...
 }
 
+void setSTSService(
+    ::envoy::config::core::v3::GrpcService_GoogleGrpc_CallCredentials_StsService
+        *sts_service,
+    const std::string &sts_port) {
+  if (!sts_service) {
+    return;
+  }
+  sts_service->set_token_exchange_service_uri("http://127.0.0.1:" + sts_port +
```
kyessenov: nit: are we OK with hard-coding IPv4 localhost here? this won't work on IPv6 machines.
Is there a reason for allowing only port changes?

bianpengyuan: @JimmyCYJ ^^^

extensions/stackdriver/metric/registry.cc
```diff
@@ -42,6 +44,20 @@ StackdriverOptions getStackdriverOptions(
                                        grpc::InsecureChannelCredentials());
     options.metric_service_stub =
         google::monitoring::v3::MetricService::NewStub(channel);
+  } else if (!sts_port.empty()) {
+    ::grpc::experimental::StsCredentialsOptions sts_options;
+    sts_options.token_exchange_service_uri =
+        "http://127.0.0.1:" + sts_port + "/token";
+    sts_options.subject_token_path = "/var/run/secrets/tokens/istio-token";
+    sts_options.subject_token_type = "urn:ietf:params:oauth:token-type:jwt";
+    sts_options.scope = "https://www.googleapis.com/auth/cloud-platform";
```
kyessenov: since these values are used in two places and not obvious, either make constants or consolidate the two blocks in one file.

bianpengyuan: Moved constant strings to one place.

extensions/stackdriver/common/utils.cc
bianpengyuan: @JimmyCYJ mind shed some lights on this?

JimmyCYJ: Yes, the server side could listen on [IPv6 address](https://github.com/istio/istio/blob/25ac96709925ecceb9ab79b6a325d031e6100733/pilot/cmd/pilot-agent/main.go#L512), is there a way to detect whether the Proxy is using IPv4 or IPv6 like Pilot-agent?

kyessenov: If gRPC supports name lookups, then hopefully `localhost` could be used. Either way, we can't hardcode 127.0.0.1 I think.

bianpengyuan: Switch to localhost. Right now at master and 1.5, sts server is running as within pilot-agent, and the only exposed option is port number. So configurability of other parts is not quite necessary.

bianpengyuan:
@JimmyCYJ @kyessenov may I get your review on this? thanks!

extensions/stackdriver/common/utils.cc
```diff
@@ -63,6 +64,30 @@ void getMonitoredResource(const std::string &monitored_resource_type,
...
+  sts_service->set_token_exchange_service_uri("http://localhost:" + sts_port +
+                                              "/token");
+  sts_service->set_subject_token_path(kSTSSubjectTokenPath);
+  sts_service->set_subject_token_type(kSTSSubjectTokenType);
+  sts_service->set_scope(kSTSScope);
+}
+
+void setSTSCallCredentialOptions(
+    ::grpc::experimental::StsCredentialsOptions *sts_options,
+    const std::string &sts_port) {
```
JimmyCYJ: Do we need to check if sts_options is nullptr?

extensions/stackdriver/log/exporter.cc
JimmyCYJ: We only turn on STS server when port is > 0, can we verify it here as well?

extensions/stackdriver/metric/registry.cc
```diff
@@ -42,6 +44,17 @@ StackdriverOptions getStackdriverOptions(
                                        grpc::InsecureChannelCredentials());
     options.metric_service_stub =
         google::monitoring::v3::MetricService::NewStub(channel);
+  } else if (!sts_port.empty()) {
```
JimmyCYJ: ditto

extensions/stackdriver/stackdriver.cc
```diff
@@ -107,6 +107,16 @@ int getExportInterval() {
   return 60;
 }
 
+// Get port of security token exchange server from node metadata, if not
+// provided, emtpy will be returned.
+std::string getSTSPort() {
+  std::string sts_port;
+  if (getValue({"node", "metadata", kSTSPortKey}, &sts_port)) {
+    return sts_port;
```
JimmyCYJ: We can also verify here, if sts_port is "0" then return "".

PR #2655, opened 2020-02-05
configurable metrics (part 2): refactor dimensions to be a vector, permit overriding

kyessenov:
To be stacked upon #2640 
- [x] change dimensions to be a vector
- [x] add separate tags to each metric family and tag-to-vector index mapping
- [x] add overrides 
- [x] integration test
- [x] test for error cases
- [x] add metric definition


kyessenov:
@gargnupur @mandarjog @bianpengyuan This is ready for review.
Two issues to be resolved:
- ABI is not returning errors, to be fixed by @jplevyak 
- creating new metrics dynamically not yet implemented

extensions/stats/plugin.cc
```diff
@@ -55,45 +121,66 @@ void clearTcpMetrics(::Wasm::Common::RequestInfo& request_info) {
...
 
   // Seed the common metric tags with the default set.
-  tags_ = IstioDimensions::defaultTags();
-
-  // Process the dimension overrides
-  for (const auto& metric : config_.metrics()) {
-    if (metric.dimensions().empty()) {
-      continue;
+  std::vector<MetricFactory> factories = DefaultMetrics();
+  Map<std::string, std::vector<MetricTag>> metric_tags;
```
gargnupur: can string here be const ref?

extensions/stats/plugin.h
```diff
@@ -316,33 +151,115 @@ class SimpleStat {
...
-  inline StringView name() const { return name_; };
-  inline bool is_tcp_metric() const { return is_tcp_metric_; }
+  inline StringView name() const { return metric_.name; };
+  inline bool is_tcp_metric() const { return is_tcp_; }
 
   // Resolve metric based on provided dimension values.
-  SimpleStat resolve(std::vector<std::string>& vals) {
-    auto metric_id = metric_.resolveWithFields(vals);
-    return SimpleStat(metric_id, value_fn_);
+  SimpleStat resolve(const IstioDimensions& instance) {
```
gargnupur: Can we add comments to this function?

mandarjog:
@kyessenov the other PR is not cherry-picked into 1.5 yet, are you going to just do 1 larger PR in 1.5 ?

extensions/stats/plugin.cc
kyessenov: STD container don't like references, right? I think you cannot use a reference as key type. In any case, this is a temporary map as it gets destroyed on exit.

extensions/stats/plugin.h
kyessenov: Sure. The code is copy-pasted from ABI to avoid creating a temporary vector just to use the lower-level call.

kyessenov:
Updated with the ABI change and added error test cases.
@mandarjog Yeah, I want to cherry pick both together in a PR.

kyessenov:
Tests fail because of #2659

kyessenov:
This is finishing the entire proposal (except for conditional matches, since that is kind of complicated and can be pushed into expressions if necessary).

extensions/stats/plugin.cc
```diff
@@ -45,6 +45,76 @@ constexpr long long kDefaultTCPReportDurationNanoseconds = 15000000000;  // 15s
...
+}
+
+// maps peer_node and request to dimensions.
+void map(IstioDimensions& instance, bool outbound,
+         const wasm::common::NodeInfo& peer_node,
+         const ::Wasm::Common::RequestInfo& request) {
+  map_peer(instance, outbound, peer_node);
+  map_request(instance, request);
+  map_unknown_if_empty(instance);
+  if (request.request_protocol == "grpc") {
```
gargnupur: nit: can use kProtocolGRPC constant here..

extensions/stats/plugin.cc
gargnupur: ahh okk it just allows const but no ref.. thanks!

PR #2658, opened 2020-02-06
feat(stats): add support for canonical service labels

douglas-reid:

This PR adds support to the `stats` extension for reporting `{source|destination}_canonical_service` labels to Istio metrics.

This is part of the approved proposal for Istio Canonical Service.  Note: the actual stats config in istio/istio bootstrap will need to be updated to match.


test/envoye2e/http_metadata_exchange/testoutput/client.yaml
```diff
@@ -102,7 +106,7 @@ static_resources:
             address:
               socket_address:
                 address: 127.0.0.1
-                port_value: 20083
+                port_value: 20103
```
kyessenov: What is this change for?

douglas-reid: ??? generated file. Got changed by running that test.

PR #2669, opened 2020-02-11
feat(edges): add support for reporting two batches (new vs full)

douglas-reid:
The contract for meshtelemetry is ~1m reporting of new edges, followed by a ~10m reporting of all seen edges.

This PR attempts to adhere closely to that contract. Of note, it repurposes the configurable duration in the proto to handle intra-epoch reporting duration.

### DIFF
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -59,8 +59,7 @@ using ::Wasm::Common::RequestInfo;
 
 constexpr char kStackdriverExporter[] = "stackdriver_exporter";
 constexpr char kExporterRegistered[] = "registered";
-constexpr int kDefaultLogExportMilliseconds = 10000;                     // 10s
-constexpr long int kDefaultEdgeReportDurationNanoseconds = 60000000000;  // 1m
+constexpr int kDefaultLogExportMilliseconds = 10000;  // 10s
 
 namespace {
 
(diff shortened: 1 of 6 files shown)


extensions/stackdriver/stackdriver.h
```diff
@@ -105,7 +108,10 @@ class StackdriverRootContext : public RootContext {
 
   long int last_edge_report_call_nanos_ = 0;
 
-  long int edge_report_duration_nanos_;
+  long int edge_new_report_duration_nanos_;
```
bianpengyuan: nit: initialize it with the 1m? Even it will definitely be set during on onConfigure, but it would still be nice to give it a default value to prevent random behavior because of future change.

extensions/stackdriver/stackdriver.cc
```diff
@@ -206,8 +206,15 @@ void StackdriverRootContext::onTick() {
   }
   if (enableEdgeReporting()) {
     auto cur = static_cast<long int>(getCurrentTimeNanoseconds());
-    if ((cur - last_edge_report_call_nanos_) > edge_report_duration_nanos_) {
-      edge_reporter_->reportEdges();
+    if ((cur - last_edge_report_call_nanos_) >
```
bianpengyuan: Is it right to using one `last report time` for two types of duration? Isn't that the longer duration one will never be triggered?

extensions/stackdriver/stackdriver.cc
```diff
@@ -168,11 +168,11 @@ bool StackdriverRootContext::onConfigure(size_t) {
   }
 
   if (config_.has_mesh_edges_reporting_duration()) {
-    edge_report_duration_nanos_ =
+    edge_new_report_duration_nanos_ =
```
bianpengyuan: Do we want to cap it to 10m?

douglas-reid: done.

extensions/stackdriver/stackdriver.cc
douglas-reid: updated logic.

extensions/stackdriver/stackdriver.h
douglas-reid: done.

PR #2670, opened 2020-02-12
test: de-duplicate config artifacts, upgrade some xDS config to v3

kyessenov:

**Release note**:

```release-note
```


test/envoye2e/driver/envoy.go
```diff
@@ -26,6 +26,10 @@ import (
 	core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 	v2 "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v2"
 
+	// Preload proto definitions
+	_ "github.com/cncf/udpa/go/udpa/type/v1"
+	_ "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
```
gargnupur: We should not need this if we use udpa struct right?

kyessenov: Yes, but we should only use UDPA struct for open-ended config, like Wasm since it keeps changing. It's ok to use Any for stable configs.

test/envoye2e/http_metadata_exchange/http_metadata_exchange_test.go
```diff
@@ -77,121 +77,11 @@ const clusterTLSContext = `tls_context:
...
-  - tag_name: "connection_security_policy"
-    regex: "(connection_security_policy=\\.=(.+?);\\.;)"
-  - tag_name: "cache"
-    regex: "(cache\\.(.+?)\\.)"
-  - tag_name: "component"
-    regex: "(component\\.(.+?)\\.)"
-  - tag_name: "tag"
-    regex: "(tag\\.(.+?);\\.)"`
+var (
+	statsConfig          = driver.LoadTestData("testdata/bootstrap/stats.yaml.tmpl")
```
gargnupur: Thanks! this makes it so cleaner!

kyessenov:
Hmmm, I can't tell why TSan sometimes flakes a lot, sometimes works.

PR #2673, opened 2020-02-13
benchmark: shared node cache

kyessenov:
This PR evaluates using a shared node cache. Currently, the flow is as follows:
- [write raw bytes] MX writes raw peer bytes and peer id
- [read raw bytes with cache] stats reads peer id, looks up in cache for parsed peer bytes

The proposal is to use the following flow:
- [write strings with cache] MX looks up in cache for parsed peer bytes, writes peer fields
- [read strings] stats reads peer fields directly

The benefit is exposing peer fields directly to all plugins without any caching, and saving on the node cache memory per plugin.

Performance result:
```
2020-02-12 16:31:45
Running bazel-bin/extensions/common/context_speed_test
Run on (32 X 2200 MHz CPU s)
CPU Caches:
  L1 Data 32K (x16)
  L1 Instruction 32K (x16)
  L2 Unified 256K (x16)
  L3 Unified 56320K (x1)
Load Average: 0.12, 0.12, 0.26
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_WriteRawBytes                300 ns          300 ns      4520378
BM_ReadRawBytesWithCache        120 ns          120 ns     11730430
BM_WriteStringsWithCache        742 ns          742 ns      1902365
BM_ReadStrings                  180 ns          180 ns      8058773
```
(these are nano-seconds)

/cc @mandarjog @bianpengyuan


kyessenov:
One perf optimization might be multi-level filter state lookup. Seems like having one global map slows us down.

bianpengyuan:
first two: 300/4520378 + 120/11730430 = 0.000076
Later two: 742/1902365 + 180/8058773 = 0.00041

Does this mean combination of the first two is better than the later two? Do I miss compute?

kyessenov:
Added flat buffer representation:
```
BM_WriteRawBytes                   314 ns          314 ns      2222310
BM_ReadRawBytesWithCache           122 ns          122 ns      5782215
BM_WriteStringsWithCache           776 ns          776 ns       897763
BM_ReadStrings                     172 ns          172 ns      3961254
BM_WriteFlatBufferWithCache        400 ns          400 ns      1718023
BM_ReadFlatBuffer                 66.6 ns         66.6 ns      9862597
```

Flat buffers are faster than read-raw-with-cache, and almost as fast to generate using a cache.

kyessenov:
@bianpengyuan Yes, using read cache is still faster. I think it's because of the map lookup inside envoy.

PR #2685, opened 2020-02-14
Add check for gRPC status in access log filter

gargnupur:

1) Check gRPC status too in accesslog filter
2) Add tracing debug logs in accesslog and metadata exchange filter

### DIFF
--- a/extensions/access_log_policy/plugin.cc
+++ b/extensions/access_log_policy/plugin.cc
@@ -62,7 +62,6 @@ bool setFilterStateValue(bool log) {
 constexpr long long kDefaultLogWindowDurationNanoseconds =
     43200000000000;  // 12h
 
-constexpr StringView kRootContextId = "accesslog_inbound";
 constexpr StringView kSource = "source";
 constexpr StringView kAddress = "address";
 constexpr StringView kConnection = "connection";
@@ -72,8 +71,7 @@ constexpr StringView kCode = "code";
 constexpr StringView kGrpcStatus = "grpc_status";
 
 static RegisterContextFactory register_AccessLogPolicy(
-    CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContext),
-    kRootContextId);
+    CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContext));
 
 bool PluginRootContext::onConfigure(size_t) {
   if (::Wasm::Common::TrafficDirection::Inbound !=
--- a/extensions/access_log_policy/plugin.cc
+++ b/extensions/access_log_policy/plugin.cc
@@ -150,17 +150,19 @@ void PluginContext::onLog() {
 bool PluginContext::isRequestFailed() {
   // Check if HTTP request is a failure.
   int64_t http_response_code = 0;
-  getValue({kResponse, kCode}, &http_response_code);
-  // If request is a failure, log it.
-  if (http_response_code != 200) {
+  if (getValue({kResponse, kCode}, &http_response_code) &&
+      http_response_code != 200) {
     return true;
   }
 
   // Check if gRPC request is a failure.
-  int64_t grpc_response_code = ::Wasm::Common::kDefaultGrpcStatusCode;
-  getValue({kResponse, kGrpcStatus}, &grpc_response_code);
-  // If request is a failure, log it.
-  if (grpc_response_code != ::Wasm::Common::kDefaultGrpcStatusCode) {
+  int64_t grpc_response_code = 0;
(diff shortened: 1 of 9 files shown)


extensions/access_log_policy/plugin.cc
```diff
@@ -62,13 +62,18 @@ bool setFilterStateValue(bool log) {
 constexpr long long kDefaultLogWindowDurationNanoseconds =
     43200000000000;  // 12h
 
+constexpr StringView kRootContextId = "accesslog_inbound";
```
mandarjog: Are we doing outbound access logs? Why root context ?

gargnupur: No, we are not... I thought its just good practice to have it

gargnupur: Talked to @bianpengyuan .. not really needed , so removing it

extensions/access_log_policy/plugin.cc
```diff
@@ -143,6 +147,26 @@ void PluginContext::onLog() {
   setFilterStateValue(false);
 }
 
+bool PluginContext::isRequestFailed() {
+  // Check if HTTP request is a failure.
+  int64_t http_response_code = 0;
+  getValue({kResponse, kCode}, &http_response_code);
+  // If request is a failure, log it.
+  if (http_response_code != 200) {
```
bianpengyuan: Not related to this pr, but should this check 2xx? Also what about 1xx and 3xx?

extensions/access_log_policy/plugin.cc
```diff
@@ -143,6 +147,26 @@ void PluginContext::onLog() {
...
+    return true;
+  }
+
+  // Check if gRPC request is a failure.
+  int64_t grpc_response_code = ::Wasm::Common::kDefaultGrpcStatusCode;
+  getValue({kResponse, kGrpcStatus}, &grpc_response_code);
```
bianpengyuan: Is the default value necessary? getValue should run non-ok if not found. Also the logic does not seem to be right. Shouldn't we check if response_code is not OK (0)?

src/envoy/tcp/metadata_exchange/metadata_exchange.cc
```diff
@@ -88,6 +88,9 @@ Network::FilterStatus MetadataExchangeFilter::onData(Buffer::Instance& data,
       // If Alpn protocol is not the expected one, then return.
       // Else find and write node metadata.
       if (read_callbacks_->connection().nextProtocol() != config_->protocol_) {
+        ENVOY_LOG(trace, "Alpn Protocol Not Found. Expected {}, Got {}",
```
bianpengyuan: Do we expect compile this into wasm? If not, then it is ok to add log aggressively. If yes, we might want to be careful on adding log on data path because it is expensive.

extensions/access_log_policy/plugin.cc
gargnupur: Every thing other than 200 is treated as an error and hence we will log that request.

extensions/access_log_policy/plugin.cc
gargnupur: For gRPC , I am not sure why but OK is value 2...
and hence if getValue return non-ok and keep grpc_response_code as 0, we will think request is always a failure and even though it is not.

gargnupur: @douglas-reid : I see you set  'grpc_status = 2', why is OK for gRPC value 2?

src/envoy/tcp/metadata_exchange/metadata_exchange.cc
gargnupur: No, we don't accept it to compile into WASM but still this would be on data path and hence I have kept severity of logs to trace and not even debug because of that...

extensions/access_log_policy/plugin.cc
bianpengyuan: I think Doug just want to make `unknown` the default, which is 2.

douglas-reid: yeah, i set it to `2` for `unknown`, but if you look at the code, if the protocol isn't `grpc` we leave it empty.

gargnupur: Thanks.. changed logic to check for OK(0) and protocol check for gRPC too

gargnupur:
@bianpengyuan , @mandarjog : Can you please review this one?

gargnupur:
@bianpengyuan , @douglas-reid , @mandarjog :
friendly ping.. it would be good to get this in..

PR #2691, opened 2020-02-19
Enable TCP Client Metrics and Remove response_code label from tcp metrics

gargnupur:
**What this PR does / why we need it**:Enable TCP Client Metrics and Remove response_code and grpc_response_status label from tcp metrics


extensions/stats/plugin.h
```diff
@@ -196,6 +196,11 @@ class StatGen {
     n.reserve(s);
     n.append(metric_.prefix);
     for (size_t i = 0; i < metric_.tags.size(); i++) {
+      // Don't add response_code and grpc_response_status labels for TCP.
+      if ((metric_.tags[i].name == "response_code" ||
```
kyessenov: Could we just skip the tag if tag value is empty? We also would need to skip setting `unknown` for TCP.

gargnupur: But what if value is empty because we couldn't find it?

gargnupur: We would still need the label for dashboards I think...

kyessenov: @douglas-reid is there a difference between missing prom tag and an empty string tag value?

douglas-reid: @kyessenov do you mean at consumption time? I don't believe there is a difference on input between missing label and empty string. for generation, as long as the regexs support empty inputs, it should be ok either way too.

gargnupur: why do we set `unknown` for HTTP? I am guessing we would need the same for TCP?
Trying to keep this change small so that it can go to 1.5.
We can evaluate and test if skipping all empty tags after this? I think that will save cost in terms of data transfer too?

douglas-reid: i'm a little confused about the discussion here. we've never set `unknown` for `response_code` or for `grpc_response_status`. For `grpc` requests, `grpc_response_status` will only be `2` (unknown) when no other code can be found (including the http response code, typically 200) and `response_code` will be `200`. For `http` requests, `grpc_response_status` is left unset (empty string).

Removing them from the TCP metrics entirely seems like the right thing to do.

gargnupur: Sorry for the confusion. I think questions here are not just for response_code but for all labels in TCP:
1) Don't set `unknown` for any unknown labels in TCP. This is not just for response_code label but for all labels in TCP.
2) If a label is empty, just don't record it rather than recording it with default 0 or unknown values.

I agree with above but I don't think we did that in 1.4, so should we just keep doing what we did for 1.4 in 1.5 now so as to not cause any dashboards regression and then make these changes for 1.6?

douglas-reid: oh, ok. yeah. i'm fine with matching the 1.4 expectations for the metrics.

gargnupur:
@kyessenov, @douglas-reid : ping.. please review the PR..

gargnupur:
@kyessenov , @douglas-reid  : Thanks for the review!

PR #2690, opened 2020-02-19
Skip tsan/asan stress test

bianpengyuan:
**What this PR does / why we need it**:
https://github.com/istio/istio/issues/21273

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #2692, opened 2020-02-19
feat(stackdriver): support canonical service labels

douglas-reid:
This PR adds support for canonical service labels to the Stackdriver plugin.  In particular, it adds support for `{source|destination}_canonical_service_{name|namespace}` and `{source|destination}_caonical_revision}`.

### DIFF
--- a/extensions/stackdriver/metric/record.cc
+++ b/extensions/stackdriver/metric/record.cc
@@ -25,6 +25,11 @@ namespace Extensions {
 namespace Stackdriver {
 namespace Metric {
 
+constexpr char kCanonicalNameLabel[] = "service.istio.io/canonical-name";
+constexpr char kCanonicalRevisionLabel[] =
+    "service.istio.io/canonical-revision";
+constexpr char kLatest[] = "latest";
+
 void record(bool is_outbound, const ::wasm::common::NodeInfo &local_node_info,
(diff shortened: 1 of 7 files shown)


bianpengyuan:
Test situation should be better after https://github.com/istio/proxy/pull/2690 is merged.

extensions/stackdriver/metric/record.cc
```diff
@@ -33,6 +38,34 @@ void record(bool is_outbound, const ::wasm::common::NodeInfo &local_node_info,
       request_info.request_protocol == ::Wasm::Common::kProtocolGRPC
           ? request_info.request_url_path
           : request_info.request_operation;
+
+  const auto &local_labels = local_node_info.labels();
+  const auto &peer_labels = peer_node_info.labels();
+
+  std::string local_canonical_name = local_node_info.workload_name();
```
bianpengyuan: How about this?
```
const auto local_name_iter = local_labels.find(kCanonicalNameLabel);
const string& local_canonical_name = local_name_iter == local_labels.end() ? 
       local_node_info.workload_name() : local_name->second;
```

I think this should compile and save an allocation.

douglas-reid: done.

douglas-reid:
PTAL.

PR #2695, opened 2020-02-20
Nuke mixer fault test.

duderino:

Removing the mixer fault test because it flakes on busy nodes due to timing issues, and it depends on envoy internals which often change.  The latter could be mitigated by upstreaming it to envoyproxy/envoy, but htuch indicated it's unlikely it would be accepted.

And it doesn't seem to actually catch bugs.  So not worth the maintenance overhead.


PiotrSikora:
@duderino could you remove to references to `mixer_fault_test` from `Makefile.core.mk`?

PR #2696, opened 2020-02-20
Nuke mixer fault test. (#2695)

PiotrSikora:


PiotrSikora:
`macos` fails because of some weird macOS / homebrew / python / hashlib interaction that I'm not keen in digging into, but I think we can ignore this failure in `release-1.2` branch. PTAL.

PR #2699, opened 2020-02-20
Make connection security policy string lowercase for prometheus

gargnupur:

Make connection security policy string lowercase for prometheus
We need this to be compatible to Istio 1.4


extensions/stats/plugin.cc
```diff
@@ -97,8 +105,8 @@ void map_request(IstioDimensions& instance,
   instance[request_protocol] = request.request_protocol;
   instance[response_code] = std::to_string(request.response_code);
   instance[response_flags] = request.response_flag;
-  instance[connection_security_policy] = std::string(
-      ::Wasm::Common::AuthenticationPolicyString(request.service_auth_policy));
+  instance[connection_security_policy] = to_lower(std::string(
```
bianpengyuan: How about using this https://github.com/abseil/abseil-cpp/blob/master/absl/strings/ascii.h#L168？

gargnupur: I thought we were not using absl stuff for wasm pluging?

bianpengyuan: String related utils should be fine. they do not involve system calls.

PR #2707, opened 2020-02-21
feat(edges): support configurable batch size

douglas-reid:
This PR adds support for a configurable batch size to the mesh telemetry reporting. It also lowers the default to 100 traffic assertions per request (from 1000).

### DIFF
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -20,7 +20,7 @@ package stackdriver.config.v1alpha1;
 import "google/protobuf/duration.proto";
 
 message PluginConfig {
-  // next id: 7
+  // next id: 8
 
   // Optional. Controls whether to export server access log.
   bool disable_server_access_logging = 1;
@@ -52,4 +52,8 @@ message PluginConfig {
   // not available from the controlplane. Disable the fallback if the host
   // header originates outsides the mesh, like at ingress.
   bool disable_host_header_fallback = 6;
+
+  // Optional. Allows configuration of the number of traffic assertions to batch
+  // into a single request. Default is 250.
+  int32 max_edges_batch_size = 7;
 }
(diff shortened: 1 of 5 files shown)


extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -68,15 +68,18 @@ void instanceFromMetadata(const ::wasm::common::NodeInfo& node_info,
 }  // namespace
 
 EdgeReporter::EdgeReporter(const ::wasm::common::NodeInfo& local_node_info,
-                           std::unique_ptr<MeshEdgesServiceClient> edges_client)
-    : EdgeReporter(local_node_info, std::move(edges_client), []() {
+                           std::unique_ptr<MeshEdgesServiceClient> edges_client,
+                           const int batch_size)
```
gargnupur: nit: afaik `const int` is not really useful

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -52,4 +52,8 @@ message PluginConfig {
   // not available from the controlplane. Disable the fallback if the host
   // header originates outsides the mesh, like at ingress.
   bool disable_host_header_fallback = 6;
+
+  // Optional. Allows configuration of the number of traffic assertions to batch
+  // into a single request. Default is 250.
```
gargnupur: Below looks like it is set to 100..

extensions/stackdriver/edges/edge_reporter.cc
douglas-reid: removed.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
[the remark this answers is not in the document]
douglas-reid: updated.

extensions/stackdriver/edges/edge_reporter.h
```diff
@@ -55,11 +57,12 @@ class EdgeReporter {
 
  public:
   EdgeReporter(const ::wasm::common::NodeInfo &local_node_info,
-               std::unique_ptr<MeshEdgesServiceClient> edges_client);
+               std::unique_ptr<MeshEdgesServiceClient> edges_client,
+               int batch_size = kDefaultAssertionBatchSize);
```
gargnupur: nit: can we just make it a required param and pass in the default when config is not set?

Because it would just increase readability as we have another EdgeReporter constructor already.

bianpengyuan:
This is also needed in 1.5 right?

douglas-reid:
@bianpengyuan yes. thank you.

PR #2715, opened 2020-02-25
Use envoy_cc_library instead of cc_library

bianpengyuan:

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stackdriver/common/BUILD
+++ b/extensions/stackdriver/common/BUILD
@@ -22,11 +22,12 @@ load(
     "envoy_cc_library",
 )
 
-cc_library(
+envoy_cc_library(
     name = "constants",
     hdrs = [
         "constants.h",
     ],
+    repository = "@envoy",
     visibility = [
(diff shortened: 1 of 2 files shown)


bianpengyuan:
Let me try to build it on my mac..

PiotrSikora:
Why this change is sent against `release-1.5` and not against `master` with a cherry-pick label?

bianpengyuan:
Yeah sorry, master fix: https://github.com/istio/proxy/pull/2722

bianpengyuan:
The reason why this is not a cherry-pick from master is because the pr (https://github.com/istio/proxy/commit/d14ae10df125e1b26b6b770e5d8c8ccfc848af30) that causes build failure is only in release-1.5, not in master, since grpc library will have a more appropriate fix for that. So I sent the fix here first to fix the build.

bianpengyuan:
@fpesce Mind approving this? This blocks 1.5 dev proxy build.

PR #2716, opened 2020-02-25
Cherry-pick JWT CVE fix into master

yangminzhu:
* Fixed JWT CVE related to exact PATH matches (#9)

* Fixed JWT CVE related to exact PATH matches

Problem: The JWT filter when matching exact paths included query parameters
which meant the JWT requirement could be bypassed by adding a "?" after the
path. The API was intended to only work for URIs.

Solution: The fix updates the match logic to only include URIs i.e. path
stripped off the query section.

Added unit tests to validate these cases.

* Fixed formatting

* Strip fragment of Path

Added unit tests to validate combination of query & fragment

* Fix lint

* Minor refactoring and more unit test cases (#11)

* Minor refactoring and more unit test cases

* Lint fixes

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/authn/origin_authenticator.cc
+++ b/src/envoy/http/authn/origin_authenticator.cc
@@ -18,6 +18,7 @@
 #include "absl/strings/match.h"
 #include "authentication/v1alpha1/policy.pb.h"
 #include "common/http/headers.h"
+#include "common/http/utility.h"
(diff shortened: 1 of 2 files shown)


yangminzhu:
Manually set `cla: yes` as @nrjpoddar has committed the same change to other branches.

PR #2721, opened 2020-02-26
Set ticker in onConfigure instead of onStart

bianpengyuan:

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -126,6 +126,10 @@ std::string getSTSPort() {
 
 bool StackdriverRootContext::onConfigure(
     std::unique_ptr<WasmData> configuration) {
+  // onStart is called prior to onConfigure
+  if (enableServerAccessLog() || enableEdgeReporting()) {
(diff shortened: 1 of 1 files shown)


bianpengyuan:
This could be the cause of https://github.com/istio/istio/issues/21505

bianpengyuan:
@kyessenov @jplevyak @douglas-reid fyi

extensions/stackdriver/stackdriver.cc
```diff
@@ -189,6 +189,12 @@ bool StackdriverRootContext::onConfigure(
 
   node_info_cache_.setMaxCacheSize(config_.max_peer_cache_size());
 
+  // All the necessary reporting objects should be initialized.
```
douglas-reid: This needs to move to the top of `onConfigure`.

douglas-reid: It should match the `release-1.5` changes:

https://github.com/istio/proxy/blob/master/extensions/stackdriver/stackdriver.cc#L140-L146.

And we should probably match `master`.

douglas-reid: OK. we are going to update 1.5 to match this placement.

PR #2730, opened 2020-02-27
cleanup: deprecate empty proto as config

kyessenov:
Empty proto is deprecated as config.
Fixes https://github.com/istio/istio/issues/21034


kyessenov:
Please help getting this merged. It's a small cleanup to stop using Empty as the config type due to collisions.

PR #2703, opened 2020-02-21
[Release-1.5] add x-goog-user-project header to Stackdriver callout

bianpengyuan:
**What this PR does / why we need it**:
release-1.4 pr: https://github.com/istio/proxy/pull/2683
issue: https://github.com/istio/istio/issues/20183

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #2731, opened 2020-02-27
add x-goog-user-project header to stackdriver call out

bianpengyuan:
This adds `x-goog-user-project` header to metrics, log and context graph call out.

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stackdriver/common/utils.cc
+++ b/extensions/stackdriver/common/utils.cc
@@ -50,6 +50,9 @@ void buildEnvoyGrpcService(
           stub_option.test_token_path.empty()
               ? ::Extensions::Stackdriver::Common::kSTSSubjectTokenPath
               : stub_option.test_token_path);
+      auto initial_metadata = grpc_service->add_initial_metadata();
+      initial_metadata->set_key("x-goog-user-project");
+      initial_metadata->set_value(stub_option.project_id);
     }
 
     grpc_service->mutable_google_grpc()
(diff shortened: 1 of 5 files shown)


mandarjog:
Is this a cherry pick up into master?

bianpengyuan:
Part of it (metric) is a cherry-pick from https://github.com/istio/proxy/pull/2703. For log and edge, we used `onCreateInitialMetadata` callback in release-1.{4,5} to inject header, but we could actually just use `initial_metadata` to do it: https://github.com/envoyproxy/envoy/blob/9186ebf374b1bd7d7c65ed1cec748377f88ec0a7/api/envoy/api/v2/core/grpc_service.proto#L225  https://github.com/envoyproxy/envoy-wasm/pull/461 removes the onCreateInitialMetadata from the ABI.

kyessenov:
Do you want to add a test to the integration test to validate the presence of the header?

bianpengyuan:
The test is already there: https://github.com/istio/proxy/pull/2731/files#diff-9a64bdacaaf291cc03cca6cf3c038174R220

kyessenov:
Yes, I'm confused how the tests passed before without this change. How did it work before?

bianpengyuan:
If you look at the newly added linked line (line 220), it searches for and verifies header value. Without .cc change in this PR, SD test will fail.

kyessenov:
Oh, OK. Sorry about the silly question, I think I lost some context how to navigate this code base...

PR #2734, opened 2020-02-28
allow unknown fields in configs

kyessenov:
To support future releases with newer config patches.


kyessenov:
@mandarjog This should be good to get in to avoid crashes due to typos in configs.

PR #2741, opened 2020-03-02
feat(stackdriver): add support for destination_subset_name

douglas-reid:
This PR adds `destination_subset_name` to the Stackdriver extension for client-side metrics. The name is taken from cluster metadata, according to the changes in https://github.com/istio/istio/pull/21611.

This PR should not be submitted until the Stackdriver schema changes have been approved.

/hold

### DIFF
--- a/extensions/stackdriver/metric/record.cc
+++ b/extensions/stackdriver/metric/record.cc
@@ -21,6 +21,21 @@
 
 using google::protobuf::util::TimeUtil;
 
+// WASM_PROLOG
+#ifndef NULL_PLUGIN
+#include "proxy_wasm_intrinsics.h"
+
+#else  // NULL_PLUGIN
+
+#include "absl/strings/str_split.h"
+#include "extensions/common/wasm/null/null_plugin.h"
+
+using Envoy::Extensions::Common::Wasm::Null::Plugin::getMessageValue;
+
+#endif  // NULL_PLUGIN
+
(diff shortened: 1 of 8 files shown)


extensions/stackdriver/metric/record.cc
```diff
@@ -61,6 +76,17 @@ void record(bool is_outbound, const ::wasm::common::NodeInfo &local_node_info,
       peer_rev_iter == peer_labels.end() ? kLatest : peer_rev_iter->second;
 
   if (is_outbound) {
+    std::string subset;
+    google::protobuf::Struct istio;
+    getMessageValue({"cluster_metadata", "filter_metadata", "istio"}, &istio);
```
mandarjog: Can you pass “subset” in the initializer list ?
@kyessenov is that supported ?

mandarjog: @gargnupur what happens if the destination cluster is empty ?
Will this call still get it or will it hit the optimization where cluster is not selected if it is empty.

gargnupur: If destination cluster is empty then struct  will be empty...
If we check for return value here, it. should return false...

douglas-reid:
This PR is on hold for a little while, as the server-side schema changes are still pending.

douglas-reid:
I believe this work is still desirable, but this PR has languished long enough to be no longer useful.

PR #2742, opened 2020-03-02
Remove not getting TCP Metrics for unknown peers and Add Context in onTick calls

gargnupur:
Remove not getting TCP Metrics for unknown peers
Also, add context in onTick. calls which makes possible to get streaminfo from context in wasm plugin

Ref: https://github.com/istio/istio/issues/21566

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -38,9 +38,9 @@ bind(
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
 # envoy-wasm commit date: Mar 3 2020
-ENVOY_SHA = "373eafdb20958a019f5556e508b5286eba88af06"
+ENVOY_SHA = "e64efd960cd8e97287764b9bae39b3a0def95046"
 
-ENVOY_SHA256 = "e7fbb8b14f1a30024a07d40019df4dfd944cbbc813be08b46a2b5fca5f495a79"
+ENVOY_SHA256 = "3bac67e31d920725b0655d2c5dd1934d63d57d542f80e324d661c6b3b536b4bf"
 
 # To override with local envoy, just pass `--override_repository=envoy=/PATH/TO/ENVOY` to Bazel or
 # persist the option in `user.bazelrc`.
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -37,10 +37,10 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
-# envoy-wasm commit date: Feb 28 2020
-ENVOY_SHA = "f4ace91833442010118a592c2dff3689d984337d"
+# envoy-wasm commit date: Mar 3 2020
+ENVOY_SHA = "373eafdb20958a019f5556e508b5286eba88af06"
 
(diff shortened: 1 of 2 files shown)


mandarjog:
per discussion this should use response flags / No Healthy upstream to conclude that mx will not happen.

extensions/stats/plugin.cc
```diff
@@ -460,8 +462,14 @@ bool PluginRootContext::report(::Wasm::Common::RequestInfo& request_info,
...
     // Otherwise, we wait for metadata exchange to happen before we report  any
     // metric.
+    // We skip this wait  if upstream is unhealthy as we won't get peer metadata
+    // in that case.
+    uint64_t response_flags = 0;
+    getValue({"response", "flags"}, &response_flags);
     if (peer_node_ptr == nullptr &&
-        peer_id != ::Wasm::Common::kMetadataNotFoundValue) {
+        peer_id != ::Wasm::Common::kMetadataNotFoundValue &&
+        !(::Wasm::Common::parseResponseFlag(response_flags)
```
kyessenov: Response flags is a bit mask so just use XOR to check for a flag.

WORKSPACE
```diff
@@ -37,10 +37,10 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
-# envoy-wasm commit date: Feb 28 2020
-ENVOY_SHA = "f4ace91833442010118a592c2dff3689d984337d"
+# envoy-wasm commit date: Mar 3 2020
+ENVOY_SHA = "373eafdb20958a019f5556e508b5286eba88af06"
```
PiotrSikora: Could you grab `e64efd960cd8e97287764b9bae39b3a0def95046` while you're at it?

kyessenov: It might be good to limit 1.5 changes for cherry-picking.

gargnupur: do workspace update. in a separate pr?

PiotrSikora: 1) This is `master` branch, so it doesn't matter in terms of cherry-picking into `release-1.5`, other than for the fact that it would get tested sooner, rather than later.
2) That change is getting into `release-1.5` along with few other WASM fixes anyway.

PiotrSikora: Why separate PR? It will just conflict with this.

extensions/stats/plugin.cc
gargnupur: Fixed

WORKSPACE
gargnupur: I had created https://github.com/istio/proxy/pull/2744 pr for updating sha, will skip it then and just update sha here

gargnupur: Updated SHA here. only

bianpengyuan:
We should probably push the wasm gen container in presubmit on the first successful build. It is too slow to rebuilt the container on every new commit.

PiotrSikora:
Do we have access to the account in presubmit jobs? I don't think we do... but we might have enabled it because of RBE?

bianpengyuan:
Yeah, good point. We don't: https://github.com/istio/test-infra/blob/master/prow/config/jobs/proxy.yaml#L29

Let me update that.

fpesce:
@bianpengyuan / @gargnupur any of you is on top of that manual cherrypick ? :D

bianpengyuan:
@fpesce  Yeah @gargnupur is going to do the manual cherry-pick.

PR #2745, opened 2020-03-04
Push wasm build container in presubmit

bianpengyuan:
wasm gen container is too slow to build and could slows down iteration for envoy sha update PR. This change makes ci push container during presubmit, so that it will only be built once during envoy sha update.

**Release note**:

```release-note
```


bianpengyuan:
Close as this won't work since presubmit does not have permission to push. The permission should not be added since it allows random pr author to have istio-testing project permission.

PR #2747, opened 2020-03-04
Fix building with gcc

iii-i:
--config=libc++ is not compatible with gcc, since gcc does not
support -stdlib option. Therefore, disable usage of libc++ when
gcc is used to compile proxy. With this patch, the following
begins to work:

    make CC=/usr/bin/gcc CXX=/usr/bin/g++

**Which issue this PR fixes** None

**Special notes for your reviewer**: None

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -37,12 +37,21 @@ BAZEL_STARTUP_ARGS := --client_debug $(BAZEL_STARTUP_ARGS)
 BAZEL_BUILD_ARGS := -s --sandbox_debug --verbose_failures $(BAZEL_BUILD_ARGS)
 endif
 
+ifeq "$(origin WITH_LIBCXX)" "undefined"
+WITH_LIBCXX := $(shell ($(CXX) --version | grep ^g++ >/dev/null && echo 0) || echo 1)
+endif
+ifeq "$(WITH_LIBCXX)" "1"
+BAZEL_CONFIG = --config=libc++
+else
+BAZEL_CONFIG =
+endif
+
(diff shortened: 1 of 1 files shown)


kyessenov:
@iii-i Can we provide more detail why you want g++ support? There's been some strong opinions on mandating libc++, cc @PiotrSikora @jplevyak .

iii-i:
@bianpengyuan nice find, I tested this only on Fedora. The new commit looks for `^g++`, which should work everywhere.

@kyessenov we are trying to build proxy on s390x, and unfortunately there are some issues with using LLVM toolchain (e.g.: lld usage is forced in a few places, but lld does not support s390x yet). While this can definitely be resolved in the future, it seems natural to support different C++ compilers, and gcc seems good enough for us at the moment.

bianpengyuan:
@kyessenov @PiotrSikora please correct me if I am wrong. IIRC https://istio.io/news/security/istio-security-2019-003/ was the reason to switch to Clang/libc++, since it has safer regex matching. Looks like envoyproxy supports gcc build, so it should be fine for istio/proxy to also include gcc as an option?

iii-i:
Hi, is there anything I need to/can do to get this merged?

lizan:
It's fine to support other architecture / compiler, as long as the default is libc++. @iii-i CI is not happy though.

iii-i:
This looks like a flaky test to me (the patch doesn't touch the code or the default build settings). Let me rebase and see if it goes away.

bianpengyuan:
Yeah the test is flaky.

PR #2748, opened 2020-03-04
Allow building proxy on s390x

iii-i:

This adds support for the s390x architecture (aka [IBM Z](https://en.wikipedia.org/wiki/IBM_Z)).

**Which issue this PR fixes**: None

**Special notes for your reviewer**: envoy does not build on s390x yet, but we are sending PRs to fix this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,8 @@ else ifeq ($(LOCAL_ARCH),aarch64)
     TARGET_ARCH ?= arm64
 else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 4),armv)
     TARGET_ARCH ?= arm
+else ifeq ($(LOCAL_ARCH),s390x)
+    TARGET_ARCH ?= s390x
 else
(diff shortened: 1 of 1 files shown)


bianpengyuan:
This file is copied from common file repo and shared by all istio repo: https://github.com/istio/common-files/blob/dde5c0862784a266c1dc114360c13b3c0da87dcb/files/Makefile#L41

Also I don't think proxy build really uses these platform options listed here. It is mostly used to decide the artifacts output path for other golang repo.

iii-i:
@lambdai no, wasm does not work at all - its build is using x86_64-only docker image. Is wasm support required to have this merged?

@bianpengyuan whoops, that's embarrassing.. this is written at the top of Makefile :-) I will submit another PR to common-files. I agree that we don't really need this variable, but not having a branch for s390x makes us hit the `$(error)`.

iii-i:
I've opened https://github.com/istio/common-files/pull/201.

PR #2765, opened 2020-03-14
fix(stackdriver): use more appropriate buckets for bytes

douglas-reid:
This PR attempts to separate the bucket definitions used for bytes distributions from the definitions used for latency measurements. Instead of 20 buckets with a scale factor of 1 and a growth factor of 2, the bytes metrics will now use 7 buckets with a scale factor of 1 and growth factor of 10.

That should lead to the following buckets:

[-Inf, 0)
[0, 10)
[10, 100)
[100, 1000)
[1000, 10000)
[10000, 100000)
[100000, 1000000)
[1000000, 10000000)
[10000000, Inf]

### DIFF
--- a/extensions/stackdriver/metric/registry.cc
+++ b/extensions/stackdriver/metric/registry.cc
@@ -141,7 +141,7 @@ StackdriverOptions getStackdriverOptions(
             .set_name(k##_v##View)                                  \
             .set_measure(k##_v##Measure)                            \
             .set_aggregation(Aggregation::Distribution(             \
-                BucketBoundaries::Exponential(8, 1, 10))) ADD_TAGS; \
+                BucketBoundaries::Exponential(7, 1, 10))) ADD_TAGS; \
     View view(view_descriptor);                                     \
     view_descriptor.RegisterForExport();                            \
   }
--- a/extensions/stackdriver/metric/registry.cc
+++ b/extensions/stackdriver/metric/registry.cc
@@ -134,6 +134,18 @@ StackdriverOptions getStackdriverOptions(
     view_descriptor.RegisterForExport();                            \
   }
 
+#define REGISTER_BYTES_DISTRIBUTION_VIEW(_v)                        \
+  void register##_v##View() {                                       \
+    const ViewDescriptor view_descriptor =                          \
+        ViewDescriptor()                                            \
+            .set_name(k##_v##View)                                  \
+            .set_measure(k##_v##Measure)                            \
+            .set_aggregation(Aggregation::Distribution(             \
+                BucketBoundaries::Exponential(8, 1, 10))) ADD_TAGS; \
+    View view(view_descriptor);                                     \
+    view_descriptor.RegisterForExport();                            \
+  }
+
 #define ADD_TAGS                                             \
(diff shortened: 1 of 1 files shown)


bianpengyuan:
Is this optimization for performance for query/precompute? I am wondering which case would be better with these new coarse buckets.

douglas-reid:
> Is this optimization for performance for query/precompute? I am wondering which case would be better with these new coarse buckets.

It is for perf along a number of dimensions. The storage space for the bytes metrics *far* (>5x) exceeds the storage space for latency metrics on the backend for the same traffic, particularly in precomputations. I am not an expert in the backend storage, but my naive stab at an explanation is that it appears in scale testing that all the requests end up in the final overflow bucket and that seems to really cause issues.

So, these buckets are not just coarser, but they cover more of the expected range.

mandarjog:
1. Seems like an easy way to reduce the total volume of timeseries distribution data produced by a constant factor.
2. What about migration? What happens to the same metrics using old and new buckets?

douglas-reid:
> 1. Seems like an easy way to reduce the total volume of timeseries distribution data produced by a constant factor.
> 2. What about migration? What happens to the same metrics using old and new buckets?

@quentinmit can you add detail here?

quentinmit:
> > 1. Seems like an easy way to reduce the total volume of timeseries distribution data produced by a constant factor.
> > 2. What about migration? What happens to the same metrics using old and new buckets?
> 
> @quentinmit can you add detail here?

Queries that involve differently-bucketed data will rebucket the data to one set of buckets. Obviously there can be some loss of precision here.

It will only reduce the total volume of data if the new buckets result in fewer buckets being filled, which does seem likely. Are you sure the last bucket is to "inf" and it's not currently just adding as many buckets as necessary to get to the size? That could explain why the data is currently so large.

douglas-reid:
> > > 1. Seems like an easy way to reduce the total volume of timeseries distribution data produced by a constant factor.
> > > 2. What about migration? What happens to the same metrics using old and new buckets?
> > 
> > 
> > @quentinmit can you add detail here?
> 
> Queries that involve differently-bucketed data will rebucket the data to one set of buckets. Obviously there can be some loss of precision here.
> 
> It will only reduce the total volume of data if the new buckets result in fewer buckets being filled, which does seem likely. Are you sure the last bucket is to "inf" and it's not currently just adding as many buckets as necessary to get to the size? That could explain why the data is currently so large.

I'm not sure. From the docs for lib:

> Creates a BucketBoundaries with num_finite_buckets with exponentially increasing boundaries starting at zero (governed by growth_factor and scale), as well as an underflow and overflow bucket... The underflow bucket covers [-inf, 0) and the overflow bucket [scale * growth_factor ^ num_finite_buckets, inf].

But perhaps the server-side does something different?

bianpengyuan:
Do we really need any bucket resolution after 10 Mb? To me seems like the last two buckets are pretty extreme (10Mb~100Mb, 100Mb~Inf).

douglas-reid:
@bianpengyuan it does seem a little excessive. i'll send a change to remove the last bucket.

PR #2767, opened 2020-03-16
fix(stackdriver): support turning off http size metrics

douglas-reid:
This PR adds configurability to the Stackdriver metrics plugin. It allows disabling of all request/response bytes metrics.

### DIFF
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -20,7 +20,7 @@ package stackdriver.config.v1alpha1;
 import "google/protobuf/duration.proto";
 
 message PluginConfig {
-  // next id: 8
+  // next id: 10
 
   // Optional. Controls whether to export server access log.
   bool disable_server_access_logging = 1;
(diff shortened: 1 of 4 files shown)


bianpengyuan:
Regarding disable outbound filter, since outbound metrics are the only telemetry reported on the way out, does it make more sense to add an option to istioctl values instead? The value will just avoid deploying the outbound filter.

douglas-reid:
@bianpengyuan that's a good question, and an interesting idea. i wonder though: doesn't that make it more of an install time option?

kyessenov:
I prefer Peter's suggestion. We could split envoyfilters into client and server side config resources. Then it's a matter of `kubectl delete envoyfilter stackdriver-outbound`.

douglas-reid:
@kyessenov that certain simplifies the code. how well does this align with the future non-EnvoyFilter-based feature enablement?

kyessenov:
We should keep separating client vs server extension configs for non-EnvoyFilter based approach.

douglas-reid:
Thanks for the great feedback so far. I've removed the toggle for client-side metrics and have added/updated tests.

bianpengyuan:
Asan failure should relate to the change.

mandarjog:
Ping ! We should get this in @douglas-reid

PR #2769, opened 2020-03-17
Partial Revert of STS changes after Istio 1.4.5 variant 1

duderino:
This is a subset of https://github.com/istio/proxy/pull/2749 which was incorrectly merged and then reverted.

It effectively removes these STS PRs from the 1.4 branch:

#2683 #2694 (just updates envoy sha, and I have to clobber this soon anyways) #2679

But retains these two PRs per @bianpengyuan:  

#2721 #2719 

As soon as this is merged I'll bring in the remaining 1.4.5 changes from our private repo and update the WORKSPACE to point to the appropriate istio/envoy SHA

PR #2683, opened 2020-02-14
Fix node metadata fetch in stackdriver filter

bianpengyuan:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #2770, opened 2020-03-17
Partial Revert of STS changes after Istio 1.4.5 variant 2

duderino:
A variant of https://github.com/istio/proxy/pull/2769 that doesn't revert the WORKSPACE istio/envoy SHA.

So this would revert:

#2683  #2679

but keep:

#2721 #2719 #2694


duderino:
After discussing with @mandarjog and @bianpengyuan, we decided to close this in favor of https://github.com/istio/proxy/pull/2771

PR #2774, opened 2020-03-17
increase life span for data in tcp cluster rewrite

yxue:
fixes https://github.com/istio/istio/issues/21676

### DIFF
--- a/src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
+++ b/src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
@@ -77,7 +77,7 @@ Network::FilterStatus TcpClusterRewriteFilter::onNewConnection() {
     } catch (...) {
       ENVOY_LOG(
           critical,
-          "tcp_cluster_rewrit: error setting data due to unknown exception");
+          "tcp_cluster_rewrite: error setting data due to unknown exception");
       throw;
     }
   }
--- a/src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
+++ b/src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
@@ -63,12 +63,25 @@ Network::FilterStatus TcpClusterRewriteFilter::onNewConnection() {
                    "tcp_cluster_rewrite: final tcp proxy cluster name {}",
                    read_callbacks_->connection(), final_cluster_name);
 
-    // The data is mutable to allow other filters to change it.
-    read_callbacks_->connection().streamInfo().filterState()->setData(
-        TcpProxy::PerConnectionCluster::key(),
-        std::make_unique<TcpProxy::PerConnectionCluster>(final_cluster_name),
-        StreamInfo::FilterState::StateType::Mutable);
+    try {
+      // The data is mutable to allow other filters to change it.
+      read_callbacks_->connection().streamInfo().filterState()->setData(
+          TcpProxy::PerConnectionCluster::key(),
+          std::make_unique<TcpProxy::PerConnectionCluster>(final_cluster_name),
+          StreamInfo::FilterState::StateType::Mutable,
+          StreamInfo::FilterState::LifeSpan::DownstreamConnection);
+    } catch (const EnvoyException& e) {
+      ENVOY_CONN_LOG(critical, "tcp_cluster_rewrite: error setting data: {}",
+                     read_callbacks_->connection(), e.what());
+      throw;
+    } catch (...) {
+      ENVOY_LOG(
+          critical,
+          "tcp_cluster_rewrit: error setting data due to unknown exception");
+      throw;
+    }
   }
+
   return Network::FilterStatus::Continue;
 }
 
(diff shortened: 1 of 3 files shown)


yxue:
Working on tests

src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
```diff
@@ -63,12 +63,25 @@ Network::FilterStatus TcpClusterRewriteFilter::onNewConnection() {
...
+          StreamInfo::FilterState::StateType::Mutable,
+          StreamInfo::FilterState::LifeSpan::DownstreamConnection);
+    } catch (const EnvoyException& e) {
+      ENVOY_CONN_LOG(critical, "tcp_cluster_rewrite: error setting data: {}",
+                     read_callbacks_->connection(), e.what());
+      throw;
+    } catch (...) {
+      ENVOY_LOG(
+          critical,
+          "tcp_cluster_rewrit: error setting data due to unknown exception");
```
duderino:
```suggestion
          "tcp_cluster_rewrite: error setting data due to unknown exception");
```

src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
```diff
@@ -63,12 +63,25 @@ Network::FilterStatus TcpClusterRewriteFilter::onNewConnection() {
...
-        std::make_unique<TcpProxy::PerConnectionCluster>(final_cluster_name),
-        StreamInfo::FilterState::StateType::Mutable);
+    try {
+      // The data is mutable to allow other filters to change it.
+      read_callbacks_->connection().streamInfo().filterState()->setData(
+          TcpProxy::PerConnectionCluster::key(),
+          std::make_unique<TcpProxy::PerConnectionCluster>(final_cluster_name),
+          StreamInfo::FilterState::StateType::Mutable,
+          StreamInfo::FilterState::LifeSpan::DownstreamConnection);
+    } catch (const EnvoyException& e) {
```
duderino: @PiotrSikora I thought Envoy didn't use exceptions on the data path.  Is this the right way to handle or should the code this is calling be changed to not through an exception?

Nice find @yxue!

duderino: Or is this the control path? :)

yxue: I think this is the data path. If the code swallow the exception here, the behavior will be unexpected because the tcp proxy depends on the data in the filter state to do routing.

src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
yxue: Done

src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
kyessenov: It throws when it's called twice with conflicting declarations (questionable design choice). The only other use I found was `sni_cluster` filter, but it matches the declaration here, so in theory it should never throw. I think it's fine to just remove the try/catch block. The danger is mis-configuration with two filters trying to declare the same filter state key in different ways.

kyessenov:
Since I don't entirely understand what this filter is for and why we need it, I'd vote for just deleting it. Who knows what other issues are lurking there.

ayj:
It is required by https://istio.io/docs/setup/install/multicluster/gateways. I'm not against removing it for 1.6. But for better or worse, people are using this even though MC is alpha and expect it to not crash their gateway.

src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
yxue: Mis configuration two filters with the same key is definitely wrong. Adding the try/catch is for debugging. Currently the exception is thrown inside the setData and difficult for pinpointing the issue. 

Current stack trace https://gist.github.com/ayj/5c07b20d651bb6be48a08b653f0ff590#file-log_trace-L604-L605

sdake:
@kyessenov here is a more detailed answer to your question of why the filter is needed: https://github.com/istio/istio/issues/21676#issuecomment-601350998

I am not sure if, given the new Envoy extensions mechanism, we can provide a path forward for existing MC users and remove the need for **tcp_rewrite_filter**, however, if that is viable, we should talk about that in the context of 1.6, not in the context of the 1.5.1 fix that is targeted for a few days from now :)

Cheers
-steve

kyessenov:
@sdake Thanks for writing it up, but I think I got lost on point 2, so not sure I am qualified to review.

sdake:
@kyessenov sure. I did leave out the detail of how the message enters the mesh via ingress, which is where point 2 starts. :) If you think of this model in reverse, that is how you connect two clusters together in a replicated control plane model.

src/envoy/tcp/tcp_cluster_rewrite/tcp_cluster_rewrite.cc
jplevyak: It does rethrow but that might be subtle to a reader.  Perhaps document that this is just to get a better error message and not for recovery?

PR #2777, opened 2020-03-20
Decide unhealthy connection on presence of no response flags

gargnupur:
Decide unhealthy connection on presence of no response flags as compared to just unhealthyhost in stats filter for TCP V2 telemetry

### DIFF
--- a/extensions/stats/plugin.cc
+++ b/extensions/stats/plugin.cc
@@ -44,8 +44,6 @@ namespace Plugin {
 namespace Stats {
 
 constexpr long long kDefaultTCPReportDurationMilliseconds = 15000;  // 15s
-// No healthy upstream.
-constexpr uint64_t kNoHealthyUpstream = 0x2;
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


mandarjog:
@gargnupur is response flag guaranteed to be set if there is a connection?
What are the possible values in the presence of a successful connection?

gargnupur:
> @gargnupur is response flag guaranteed to be set if there is a connection?
> What are the possible values in the presence of a successful connection?

@mandarjog :  responseflag being set implies there is a failure in connection and value of 0 implies successful connection. So if not set, implies things are going fine

gargnupur:
@mandarjog , @kyessenov , @bianpengyuan : PTAL.. it would be good to get this in..

kyessenov:
Response flag is not just for errors. A valid response (e.g. direct response) would carry a response flag.

bianpengyuan:
Ok, it would still be nice to capture all response flags, since it represents a special state.

kyessenov:
Yes, we should report all connections, not just the ones with the metadata exchange. It might pollute some initial data when the response is not received yet, but it's better than not reporting at all.

gargnupur:
@kyessenov, @bianpengyuan : yes, that is the hope with this change that, we don't drop telemetry because MX didn't happen...  and I have one more fear that these requests will start collecting in the queue and not get deleted as they don't get reported which could lead to slow memory leak.

PR #2780, opened 2020-03-24
Flush log entries in on done

bianpengyuan:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #2782, opened 2020-03-24
fix and deflake SD reload test

bianpengyuan:
Several changes to SD reload test:
* In https://github.com/istio/proxy/pull/2780, [the last revision](https://github.com/istio/proxy/pull/2780/commits/f4ceb426f525f9c1f80afec1d02694db047614d8) actually makes test not verify vm reload. This PR makes it so that VM id is modified with lds version, which will really trigger a VM reload.
* Make SD listener reload test only update listener config once. Even two config updates could make test flaky, so lower it to update only once.
* Skip both SD listener reload test and vm reload test when ASan/TSan due to known flake caused by slowness.


bianpengyuan:
@kyessenov ptal, thanks! Test in last pr is not verifying vm reload because of a bad revision. This PR fix it.

kyessenov:
Ok, I was concerned that we don't actually test reloads, so wanted some indication that without the fix the reloads are broken.

bianpengyuan:
Yeah the last revision actually revokes verification of VM reload. Sorry about that :p. For this PR I've for sure tested the latest commit would fail without onDone logic.

PR #2783, opened 2020-03-25
Improve Wasm plugins stats

bianpengyuan:
Several improvements to stats from Wasm plugins:
* Add stats for sd edge report call.
* Add back stats of sd plugin filter logging call out, which is removed by https://github.com/istio/proxy/pull/2780. I thought gRPC native stats is already available, but turns out it is actually not.
* Make `wasm_filter` the first tag for all stats.

After this PR merged, I will add `wasm` prefix to default stats match and add `wasm_filter` to tag match, so that stats in this PR and ones from upstream (wasm_vm, etc) could be exposed by default.

### DIFF
--- a/extensions/stackdriver/common/metrics.cc
+++ b/extensions/stackdriver/common/metrics.cc
@@ -1,0 +1,59 @@
+/* Copyright 2020 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "extensions/stackdriver/common/metrics.h"
+
+#ifdef NULL_PLUGIN
+namespace Envoy {
+namespace Extensions {
+namespace Common {
+namespace Wasm {
+namespace Null {
+namespace Plugin {
+
+#endif
+
+namespace Extensions {
+namespace Stackdriver {
+namespace Common {
+
+// NOTE: every call to this function will alloc a new metric object. Though this
+// object is the same in all calls, it cannot be extracted out to a global variable.
+// The metric object caches all fully resolved metrics and won't make define
+// metrics abi call when the same resolved metrics are defined. In Null VM case,
+// since global objects are shared between root contexts in base VM and in thread
+// local VM, and the same metrics definition code path will excercised in both VMs (we don't really have a way to distinguish base VM and thread local VM), the cache will make thread local root context miss metric definition
+// at host side. In real Wasm VM, this can be extracted to a global var.
+uint32_t newExportCallMetric(const std::string& type, bool success) {
+  Metric export_call(MetricType::Counter, "stackdriver_export_call",
+                    {MetricTag{"wasm_filter", MetricTag::TagType::String},
+                      MetricTag{"type", MetricTag::TagType::String},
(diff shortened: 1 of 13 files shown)


extensions/stackdriver/edges/mesh_edges_service_client.cc
```diff
@@ -50,13 +50,23 @@ MeshEdgesServiceClientImpl::MeshEdgesServiceClientImpl(
     RootContext* root_context,
     const ::Extensions::Stackdriver::Common::StackdriverStubOption& stub_option)
     : context_(root_context) {
-  success_callback_ = [](size_t) {
+  Metric export_call(MetricType::Counter, "export_call",
```
douglas-reid: should the metric be defined elsewhere, as it is used in the log exporter with the same definition?

bianpengyuan: Make sense. Done.

bianpengyuan:
@istio/wg-policies-and-telemetry-maintainers May I get some review on this PR? Thanks!

extensions/stackdriver/common/metrics.cc
```diff
@@ -0,0 +1,62 @@
...
+// NOTE: every call to this function will alloc a new metric object. Though this
+// object is the same in all calls, it cannot be extracted out to a global
+// variable. The metric object caches all fully resolved metrics and won't make
+// define metrics abi call when the same resolved metrics are defined. In Null
+// VM case, since global objects are shared between root contexts in base VM and
+// in thread local VM, and the same metrics definition code path will excercised
+// in both VMs (we don't really have a way to distinguish base VM and thread
+// local VM), the cache will make thread local root context miss metric
+// definition at host side. In real Wasm VM, this can be extracted to a global
+// var.
```
kyessenov: I'm not sure I understand. Is there a performance cost and difference between null VM and v8? I thought the cache is on the host side.

bianpengyuan: Performance is not a concern, these function will only be called during initialization. Essentially what causes problem is the asymmetric ownership of metric object: Inside VM metric is not attached to a specific context, but it attaches to a VM/root context in host. This should be fine in a real wasm VM.

I originally make it a static global var `static Metric export_call(...)`, and inside the metric object, there is a map to keep currently fully resolved metrics: https://github.com/envoyproxy/envoy-wasm/blob/f6d0282f7ab42dfd215ae59112edbab3e70f2d3b/api/wasm/cpp/proxy_wasm_api.h#L831 What happened is that in nullvm case, thread local VM and main thread base VM share the same global metric object, [the metric id map](https://github.com/envoyproxy/envoy-wasm/blob/f6d0282f7ab42dfd215ae59112edbab3e70f2d3b/api/wasm/cpp/proxy_wasm_api.h#L831) is filled during base VM root context creation and [blocks](https://github.com/envoyproxy/envoy-wasm/blob/f6d0282f7ab42dfd215ae59112edbab3e70f2d3b/api/wasm/cpp/proxy_wasm_api.h#L917) proxy_define_metric ABI call in thread local VM root context, where metric is really incremented. It is not going to be a problem for real wasm VM because of memory isolation.

Hope this makes sense, let me see how can i make the comment more clear.

bianpengyuan: I updated the comment. Hope it is clearer than the original one, or at least not making it more confusing. :p

extensions/stackdriver/common/metrics.cc
```diff
@@ -0,0 +1,61 @@
...
+namespace Stackdriver {
+namespace Common {
+
+uint32_t newExportCallMetric(const std::string& type, bool success) {
+  // NOTE: export_call cannot be a static global object, because in Nullvm,
+  // global metric is shared by base VM and thread local VM, but at host side,
+  // metrics are attached to a specific VM/root context. Since (1) metric object
+  // keeps an internal map which records all fully resolved metrics and avoid
+  // define metric ABI call when the same metric are seen (2) base VM always
+  // initiliazing before thread local VM, sharing a global metric object between
```
gargnupur: nit: initializes?

PR #2784, opened 2020-03-25
Make wasm sdk image configurable

bianpengyuan:

**Release note**:

```release-note
```

### DIFF
--- a/extensions/metadata_exchange/build_wasm.sh
+++ b/extensions/metadata_exchange/build_wasm.sh
@@ -16,6 +16,7 @@
 
 set -e
 
-source ../../scripts/build_wasm.inc
+WS=$(bazel info workspace)
+source ${WS}/scripts/build_wasm.inc
 
 build_wasm
--- a/extensions/metadata_exchange/build_wasm.sh
+++ b/extensions/metadata_exchange/build_wasm.sh
@@ -16,5 +16,6 @@
 
 set -e
 
-docker run -e uid="$(id -u)" -e gid="$(id -g)" -v $PWD:/work -w /work -v $(realpath $PWD/../../extensions):/work/extensions gcr.io/istio-testing/wasmsdk:v3 bash /build_wasm.sh
+WASM_SDK_IMAGE=${WASM_SDK_IMAGE:=gcr.io/istio-testing/wasmsdk}
+docker run -e uid="$(id -u)" -e gid="$(id -g)" -v $PWD:/work -w /work -v $(realpath $PWD/../../extensions):/work/extensions ${WASM_SDK_IMAGE}:v3 bash /build_wasm.sh
 rmdir extensions
--- a/extensions/metadata_exchange/build_wasm.sh
+++ b/extensions/metadata_exchange/build_wasm.sh
@@ -17,3 +17,5 @@
 set -e
 
 source ../../scripts/build_wasm.inc
+
+build_wasm
--- a/extensions/metadata_exchange/build_wasm.sh
+++ b/extensions/metadata_exchange/build_wasm.sh
@@ -16,6 +16,11 @@
 
 set -e
 
+# Get SHA of envoy-wasm repo and use that as wasm sdk image tag.
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
+WORKSPACE=${ROOT}/WORKSPACE
+WASM_SDK_TAG="$(grep -Pom1 "^ENVOY_SHA = \"\K[a-zA-Z0-9]{40}" "${WORKSPACE}")"
 WASM_SDK_IMAGE=${WASM_SDK_IMAGE:=gcr.io/istio-testing/wasmsdk}
-docker run -e uid="$(id -u)" -e gid="$(id -g)" -v $PWD:/work -w /work -v $(realpath $PWD/../../extensions):/work/extensions ${WASM_SDK_IMAGE}:v3 bash /build_wasm.sh
+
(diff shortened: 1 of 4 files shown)


scripts/generate-wasm.sh
```diff
@@ -46,14 +46,14 @@ done
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
 WORKSPACE=${ROOT}/WORKSPACE
 ENVOY_SHA="$(grep -Pom1 "^ENVOY_SHA = \"\K[a-zA-Z0-9]{40}" "${WORKSPACE}")"
-IMAGE=gcr.io/istio-testing/wasmsdk
+WASM_SDK_IMAGE=${WASM_SDK_IMAGE:=gcr.io/istio-testing/wasmsdk}
 TAG=${ENVOY_SHA}
 
 # Try pull wasm builder image.
-docker pull ${IMAGE}:${TAG} || echo "${IMAGE}:${TAG} does not exist"
+docker pull ${WASM_IMAGE}:${TAG} || echo "${WASM_SDK_IMAGE}:${TAG} does not exist"
```
gargnupur: This should be WASM_SDK_IMAGE?

scripts/generate-wasm.sh
```diff
@@ -69,15 +69,15 @@ if [[ "$(docker images -q ${IMAGE}:${TAG} 2> /dev/null)" == "" ]]; then
...
   if [[ ${PUSH_DOCKER_IMAGE} == 1 ]]; then
-    docker push ${IMAGE}:${TAG} || "fail to push to gcr.io/istio-testing hub"
+    docker push ${WASM_SDK_IMAGE}:${TAG} || "fail to push to gcr.io/istio-testing hub"
   fi
 fi
 
 # Regenerate all wasm plugins and compare diffs
 # Tag image to v3, which is what used by all build wasm script.
-docker tag ${IMAGE}:${TAG} ${IMAGE}:v3
+docker tag ${WASM_SDK_IMAGE}:${TAG} ${WASM_SDK_IMAGE}:v3
```
kyessenov: We should probably use istio/proxy version instead of v3. We may need to re-compile old Wasm modules from older releases for a fix.

scripts/generate-wasm.sh
bianpengyuan: Good catch. Done.

scripts/generate-wasm.sh
bianpengyuan: Yeah the tag leads to confusion here. The wasm sdk build script actually does retag: https://github.com/istio/proxy/blob/96399ff4f54adac9152a07a0674c5e40cfbb09c3/scripts/generate-wasm.sh#L80 and sdk image is actually versioned with envoy sha.

I've updated the script to use envoy sha as in wasm gen script. I think it would help to clarify the thing a bit.

extensions/metadata_exchange/build_wasm.sh
```diff
@@ -16,5 +16,11 @@
 
 set -e
 
-docker run -e uid="$(id -u)" -e gid="$(id -g)" -v $PWD:/work -w /work -v $(realpath $PWD/../../extensions):/work/extensions gcr.io/istio-testing/wasmsdk:v3 bash /build_wasm.sh
+# Get SHA of envoy-wasm repo and use that as wasm sdk image tag.
```
lizan: Can we extract this part to a common util script?

bianpengyuan: Done.

scripts/build_wasm.inc
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Get SHA of envoy-wasm repo and use that as wasm sdk image tag.
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+WORKSPACE=${ROOT}/WORKSPACE
+WASM_SDK_TAG="$(grep -Pom1 "^ENVOY_SHA = \"\K[a-zA-Z0-9]{40}" "${WORKSPACE}")"
+WASM_SDK_IMAGE=${WASM_SDK_IMAGE:=gcr.io/istio-testing/wasmsdk}
+
+docker run -e uid="$(id -u)" -e gid="$(id -g)" -v $PWD:/work -w /work -v $(realpath $PWD/../../extensions):/work/extensions ${WASM_SDK_IMAGE}:${WASM_SDK_TAG} bash /build_wasm.sh
```
mandarjog: Include / sourced files should not have side efffects.
You can make define a build_wasm function and then call it from the including script.

mandarjog:
@PiotrSikora where do we configure memory allocated to every vm ?

scripts/build_wasm.inc
bianpengyuan: Done.

bianpengyuan:
Ping, could anyone help me approve this? thanks!

bianpengyuan:
@gargnupur @kyessenov ptal. thanks!

extensions/metadata_exchange/build_wasm.sh
```diff
@@ -16,5 +16,6 @@
 
 set -e
 
-docker run -e uid="$(id -u)" -e gid="$(id -g)" -v $PWD:/work -w /work -v $(realpath $PWD/../../extensions):/work/extensions gcr.io/istio-testing/wasmsdk:v3 bash /build_wasm.sh
-rmdir extensions
+source ../../scripts/build_wasm.inc
```
kyessenov, line 19: nit: we should not assume that we are in the directory of the script.

PR #2785, opened 2020-03-25
Add configurable monitoring endpoint

bianpengyuan:
Add an option to override default SD monitoring endpoint. There are two ways to use it:

1) Use [bootstrap customization](https://github.com/istio/istio/tree/master/samples/custom-bootstrap). This is suitable for per pod enablement. Specifically apply a bootstrap customization configmap similar to [this example](https://github.com/istio/istio/blob/018d5f92acdc19481d505637ca70c3174a338888/tests/integration/telemetry/stackdriver/testdata/custom_bootstrap.yaml.tmpl), with metadata section being
```
"metadata": {
    "STACKDRIVER_MONITORING_ENDPOINT": "staging_endpoint",
}
```
Then add override annotation to a pod: `sidecar.istio.io/bootstrapOverride: "stackdriver-bootstrap-config"`

2) To apply this option globally, a more involved approach is to edit side injection config map and add `ISTIO_META_STACKDRIVER_MONITORING_ENDPOINT` env var to sidecar template, which will be relayed to sidecar node metadata, then restart sidecar injector and workload pod to pick the sidecar spec.

### DIFF
--- a/extensions/stackdriver/metric/registry.cc
+++ b/extensions/stackdriver/metric/registry.cc
@@ -108,9 +108,18 @@ StackdriverOptions getStackdriverOptions(
         std::make_unique<GoogleUserProjHeaderInterceptorFactory>(
             options.project_id);
     creators.push_back(std::move(header_factory));
+    // When STS is turned on, first check if secure_endpoint is set or not,
+    // which indicates whether this is for testing senario. If not set, check
+    // for monitoring_endpoint override, which indicates a different SD backend
+    // endpoint, such as staging.
+    std::string monitoring_endpoint = stub_option.default_endpoint;
+    if (!stub_option.secure_endpoint.empty()) {
+      monitoring_endpoint = stub_option.secure_endpoint;
+    } else if (!stub_option.monitoring_endpoint.empty()) {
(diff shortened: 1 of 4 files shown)


extensions/stackdriver/metric/registry.cc
```diff
@@ -125,6 +125,11 @@ StackdriverOptions getStackdriverOptions(
                                        grpc::InsecureChannelCredentials());
     options.metric_service_stub =
         google::monitoring::v3::MetricService::NewStub(channel);
+  } else if (!stub_option.monitoring_endpoint.empty()) {
+    auto channel = ::grpc::CreateChannel(stub_option.monitoring_endpoint,
```
mandarjog: So this is either default creds or sts?
This is a good first step but I think we need both.

bianpengyuan: Done.

douglas-reid:
Can we update the PR description and add an example of how to use in there (just for clarity)?

bianpengyuan:
@douglas-reid Added the instruction to use the option.

PR #2786, opened 2020-03-25
Add tracing support in fake stackdriver

gargnupur:
Add tracing support in fake stackdriver


test/envoye2e/driver/fake_stackdriver.go
```diff
@@ -191,9 +204,111 @@ func (s *LoggingServer) GetLogEntries(w http.ResponseWriter, req *http.Request)
...
+		var err error
+		if spanID, err = strconv.Atoi(span.SpanId); err != nil {
+			log.Printf("err: %v", err)
+			continue
+		}
+		if span.ParentSpanId == "" {
+			traceID, _ := randomHex(32)
+			m = append(m, &cloudtracev1.Trace{
+				ProjectId: req.Name,
+				TraceId:   traceID,
```
bianpengyuan: Trace ID should already be embedded in [span name](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2#google.devtools.cloudtrace.v2.Span), no? I think it needs to use that to identify spans that belong to the same trace.

test/envoye2e/driver/fake_stackdriver.go
```diff
@@ -191,9 +204,111 @@ func (s *LoggingServer) GetLogEntries(w http.ResponseWriter, req *http.Request)
...
+			continue
+		}
+		if parentSpanID, err = strconv.Atoi(span.ParentSpanId); err != nil {
+			log.Printf("err: %v", err)
+			continue
+		}
+
+		for _, trace := range m {
+			for _, parentSpan := range trace.Spans {
+				if parentSpan.SpanId == uint64(parentSpanID) {
```
bianpengyuan: This logic might not work properly, since batch with child span could arrive at server earlier than the parent span, which will cause the span being dropped. So I think you'd have to use trace id inside span name to find out spans belong to the same trace.

test/envoye2e/driver/fake_stackdriver.go
gargnupur: Thanks.. I didn't know that.. this would simplify logic for identifying child spans too..

test/envoye2e/driver/fake_stackdriver.go
gargnupur: Replied above.. thanks!

test/envoye2e/driver/fake_stackdriver.go
```diff
@@ -191,9 +203,98 @@ func (s *LoggingServer) GetLogEntries(w http.ResponseWriter, req *http.Request)
...
+	return uint64(dec), nil
+}
+
+// BatchWriteSpans implements BatchWriteSpans method.
+func (s *TracesServer) BatchWriteSpans(ctx context.Context, req *cloudtracev2.BatchWriteSpansRequest) (*empty.Empty, error) {
+	log.Printf("receive BatchWriteSpansRequest %+v", *req)
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	// Map that stores individual traces.
+	m := make(map[string]*cloudtracev1.Trace)
```
bianpengyuan: This does not seem to be right. Spans from one trace won't be included in the same batchWriteSpansRequest. I think what you probably should do is keep the map in TracesServer, add new span to it on every batch write reqeust, and create a `listTracesResp` with the map at the end of this method or in the ListTraces method. This way response would have all the spans received correctly organized (spans from the same trace end up in the same trace object in ListTraceResp).

gargnupur: Thanks.. so if spans from one trace can come in multiple requests.. it can happen that List call is called before all the spans are received...
In that case, before sending a trace in listtrace we should make sure that atleast parent trace is received?

test/envoye2e/driver/fake_stackdriver.go
```diff
@@ -191,9 +204,98 @@ func (s *LoggingServer) GetLogEntries(w http.ResponseWriter, req *http.Request)
...
+	}
+	return dec, nil
+}
+
+// BatchWriteSpans implements BatchWriteSpans method.
+func (s *TracesServer) BatchWriteSpans(ctx context.Context, req *cloudtracev2.BatchWriteSpansRequest) (*empty.Empty, error) {
+	log.Printf("receive BatchWriteSpansRequest %+v", *req)
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	s.traceMap = make(map[string]*cloudtracev1.Trace)
```
bianpengyuan: This should not be reset on every request.

test/envoye2e/driver/fake_stackdriver.go
```diff
@@ -191,9 +204,98 @@ func (s *LoggingServer) GetLogEntries(w http.ResponseWriter, req *http.Request)
 	}
 }
 
+// ListTraces implements ListTraces method.
+func (s *TracesServer) ListTraces(context.Context, *cloudtracev1.ListTracesRequest) (*cloudtracev1.ListTracesResponse, error) {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	for traceID, trace := range s.traceMap {
+		s.listTracesResp.Traces = append(s.listTracesResp.Traces, trace)
+		delete(s.traceMap, traceID)
```
bianpengyuan: No need to delete it I think. This is just for testing, no need to prevent leak.

bianpengyuan:
Also do you need help to push the image? CI does not have automatic set up to push the fakesd image.

gargnupur:
> Sorry for the delay. I won't get any github notification on new commit to a PR, so better reply to the comments if you want to turn around the PR.

No problem at all.. sure.. from next time :)

gargnupur:
> Also do you need help to push the image? CI does not have automatic set up to push the fakesd image.

@bianpengyuan : yes that would be great! I don't think I have access to push the image to istio-testing

bianpengyuan:
I will push after this is merged.

gargnupur:
@bianpengyuan : please help update the fake stackdriver image in istio-testing repo.

bianpengyuan:
Done: gcr.io/istio-testing/fake-stackdriver:3.0

PR #2787, opened 2020-03-26
deps: update go-control-plane, switch to v3 xDS

kyessenov:

v2 xDS is about to be frozen. Opening up the testing to v3 before that happens.


kyessenov:
PTAL @gargnupur

kyessenov:
fyi @howardjohn this is what v3 xDS update looks like

test/envoye2e/stackdriver_plugin/stackdriver_plugin_test.go
```diff
@@ -32,83 +32,104 @@ import (
...
-  config:
-    config:
-      vm_config:
-        runtime: "envoy.wasm.runtime.null"
-        code:
-          local: { inline_string: "envoy.wasm.metadata_exchange" }
-      configuration: "test"
+  typed_config:
+    "@type": type.googleapis.com/udpa.type.v1.TypedStruct
+    type_url: envoy.extensions.filters.http.wasm.v3.Wasm
```
howardjohn: not really related to this PR, but do we need to import these APIs into pilot so we can use them in EnvoyFilters?

kyessenov: Eventually. They can still change.

kyessenov:
May I get a review @gargnupur @bianpengyuan . It's easier to review if you hide whitespaces in the review options.

Makefile.core.mk
```diff
@@ -81,7 +81,7 @@ build_wasm:
 check_wasm:
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_DEV) //src/envoy:envoy
 	./scripts/generate-wasm.sh -b
-	env ENVOY_PATH=$(BAZEL_ENVOY_PATH) GO111MODULE=on WASM=true go test ./test/envoye2e/stats/...
+	env ENVOY_PATH=$(BAZEL_ENVOY_PATH) GO111MODULE=on WASM=true go test ./test/envoye2e/stats_plugin/...
```
bianpengyuan, line 84: Why switch to stats_plugin? I don't think it has any tests for v8. Also since you are here, could you update the `skipWasm` logic:
https://github.com/istio/proxy/blob/706c98ff8b7661e70ceb6e3ce1909ea62005ebd3/test/envoye2e/stats/stats_xds_test.go#L138

Not sure why I wrote condition as that but it will make the test being bypassed when running with nullvm. It should be
```
if os.Getenv("WASM") == "" && runtime == "envoy.wasm.runtime.v8" {
...
}
```

So that it will only bypass when wasm file is not generated and run with v8.

kyessenov: Fixed the test. I'm trying to organize tests by a plugin. We'll need to test others with v8 so we need a better solution that separating tests by runtime. It only takes a few seconds to re-test stats tests and they're not flakey.

bianpengyuan: Ok, but you should not switch test target from `/stats` to `/stats_plugin` for now, right? I don't think there is any test under`/stats_plugin` using v8, or do I miss them somehow?

kyessenov: I moved the test file to stats_plugin using git mv.

bianpengyuan: Ah, ok, I missed the it. thanks!

PR #2790, opened 2020-03-29
call proxy_done only when there is no inflight call

bianpengyuan:
~~Probably the root cause of https://github.com/envoyproxy/envoy-wasm/issues/465~~ actually the test we have unlike trigger the bug fixed in this PR. So root cause of https://github.com/envoyproxy/envoy-wasm/issues/465 is still unclear.


extensions/stackdriver/log/exporter.cc
```diff
@@ -48,7 +48,8 @@ ExporterImpl::ExporterImpl(
   context_ = root_context;
   success_callback_ = [this](size_t) {
     logDebug("successfully sent Stackdriver logging request");
-    if (is_on_done_) {
+    in_flight_export_call_ -= 1;
+    if (in_flight_export_call_ <= 0 && is_on_done_) {
```
mandarjog: < 0 should be an assertion no?
Not suggesting a real assert, but at least a warning ?

bianpengyuan: Added a warn log about negative in flight request.

bianpengyuan:
https://github.com/envoyproxy/envoy-wasm/issues/465 starts to make test with slow proxy flaky. There seems to be some timing issue between envoy exit and on done callback. Does envoy host guarantee that deferred deletion will wait for async call to finish?

bianpengyuan:
@mandarjog ptal, thanks!

bianpengyuan:
@lambdai thanks! fixed. Though this would not cause not actual bug since its value is always set before calling out (https://github.com/istio/proxy/blob/b2080ee197a0586dcc6613f60b35882315f8e1db/extensions/stackdriver/log/exporter.cc#L82).

extensions/stackdriver/log/exporter.cc
```diff
@@ -81,10 +89,15 @@ void ExporterImpl::exportLogs(
...
-                             kGoogleWriteLogEntriesMethod, *req,
-                             kDefaultTimeoutMillisecond, success_callback_,
-                             failure_callback_);
+    auto result = context_->grpcSimpleCall(
+        grpc_service_string_, kGoogleLoggingService,
+        kGoogleWriteLogEntriesMethod, *req, kDefaultTimeoutMillisecond,
+        success_callback_, failure_callback_);
+    if (result != WasmResult::Ok) {
+      LOG_WARN("failed to make stackdriver logging export call");
+      break;
```
lambdai, line 98: It is not clear to me if we should use break or continue. Both are good in terms of the counter. I will let you make the final call.

bianpengyuan: grpcSimpleCall returns non-ok status if there is something bad in Grpc service configuration, which is unlikely going to work for the following requests.

bianpengyuan:
@mandarjog @gargnupur @kyessenov ptal, thanks!

PR #2792, opened 2020-04-03
Fix telemetry v2 for blackhole and passthrough clusters

jacob-delgado:

Fix blackhole and passthrough telemetry regression introduced in telemetry v2.

Make sure the destination_service_name label for istio_request metrics are set apppropriately to BlackHoleCluster and PassthroughCluster

Check the name of the route to see if it is associated with blackhole
and passthrough clusters. Change the destination_service_name
label appropriately

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/istio/istio/issues/21385

Associated with https://github.com/istio/istio/pull/22693

**Release note**:

```release-note
Fix http blackhole and passthrough regression for telemetry v2
```

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -45,6 +45,8 @@ namespace Common {
 
 const char kBlackHoleCluster[] = "BlackHoleCluster";
 const char kPassThroughCluster[] = "PassthroughCluster";
+const char kBlackHoleRouteName[] = "black_hole_route";
+const char kPassThroughRouteName[] = "pass_through_route";
 const char kInboundPassthroughClusterIpv4[] = "InboundPassthroughClusterIpv4";
 const char kInboundPassthroughClusterIpv6[] = "InboundPassthroughClusterIpv6";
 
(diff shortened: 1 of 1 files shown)


extensions/common/context.cc
```diff
@@ -106,6 +108,16 @@ void getDestinationService(const std::string& dest_namespace,
...
                        : "unknown";
 
+  // override the cluster name if this is being sent to the
+  // blackhole or passthrough cluster
+  std::string route_name;
+  getValue({"route_name"}, &route_name);
+  if (route_name == kBlackHoleRouteName) {
+      cluster_name = kBlackHoleCluster;
+  } else if (route_name == kPassThroughRouteName) {
+      cluster_name = kPassThroughCluster;
```
nrjpoddar, line 118: Let's get this in for now and unblock users.

We can implement the generic metadata approach as a follow-on.

jacob-delgado: I'll write an issue for the metadata approach and put it for 1.6 milestone.

PR #2794, opened 2020-04-03
replace protobuffers with flatbuffers at runtime

kyessenov:
This PR replaces `NodeInfo` protobuf with a flatbuffer. Flatbuffers allow zero-cost access to structured data (no serialization, no copying), which suits well for Wasm data transfer API. It also allows us to remove all the caches on the consuming side (stats, stackdriver), and move the cache to the production side (metadata_exchange). The write cache could be further eliminated if we used flatbuffers on the wire.

Micro-benchmark shows about the same cost of using flatbuffers vs cache-enabled protobuffers:
1. Read protobuf with a cache: 110ns, read flatbuffer without a cache: 102ns
2. Write protobuf raw bytes: 303ns, write parsed flatbuffer with a cache: 370ns.

The end goal of the effort is to make the node info a public API so that extensions can consume it directly without complex caches. This also applies to native, built-in Envoy extensions via reflection.


kyessenov:
@PiotrSikora @jplevyak please help resolving libc++ ABI issue. I don't know what it wants from the ABI. It works fine for me locally.

lizan:
@kyessenov try add `--host_linkopts=-pthread` to bazel.

extensions/common/node_info.fbs
```diff
@@ -0,0 +1,20 @@
+namespace Wasm.Common;
```
lizan, line 1: Can this file generated from `flatc`? It can take `.proto` as input and covert to `fbs`.

kyessenov: I need the key attribute. I can try it but it's a bit more opaque to through proto.

prow/proxy-common.inc
```diff
@@ -28,7 +28,7 @@ GOPATH=/home/prow/go
 ROOT=/go/src
 
 # Configure available resources and disable IPv6 tests.
-export BAZEL_BUILD_ARGS="--local_ram_resources=12288 --local_cpu_resources=8 --verbose_failures --test_env=ENVOY_IP_TEST_VERSIONS=v4only --test_output=errors"
+export BAZEL_BUILD_ARGS="--local_ram_resources=12288 --local_cpu_resources=8 --verbose_failures --test_env=ENVOY_IP_TEST_VERSIONS=v4only --test_output=errors --host_linkopt=-pthread"
```
PiotrSikora, line 31: That should be added to `.bazelrc`, otherwise it's only applied to Prow.

extensions/common/context.h
```diff
@@ -202,5 +212,22 @@ google::protobuf::util::Status extractNodeMetadataValue(
...
+  void updateSize(uint32_t max_size) { max_size_ = max_size; }
+  // Updates the filter state for "key" using this cache or the provided peer id
+  // and content.
+  bool updatePeer(absl::string_view key, absl::string_view peer_id,
+                  absl::string_view peer_content);
+  // Override for custom behavior
+  virtual void updateState(absl::string_view key, absl::string_view value) = 0;
+
+ protected:
+  std::unordered_map<std::string, std::string> cache_;
```
lambdai, line 228: any reason not to use flatmap?

extensions/common/context.h
```diff
@@ -202,5 +212,22 @@ google::protobuf::util::Status extractNodeMetadataValue(
     const google::protobuf::Struct& node_metadata,
     google::protobuf::Struct* metadata);
 
+class FlatNodeCache {
+ public:
+  FlatNodeCache() {}
+  virtual ~FlatNodeCache() {}
```
lambdai, line 218: nit: = default;

extensions/common/context.h
```diff
@@ -202,5 +212,22 @@ google::protobuf::util::Status extractNodeMetadataValue(
     const google::protobuf::Struct& node_metadata,
     google::protobuf::Struct* metadata);
 
+class FlatNodeCache {
+ public:
+  FlatNodeCache() {}
```
lambdai, line 217: nit: = default;
also, if updateSize is required it is probably better to set max_size in ctor. Otherwise, set max_size_ to a reasonable value to avoid indeterministic value.

extensions/common/context.cc
```diff
@@ -355,5 +433,43 @@ google::protobuf::util::Status extractNodeMetadataValue(
   return google::protobuf::util::Status(google::protobuf::util::error::OK, "");
 }
 
+bool FlatNodeCache::updatePeer(absl::string_view key, absl::string_view peer_id,
+                               absl::string_view peer_content) {
+  std::string id = std::string(peer_id);
```
lambdai, line 438: if flatmap is used, cache_ should support promoting from string_view to string

mandarjog:
@kyessenov fb write is more expensive than Porto write. If cache is handed on the production side, hopefully we will not do cache write very often.
Once the peers that connect to a specific proxy stabilize, we need not do any writes at all. This is great. Json/ struct support for fb is not very good languages other than c++ though.

extensions/common/context.cc
bianpengyuan: IIRC, flatmap cannot easily run in a envoy wasm module, due to missing system call. https://github.com/istio/proxy/commit/35f3727fdd2839146bae2dcc05949b58dd38eead#diff-70887b3a407a875b71e9cb77b627856eL366

kyessenov:
We could also use flatbuffers on the wire but that's more controversial. We wouldn't need any caches at all. JSON doesn't map integers well, so it's not really useful for low-level ABI.

extensions/common/context.cc
kyessenov: cc @jplevyak. I know that absl flat map is faster here but we can't use it.

extensions/stats/plugin.cc
```diff
@@ -47,54 +47,63 @@ constexpr long long kDefaultTCPReportDurationMilliseconds = 15000;  // 15s
 
 namespace {
 
+// Efficient way to assign flatbuffer to a vector of strings
+#define FB_ASSIGN(name, v) \
```
bianpengyuan, line 51: This will make istio_dimensions_ take over ownership of a subpart of flat node? So this assumes each dimension will only be set once in each onLog call, right?

kyessenov:
@bianpengyuan I think the main difference with Wasm measurement is the memcopy of the flatbuffer into the VM. The rest should be the same (modulo compute differences with the VM). Note we also save on memcopy of the peer ID in some cases, so it's really not adding another memcopy from what I can tell. FWIW flatbuffer library is significantly simpler than protobuf so it should be in theory faster inside the VM (due to reduction in code cache sizes, less bloat). It's effectively a C struct with an extra vtable.

extensions/stats/plugin.cc
kyessenov: It will copy into the dimension vector. The vector owns all its content. We're just saving on local allocation (which C++17 gives for free awkward for now). It would be interesting to avoid even that copy but that requires extra care in ensuring the longevity of the flatbuffers.

extensions/metadata_exchange/plugin.cc
```diff
@@ -93,23 +93,29 @@ bool PluginRootContext::onConfigure(size_t) {
...
+  json_options.ignore_unknown_fields = true;
+  metadata_exchange::PluginConfig config;
+  const auto status = google::protobuf::util::JsonStringToMessage(
+      configuration->toString(), &config, json_options);
+  if (!status.ok()) {
+    logWarn(absl::StrCat("cannot parse plugin configuration JSON string ",
+                         configuration->toString()));
+    return false;
+  }
+  updateSize(config.max_peer_cache_size());
```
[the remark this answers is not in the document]
kyessenov: Is there a custom default apart from 0? It looks like it defaults to 0 if not set.

kyessenov:
plugin.wasm for stats is 1.1M, mx is 1022K

kyessenov:
I'd guess that to complete remove proto we'd need to move the extraction function to a host.

extensions/metadata_exchange/plugin.cc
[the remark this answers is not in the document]
kyessenov: What do you think should the default be?

bianpengyuan:
Yeah I thought maybe stats module will be smaller, but seems like we are still using protobuf for config parsing. We could get rid of it though. It does not bring value.

extensions/metadata_exchange/plugin.cc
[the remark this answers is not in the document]
kyessenov: Set it to 500 and used the wrapper to detect presence of the override. Thanks for noticing it!

kyessenov:
I think it's possible to eliminate protobuf entirely. We'd need to 1) fix up some protos on ABI (@jplevyak / @PiotrSikora ) for gRPC service etc; 2) pick a reasonable JSON library since we use proto for JSON parsing.

extensions/common/context.h
kyessenov: removed the super class altogether since it's only used once.

extensions/metadata_exchange/plugin.h
```diff
@@ -63,11 +64,16 @@ class PluginRootContext : public RootContext {
...
   StringView nodeId() { return node_id_; };
+  bool updatePeer(StringView key, StringView peer_id, StringView peer_header);
 
  private:
   void updateMetadataValue();
   std::string metadata_value_;
   std::string node_id_;
+
+  // maps peer ID to the decoded peer flat buffer
+  std::unordered_map<std::string, std::string> cache_;
```
gargnupur, line 75: should this be an ordered map in terms of when the entry was added, so that when cache is resized it can be FIFO?

kyessenov: Yeah, I just copied what was there in node_info_cache. I agree that it should not randomly remove 1/4 elements. I can take that as a task for later.

kyessenov: I split the root context for MX into outbound and inbound to at least get partitioned cache.

kyessenov:
All tests passed, so it's up to reviewers to approve it.

extensions/common/node_info.fbs
kyessenov: flatc doesn't seem to support protobuf map. I think we can just delete proto version after the transition is doe.

kyessenov:
Will follow up with clean up of proto code after a few weeks. The goal would be to get rid of libprotobuf in stats filter.

PR #2795, opened 2020-04-07
feat(stackdriver): use app container for k8s_container when appropriate

douglas-reid:
Takes advantage of the `APP_CONTAINERS` metadata that has been added to allow the Stackdriver extension to choose the application container for `k8s_container` Monitored Resources. This will more appropriately match traffic to the actual serving container in the vast majority of use cases.

Potential refinements beyond this PR:
- filter out traffic to known Istio proxy ports to be `istio-proxy` (health check, admin).
- use `unknown` in cases where the the num of app containers is >1.


extensions/stackdriver/common/utils.cc
```diff
@@ -22,6 +22,19 @@ namespace Extensions {
 namespace Stackdriver {
 namespace Common {
 
+namespace {
+
+const std::string getContainerName(
```
bianpengyuan, line 27: nit: make return value `const std::string&` and make type of returned value (line 116) to be `auto&` to save a copy.

bianpengyuan:
Oh looks like char[] cannot be returned as string&.. Sorry for the bad suggestion! Let's switch it back to the original code.

douglas-reid:
> Oh looks like char[] cannot be returned as string&.. Sorry for the bad suggestion! Let's switch it back to the original code.

Simple enough. Done.

extensions/stackdriver/common/utils.cc
kyessenov: we don't even need a function for this since it's used in one place only.

kyessenov:
Letting it merge so I can merge conflicts.

PR #2797, opened 2020-04-09
Attempt to fix post-submit #2

kyessenov:
re-enable RBE
build before pushing so that `k8-fastbuild` is populated with generated files.


kyessenov:
cc @bianpengyuan if RBE is set up the same way on pre-post submits, then the issue I think is that we switch bazel options in post-submit to opt/dbg inside scripts. the easy workaround is to do just a normal build first (fastbuild).

kyessenov:
This fixed post-submit https://prow.k8s.io/view/gcs/istio-prow/logs/release_proxy_postsubmit/122

PR #2798, opened 2020-04-09
fix(edges): increase timeout for meshtelemetry API calls

douglas-reid:
Move the default timeout for gRPC calls into meshtelemetry to `60s` from `10s`.

Reports of `DEADLINE_EXCEEDED` returned from the service under moderate load indicate that `10s` is not adequate. Calls to meshtelemetry are not user-facing, and they consist of batches of data that need to be processed. Given that the calls are every `1m` and every `10m`, having a longer deadline seems reasonable.

### DIFF
--- a/extensions/stackdriver/edges/mesh_edges_service_client.cc
+++ b/extensions/stackdriver/edges/mesh_edges_service_client.cc
@@ -38,7 +38,7 @@ using Envoy::Extensions::Common::Wasm::Null::Plugin::StringView;
 constexpr char kMeshEdgesService[] =
     "google.cloud.meshtelemetry.v1alpha1.MeshEdgesService";
 constexpr char kReportTrafficAssertions[] = "ReportTrafficAssertions";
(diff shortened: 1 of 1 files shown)


douglas-reid:
> Is it still an issue that this time out is not respected by the runtime?

I can't find any solid evidence that the timeout is not respected by the runtime. That was my original guess, but I think it was just an artifact of the test setup.

Playing around with it locally, I see the expected deadline behavior.

The logs I have access to only show the error, but not the duration, so it maybe that `5s` was a rough guess based on charts. Since there is also server-side re-batching, it is unclear if one request was split into two or more requests that each took ~5s or similar.

mandarjog:
10s for a call that is not very expensive on the server side does not seem correct.
Sending 10k data should not take 10s.

douglas-reid:
@mandarjog perhaps, but we were seeing a bunch of deadline exceededs for a period of hours. these are batched operations that are then expanded and rebatched and so forth. the backend might be more expensive than we think. `60s` aligns with internal defaults for the underlying systems.

PR #2799, opened 2020-04-13
Request classification proto

mandarjog:
Add: proto definition for request classification.
Add: two examples for use.

Note: the configuration must be json at the point of delivery, not yaml.
Implementation not done yet.

### DIFF
--- a/extensions/classify/config.proto
+++ b/extensions/classify/config.proto
@@ -1,0 +1,158 @@
+/* Copyright 2020 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package classify;
+
+
+// Configuration to classify http traffic.
+// Traffic classification can be done on mutiple dimensions using request and
+// response headers and other measurements. 
+// The plugin must be configured as request, response or log based on
+// which other plugins need to consume the classfication results.
+// For exmaple: If Stats plugin uses `operation` dimension produced by the classification
+// plugin, then it must be executed in LOG phase *before* the Stats plugin.
+message PluginConfig {
+  // next id: 3
+  // Plugin phase when classification is done.
+  PluginPhase phase = 1;
+
+  // Request classification along multiple dimensions.
+  repeated Dimension dimensions = 2;
+}
+
+// Dimension configuration produces the named dimension with a value based
+// on the match clause in the value. 
+// The value matches are processed sequentially and the first value to match is assigned.
+message Dimension {
+  // For example `operation` is a dimension.
+  string name = 1;
+
+  // At most one of the following values is assigned to the dimension. 
+  repeated Value values = 2;
+}
+
+// Value assigned to the dimension if the request matches.
+message Value {
+  // Assign `value` to the enclosing dimension if match is true.
+  string value = 1;
+
+  // Match specification.
+  // Empty match always matches, which can be used to specify a default value when placed
+  // at the end of the list.
+  Match match = 2;
+}
+
+// Match condition based on request and response information.
+// Every match condition must match for the result to be true.
+message Match {
+  // invert the match condition.
+  bool invert = 1;
+
+  // Match uri.
+  StringMatch uri = 2;
+
+  // Match scheme: http or https.
+  string scheme = 3;
+
+  // Match method: GET, POST, HEAD, PUT, DELETE, ...
+  string method = 4;
+
+  // Match request headers.
+  // Not implemented.
+  // $hide_from_docs
+  map<string, StringMatch> request_headers = 5;
+  
+  // Match response headers.
+  // Not implemented.
+  // $hide_from_docs
+  map<string, StringMatch> response_headers = 6;
+
+  // Match query parameters.
+  // Not implemented.
+  // $hide_from_docs
+  map<string, StringMatch> query_params = 7;
+
+  // Match response code.
+  IntMatch response_code = 8;
+
+  // Match request size.
+  // Not implemented.
+  // $hide_from_docs
+  IntMatch request_size = 9;
+
+  // Match response size.
+  // Not implemented.
+  // $hide_from_docs
+  IntMatch response_size = 10;
+}
+
+// Describes how to match a given string.
+message StringMatch {
+  // default is case sensitive.
+  bool case_insensitive = 1;
+
+  oneof match_type {
+
+    // exact string match
+    string exact = 2;
+
+    // prefix-based match
+    string prefix = 3;
+
+    // RE2 style regex-based match (https://github.com/google/re2/wiki/Syntax).
+    string regex = 4;
+  }
+}
+
+// Describes how to match a given Integer.
+message IntMatch {
+  oneof match_type {
+
+    // exact Int match
+    int64 exact = 1;
+
+    // Range match
+    IntRange range = 2;
+  }
+}
+
+// IntRange matches a range of integers.
+message IntRange {
+  // Greater than or equal to this number.
+  int64 ge = 1;
+
+  // Less than or equal to this number.
+  int64 le = 2;
+}
+
+// PluginPhase determines when the classification is performed.
+enum PluginPhase {
+  // Classification is performed after the response is sent back to the client.
+  // All information about the request is available in this phase,
+  // and classification does not add to the latency observed by the client.
+  LOG = 0;
+  // Classification is performed during request processing, and adds to
+  // the request latency as observed by the client.
+  // Only request related information can be used in this phase.
+  // For example: you may not use response headers here.
+  // Attemping to use response headers in request phase will never result in a match.
+  // Not Implemented
+  // $hide_from_docs
+  REQUEST = 1;
+
+  // ADD RESPONSE=2 if there is a need.
+}
--- a/extensions/classify/config.proto
+++ b/extensions/classify/config.proto
@@ -17,18 +17,19 @@ syntax = "proto3";
 
 package classify;
 
+import "google/protobuf/duration.proto";
 
 // Configuration to classify http traffic.
 // Traffic classification can be done on mutiple dimensions using request and
 // response headers and other measurements. 
 // The plugin must be configured as request, response or log based on
 // which other plugins need to consume the classfication results.
-// For exmaple: If Stats plugin uses `operation` dimension produced by the classification
+// For example: If Stats plugin uses `operation` dimension produced by the classification
 // plugin, then it must be executed in LOG phase *before* the Stats plugin.
 message PluginConfig {
   // next id: 3
   // Plugin phase when classification is done.
-  PluginPhase phase = 1;
+  Phase phase = 1;
 
   // Request classification along multiple dimensions.
   repeated Dimension dimensions = 2;
@@ -62,14 +63,14 @@ message Match {
   // invert the match condition.
   bool invert = 1;
 
-  // Match uri.
-  StringMatch uri = 2;
+  // Match path.
+  repeated StringMatch path = 2;
 
-  // Match scheme: http or https.
-  string scheme = 3;
+  // Match scheme: http, https or unset.
+  Scheme scheme = 3;
 
   // Match method: GET, POST, HEAD, PUT, DELETE, ...
-  string method = 4;
+  repeated string method = 4;
 
   // Match request headers.
   // Not implemented.
@@ -87,17 +88,22 @@ message Match {
   map<string, StringMatch> query_params = 7;
 
   // Match response code.
-  IntMatch response_code = 8;
+  repeated IntMatch response_code = 8;
 
   // Match request size.
   // Not implemented.
   // $hide_from_docs
-  IntMatch request_size = 9;
+  repeated IntMatch request_size = 9;
 
   // Match response size.
   // Not implemented.
   // $hide_from_docs
-  IntMatch response_size = 10;
+  repeated IntMatch response_size = 10;
+
+  // Match response duration.
+  // Not implemented.
+  // $hide_from_docs
+  repeated DurationMatch response_duration = 11;
 }
 
 // Describes how to match a given string.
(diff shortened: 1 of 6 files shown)


mandarjog:
@douglas-reid @kyessenov @nrjpoddar PTAL the proto definitions and examples.

mandarjog:
@bianpengyuan look like a new lint rule?

‘ test/envoye2e/driver/fake_stackdriver.go:48:2: ST1003: struct field listTsResp should be listTSResp (stylecheck)
74
	listTsResp   monitoringpb.ListTimeSeriesResponse’

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+
+// Dimension configuration produces the named dimension with a value based
+// on the match clause in the value. 
+// The value matches are processed sequentially and the first value to match is assigned.
+message Dimension {
+  // For example `operation` is a dimension.
+  string name = 1;
+
+  // At most one of the following values is assigned to the dimension. 
+  repeated Value values = 2;
```
douglas-reid, line 45: I wonder: would this be clearer if it was `repeated Classifier classifiers = 2` ?

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Match condition based on request and response information.
+// Every match condition must match for the result to be true.
+message Match {
+  // invert the match condition.
+  bool invert = 1;
+
+  // Match uri.
+  StringMatch uri = 2;
```
douglas-reid, line 66: Should this be split into two: `host` and `path`  ?  Or even three: `host`, `path`, `query_params` ?

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+  // $hide_from_docs
+  map<string, StringMatch> query_params = 7;
+
+  // Match response code.
+  IntMatch response_code = 8;
+
+  // Match request size.
+  // Not implemented.
+  // $hide_from_docs
+  IntMatch request_size = 9;
```
douglas-reid, line 95: if we are classifying by size, will we ever want to enable classifying by durations ?

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+message Match {
+  // invert the match condition.
+  bool invert = 1;
+
+  // Match uri.
+  StringMatch uri = 2;
+
+  // Match scheme: http or https.
+  string scheme = 3;
+
```
douglas-reid, line 70: do we care about classifying by HTTP version at all?

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+// Value assigned to the dimension if the request matches.
+message Value {
+  // Assign `value` to the enclosing dimension if match is true.
+  string value = 1;
+
+  // Match specification.
+  // Empty match always matches, which can be used to specify a default value when placed
+  // at the end of the list.
+  Match match = 2;
```
douglas-reid, line 56: Do we have a way / plan to handle overlapping matches ?  Is it first to match is used?

extensions/classify/testdata/operation.json
```diff
@@ -0,0 +1,37 @@
...
+              "regex": "^/reviews/\d*"
+            },
+            "method": "GET"
+          }
+        },
+        {
+          "value": "ListReviews",
+          "match": {
+            "uri": {
+              "exact": "/reviews/"
```
douglas-reid, line 20: in testing this: we should add a testcase for something like `/books/{id}/chapter/{id}`

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+  // Less than or equal to this number.
+  int64 le = 2;
+}
+
+// PluginPhase determines when the classification is performed.
+enum PluginPhase {
+  // Classification is performed after the response is sent back to the client.
+  // All information about the request is available in this phase,
+  // and classification does not add to the latency observed by the client.
+  LOG = 0;
```
douglas-reid, line 147: `LOG` and `REQUEST` seem very envoy-specific.  Should we use something like:

`AFTER_RESPONSE`, `ON_REQUEST`, `ON_RESPONSE`, or something?

douglas-reid, line 147: Do we care about style guide recommendations here?

```
PHASE_UNSPECIFIED = 0; // Defaults to after response
ON_REQUEST_PHASE = 1;
AFTER_RESPONSE_PHASE = 2;
```

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+message IntRange {
+  // Greater than or equal to this number.
+  int64 ge = 1;
+
+  // Less than or equal to this number.
+  int64 le = 2;
+}
+
+// PluginPhase determines when the classification is performed.
+enum PluginPhase {
```
douglas-reid, line 143: If we don't call this `ClassificationPhase`, I'd be in favor of dropping `Plugin`, and having this just be `Phase`.

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+    // exact Int match
+    int64 exact = 1;
+
+    // Range match
+    IntRange range = 2;
+  }
+}
+
+// IntRange matches a range of integers.
+message IntRange {
```
douglas-reid, line 134: What do you think about using `from` and `to` or `minimum` and `maximum` instead of `ge` and `le` here ?

douglas-reid, line 134: As this is the only range we support, should this just be `Range` ?

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package classify;
```
kyessenov, line 18: Please use `istio` in the package prefix, since this is probably istio-specific for now.
Meta-comment: no need to use proto for extension configuration. You could just use JSON/OpenAPI.

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+
+
+// Configuration to classify http traffic.
```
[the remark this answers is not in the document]
kyessenov: nit: capital `HTTP`.

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+// Traffic classification can be done on mutiple dimensions using request and
+// response headers and other measurements. 
+// The plugin must be configured as request, response or log based on
```
kyessenov, line 24: There is no response based plugins as far as I know. All plugins implement the same callbacks, except for access log based ones.

extensions/classify/config.proto
```diff
@@ -0,0 +1,158 @@
...
+// which other plugins need to consume the classfication results.
+// For exmaple: If Stats plugin uses `operation` dimension produced by the classification
+// plugin, then it must be executed in LOG phase *before* the Stats plugin.
+message PluginConfig {
+  // next id: 3
+  // Plugin phase when classification is done.
+  PluginPhase phase = 1;
```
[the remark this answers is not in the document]
kyessenov: Is it really expensive to run this in onRequestHeaders always? It might be pre-mature optimization to try to move it to the logging phase.

extensions/classify/config.proto
kyessenov: This feels like yet-another AST for a condition? Why not just use CEL expression instead of creating one?

extensions/classify/config.proto
kyessenov: I wonder if this could be implemented in a way that is transparent to the phase. Maybe some kind of lazily computed property.

extensions/classify/config.proto
mandarjog: good point, updated names.

extensions/classify/config.proto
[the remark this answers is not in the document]
mandarjog: It is not an optimization.

1. It depends on what information is available when. (response codes, durations are not available on_request)
2. Which downstream plugin needs this information.

For example, you can have a `operation` classifier deployed as `ON_REQUEST` which can be used by authz and subsequently by stats.
If you also want to use responseClassification it should happen `before` stats filter gets called.

@kyessenov are you suggesting that we do it automagically, and the implementation decides when classification takes place?

extensions/classify/config.proto
```diff
@@ -0,0 +1,198 @@
...
+
+  // Match specification.
+  // Empty match always matches, which can be used to specify a default value when placed
+  // at the end of the list.
+  Match match = 2;
+}
+
+// Match condition based on request and response information.
+// Every match condition must match for the result to be true.
+message Match {
```
mandarjog, line 62: @kyessenov are you suggesting that we replace Match message with a CEL expression?
I would still have a list of those which would be more readable.

I think all the headers are already available in the CEL, are any measurements like response duration available there?
We can always add.

extensions/classify/config.proto
nrjpoddar: Lines 25:26 is quite hard to follow, may be add this:

This filter generates attributes based on the configured classification and request/response being processed. These attributes can be consumed by downstream filters in the chain. For e.g. Stats plugin uses `operation` dimension produced by the classification plugin, then it must be executed in LOG phase *before* the Stats plugin.

extensions/classify/config.proto
```diff
@@ -0,0 +1,198 @@
...
+  Phase phase = 1;
+
+  // Request classification along multiple dimensions.
+  repeated Dimension dimensions = 2;
+}
+
+// Dimension configuration produces the named dimension with a value based
+// on the match clause in the value. 
+// The value matches are processed sequentially and the first value to match is assigned.
+message Dimension {
```
nrjpoddar, line 41: Dimension encapsulates the configuration of a named attribute which is emitted by the filter if the conditions are met.

extensions/classify/config.proto
```diff
@@ -0,0 +1,198 @@
...
+  // For example `operation` is a dimension.
+  string name = 1;
+
+  // At most one of the following values is assigned to the dimension. 
+  repeated Value values = 2;
```
nrjpoddar, line 46: The values are evaluated in order and the first condition to match wins.

extensions/classify/config.proto
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// Value assigned to the dimension if the request matches.
+message Value {
+  // Assign `value` to the enclosing dimension if match is true.
```
nrjpoddar, line 51: Value of the dimension if the match condition is met

extensions/classify/config.proto
```diff
@@ -0,0 +1,198 @@
...
+  // Request classification along multiple dimensions.
+  repeated Dimension dimensions = 2;
+}
+
+// Dimension configuration produces the named dimension with a value based
+// on the match clause in the value. 
+// The value matches are processed sequentially and the first value to match is assigned.
+message Dimension {
+  // For example `operation` is a dimension.
+  string name = 1;
```
nrjpoddar, line 43: Name of the dimension for e.g. "operation"

extensions/classify/config.proto
```diff
@@ -0,0 +1,198 @@
...
+  // Match scheme: http, https or unset.
+  Scheme scheme = 3;
+
+  // Match method: GET, POST, HEAD, PUT, DELETE, ...
+  repeated string method = 4;
+
+  // Match request headers.
+  // Not implemented.
+  // $hide_from_docs
+  map<string, StringMatch> request_headers = 5;
```
nrjpoddar, line 78: Since you have to copy these fields from VirtualService API instead of importing, wdyt about using CEL and listing available attributes?

Have a copy of the API field is kind of a maintenance nightmare with accumulating tech debt of keeping parity.

extensions/classify/config.proto
```diff
@@ -0,0 +1,198 @@
...
+}
+
+enum Scheme {
+  // Matches any
+  UNSPECIFIED = 0;
+  HTTP = 1;
+  HTTPS = 2;
+}
+
+// Phase determines when the classification is performed.
```
nrjpoddar, line 180: Does Phase capture intent of the user or actually places the filter at the right place i.e. do users still need to create EnvoyFilter with the current placement settings for this new filter?

extensions/classify/testdata/operation.json
```diff
@@ -0,0 +1,37 @@
...
+            },
+            "method": "GET"
+          }
+        },
+        {
+          "value": "CreateReview",
+          "match": {
+            "uri": {
+              "exact": "/reviews/"
```
gargnupur, line 29: nit: add a match test for prefix?

extensions/classify/config.proto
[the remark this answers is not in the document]
kyessenov: You should add presence indicator for information that is missing. You don't always have those attributes even on logging callback (e.g. abrupt request termination).

extensions/classify/config.proto
kyessenov: We need to be very precise what this means if this is going to be used for authz. Path, canonicalized path, query params all make a huge difference and can cause a CVE if there's some misunderstanding.

extensions/classify/config.proto
kyessenov: Yeah, you can follow the same pattern as authz in Envoy. Have an AST predicate with a CEL expression for more complicated logic (ORs, ANDs, functions).

extensions/classify/config.proto
kyessenov: h2 allows bi-directional streaming. So there's no clear boundary between when request ends and response starts.

extensions/classify/config.proto
mandarjog: I will try out a CEL only api and see how it works. The proto defn will be more. Concise and kuat and I have been discussing that.
The user experience and validation is the issue.

I am not worried about maintaining parity with virtual service.
Neither APIs matching properties are expected to change often.

extensions/classify/config.proto
mandarjog: User needs to place the filter correctly in relation to other filters.
Phase dictates the filter callback when classification is performed.

extensions/classify/config.proto
mandarjog: @kyessenov specifying cel in json with escaping “ will be a pain so we need envoy filter api to do the conversations.

extensions/classify/config.proto
nrjpoddar: Isn't that too low level for any user?

Why can't we have the Phase fixed for the most common use cases as most of the end users won't understand this.

nrjpoddar:
Just as a thought exercise, how would we make this protocol agnostic. This ties a bit into my RFC for meta routing. Since HTTP mostly is a superset of attributes and your API is already abstract, the only thing making it more specific to HTTP is using the name "Request/Response"

@douglas-reid @mandarjog FYI

mandarjog:
Peer attributes, self attributes and protocol attributes are the 3 types of attributes available. I am making an alternate PR with CEL only in that case all attributes are avl inside CEL, but there will be no request response nomenclature in the proto itself.

For kafka: we could have new cel attributes that you can use.

extensions/classify/config.proto
mandarjog: You cannot have a fixed phase. If I want to use the output of classification for responseClass in the stats filter then the classification must be performed in LOG phase when responseCodes are available.

Alternatively if you want to use the classification in an authz filter you cannot perform classification in the LOG phase, it will be too late and authz will not have the request classification that is needed.

We can do some magic to figure this out, I think we should start simple and explicit .

kyessenov: Can the filter guess based on its position? It's not a good experience to both specify the filter position in EnvoyFilter, and then specify it again in the phase of the filter config.

mandarjog:
@kyessenov @douglas-reid @nrjpoddar @gargnupur I have created a new PR https://github.com/istio/proxy/pull/2801 which is CEL based, please review.

mandarjog:
closing this in favour of #2801

PR #2800, opened 2020-04-13
authn: wasm implementation

Shikugawa:
Draft of AuthN wasm implementation.

**Which issue this PR fixes**
To resolve https://github.com/istio/istio/issues/15772

We can't build it into wasm because there is no suitable build system. I think that this shouldn't be built with make. Instead of this, we should apply bazelbuild to build wasm filter.

**Release note**:

```release-note
```


mandarjog:
@yangminzhu @diemtvu is this how you guys want to do it ?

diemtvu:
That's fast. Thanks for the PR.

@yangminzhu, I think we should simplify this filter. At least have a new struct/proto for configuration, instead of reuse (alpha) proto. All the conditional validations and principal binding can also be removed. I guess it's best to have a short design for the new filter. Wdyt?

yangminzhu:
> @yangminzhu, I think we should simplify this filter. At least have a new struct/proto for configuration, instead of reuse (alpha) proto. All the conditional validations and principal binding can also be removed. I guess it's best to have a short design for the new filter. Wdyt?

Yes, one of the major issues in today's istio_authn filter is its API is unnecessary complicated as it was originally designed for the old v1alpha1 APIs, the others are some unnecessary principal/identity transformations in multiple data structures. We should definitely have a new and simpler API if we're moving to a new filter implementation.

+1 for a short design doc on this.

yangminzhu:
@lizan @Shikugawa 

Can we have a redesign of the filter API? A one page doc of the new API would be very helpful, the new API could be very simplified as it only needs to support the beta Peer/RequestAuthN policy. Thanks.

Shikugawa:
@yangminzhu There is no filter API changes in this PR. I will send another PR even if we need to change it. I'd like this PR only to be re-implementation of WASM. Should we re-design it in this?

yangminzhu:
> @yangminzhu There is no filter API changes in this PR. I will send another PR even if we need to change it. I'd like this PR only to be re-implementation of WASM. Should we re-design it in this?

well, that might be okay but the most important benefit of rewriting the authn filter in WASM is it cleans up the internal API and removes the tech-debt, right? If we just have the WASM implementation of the old API, we don't get much benefits with all these efforts (yet still need to test the stability of the WASM). And if we have the new API, much of the code can just be removed/simplified and thus not very worth for a through review for now.

Shikugawa:
@yangminzhu Right. I see. I will consider the new API. Thanks.

Shikugawa:
@yangminzhu @bianpengyuan For now, building problem is resolved so that reopened to work this. But, this PR is too large to be hard to review. I prefer split out this large PR and merge incrementally to this. What do you think about it?

zirain:
outdated

PR #2801, opened 2020-04-14
CEL based v2 attribute generation: Request Classification.

mandarjog:
Add: proto definition for request classification.
Add: two examples for use.
Add: Implementation
Add: Unit tests.

Note: the configuration must be json at the point of delivery, not yaml.

An Alternate to #2799


extensions/classify/config.proto
```diff
@@ -0,0 +1,98 @@
...
+  // Example: `istio.operationId`
+  string output_attribute = 2;
+
+  // A Classification may fail to evaluate when an attribute is not available.
+  // Example: `response.code` may not be available when a request ends abruptly.
+  // When evalution failes, the plugin will return an error to the client.
+  // This is the correct behaviour if the result of the classification is used by an authz plugin.
+  // If the result of the classifier is used by stats, you may just want to log an error but not
+  // fail the request. If `ignore_errors=true` then the classifier plugin will not fail the request.
+  bool ignore_errors = 3;
```
nrjpoddar, line 47: This is interesting and can have broad security implications if any security features are dependent on these attributes.

Should this be always set to "false" and not even exposed as a field i.e. "attribute" is best effort with defaults. Downstream filters should be created in a way that they handle unset/default/actual value.

extensions/classify/config.proto
```diff
@@ -0,0 +1,98 @@
...
+  // Note: Use anchors {^, $} to ensure that the regex evaluates efficiently.
+  // Note: request.url_path is normalized and stripped of query params.
+  //
+  // a Read only operation on books
+  // { value: ReadOnlyBooks, 
+  //   condition: request.url_path.startsWith("/books/") && in(request.method, ["GET", "HEAD"])}
+  string condition = 1;
+
+  // If condition evaluates to true, return the `value`.
+  string value = 2;
```
nrjpoddar, line 77: should we have a concept of "default"?

extensions/classify/config.proto
```diff
@@ -0,0 +1,98 @@
...
+// Configuration to classify traffic.
+message PluginConfig {
+  // next id: 2
+  // Multiple independent classification configurations.
+  repeated Classification classifications = 1;
+}
+
+// Classification uses [well known attributes](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/rbac_filter#condition) as inputs and produces
+// a new attribute that denotes a specific classification.
+// The new attribute can be used by a downstream plugin.
```
nrjpoddar, line 29: plugin vs filter which is preferable?

nrjpoddar:
This is basically bringing Mixer back in a much more performant way (hopefully). :)

extensions/classify/config.proto
```diff
@@ -0,0 +1,98 @@
...
+
+syntax = "proto3";
+
+package istioclassify;
+
+// Configuration to classify traffic.
+message PluginConfig {
+  // next id: 2
+  // Multiple independent classification configurations.
+  repeated Classification classifications = 1;
```
nrjpoddar, line 24: We still need to add documentation around the behavior of overlapping "output_attribute" i.e. what happens if the config has `operation` twice. Does the last one win?

nrjpoddar: Classify is too narrow a use case for this implementation since this can be used in an abstract way.

Other name suggestion:
repeated AttributeConfig attributes = 1;

extensions/classify/config.proto
```diff
@@ -0,0 +1,98 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package istioclassify;
```
douglas-reid, line 18: any reason not to make this `istio.classify` ?  or `istio.attributegen` ?

extensions/classify/config.proto
douglas-reid: `AttributeGeneration` would match the v1 name for these sorts of extensions. Not sure if that helps or not.

extensions/classify/config.proto
```diff
@@ -0,0 +1,98 @@
...
+  // Phase denotes plugin lifecycle phase when the classification is evaluated. 
+  Phase phase = 1;
+
+  // The following attribute is populated on successfully running the classification.
+  // Example: `istio.operationId`
+  string output_attribute = 2;
+
+  // A Classification may fail to evaluate when an attribute is not available.
+  // Example: `response.code` may not be available when a request ends abruptly.
+  // When evalution failes, the plugin will return an error to the client.
```
douglas-reid, line 43:
```suggestion
  // When evalution fails, the plugin will return an error to the client.
```

extensions/classify/config.proto
douglas-reid: If possible, i'd prefer to have some distinction between failure and unset. maybe `istio.attributeGen.status` should be an attribute ;p.

extensions/classify/config.proto
```diff
@@ -0,0 +1,98 @@
...
+//
+// Example: A classifier may produce an attribute called `istio.operationId` which is analogous to the `operationId` in the OpenAPI Spec.
+// The Stats plugin will run after the classifier and use `istio.operationId` to populate a dimension. 
+message Classification {
+  // Phase denotes plugin lifecycle phase when the classification is evaluated. 
+  Phase phase = 1;
+
+  // The following attribute is populated on successfully running the classification.
+  // Example: `istio.operationId`
+  string output_attribute = 2;
```
douglas-reid, line 39: do we have guidance on attribute namespaces?  like `istio.*` is reserved ?

extensions/classify/config.proto
mandarjog: This exactly gives us the difference between failure to evaluate and evaluated to empty.
We could put the onus of failing the request of the consumer of the attribute and not the producer.

extensions/classify/config.proto
mandarjog: default is just a `Match clause` with no condition.
I had it in the other PR. forgot to copy it.

extensions/classify/config.proto
mandarjog: ack

extensions/classify/config.proto
mandarjog: We call it plugin in the Wasm paralance, so I prefer plugin.

extensions/classify/config.proto
mandarjog: istio.attributegen sounds good.

extensions/classify/config.proto
mandarjog: I agree classification is too narrow, this is now an APA from v1.

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,113 @@
...
+  //
+  // a Read only operation on books
+  // { value: ReadOnlyBooks, 
+  //   condition: request.url_path.startsWith("/books/") && in(request.method, ["GET", "HEAD"])}
+  //
+  // An empty condition evaluates to `true` and should be used to provide a default value.
+  string condition = 1;
+
+  // If condition evaluates to true, return the `value`.
+  string value = 2;
```
mandarjog, line 91: At this time, value is a constant.
Should also add output_expr or wait until we need it?

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,113 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package istio.attributegen;
+
+// Configuration to generate attributes based on existing attributes.
```
nrjpoddar, line 20: Nit: This is a bit difficult to understand, suggestion: 

Configuration to generate synthetic attributes based on attributes of the proxied traffic

Something like this might be more clear.

extensions/attributegen/config.proto
nrjpoddar: I would add it later if requested.

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Phase determines when the classification is performed.
+enum Phase {
+  option allow_alias = true;
+  // AttributeGeneration is performed after the response is sent back to the client.
+  // All information about the request is available in this phase,
+  // and classification does not add to the latency observed by the client.
+  PHASE_UNSPECIFIED = 0;
+  AFTER_RESPONSE = 0;
```
nrjpoddar, line 101: Using response/request is fine for now and will work for HTTP, how would you name it so that’s it’s suitable for other protocols?

I have used “transaction” as an abstract way to define this in the past but Istio doesn’t really use transaction in the current vocabulary.

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,113 @@
...
+// AttributeGeneration uses [well known attributes]
+// (https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/rbac_filter#condition)
+// as inputs and produces a new attribute.
+// The new attribute can be used by a downstream plugin.
+//
+// Example: AttributeGen may produce an attribute called `istio.operationId` which is analogous to
+// the `operationId` in the OpenAPI Spec.
+// The Stats plugin will run after this plugin and use `istio.operationId` to populate a dimension.
+//
+// If multiple AttributeGenerations produce the same attribute, the result of the last generation
```
douglas-reid, line 36: Using last is fine, but I recall that in Mixer once an attribute had a value, any subsequent updates were ignored.

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,113 @@
...
+  Phase phase = 1;
+
+  // The following attribute is populated on a successful match.
+  // Example: `istio.operationId`
+  // `istio.` attribute namespace is reserved by Istio.
+  //
+  // AttributeGeneration may fail to evaluate when an attribute is not available.
+  // Example: `response.code` may not be available when a request ends abruptly.
+  // When evaluation fails the plugin will set an error_attribute to the error string.
+  // error_attribute = ${output_attribute}_error
```
douglas-reid, line 50: interesting. if i have multiple generators that each produce the same `output_attribute` and only one fails, do i see both an error attribute and a usable attribute ?

extensions/attributegen/config.proto
mandarjog: REQUEST, REPLY and LOG can be mapped to most other protocols.

1. memcache / redis all 3 phases are clear
2. mysql  same.
3.kafka  REQUEST (Message). and LOG are ok. There is no reply for payload.
4. grpc stream is like kafka, and bidi-grpc is like 2 kafkas... 

can you suggest something else? @kyessenov  ?

nrjpoddar: Can we remove this option for now since we only support one callback insertion point.
I just think it's too much for a user to configure both filter & callback placements. 

Long term I was thinking for filters & extensions we should defined "classes/categories" and based on the class your placement in filter chain is more or less. This same category can also dictate the phase.

mandarjog: The code supports both insertion points, it is just hidden from the docs.
I am ok hiding Phase completely if you want, but if you need to classify and use the attribute for policy decisions the *must* specify the callback not just the insertion order.

We can recast phases as filter `functions`, as it is done in the extensions API to make it less implementation specific.
This does determine the plugin ordering, but the plugin also needs to know where it will run.

```proto
enum Function {
    UNSPECIFIED_FUNCTION = 0;
    TELEMETRY = 1;
    PRE_PROCESSING = 2;
    AUTHORIZATION = 3;
    POST_PROCESSING = 4;
}
```

nrjpoddar: Yes! These names make much more sense to an end user. With some documentation around these terms like Pre processing and post processing and examples I think it’s very reasonable and good UX.

I don’t mean to be a bottleneck on naming but I believe getting it right now will save us a lot of headaches in the future.

@dougreid thoughts?

kyessenov: Unfortunately, this categorization is rather simplistic. Even the native filter mix and match functions in the same filters (filter does auth, telemetry, and pre-processing at the same time). It's a lot of work to try to re-organize existing native extensions to follow this paradigm, so I'm not sure we want to commit to that in this filter. I think we want this low-level be as close to the metal as to what actually happens in Envoy, and add some sugar on top if that irks some people the wrong way.

kyessenov: Put it the other way - I don't dismiss UX, but it should reflect the reality :)

mandarjog: Being close to the implementation was my goal initially since this is a filter and xds is not supposed to human consumable or producible. 

However we are caught in the middle because the xds APi *is* exposed as is to the end user for now and so UX discussion comes in.
Having a higher level abstraction like TELEMETRY, et al without the higher level extension mechanism to put things in the correct place only gets us half way.

We have 2 options
1. Stick with what we have since it maps cleanly to envoy. We can de emphasize the phase is the doc and examples.
2. Remove the option completely. but we will be in the same spot as today when someone asks for the feature in the future.

mandarjog:
@kyessenov @bianpengyuan may I get a review on this?
I will add a few more tests with more matching rules, but otherwise this is ready for review.

extensions/attributegen/config.proto
nrjpoddar: My suggestion here would be to hide it until someone asks for it, or we have more understanding of requirements.

@kyessenov I agree about pre-mature APIs, hence why expose something that we are not sure of for now.

mandarjog: Ok done, I hid the option from the proto.

extensions/attributegen/plugin_test.cc
```diff
@@ -0,0 +1,423 @@
...
+
+  Http::TestRequestHeaderMapImpl request_headers{{":path", "/books"},
+                                                 {":method", "GET"}};
+  Http::TestResponseHeaderMapImpl response_headers{{":status", "404"}};
+
+  verify_request(request_headers, response_headers, attribute, true, false,
+                 "ListBooks");
+}
+
+// CEL is unable to get access to response status code.
```
mandarjog, line 397: @kyessenov @jplevyak CEL is unable to get access to `response.code` .
This works fine in an actual full manual verification, but does not work in this test.
Setting response header `{":status", "404"}` is not populating response.code internally.

extensions/attributegen/config.proto
douglas-reid: If we are going to expand it later, should this be:

`Value value = 2;`

Using a message would allow us flexibility.

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,124 @@
...
+// as inputs and produces a new attribute.
+// The new attribute can be used by a downstream plugin.
+//
+// Example: AttributeGen may produce an attribute called `istio.operationId`
+// which is analogous to the `operationId` in the OpenAPI Spec. The Stats plugin
+// will run after this plugin and use `istio.operationId` to populate a
+// dimension.
+//
+// If multiple AttributeGenerations produce the same attribute, the result of
+// the last generation will be visible to downstream filters.
```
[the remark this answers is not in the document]
mandarjog: The primary attributes *cannot* be overridden at this time.

The synthetic attributes are set in the filterState and those *can* be overridden.

extensions/attributegen/config.proto
mandarjog: It adds to the level of indentation :-( but these are not supposed to be written by a human !

mandarjog: could be do a one_of (value and value_expr)?

mandarjog:
@nrjpoddar I just went through you comments in the previous PR again, they were already factored in. Please point me if I have missed something.

1. I have hidden phase from the proto.
2. I think `Value value` will create one more level of indent, and we can address it by a one_of later.

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,125 @@
...
+  // The value specified by the successful match is assgined to the
+  // output_attribute.
+  repeated Match match = 3;
+}
+
+// If the condition evaluates to true then the Match returns the specified
+// value.
+message Match {
+  // Condition is a [CEL
+  // expressions](https://github.com/google/cel-spec/blob/master/doc/langdef.md)
```
douglas-reid, line 71:
```suggestion
  // expression](https://github.com/google/cel-spec/blob/master/doc/langdef.md)
```

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,125 @@
...
+  //
+  // a GET on /shelves/{id1}/books/{id2}
+  // { value: GetBook,
+  //   condition:
+  //   request.url_path.matches('^/shelves/[[:alnum:]]*/books/[[:alnum:]]*$') &&
+  //   request.method == 'GET' }
+  // Note: CEL uses [re2](https://github.com/google/re2/wiki/Syntax) regex
+  // library.
+  //       Use anchors {^, $} to ensure that the regex evaluates efficiently.
+  // Note: request.url_path is normalized and stripped of query params.
```
douglas-reid, line 90:
```suggestion
  // Note: `request.url_path` is normalized and stripped of query params.
```

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,125 @@
...
+  // Condition is a [CEL
+  // expressions](https://github.com/google/cel-spec/blob/master/doc/langdef.md)
+  // You may use [well known attributes]
+  // (https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/rbac_filter#condition)
+  // in the CEL expression.
+  //
+  // Examples:
+  //
+  // a GET on /books
+  // { value: ListBooks,
```
douglas-reid, line 79: should these be in pre-formatted blocks?

extensions/attributegen/config.proto
```diff
@@ -0,0 +1,125 @@
...
+  // An empty condition evaluates to `true` and should be used to provide a
+  // default value.
+  string condition = 1;
+
+  // If condition evaluates to true, return the `value`.
+  string value = 2;
+}
+
+// Phase determines when the classification is performed.
+enum Phase {
```
douglas-reid, line 106: i can't recall. does the entire message need to be hidden from docs too?

extensions/attributegen/config.proto
mandarjog: ack

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,147 @@
...
+    if (phase != ag.phase()) {
+      continue;
+    }
+
+    std::string val;
+    auto eval_status = ag.evaluate(&val);
+    if (!eval_status) {
+      LOG_DEBUG(absl::StrCat("Failed eval. setting ", ag.error_attribute()));
+      // eval failed set error attribute
+      setFilterState(ag.error_attribute(), "1");
```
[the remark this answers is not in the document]
mandarjog: Good point about `_error` attribute overshadowing an attribute that is already there.
However since this is part of the contract and API, use will hopefully see this.

We always *need* to filterstate or error so that downstream plugins can distinguish between an error and no-match.

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,147 @@
...
+  return initAttributeGen(config);
+}
+
+bool PluginRootContext::initAttributeGen(
+    const istio::attributegen::PluginConfig& config) {
+  // Clear prior host resources.
+  // TODO(mjog) see if this still needed since onDone is always called.
+  cleanupAttributeGen();
+
+  for (auto agconfig : config.attributes()) {
```
kyessenov, line 62: nit: C++ style is to use long name, attribute_gen_config. This is not golang.

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,147 @@
...
+    EvalPhase phase = on_log;
+    if (agconfig.phase() == istio::attributegen::ON_REQUEST) {
+      phase = on_request;
+    }
+    auto ag = AttributeGenerator(phase, agconfig.output_attribute());
+    for (auto matchconfig : agconfig.match()) {
```
kyessenov, line 68: const auto&

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,147 @@
...
+      uint32_t token = 0;
+      if (matchconfig.condition().empty()) {
+        ag.add_match(Match("", 0, matchconfig.value()));
+        continue;
+      }
+      if (createExpression(matchconfig.condition(), &token) != WasmResult::Ok) {
+        LOG_WARN(absl::StrCat("Cannot create expression: <",
+                              matchconfig.condition(), "> for ",
+                              agconfig.output_attribute()));
+        return false;
```
kyessenov, line 78: why fail fast and not ignore?

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,147 @@
...
+void PluginRootContext::attributeGen(EvalPhase phase) {
+  for (const auto& ag : gen_) {
+    if (phase != ag.phase()) {
+      continue;
+    }
+
+    std::string val;
+    auto eval_status = ag.evaluate(&val);
+    if (!eval_status) {
+      LOG_DEBUG(absl::StrCat("Failed eval. setting ", ag.error_attribute()));
```
kyessenov, line 107: This is noisy. Maybe use trace logging for per-request logging.

extensions/attributegen/plugin.cc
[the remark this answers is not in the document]
kyessenov: It's unclear why we need error filter state. Filter state is expensive, and just putting "1" there is wasteful.

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+
+  std::string value() const { return value_; };
+
+  void cleanup() {
+    if (!condition_.empty()) {
+      exprDelete(condition_token_);
+    }
+  }
+
+ protected:
```
kyessenov, line 92: why protected? there's nothing virtual here.

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+      return true;
+    }
+
+    bool matched;
+    if (!evaluateExpression(condition_token_, &matched)) {
+      LOG_WARN(absl::StrCat("Failed to evaluate expression: ", condition_));
+      return {};
+    }
+    if (matched) {
+      LOG_DEBUG(absl::StrCat("Matched: ", condition_));
```
kyessenov, line 79: Too noisy.

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+    }
+    return matched;
+  };
+
+  std::string value() const { return value_; };
+
+  void cleanup() {
```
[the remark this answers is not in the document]
kyessenov: +1

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+        return true;
+      }
+    }
+    return false;
+  }
+
+  EvalPhase phase() const { return phase_; }
+  std::string output_attribute() const { return output_attribute_; }
+  std::string error_attribute() const { return error_attribute_; }
+  void add_match(Match match) { matches_.push_back(match); }
```
kyessenov, line 127: Why does this need to be mutable? It's created once and const after that.

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+};
+
+enum EvalPhase { on_log, on_request };
+
+class AttributeGenerator {
+ public:
+  explicit AttributeGenerator(EvalPhase phase, std::string output_attribute)
+      : phase_(phase),
+        output_attribute_(output_attribute),
+        error_attribute_(absl::StrCat(output_attribute, "_error")) {}
```
kyessenov, line 107: uninitialized member `matches_`

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+};
+
+// PluginRootContext is the root context for all streams processed by the
+// thread. It has the same lifetime as the worker thread and acts as target
+// for interactions that outlives individual stream, e.g. timer, async calls.
+class PluginRootContext : public RootContext {
+ public:
+  PluginRootContext(uint32_t id, StringView root_id)
+      : RootContext(id, root_id) {}
+  ~PluginRootContext() = default;
```
kyessenov, line 149: this is tautological?

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+
+// Per-stream context.
+class PluginContext : public Context {
+ public:
+  explicit PluginContext(uint32_t id, RootContext* root) : Context(id, root) {}
+
+  void onLog() override { rootContext()->attributeGen(on_log); };
+
+  FilterHeadersStatus onRequestHeaders(uint32_t) override {
+    rootContext()->attributeGen(on_request);
```
kyessenov, line 177: There's too much code in the header.  Please move some code into .cc file. That will make imports smaller, too.

extensions/attributegen/plugin_test.cc
```diff
@@ -0,0 +1,501 @@
...
+  Http::TestRequestHeaderMapImpl request_headers{{":path", "/books"},
+                                                 {":method", "GET"}};
+  Http::TestResponseHeaderMapImpl response_headers{{":status", "200"}};
+
+  verify_request(request_headers, response_headers, attribute, true, false,
+                 "ListBooks");
+}
+
+TEST_P(AttributeGenFilterTest, OperationFileListNoMatch) {
+  const std::string attribute = "istio.operationId";
```
kyessenov, line 405: move into a compilation unit global

extensions/attributegen/plugin_test.cc
```diff
@@ -0,0 +1,501 @@
...
+  const std::string attribute = "istio.responseClass";
+
+  setupConfig({.mock_logger = false,
+               .plugin_config_file =
+                   "responseCode.json"});  // testdata/responseCode.json
+
+  Http::TestRequestHeaderMapImpl request_headers{{":path", "/books"},
+                                                 {":method", "GET"}};
+  Http::TestResponseHeaderMapImpl response_headers{{":status", "504"}};
+  verify_request(request_headers, response_headers, attribute, true, false,
```
kyessenov, line 487: verityRequest per style guide

extensions/attributegen/testdata/server.yaml
```diff
@@ -0,0 +1,236 @@
+node:
+  id: test-server-ratings
```
kyessenov, line 2: Please use the test framework to add an integration test. I don't really trust the mocking framework. That would help save on duplication of the JSONs in this repo, since you can use one of the templates.

extensions/attributegen/plugin.cc
mandarjog: I do not follow, what do you want to ignore?

extensions/attributegen/plugin.cc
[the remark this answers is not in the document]
mandarjog: Please suggest something else that will help the consumer of this attribute distinguish between attribute not set because condition did not match, and attribute could not be evaluated.
The consuming plugin may be authz and it is important to distinguish between the two.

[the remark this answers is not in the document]
kyessenov: If it's not set, then it's not set. The policy would just evaluate to an error, or use presence the checks for extra control in rare cases.

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,147 @@
...
+      // eval failed set error attribute
+      setFilterState(ag.error_attribute(), "1");
+      continue;
+    }
+
+    if (!eval_status.value()) {
+      continue;
+    }
+    LOG_DEBUG(absl::StrCat("Setting ", ag.output_attribute(), " --> ", val));
+    setFilterState(ag.output_attribute(), val);
```
gargnupur, line 117: I think default span for filterstate is "filterchain", for this plugin, shouldn't it be longer "downstream connection"?
(Not sure if it's available in ABI to do that yet.. would need access to WasmStatePrototype I think)

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,147 @@
...
+  std::unique_ptr<WasmData> configuration = getConfiguration();
+  // Parse configuration JSON string.
+  JsonParseOptions json_options;
+  json_options.ignore_unknown_fields = true;
+  istio::attributegen::PluginConfig config;
+  Status status =
+      JsonStringToMessage(configuration->toString(), &config, json_options);
+  if (status != Status::OK) {
+    LOG_WARN(absl::StrCat(
+        "Cannot parse plugin configuration JSON string [YAML not supported]: ",
```
gargnupur, line 48: nit: add name of this plugin in error message?

extensions/attributegen/plugin_test.cc
```diff
@@ -0,0 +1,501 @@
+/* Copyright 2019 Istio Authors. All Rights Reserved.
```
gargnupur, line 1: nit:2020

extensions/attributegen/plugin.cc
kyessenov: Generally we never want to reject config unless something terrible happens since it immediately crashes envoy in some cases. Seems like one bad expression is not a sufficient reason to crash.

extensions/attributegen/plugin.cc
mandarjog: Good question, @kyessenov  @jplevyak ?

kyessenov: Yes, it's per filter chain (individual pass per request) if you don't declare the property. The ABI is available, see #2817 .

extensions/attributegen/config.proto
mandarjog: yes the entire message is hidden.

extensions/attributegen/config.proto
mandarjog: ack.

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+namespace AttributeGen {
+
+using StringView = absl::string_view;
+
+using google::protobuf::util::JsonParseOptions;
+using google::protobuf::util::Status;
+
+class Match {
+ public:
+  explicit Match(std::string condition, uint32_t condition_token,
```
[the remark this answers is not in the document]
mandarjog: ack.

extensions/attributegen/plugin.h
mandarjog: protected all logs under `if (debug_)` or trace.

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+      LOG_WARN(absl::StrCat("Failed to evaluate expression: ", condition_));
+      return {};
+    }
+    if (matched) {
+      LOG_DEBUG(absl::StrCat("Matched: ", condition_));
+    }
+    return matched;
+  };
+
+  std::string value() const { return value_; };
```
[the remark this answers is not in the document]
mandarjog: ack.

extensions/attributegen/plugin.h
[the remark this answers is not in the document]
mandarjog: only rootContext has cleanup now.
There was no need for destructors.

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+
+ protected:
+  // Only used for debug messages after construction.
+  std::string condition_;
+  // Expression token associated with the condition.
+  uint32_t condition_token_;
+  std::string value_;
+};
+
+enum EvalPhase { on_log, on_request };
```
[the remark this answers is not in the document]
mandarjog: ack.

extensions/attributegen/plugin.h
mandarjog: fixed.

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+      if (eval_status.value()) {
+        *val = match.value();
+        return true;
+      }
+    }
+    return false;
+  }
+
+  EvalPhase phase() const { return phase_; }
+  std::string output_attribute() const { return output_attribute_; }
```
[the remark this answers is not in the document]
mandarjog: done.

extensions/attributegen/plugin.h
mandarjog: removed.

extensions/attributegen/plugin.h
mandarjog: removed.

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,205 @@
...
+
+  bool onConfigure(size_t) override;
+  // is called even if
+  bool onDone() override {
+    cleanupAttributeGen();
+    return true;
+  };
+  void attributeGen(EvalPhase);
+
+ protected:
```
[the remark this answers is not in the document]
mandarjog: private now.

extensions/attributegen/plugin.h
mandarjog: moved to .cc

extensions/attributegen/plugin.cc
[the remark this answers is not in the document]
mandarjog: it is no more wasteful than setting the attribute itself.

mandarjog:
@kyessenov @bianpengyuan @gargnupur I have addressed code review comments.
There are unit tests. I don't want to add integration tests to the same PR and make it bigger.

extensions/attributegen/Makefile
```diff
@@ -0,0 +1,17 @@
...
+PROTO_SRCS = config.pb.cc
+COMMON_SRCS = 
+
+all: plugin.wasm
+
+%.wasm %.wat: %.cc ${CPP_API}/proxy_wasm_intrinsics.h ${CPP_API}/proxy_wasm_enums.h ${CPP_API}/proxy_wasm_externs.h ${CPP_API}/proxy_wasm_api.h ${CPP_API}/proxy_wasm_intrinsics.js ${CPP_CONTEXT_LIB}
+	protoc config.proto --cpp_out=.
+	em++ -s STANDALONE_WASM=1 -s EMIT_EMSCRIPTEN_METADATA=1 -s EXPORTED_FUNCTIONS=['_malloc'] --std=c++17 -O3 -I${CPP_API} -I${CPP_API}/google/protobuf -Iextensions/common -I. -I/usr/local/include -I${ABSL} -I. --js-library ${CPP_API}/proxy_wasm_intrinsics.js $*.cc ${CPP_API}/proxy_wasm_intrinsics.pb.cc ${PROTO_SRCS} ${COMMON_SRCS} ${CPP_CONTEXT_LIB} ${ABSL_CPP} ${CPP_API}/libprotobuf.a -o $*.wasm
+	rm -f $*.wast
+	rm -f extensions/attributegen/config.pb.*
```
kyessenov, line 16: rm -f config.pb.*

extensions/attributegen/plugin.h
```diff
@@ -0,0 +1,172 @@
...
+  const std::string value_;
+};
+
+enum EvalPhase { OnLog, OnRequest };
+
+class AttributeGenerator {
+ public:
+  explicit AttributeGenerator(EvalPhase phase,
+                              const std::string& output_attribute,
+                              const std::vector<Match> matches)
```
kyessenov, line 82: `const std::vector<Match>&` or `&&`.

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,233 @@
...
+  const std::string function = "expr_evaluate";
+  char* out = nullptr;
+  size_t out_size = 0;
+  auto result = proxy_call_foreign_function(
+      function.data(), function.size(),
+      reinterpret_cast<const char*>(&condition_token_), sizeof(uint32_t), &out,
+      &out_size);
+  if (result != WasmResult::Ok) {
+    LOG_TRACE(absl::StrCat("Failed to evaluate expression:[", condition_token_,
+                           "] ", condition_, " result: ", toString(result)));
```
kyessenov, line 54: if out not null, free out.

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,233 @@
...
+  istio::attributegen::PluginConfig config;
+  Status status =
+      JsonStringToMessage(configuration->toString(), &config, json_options);
+  if (status != Status::OK) {
+    LOG_WARN(absl::StrCat(
+        "Cannot parse 'attributegen' plugin configuration JSON string [YAML is "
+        "not supported]: ",
+        configuration->toString()));
+    incrementMetric(config_errors_, 1);
+    return false;
```
kyessenov, line 105: it'll crash immediately at this point.

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,233 @@
...
+    std::string val;
+    auto eval_status = attribute_generator.evaluate(&val);
+    if (!eval_status) {
+      incrementMetric(runtime_errors_, 1);
+
+      // eval failed set error attribute
+      // __error attribute should be used by downstream plugins to distinguish
+      // between
+      // 1. No conditions matched.
+      // 2. Error evaluating a condition.
```
kyessenov, line 190: Why does it matter to downstream plugins? Either way, there is no value so the outcome is the same.

kyessenov:
Mostly looks good apart from a few comments. Will wait for other to finish the review.

extensions/attributegen/plugin.cc
bianpengyuan: +1. Can you elaborate some cases that you think it would be useful?

douglas-reid: checking for errors is often easier than remembering to reason about absence of values, imho. any policy-like extensions relying on attribute-gen may only want to operate on a full set. is there another way to signal error in generation?

extensions/attributegen/plugin_test.cc
```diff
@@ -0,0 +1,489 @@
...
+                     Http::TestResponseHeaderMapImpl& response_headers,
+                     const std::string& attribute, bool found, bool error,
+                     const std::string& value = "") {
+    auto fs = makeTestRequest(request_headers, response_headers);
+
+    ASSERT_EQ(fs->hasData<WasmState>(attribute + "__error"), error);
+    ASSERT_EQ(fs->hasData<WasmState>(attribute), found)
+        << absl::StrCat(attribute, "=?", value);
+    if (found) {
+      ASSERT_EQ(fs->getDataReadOnly<WasmState>(attribute).value(), value)
```
kyessenov, line 297: Since 1.14.1 PR merged, you want to update to `"wasm." + attribute` here I think.

extensions/attributegen/plugin.cc
```diff
@@ -0,0 +1,233 @@
...
+            " if (", matchconfig.condition(), ") -> ", matchconfig.value()));
+      }
+
+      tokens_.push_back(token);
+      matches.push_back(
+          Match(matchconfig.condition(), token, matchconfig.value()));
+    }
+    gen_.push_back(AttributeGenerator(
+        phase, attribute_gen_config.output_attribute(), std::move(matches)));
+    matches.clear();
```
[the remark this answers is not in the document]
mandarjog: I just checked the c++ standard, and it says the object is in a valid but unspecified state.
You may use any operation that does not require a pre-condition.
So `.clear()` will get it into a known good state.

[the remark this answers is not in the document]
bianpengyuan: Ok, why not declare a new vector at beginning of loop? seems easier to follow than using the same vector repeatedly.

mandarjog:
@bianpengyuan @kyessenov @douglas-reid all comments are addressed and `__error` attribute is removed.

bianpengyuan:
Woo, this was merged fast.. Wanted to block it if you want to fix the `matches` vector or wait for someone else to take another look. Anyway :)

PR #2803, opened 2020-04-15
add gen doc for extension protos

richardwxn:
initial effort to resolve: https://github.com/istio/istio/issues/22909

### DIFF
--- a/extensions/metadata_exchange/config.proto
+++ b/extensions/metadata_exchange/config.proto
@@ -15,10 +15,12 @@
 
 syntax = "proto3";
 
+// clang-format off
 // $title: Metadata Exchange Config
 // $description: Configuration for Metadata Exchange Filter.
-// $location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.html
+// $location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.html.
 // $weight: 20
+// clang-format on
 
 package metadata_exchange;
 
--- a/extensions/metadata_exchange/config.proto
+++ b/extensions/metadata_exchange/config.proto
@@ -15,10 +15,9 @@
 
 syntax = "proto3";
 
-// $schema: metadata_exchange
 // $title: Metadata Exchange Config
 // $description: Configuration for Metadata Exchange Filter.
-// $location: https://istio.io/docs/reference/config/metadata_exchange.html
+// $location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.html
 // $weight: 20
 
 package metadata_exchange;
--- a/extensions/metadata_exchange/config.proto
+++ b/extensions/metadata_exchange/config.proto
@@ -15,6 +15,12 @@
 
 syntax = "proto3";
 
+// $schema: metadata_exchange
+// $title: Metadata Exchange Config
+// $description: Configuration for Metadata Exchange Filter.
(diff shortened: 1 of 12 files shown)


extensions/metadata_exchange/config.proto
```diff
@@ -15,6 +15,12 @@
 
 syntax = "proto3";
 
+// $schema: metadata_exchange
```
jasonwzm, line 18: You can remove the schema here, as it is not being used now.

extensions/metadata_exchange/metadata_exchange.pb.html
```diff
@@ -0,0 +1,67 @@
+---
+title: metadata_exchange
```
jasonwzm, line 2: It seems protoc-gen-docs does not interpret the front matter you put in proto here.

kyessenov:
Should we copy these protos to istio/api instead? I'm not a fan of committing generated files into an active repo. Additionally, stats extension will likely not use its proto for config at all, so it becomes just a schema.

mandarjog:
@kyessenov are you suggesting copying the generated output to the api repo?
It can directly go to the docs site as well.

kyessenov:
Copying both protos and generated files.

mandarjog:
I want the schema to be authoritative here, so as long as that is satisfied, we are ok.

mandarjog:
@ericvn both Carolyn and Xinnan meet the requirements. They have a lot of code in the tools repo, and they should be able to approve that code now.

kyessenov:
But stats config.proto is soon no longer authoritative. It'll exist solely as schema and not be referenced in code.

richardwxn:
this is first attempt to establish the flow, if any proto file is no longer authoritative or some other proto files needed to be added, please comment here

richardwxn:
@kyessenov Can you clarify a bit more? If stats config.proto is not used anywhere, then we don't need to gen doc for it

kyessenov:
@richardwxn config.proto is no longer used after #2807 . The config is parsed by hand but it matches what would the proto do. So there is no need for this proto to live here, but it can act as a schema definition.

mandarjog:
@kyessenov we are still committing to maintaining proto as the authoritative schema, right?

kyessenov:
@mandarjog I'm not sure about that. Proto has weird things like duration as either string or struct that are a burden to maintain. Using map<string,string> prevents using flatbuffers as a schema as well.

kyessenov:
Why can't we just produce OpenAPI to begin with? Schema is disconnected from code already, and OpenAPI is a much nicer experience for declarative API development. All the tooling in istio/api ends up making OpenAPI anyways, and it's easy to generate HTML from OpenAPI.

mandarjog:
So you want to maintain openAPI as the authoritative source? We need to ensure that it keeps up-to-date with the actual structure used in code.

kyessenov:
Yes. OpenAPI as documentation that is manually kept in sync with custom code in extensions to keep extensions small in size.

mandarjog:
We decided that for 1.6, proto will remain the schema description language. The extensions are free to parse it as json or any other format.

Docs will start fetching from istio/proxy repo as well.

richardwxn:
I fixed the location tag issue by specifying per-file flag, @jasonwzm if that looks good to you feel free to remove the hold

PR #2804, opened 2020-04-15
feat(edges): add support for canonical service labels 

douglas-reid:
`WorkloadInstance` has been updated to support the canonical service labels server-side.

This PR updates the client definition and adds logic to populate that information. It will be used in Topology views related to canonical services.

As part of this PR, a few related consts are pulled out to a common location.


extensions/common/context.h
```diff
@@ -54,6 +54,11 @@ const std::string kProtocolHTTP = "http";
 const std::string kProtocolGRPC = "grpc";
 const std::string kProtocolTCP = "tcp";
 
+const char kCanonicalServiceLabelName[] = "service.istio.io/canonical-name";
+const char kCanonicalServiceRevisionLabelName[] =
+    "service.istio.io/canonical-revision";
+const std::string kLatest = "latest";
```
kyessenov, line 60: Can you please adhere to the modern standard (see ToW #140):
```c++
inline constexpr absl::string_view kLatest = "latest";
```

The problem with this code is that it duplicates variables every time this header is included.

kyessenov: For protos, you can then use `proto.set_field(kLatest.data()` or `proto.set_field(kLatest.data(), kLatest.size())`

douglas-reid: string_view didn't work with the flatbuffers api. trial by experimentation led to `const char[]`.

douglas-reid: but i can update `kLatest`

kyessenov: Just use `kLatest.data()`. Eventually it'll take string_view directly.

kyessenov: string_view is slightly faster to convert to a string since it avoids the scan for `\0`.

douglas-reid: btw, for anyone looking for the absl tip of the week link, here is the one I think was meant: https://abseil.io/tips/168

douglas-reid: adding `inline` resulted in failures. are we 100% on c++17 (because my build process isn't)?

douglas-reid:
> Can we add the new field into integration test?

Sorry, missed the file. Should be there now.

extensions/common/context.h
kyessenov: You can drop inline but use constexpr. The build is probably C++11/14.

douglas-reid: should be done now.

extensions/stats/plugin.cc
```diff
@@ -70,24 +70,24 @@ void map_node(IstioDimensions& instance, bool is_source,
...
 
-      auto rev =
-          source_labels->LookupByKey("service.istio.io/canonical-revision");
+      auto rev = source_labels->LookupByKey(
+          ::Wasm::Common::kCanonicalServiceRevisionLabelName.data());
       if (rev) {
         FB_ASSIGN(source_canonical_revision, rev->value());
       } else {
-        instance[source_canonical_revision] = "latest";
+        instance[source_canonical_revision] = ::Wasm::Common::kLatest.data();
```
kyessenov, line 84: This is expecting a string. Can you just pass a string view to avoid the scan? The compiler should auto convert from string_view to string.

douglas-reid:
```
extensions/stats/plugin.cc:84:45: error: no viable overloaded '='
        instance[source_canonical_revision] = ::Wasm::Common::kLatest;
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/basic_string.h:665:7: note: candidate function not viable: no known conversion from 'const absl::string_view' to 'const std::__cxx11::basic_string<char>' for 1st argument
      operator=(const basic_string& __str)
      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/basic_string.h:704:7: note: candidate function not viable: no known conversion from 'const absl::string_view' to 'const char *' for 1st argument
      operator=(const _CharT* __s)
      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/basic_string.h:715:7: note: candidate function not viable: no known conversion from 'const absl::string_view' to 'char' for 1st argument
      operator=(_CharT __c)
      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/basic_string.h:732:7: note: candidate function not viable: no known conversion from 'const absl::string_view' to 'std::__cxx11::basic_string<char>' for 1st argument
      operator=(basic_string&& __str)
      ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/basic_string.h:795:7: note: candidate function not viable: no known conversion from 'const absl::string_view' to 'initializer_list<char>' for 1st argument
      operator=(initializer_list<_CharT> __l)
```

douglas-reid: I'm beginning to have concerns about `absl::string_view` as it requires `.data()` calls when using with FlatBuffers (as far as I can tell).

From the docs:

> Note that `string_view::data()` may contain embedded nul
  // characters, but the returned buffer may or may not be NUL-terminated;
  // therefore, do not pass `data()` to a routine that expects a NUL-terminated
  // string.

I think we should use `char[]` here.

kyessenov: Not a big deal. It'll get much cleaner with C++17.

douglas-reid: sorry, what do you think we need here to get this in?

PR #2805, opened 2020-04-15
feat(node_info): add cluster_id to node_info

douglas-reid:
Populate node_info from from `CLUSTER_ID` values in the envoy node metadata (and exchanged via metadata exchange).

This PR will enable cluster-focused telemetry allowing monitoring of cross-cluster traffic, etc. Users will be able to opt-in to stats dimensioned by cluster_id when so desired.

Related PRs:
- [x]  https://github.com/istio/istio/pull/22975
- [x] https://github.com/istio/istio/pull/23095


mandarjog:
In single cluster deployments this seems like a waste, even though it gets us ready for multi cluster. Anything we can do?

douglas-reid:
> In single cluster deployments this seems like a waste, even though it gets us ready for multi cluster. Anything we can do?

we could not include the labels when the cluster ids are `Kubernetes`.  Or we could make this an optional config that gets set in the envoyfilters when multicluster is enabled at install time or similar?  Thoughts?

kyessenov:
1. Why doesn't OWNER field include the cluster in the URI?
2. It's fine to add a field and leave it unpopulated. It might cause slight perf cost due to regex.

kyessenov:
I think you need to patch configurable metric desired output to include cluster labels.

douglas-reid:
> 1. Why doesn't OWNER field include the cluster in the URI?

OWNER matches (for k8s) the kube resource URI, which doesn't include cluster name.

mandarjog:
@douglas-reid or @bianpengyuan please check the tsan failure.

bianpengyuan:
Not sure why connection is broken in the test. Probably should enable debugging log in CI to get better information.

bianpengyuan:
Let me try to reproduce the test failure locally.

bianpengyuan:
Let's skip if this rerun still does not work. I will spend some time to investigate.

douglas-reid:
@bianpengyuan yeah, it all passes locally for me.

nrjpoddar:
@douglas-reid can we make this optional instead of enabled by default?

douglas-reid:
@nrjpoddar this is a topic for discussion tomorrow for this very reason.

douglas-reid:
@kyessenov do we still need the node_info flatbuffer and context.cc parts of this PR, given that we are moving to opt-in with customizable metrics?  If not, I think I can close this entirely?

kyessenov:
Yes, we do need the schema update for node info to make it available.

douglas-reid:
@kyessenov OK. PTAL. I've removed all other bits of the PR.

PR #2806, opened 2020-04-16
Consolidate proxy tests to use step based framework

bianpengyuan:
Several notable changes:
* Instead of the specify a const string in env package, add a test inventory to specify tests for dedicated test port.
* Merged several tests.
    * merge original stackdriver plugin test into `TestStackdriverPayload` by adding context graph and latency verification there.
    * merge all access log filter test into `TestStackdriverAccessLog`.
    * merge tcp metadata exchange and tcp metadata exchange on tick into `TestTCPMetadataExchange`.
    * delete TestStatsPlugin since it is a duplication.
    * merge `TestStatsHHFallback` into `TestStatsPayload`.

Lots of clean up could be done after this PR. Many code in env package won't be needed any more. The only test that has not yet been moved is `TestHttpMetadataExchange` since it is skipped right now. I might remove the test when doing the clean up in follow up, since we will need to rewrite it anyway.


bianpengyuan:
All tests passed. PTAL.

test/envoye2e/inventory.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	env "istio.io/proxy/test/envoye2e/env"
+)
+
+var ProxyE2ETests *env.TestInventory
+
+func init() {
+	ProxyE2ETests = &env.TestInventory{
+		Tests: []string{
```
kyessenov, line 25: This is nice. We could probably parallellize sub-tests as well. I wonder if this could be auto-generated later with `go list`

test/envoye2e/driver/tcp.go
```diff
@@ -0,0 +1,133 @@
...
+		var conn net.Conn
+		var err error
+		conn, err = net.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", port))
+		if err != nil {
+			log.Println("Will wait 200ms and try again.")
+			time.Sleep(200 * time.Millisecond)
+			continue
+		}
+		// send to socket
+		fmt.Fprintf(conn, "ping"+"\n")
```
kyessenov, line 74: nit: why not `"ping\n"`?

test/envoye2e/driver/tcp.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return errors.New("timeout waiting for TCP server to be ready")
+}
+
+func handleConnection(conn net.Conn, prefix string) {
+	defer conn.Close()
+	reader := bufio.NewReader(conn)
+	for {
+		// read client request data
+		bytes, err := reader.ReadBytes(byte('\n'))
```
kyessenov, line 93: nit: I think `ReadString('\n')` is cleaner.

testdata/metric/client_request_total.yaml.tmpl
```diff
@@ -31,21 +31,21 @@ metric:
...
     value: ratings
   - name: destination_canonical_revision
     value: version-1
   - name: destination_service_name
-    value: 127.0.0.1:{{ .Vars.ClientPort }}
+    value: server
   - name: destination_service_namespace
     value: default
   - name: request_protocol
     value: http
```
kyessenov, line 44: This looks like an issue: it should be `grpc`. Maybe file an issue?

test/envoye2e/inventory.go
bianpengyuan: Yeah, seems like a good code gen target. Added a todo for it.

test/envoye2e/driver/tcp.go
bianpengyuan: Done.

test/envoye2e/driver/tcp.go
bianpengyuan: Done.

testdata/metric/client_request_total.yaml.tmpl
bianpengyuan: In the grpc test we only verify server side metrics. I updated the test to also verify client side metrics and now this `request_protocol` field is filled selectively based on presence of GrpcResponseCode.

testdata/metric/tcp_server_connection_open.yaml.tmpl
```diff
@@ -0,0 +1,48 @@
+name: istio_tcp_connections_opened_total
+type: COUNTER
+metric:
+- counter:
+    value: 10
+  label:
```
kyessenov, line 6: nit: we could probably template the labels to avoid repetition.

test/envoye2e/stackdriver_plugin/stackdriver_xds_test.go
```diff
@@ -138,45 +139,110 @@ filter_chains:
...
+        typed_config:
+          "@type": type.googleapis.com/udpa.type.v1.TypedStruct
+          type_url: envoy.extensions.filters.http.wasm.v3.Wasm
+          value:
+            config:
+              vm_config:
+                runtime: "envoy.wasm.runtime.null"
+                code:
+                  local: { inline_string: "envoy.wasm.metadata_exchange" }
+              configuration: "{ max_peer_cache_size: 20 }"
```
kyessenov, line 167: nit: we should probably use mx_outbound and mx_inbound root configs

kyessenov:
TSAN might need longer timeouts in places. I think the TCP dial needs a timeout and repeat loop.

bianpengyuan:
Thanks! I will do some clean up and dedup in following PRs. TSAN failure is because missing sleep step after envoy starts in tcp metadata exchange test.

PR #2807, opened 2020-04-16
Remove protobuf from stats extension

kyessenov:

- split context utilities into non-protobuf and protobuf ones;
- add a JSON library (`extensions/common/json.h`) to parse JSONs without protobuf and without exceptions (exceptions are not supported by Wasm)
- change metadata_exchange to consume JSON (reduction in size from 1.1M to 800KB)
- change stats to consume JSON (reduction in size from 1.1M to 228KB), additional ~30KB could be shaved off if we don't parse duration strings
- update tests to normalize config as strict JSON


kyessenov:
FYI @2807

kyessenov:
FYI @jplevyak

kyessenov:
We could use flatbuffers JSON parser but it doesn't support map<string, string>, so stats config does not map the way it is.

kyessenov:
Can we change `dimensions: map<string, string>` field in stats extension config to be a vector instead? That way I can re-use flatbuffers JSON parser, and we get a schema.

WDYT @mandarjog @nrjpoddar

kyessenov:
Flatbuffers JSON parser adds additional 300KB to the binary, so it's not particularly space efficient. I think we're good keeping raw JSON.

mandarjog:
Reduction to stats is big, mx is limited.
So we are keeping map<string, string> ?
I am ok moving to a list instead from a map. This API is not widely used yet, so we can change it now and generally starting using lists everywhere.

kyessenov:
MX still links protobuf, so it's expected. Basically it takes 200KB to link a proto with just one field, that's how bad it is :D 
Flatbuffers IDL parser is rather big (it's shared with flatc) so let's not use it, and just rely on raw JSON like it's done here. We can keep dimensions as map.

kyessenov:
@jplevyak I've tried to use libprotobuf-lite.a but it appears JSON util doesn't work with it:
```
/usr/local/include/google/protobuf/util/json_util.h:107:30: note: candidate function not viable: no known conversion from 'stats::PluginConfig *' to 'google::protobuf::Message *' for 2nd argument
PROTOBUF_EXPORT util::Status JsonStringToMessage(
```

It looks like JSON is not supported by protobuf lite.

bianpengyuan:
Are you planning to get this into 1.6?

kyessenov:
Yeah, let me fix merge conflicts.

kyessenov:
This is ready for review.

extensions/common/json_util.cc
```diff
@@ -0,0 +1,114 @@
...
+
+#include "extensions/common/json_util.h"
+
+#include "absl/strings/numbers.h"
+
+namespace Wasm {
+namespace Common {
+
+::nlohmann::json JsonParse(absl::string_view str) {
+  return ::nlohmann::json::parse(str, nullptr, false);
```
bianpengyuan, line 24: I guess we will need to extend this to support json marshaling in the future, will that be easy?

extensions/stats/plugin.cc
```diff
@@ -263,74 +267,103 @@ void PluginRootContext::initializeDimensions() {
...
+        auto type =
+            JsonGetField<absl::string_view>(definition, "type").value_or("");
+        if (type == "GAUGE") {
+          factory.type = MetricType::Gauge;
+        } else if (type == "HISTOGRAM") {
+          factory.type = MetricType::Histogram;
+        }
+        return true;
+      })) {
+    LOG_WARN("failed to parse 'dimensions'");
```
bianpengyuan, line 296: nit: failed to parse 'definitions'

extensions/stats/plugin.cc
```diff
@@ -377,28 +413,25 @@ void PluginRootContext::initializeDimensions() {
   build.record(
       1, "proxy",
       absl::StrCat(flatbuffers::GetString(local_node.istio_version()), ";"));
-}
+  return true;
+}  // namespace Stats
```
bianpengyuan, line 417: nit: seems to be wrong comment. This is not name of Stats namespace

mandarjog:
@bianpengyuan ping !

extensions/common/context.cc
```diff
@@ -198,51 +198,50 @@ TrafficDirection getTrafficDirection() {
...
+void extractEmptyNodeFlatBuffer(std::string* out) {
+  flatbuffers::FlatBufferBuilder fbb;
+  FlatNodeBuilder node(fbb);
+  auto data = node.Finish();
+  fbb.Finish(data);
+  out->assign(reinterpret_cast<const char*>(fbb.GetBufferPointer()),
+              fbb.GetSize());
+}
+
+bool extractPartialLocalNodeFlatBuffer(std::string* out) {
```
bianpengyuan, line 210: This is partial only because the extensions uses the method does not need the full node info? The reason in the comment is `avoid parsing a protobuf from the host`. Not clear why that makes this need to be partially populated

extensions/common/json_util.cc
kyessenov: Can you elaborate? There is `json::dump()` that prints out JSONs

extensions/stats/plugin.cc
kyessenov: good catch, thanks.

extensions/stats/plugin.cc
kyessenov: indeed, it's wrong.

extensions/common/json_util.cc
bianpengyuan: Like marshaling a struct to json or build a json string for outbound http request (OPA for example). Not sure if the library provides handy functions for this.

extensions/common/context.cc
kyessenov: This is because we can't parse the list of labels without using protobuf. I linked to the issue. We'd need a different encoding for lists and maps but that's another ABI change. The partial node info is sufficient for the stats extension but not stackdriver extension, for example.

extensions/common/json_util.cc
kyessenov: Building JSON is very easy with the library. E.g.:
```
json j2 = {
  {"pi", 3.141},
  {"happy", true},
  {"name", "Niels"},
  {"nothing", nullptr},
  {"answer", {
    {"everything", 42}
  }},
  {"list", {1, 0, 2}},
  {"object", {
    {"currency", "USD"},
    {"value", 42.99}
  }}
};
```

extensions/common/context.cc
bianpengyuan: Ok I see.

extensions/stats/plugin.cc
```diff
@@ -272,9 +272,14 @@ bool PluginRootContext::initializeDimensions(const json& j) {
         auto value =
             JsonGetField<std::string>(definition, "value").value_or("");
         if (name.empty() || value.empty()) {
-          return true;
+          LOG_WARN("empty name or value in  'definitions'");
```
bianpengyuan, line 275: Want to make it crash in these two cases?

kyessenov: Nope, changed to never return false. It'll just halt iteration but shouldn't break.

bianpengyuan: Wait, I am confused.. Shouldn't it return true here in order to halt the iteration?

kyessenov: The visitor call automatically halts iteration if any element returns false.

bianpengyuan: Oh sorry, I thought the last commit is to remove `return false` from the second last commit. nvm.

PR #2817, opened 2020-04-21
update envoy to 1.14.1

kyessenov:

- update Wasm to use the new declare/get property
- fix tests
- add a test using peer info

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -37,18 +37,23 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
-# envoy-wasm commit time: 4/20/20
+# Commit time: 4/20/20
+# Used by scripts/generate-wasm.sh
 ENVOY_SHA = "54db920505bc313e37d86921746dd1bf46dc1aa5"
 
 ENVOY_SHA256 = "d1b5546f830e1722bf1af6fde551a624346abf02540c3638e5a41941ff16c036"
 
+ENVOY_ORG = "istio"
+
+ENVOY_REPO = "envoy"
+
 # To override with local envoy, just pass `--override_repository=envoy=/PATH/TO/ENVOY` to Bazel or
 # persist the option in `user.bazelrc`.
 http_archive(
-    name = "envoy",
+    name = ENVOY_REPO,
     sha256 = ENVOY_SHA256,
-    strip_prefix = "envoy-" + ENVOY_SHA,
-    url = "https://github.com/istio/envoy/archive/" + ENVOY_SHA + ".tar.gz",
+    strip_prefix = ENVOY_REPO + "-" + ENVOY_SHA,
+    url = "https://github.com/" + ENVOY_ORG + "/" + ENVOY_REPO + "/archive/" + ENVOY_SHA + ".tar.gz",
 )
 
 load("@envoy//bazel:api_binding.bzl", "envoy_api_binding")
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -37,18 +37,18 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
-# envoy-wasm commit time: Wed Mar 20 20:46:00 2020 -0700
-ENVOY_SHA = "d29f7a659ba736aab97697a7bcfc69a71bc66b66"
+# envoy-wasm commit time: 4/20/20
+ENVOY_SHA = "54db920505bc313e37d86921746dd1bf46dc1aa5"
 
-ENVOY_SHA256 = "ffc2b25af02242a95bf0e65b2f9c4ac0248fb07ade6b5bb3517be934340dfec9"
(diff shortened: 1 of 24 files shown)


extensions/stackdriver/stackdriver.cc
```diff
@@ -346,8 +344,7 @@ inline bool StackdriverRootContext::enableEdgeReporting() {
 
 bool StackdriverRootContext::shouldLogThisRequest() {
   std::string shouldLog = "";
-  if (!getValue({"filter_state", ::Wasm::Common::kAccessLogPolicyKey},
-                &shouldLog)) {
+  if (!getValue({::Wasm::Common::kAccessLogPolicyKey}, &shouldLog)) {
```
bianpengyuan, line 347: I don't see this was declared as property anywhere. We should have test for this, wondering why test does not fail.

src/envoy/tcp/metadata_exchange/metadata_exchange.cc
```diff
@@ -269,35 +270,55 @@ void MetadataExchangeFilter::tryReadProxyData(Buffer::Instance& data) {
...
+      reinterpret_cast<const char*>(fbb.GetBufferPointer()), fbb.GetSize()));
+
+  auto key = config_->filter_direction_ == FilterDirection::Downstream
+                 ? ::Wasm::Common::kDownstreamMetadataKey
+                 : ::Wasm::Common::kUpstreamMetadataKey;
   read_callbacks_->connection().streamInfo().filterState()->setData(
-      key,
-      std::make_unique<::Envoy::Extensions::Common::Wasm::WasmState>(value),
-      StreamInfo::FilterState::StateType::Mutable,
+      absl::StrCat("wasm.", key), std::move(state),
```
bianpengyuan, line 306: Why attach the prefix here? Is it that Wasm runtime will attach the `wasm.` prefix by default so here it is to make the behavior consistent?

extensions/common/context.h
```diff
@@ -30,20 +30,16 @@ using StringView = absl::string_view;
 // Node metadata
 constexpr StringView WholeNodeKey = ".";
 
-constexpr StringView kUpstreamMetadataIdKey =
-    "envoy.wasm.metadata_exchange.upstream_id";
-constexpr StringView kUpstreamMetadataKey =
-    "envoy.wasm.metadata_exchange.upstream";
+constexpr StringView kUpstreamMetadataIdKey = "upstream_peer_id";
```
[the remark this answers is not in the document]
kyessenov: We want to hide the implementation detail of filter_state from Wasm ABI. If we had dots there, then it's difficult to guess whether the dot belongs to the name of the filter state key, or a field traversal. Using a flat namespace `wasm.variable_name` fits most cases for now. The prefix is to make sure we don't collide with existing `envoy.*` variables.

WORKSPACE
```diff
@@ -37,18 +37,18 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/envoyproxy/envoy-wasm/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
-# envoy-wasm commit time: Wed Mar 20 20:46:00 2020 -0700
-ENVOY_SHA = "d29f7a659ba736aab97697a7bcfc69a71bc66b66"
+# envoy-wasm commit time: 4/20/20
```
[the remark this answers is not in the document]
bianpengyuan: actually nvm. I think we might be going to switch back to envoy-wasm some time later, so let's keep it that way.

extensions/stackdriver/stackdriver.cc
kyessenov: It's set by access log policy extension. Declaration is not necessary, it'll default to a default declaration https://github.com/envoyproxy/envoy-wasm/commit/80e1a645c25f6243bfacf0eb5b9e96eb2e767aff#diff-965399965695ac238aae83e5d9bd5666R42. Get/set/declare now all use the same identifier for consistency.

src/envoy/tcp/metadata_exchange/metadata_exchange.cc
kyessenov: Yes. We should really be using Wasm ABI for this, so there's a lot of boilerplate.

extensions/stackdriver/stackdriver.cc
bianpengyuan: I see. `property` declaration is only needed for CEL expression. Find a typo during code reading: https://github.com/envoyproxy/envoy-wasm/blob/f80bba648062e95b98497b84ba20a8165828b082/source/extensions/common/wasm/context.cc#L428

kyessenov: @jplevyak might want to fix the macro pollution ^^^

kyessenov: getProperty should also be able to read inside buffers without using CEL compiler.

extensions/stackdriver/edges/mesh_edges_service_client.cc
```diff
@@ -79,9 +79,11 @@ void MeshEdgesServiceClientImpl::reportTrafficAssertions(
   LOG_TRACE("mesh edge services client: sending request '" +
             request.DebugString() + "'");
 
-  context_->grpcSimpleCall(
-      grpc_service_, kMeshEdgesService, kReportTrafficAssertions, request,
-      kDefaultTimeoutMillisecond, success_callback_, failure_callback_);
+  HeaderStringPairs initial_metadata;
+  context_->grpcSimpleCall(grpc_service_, kMeshEdgesService,
```
mandarjog, line 83: Is this because of an underlying API change?

kyessenov: Yeah, the initial metadata callback was not usable until this.

PR #2821, opened 2020-04-22
Remove directional root context in metadata exchange filter

bianpengyuan:
Not sure why metadata exchange filter has directional root context defined. If there is indeed some reason to have it, we need to fix the name of static global variable (s/register_Stats/register_MetadataExchange).

### DIFF
--- a/extensions/metadata_exchange/plugin.cc
+++ b/extensions/metadata_exchange/plugin.cc
@@ -64,14 +64,6 @@ bool serializeToStringDeterministic(const google::protobuf::Message& metadata,
 static RegisterContextFactory register_MetadataExchange(
     CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContext));
 
-static RegisterContextFactory register_StatsOutbound(
-    CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContextOutbound),
-    "mx_outbound");
-
-static RegisterContextFactory register_StatsInbound(
-    CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContextInbound),
-    "mx_inbound");
-
 void PluginRootContext::updateMetadataValue() {
   google::protobuf::Struct node_metadata;
   if (!getMessageValue({"node", "metadata"}, &node_metadata)) {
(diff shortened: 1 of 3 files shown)


kyessenov:
There are two roots to keep the caches separate. We could combine them into a two-level cache in one root, which seems to be cleaner. I could do that in a follow-up.

extensions/metadata_exchange/plugin.cc
```diff
@@ -64,14 +64,6 @@ bool serializeToStringDeterministic(const google::protobuf::Message& metadata,
 static RegisterContextFactory register_MetadataExchange(
     CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContext));
 
-static RegisterContextFactory register_StatsOutbound(
-    CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContextOutbound),
-    "mx_outbound");
-
-static RegisterContextFactory register_StatsInbound(
```
kyessenov, line 71: Thanks for catching that. The duplicate name is harmless since it's namespaced I think.

bianpengyuan: Is static variable follows the namespace scope? This seems to be causing problem in the integration test. The test fails when specifying mx_inbound/mx_outbound as root context, complaining it cannot find the root context. I think it might be race between these and the same static variables in stats plugin.

kyessenov: Might be some UB. Let's not try our luck and just clean up the names.

kyessenov: Do we need two stats roots if we don't have a cache there?

bianpengyuan: It uses the root context indication to distinguish listener direction at data path. It is not necessary, but need some code change and cares to make sure nothing breaks.

bianpengyuan: It will be a breaking change someone is not using the default installation. We could remove the usage of root id in the code, as well as from default configuration, but might need to keep the declaration in code just for safety and remove it later with some announcement. So not a free clean up :)

PR #2822, opened 2020-04-24
Add request host header to metric dimension.

bianpengyuan:
Add request host header to metric tag. It follows `use_host_header` option, and set to `unknown` when host header fallback is disabled.

@jshaughn

ref: https://github.com/istio/istio/issues/21820

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -99,7 +99,8 @@ void extractServiceName(const std::string& host,
 //   part as destination service name. Otherwise, fallback to use destination
 //   host for destination service name.
 void getDestinationService(const std::string& dest_namespace,
-                           bool use_host_header, std::string* dest_svc_host,
+                           bool use_host_header, std::string* request_host,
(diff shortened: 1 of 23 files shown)


douglas-reid:
It feels like we should follow the rough policy that we followed with the other additions for this release:

- add the information so that it is readily available
- add the dimension to the bootstrap config so that it can easily be used
- document how to add it to metrics w/ configurable metrics functionality

Then once we see how it works, etc., decide on adding it to the default set.

kyessenov:
There is a spot to document stats extensions https://github.com/istio/istio.io/pull/7137.
We can make it more concrete by enumerating various use cases (SNI, request_host, request_operation, multi cluster).

bianpengyuan:
Ok thanks! Make a lot of sense. Then the right change is to add the additional tag to bootstrap instead of doing it here. I will close this one then.

bianpengyuan:
@kyessenov sg, I will add it here.

PR #2824, opened 2020-04-24
Add documentation for access log plugin

gargnupur:
Add documentation for access log plugin


extensions/access_log_policy/config/v1alpha1/access_log_policy_config.proto
```diff
@@ -15,10 +15,19 @@
 
 syntax = "proto3";
 
+// $title: AccessLogPolicy Config
+// $description: Configuration for AccessLogPolicy Filter.
+// $location:
+// https://istio.io/docs/reference/config/proxy_extensions/accesslogpolicy.html
+// $weight: 20
+// This filter logs all error requests. Success requests are repeatedly logged
```
mandarjog, line 23: This filter is used in conjunction with another filter.
Can you add more details here?
Which attribute is produced, and how another filter is expected to use it.

gargnupur: Done.

gargnupur:
@mandarjog , @bianpengyuan : it's ready for review again..

bianpengyuan:
Block in case @mandarjog want to take a look.

extensions/access_log_policy/config/v1alpha1/access_log_policy_config.proto
mandarjog: Accesslog Policy plugin is a stateful http log sampler. It decides whether a request is logged based on the following rules.
1. All requests resulting in errors are logged.
2. First successful request within `log_window_duration ` from a specific source ip (source principal) is  logged.

The plugin records its decision in the `access_log_policy` with a value of  `"no"`. A downstream plugin may honor the  the attribute. For example, `Stackdriver` plugin will not produce an access log entry if this attribute is set.

gargnupur: Thanks! Changed.. this is much better :)

mandarjog:
@gargnupur @kyessenov this Plug-in predates our “istio.” Convention for injecting new attributes. So for this has just been an internal contract between stack driver and accesslog policy.

Here is what I suggest 
1. Change the name of the attribute to “istio.accesslog_sampled” 
2. Update the name in the SD plugin
3. Update the proto doc to talons our “attribute”, at present it does not mention this word.

extensions/metadata_exchange/config.pb.html
```diff
@@ -1,7 +1,7 @@
 ---
 title: Metadata Exchange Config
 description: Configuration for Metadata Exchange Filter.
-location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.html
+location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.html.
```
richardwxn, line 4: is this changed added manually or updated by make extensions-docs?

gargnupur: updated using make extension-docs.. I can revert this change?

gargnupur:
@mandarjog , @kyessenov : added "istio." in access log attribute ..
PTAL

PR #2831, opened 2020-04-27
fix location tag, update proto format

richardwxn:
1. location tag and url should be same line
2. location url should be placed together under proxy_extensions subdir
3. move position of next id block, otherwise it would be added to the doc.

### DIFF
--- a/extensions/access_log_policy/config/v1alpha1/access_log_policy_config.pb.html
+++ b/extensions/access_log_policy/config/v1alpha1/access_log_policy_config.pb.html
@@ -1,9 +1,9 @@
 ---
 title: AccessLogPolicy Config
 description: Configuration for AccessLogPolicy Filter.
+location: https://istio.io/docs/reference/config/proxy_extensions/accesslogpolicy.html
 layout: protoc-gen-docs
 generator: protoc-gen-docs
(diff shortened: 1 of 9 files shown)


mandarjog:
@richardwxn you need to run ./scripts/check-style to ensure correct format.

richardwxn:
@mandarjog Ok I fixed that, but actually the issue is that I we always need to add // clang-format off to the header part, otherwise the location tag line would be broken into two by the check-style script, which would make it fail to go to correct location in istio.io

mandarjog:
I think it is ok to turn if off for that reason. 
Can we add a comment above that says why it is turned off? A subsequent PR is fine.

mandarjog:
@richardwxn can you check why cherry pick is failing ?

mandarjog:
@bianpengyuan @richardwxn has code diverged in 1.6 already ? It should not be .

richardwxn:
@mandarjog actually not, the conflict is not related to the proto, but it is because git hard to resolve the conflicts in the comment section(edited header for the doc)

richardwxn:
fyi https://github.com/istio/proxy/pull/2833 @mandarjog

PR #2837, opened 2020-05-01
Make magic mismatch condition a warning and actionable

mandarjog:
This will emit actionable messages if https://github.com/istio/istio/issues/23401 condition recurs.


src/envoy/tcp/metadata_exchange/metadata_exchange.cc
```diff
@@ -227,7 +227,9 @@ void MetadataExchangeFilter::tryReadInitialProxyHeader(Buffer::Instance& data) {
       MetadataExchangeInitialHeader::magic_number) {
     config_->stats().initial_header_not_found_.inc();
     setMetadataNotFoundFilterState();
-    ENVOY_LOG(trace, "Alpn Protocol Matched. Magic not matched.");
+    ENVOY_LOG(warn,
+              "Incorrect istio-peer-exchange Alpn magic. Peer missing tcp "
```
douglas-reid, line 231: should `Alpn` be `ALPN` and `tcp` be `TCP` ?

PR #2841, opened 2020-05-07
fix server yaml and failure log

mandarjog:
In case of evaluation failure, log the expression that failed to eval.

### DIFF
--- a/extensions/attributegen/testdata/server.yaml
+++ b/extensions/attributegen/testdata/server.yaml
@@ -104,14 +104,14 @@ static_resources:
           - name: istio.attributegen
             typed_config:
               '@type': type.googleapis.com/udpa.type.v1.TypedStruct
-              type_url: type.googleapis.com/envoy.config.filter.http.wasm.v2.Wasm
+              type_url: type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm
               value:
                 config:
                   configuration: |
-                    {
+                    { "debug": "true",
                       "attributes": [
                         {
-                          "output_attribute": "istio.operationId",
+                          "output_attribute": "istio_operationId",
                           "match": [
                             {
                               "value": "GetStatus",
@@ -120,7 +120,7 @@ static_resources:
                           ]
                         },
                         {
-                          "output_attribute": "istio.responseClass",
+                          "output_attribute": "istio_responseClass",
                           "match": [
                             {
                               "value": "2xx",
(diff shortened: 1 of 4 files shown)


extensions/attributegen/testdata/server.yaml
```diff
@@ -104,14 +104,14 @@ static_resources:
...
+              type_url: type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm
               value:
                 config:
                   configuration: |
-                    {
+                    { "debug": "true",
                       "attributes": [
                         {
-                          "output_attribute": "istio.operationId",
+                          "output_attribute": "istio_operationId",
```
douglas-reid, line 114: is there a separate PR for fixing the docs re: namespacing of attributes?

extensions/stats/plugin.cc
```diff
@@ -587,8 +587,15 @@ bool PluginRootContext::report(::Wasm::Common::RequestInfo& request_info,
...
+      std::string expression_str;
+      for (const auto& expression : input_expressions_) {
+        if (expression.second == i) {
+          expression_str = expression.first;
+          break;
+        }
+      }
+      LOG_TRACE(absl::StrCat("Failed to evaluate expression: <", expression_str,
+                             "> at slot: ", std::to_string(i)));
       istio_dimensions_[count_standard_labels + i] = "";
```
douglas-reid, line 599: i'm beginning to wonder if `""` is the best value here. didn't we have something elsewhere in this that was auto-converting `""` to `"unknown"` ? are we now inconsistent in missing/errors for dimensions?

extensions/stats/plugin.cc
```diff
@@ -587,8 +587,15 @@ bool PluginRootContext::report(::Wasm::Common::RequestInfo& request_info,
...
-      LOG_TRACE(absl::StrCat("Failed to evaluate expression at slot: " +
-                             std::to_string(i)));
+      std::string expression_str;
+      for (const auto& expression : input_expressions_) {
+        if (expression.second == i) {
+          expression_str = expression.first;
+          break;
+        }
+      }
+      LOG_TRACE(absl::StrCat("Failed to evaluate expression: <", expression_str,
```
douglas-reid, line 597: is the `slot` information useful? i wouldn't know what to do with it.  is there a way to say which dimension the expression was intended to populate ?

mandarjog: Updated, adding dimension is a pain, i will take a fresh look tomorrow.

extensions/stats/plugin.cc
mandarjog: do we want to change this to unknown or unmapped?

extensions/attributegen/testdata/server.yaml
mandarjog: Will send another PR.

mandarjog:
@bianpengyuan updated to align with attributegen that keep token and expression in a struct.

extensions/stats/plugin.cc
bianpengyuan: `unknown` sounds better to me. It will be a good indicator that config takes effect, but evaluation fails.

bianpengyuan:
Need to update this line to unknown: https://github.com/istio/proxy/blob/147136932569f02b9959e7915ebeac55881bddca/testdata/metric/client_request_total_customized.yaml.tmpl#L23

PR #2843, opened 2020-05-08
Add TCP Metrics in Stackdriver

gargnupur:

Add TCP Metrics in Stackdriver

### DIFF
--- a/extensions/stackdriver/common/constants.h
+++ b/extensions/stackdriver/common/constants.h
@@ -29,6 +29,23 @@ constexpr char kClientResponseBytesMeasure[] = "client/response_bytes_measure";
 constexpr char kClientRoundtripLatenciesMeasure[] =
     "client/roundtrip_latencies_measure";
 
+constexpr char kServerConnectionsOpenCountMeasure[] =
+    "server/connection_open_count_measure";
+constexpr char kServerConnectionsCloseCountMeasure[] =
+    "server/connection_close_count_measure";
+constexpr char kServerReceivedBytesCountMeasure[] =
+    "server/received_bytes_count_measure";
+constexpr char kServerSentBytesCountMeasure[] =
+    "server/sent_bytes_count_measure";
+constexpr char kClientConnectionsOpenCountMeasure[] =
+    "client/connection_open_count_measure";
+constexpr char kClientConnectionsCloseCountMeasure[] =
+    "client/connection_close_count_measure";
+constexpr char kClientReceivedBytesCountMeasure[] =
+    "client/received_bytes_count_measure";
+constexpr char kClientSentBytesCountMeasure[] =
+    "client/sent_bytes_count_measure";
+
 // View names of metrics.
 constexpr char kServerRequestCountView[] = "server/request_count";
 constexpr char kServerRequestBytesView[] = "server/request_bytes";
@@ -38,6 +55,19 @@ constexpr char kClientRequestCountView[] = "client/request_count";
 constexpr char kClientRequestBytesView[] = "client/request_bytes";
 constexpr char kClientResponseBytesView[] = "client/response_bytes";
 constexpr char kClientRoundtripLatenciesView[] = "client/roundtrip_latencies";
+
+constexpr char kServerConnectionsOpenCountView[] =
+    "server/connection_open_count";
+constexpr char kServerConnectionsCloseCountView[] =
+    "server/connection_close_count";
+constexpr char kServerReceivedBytesCountView[] = "server/received_bytes_count";
+constexpr char kServerSentBytesCountView[] = "server/sent_bytes_count";
+constexpr char kClientConnectionsOpenCountView[] =
+    "client/connection_open_count";
+constexpr char kClientConnectionsCloseCountView[] =
+    "client/connection_close_count";
+constexpr char kClientReceivedBytesCountView[] = "client/received_bytes_count";
+constexpr char kClientSentBytesCountView[] = "client/sent_bytes_count";
 
 // Prefix for Istio metrics.
 constexpr char kIstioMetricPrefix[] = "istio.io/service/";
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -26,7 +26,7 @@ package stackdriver.config.v1alpha1;
 
 import "google/protobuf/duration.proto";
 
-// next id: 10
+// next id: 9
 
 message PluginConfig {
   // Optional. Controls whether to export server access log.
@@ -68,7 +68,4 @@ message PluginConfig {
   // metrics for HTTP traffic. Defaults to false (request and response size
   // metrics are enabled).
   bool disable_http_size_metrics = 8;
-
-  // Optional. Allows enabling of reporting of the metrics for TCP traffic.
-  bool enable_tcp_metrics = 9;
 }
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -26,7 +26,7 @@ package stackdriver.config.v1alpha1;
 
 import "google/protobuf/duration.proto";
 
-// next id: 9
+// next id: 10
 
 message PluginConfig {
   // Optional. Controls whether to export server access log.
@@ -68,4 +68,7 @@ message PluginConfig {
   // metrics for HTTP traffic. Defaults to false (request and response size
   // metrics are enabled).
   bool disable_http_size_metrics = 8;
+
+  // Optional. Allows enabling of reporting of the metrics for TCP traffic.
+  bool enable_tcp_metrics = 9;
 }
--- a/extensions/stackdriver/metric/record.cc
+++ b/extensions/stackdriver/metric/record.cc
@@ -93,61 +93,38 @@ TagKeyValueList getOutboundTagMap(
     const ::Wasm::Common::FlatNode& peer_node_info,
     const ::Wasm::Common::RequestInfo& request_info) {
   TagKeyValueList outboundMap = {
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          meshUIDKey(), flatbuffers::GetString(local_node_info.mesh_id()))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          requestProtocolKey(), std::string(request_info.request_protocol))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          serviceAuthenticationPolicyKey(),
-          std::string(::Wasm::Common::AuthenticationPolicyString(
-              request_info.service_auth_policy)))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationServiceNameKey(),
-          std::string(request_info.destination_service_name))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationServiceNamespaceKey(),
-          flatbuffers::GetString(peer_node_info.namespace_()))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationPortKey(), std::to_string(request_info.destination_port))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          sourcePrincipalKey(), std::string(request_info.source_principal))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          sourceWorkloadNameKey(),
-          flatbuffers::GetString(local_node_info.workload_name()))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          sourceWorkloadNamespaceKey(),
-          flatbuffers::GetString(local_node_info.namespace_()))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          sourceOwnerKey(), Common::getOwner(local_node_info))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationPrincipalKey(),
-          std::string(request_info.destination_principal))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationWorkloadNameKey(),
-          flatbuffers::GetString(peer_node_info.workload_name()))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationWorkloadNamespaceKey(),
-          flatbuffers::GetString(peer_node_info.namespace_()))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationOwnerKey(), Common::getOwner(peer_node_info))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationCanonicalServiceNameKey(),
-          getPeerCanonicalName(peer_node_info))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationCanonicalServiceNamespaceKey(),
-          flatbuffers::GetString(peer_node_info.namespace_()))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          destinationCanonicalRevisionKey(),
-          getPeerCanonicalRev(peer_node_info))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          sourceCanonicalServiceNameKey(),
-          getLocalCanonicalName(local_node_info))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          sourceCanonicalServiceNamespaceKey(),
-          flatbuffers::GetString(local_node_info.namespace_()))},
-      {std::make_pair<opencensus::tags::TagKey, std::string>(
-          sourceCanonicalRevisionKey(),
-          getLocalCanonicalRev(local_node_info))}};
+      {meshUIDKey(), flatbuffers::GetString(local_node_info.mesh_id())},
+      {requestProtocolKey(), std::string(request_info.request_protocol)},
+      {serviceAuthenticationPolicyKey(),
+       std::string(::Wasm::Common::AuthenticationPolicyString(
+           request_info.service_auth_policy))},
+      {destinationServiceNameKey(),
+       std::string(request_info.destination_service_name)},
+      {destinationServiceNamespaceKey(),
+       flatbuffers::GetString(peer_node_info.namespace_())},
+      {destinationPortKey(), std::to_string(request_info.destination_port)},
+      {sourcePrincipalKey(), std::string(request_info.source_principal)},
+      {sourceWorkloadNameKey(),
+       flatbuffers::GetString(local_node_info.workload_name())},
+      {sourceWorkloadNamespaceKey(),
+       flatbuffers::GetString(local_node_info.namespace_())},
+      {sourceOwnerKey(), Common::getOwner(local_node_info)},
+      {destinationPrincipalKey(),
+       std::string(request_info.destination_principal)},
+      {destinationWorkloadNameKey(),
+       flatbuffers::GetString(peer_node_info.workload_name())},
+      {destinationWorkloadNamespaceKey(),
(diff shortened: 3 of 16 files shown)


extensions/stackdriver/common/constants.h
```diff
@@ -29,6 +29,23 @@ constexpr char kClientResponseBytesMeasure[] = "client/response_bytes_measure";
 constexpr char kClientRoundtripLatenciesMeasure[] =
     "client/roundtrip_latencies_measure";
 
+constexpr char kServerConnectionsOpenCountMeasure[] =
+    "server/connection_open_count_measure";
+constexpr char kServerConnectionsCloseCountMeasure[] =
+    "server/connection_close_count_measure";
```
mandarjog, line 35: Are all these existing names used in mixer v1?

gargnupur: For the size one's yes, but mixer v1 didn't have the connection count metrics, so name is same as here: https://istio.io/docs/reference/config/policy-and-telemetry/metrics/
@douglas-reid is helping update the schema here..

extensions/stackdriver/stackdriver.h
```diff
@@ -114,19 +120,52 @@ class StackdriverRootContext : public RootContext {
...
-  StackdriverContext(uint32_t id, RootContext* root) : Context(id, root) {}
+  StackdriverContext(uint32_t id, RootContext* root)
+      : Context(id, root), is_tcp_(false), context_id_(id) {
+    request_info_ = std::make_shared<::Wasm::Common::RequestInfo>();
+  }
   void onLog() override;
 
+  FilterStatus onNewConnection() override {
+    is_tcp_ = true;
+    request_info_->tcp_connections_opened++;
```
bianpengyuan, line 140: I'd prefer not let stream context own request_info_. Code is clear that way instead of passing around. You can call `addToTCPRequestQueue` and let root context initialize a request info there.

extensions/stackdriver/stackdriver.cc
```diff
@@ -292,44 +316,76 @@ bool StackdriverRootContext::onDone() {
...
+      continue;
+    }
+    record(*item.second, true);
+  }
+  tcp_request_queue_.clear();
   return done;
 }
 
-void StackdriverRootContext::record() {
+bool StackdriverRootContext::record(::Wasm::Common::RequestInfo& request_info,
```
bianpengyuan, line 330: Could you separate out TCP record logic to a different function (`recordTCP`)? And you can move the loop in onTick to that function.

extensions/stackdriver/common/constants.h
gargnupur: Thanks.. will change default to off for the metrics to be reported...

extensions/stackdriver/stackdriver.h
gargnupur: @bianpengyuan : I am open to suggestions but why is that? I feel it belongs here, because we edit, size and connection total metrics in stream context. It's lifecycle is tied to stream's lifecycle this way...

extensions/stackdriver/stackdriver.cc
gargnupur: Sure..

gargnupur: +1 for moving the logic of TCP to a different function..
but moving the loop in onTick to that function might be tricky.. we call record at the end of a connection too, to make sure this connection does gets recorded if ticker didn't pick it up.. 
will add a different function for loop that onTick can use..so that code their is concise..

@mandarjog : thinking about this, I remember you mentioned that we always want to capture connection open.. currently, we call record on ticker and when connection ends. Is that enough? I feel if connection is going to last for long, ticker will most likely pick it up and then connection end record is always there...

bianpengyuan: Ok I see. Yeah then keeping the loop out of the record call makes sense to me.

Record on open would introduce some overhead, so better avoid doing that.

mandarjog: We need to record at begin , ticker and end  for tcp. When keep alives are used we should see one tcp connection and many http on top of it.

bianpengyuan: We already record at beginning. We only record connection open event at new connection now (probably will need to record the time as well, but that could happen later), instead of the full connection information. Those information could be backfilled ontick or onclose, which is the current behavior and I think it is good enough.

extensions/stackdriver/stackdriver.h
bianpengyuan: I am hoping not to pass the pointer around if it could belong to one place. HTTP request info could be construct on demand in root context. TCP connection info is already stored as a vector in root context, and stream context could just call root context functions to increment different fields.

This code used to store request info in each stream and it was removed in favor of current way for code simplicity.

extensions/stackdriver/common/constants.h
gargnupur: Fixed

extensions/stackdriver/stackdriver.h
gargnupur: Fixed..PTAL

extensions/stackdriver/stackdriver.cc
```diff
@@ -332,6 +365,55 @@ void StackdriverRootContext::record() {
...
+  const bool outbound = isOutbound();
+  const auto& metadata_key =
+      outbound ? kUpstreamMetadataKey : kDownstreamMetadataKey;
+  const auto& metadata_id_key =
+      outbound ? kUpstreamMetadataIdKey : kDownstreamMetadataIdKey;
+  std::string peer_id;
+  getValue({metadata_id_key}, &peer_id);
+  std::string peer;
+  bool peer_found = getValue({metadata_key}, &peer);
+  const ::Wasm::Common::FlatNode& peer_node =
```
gargnupur, line 378: will try to refactor this into common code with record...

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -68,4 +68,7 @@ message PluginConfig {
   // metrics for HTTP traffic. Defaults to false (request and response size
   // metrics are enabled).
   bool disable_http_size_metrics = 8;
+
+  // Optional. Allows enabling of reporting of the metrics for TCP traffic.
+  bool enable_tcp_metrics = 9;
```
bianpengyuan, line 73: I am wondering do we need this option. We can always record and send the TCP metrics when this this filter is added to a network filter chain. The other thing that this option controls is whether to set up timer. I think we can also just start timer regardless of this option or other enablement. i.e. replace
```
if (enableServerAccessLog() || enableEdgeReporting() || enableTCPMetrics()) {
  proxy_set_tick_period_milliseconds(getLoggingExportIntervalMilliseconds());
} else {
  ...
}
```
with
```
proxy_set_tick_period_milliseconds(getLoggingExportIntervalMilliseconds());
```

gargnupur: What about the scenario if we want to get TCP connection metrics for HTTP?

extensions/stackdriver/metric/record.cc
```diff
@@ -28,7 +28,8 @@ namespace Metric {
 void record(bool is_outbound, const ::Wasm::Common::FlatNode& local_node_info,
             const ::Wasm::Common::FlatNode& peer_node_info,
             const ::Wasm::Common::RequestInfo& request_info,
-            bool record_http_size_metrics) {
+            bool record_http_size_metrics, bool record_tcp_metrics,
```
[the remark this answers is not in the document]
gargnupur: Needed 2 to distinguish between the scenario for whether it's tcp request or not and whether tcp requests metrics. are enabled or not.. I think I will actually make a separate function for TCP.. that way TagMap can be differentiated too

extensions/stackdriver/metric/record.cc
```diff
@@ -108,19 +109,34 @@ void record(bool is_outbound, const ::Wasm::Common::FlatNode& local_node_info,
...
+      if (record_tcp_metrics) {
+        opencensus::stats::Record(
+            {{clientConnectionsOpenCountMeasure(),
+              request_info.tcp_connections_opened},
+             {clientConnectionsCloseCountMeasure(),
+              request_info.tcp_connections_closed},
+             {clientReceivedBytesCountMeasure(),
+              request_info.tcp_received_bytes},
+             {clientSentBytesCountMeasure(), request_info.tcp_sent_bytes}},
+            tagMap);
```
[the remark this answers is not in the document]
gargnupur: I think in our current stackdriver metric schema, we do... but will change this to 2 functions to allow for separate tagMap as we don't want response_code in TCP.

extensions/stackdriver/stackdriver.cc
```diff
@@ -311,34 +343,82 @@ void StackdriverRootContext::record() {
...
     logger_->addLogEntry(request_info, peer_node);
   }
   if (enableEdgeReporting()) {
     std::string peer_id;
-    if (!getValue({::Wasm::Common::kDownstreamMetadataIdKey}, &peer_id)) {
-      LOG_DEBUG(absl::StrCat(
-          "cannot get metadata for: ", ::Wasm::Common::kDownstreamMetadataIdKey,
-          "; skipping edge."));
+    if (!getPeerId(peer_id)) {
+      LOG_DEBUG(absl::StrCat("cannot get metadata for: ", peer_id,
```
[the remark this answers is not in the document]
gargnupur: Thanks.. will fix..

extensions/stackdriver/metric/record.cc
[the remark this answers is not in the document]
bianpengyuan: But this function only execute the tcp record logic when both booleans are false, right?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
bianpengyuan: That will also require to inject this filter into listener filter chains.

extensions/stackdriver/metric/record.cc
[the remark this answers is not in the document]
bianpengyuan: they are actually different. See https://cloud.google.com/monitoring/api/metrics_istio. TCP metrics does not have request_operation, response_code, api labels.

extensions/stackdriver/stackdriver.cc
```diff
@@ -311,34 +343,82 @@ void StackdriverRootContext::record() {
...
+  }
+  // Record TCP Metrics.
+  ::Extensions::Stackdriver::Metric::record(
+      outbound, local_node, peer_node, request_info,
+      !config_.disable_http_size_metrics(), enableTCPMetrics(),
+      true /* is_tcp */);
+
+  // Add LogEntry to Logger. Log Entries are batched and sent on timer
+  // to Stackdriver Logging Service.
+  if (enableServerAccessLog() && shouldLogThisRequest()) {
```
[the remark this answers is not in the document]
gargnupur: Just for this PR.. In Access Log PR.. this will be changed to enableTCPServerAccessLog... we currently get access logs for TCP connection at the end and thus didn't want to remove that.

extensions/stackdriver/stackdriver.cc
```diff
@@ -311,34 +343,82 @@ void StackdriverRootContext::record() {
...
+      *flatbuffers::GetRoot<::Wasm::Common::FlatNode>(
+          peer_found ? peer.data() : empty_node_info_.data());
+  const ::Wasm::Common::FlatNode& local_node = getLocalNode();
+  const ::Wasm::Common::FlatNode& destination_node_info =
+      outbound ? peer_node : local_node;
+
+  if (tcp_request_queue_[id] == nullptr) {
+    return false;
+  }
+  ::Wasm::Common::RequestInfo request_info = *tcp_request_queue_[id];
```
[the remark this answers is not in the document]
gargnupur: Thanks.. will fix

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: Agreed to that point.. but I thought we wanted to disable metrics till we have a stackdriver schema done for TCP.. and we can control addition of this filter via a bool in Istio but then we will not be able to get TCP access log too, if we control just addition of this filter in listener filter chains.. it would be good to have an option to enable metrics and logs separately..

extensions/stackdriver/stackdriver.h
```diff
@@ -118,19 +139,52 @@ class StackdriverRootContext : public RootContext {
...
 // the request stream itself.
 class StackdriverContext : public Context {
  public:
-  StackdriverContext(uint32_t id, RootContext* root) : Context(id, root) {}
+  StackdriverContext(uint32_t id, RootContext* root)
+      : Context(id, root), is_tcp_(false), context_id_(id) {}
   void onLog() override;
 
+  FilterStatus onNewConnection() override {
+    if (!getRootContext()->initialized()) {
```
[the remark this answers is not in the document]
gargnupur: @bianpengyuan : looks like initialized_ variable gets set in rootcontext in onConfigure. Do we have ordering that streamcontext constructor will get called after rootcontext's onConfigure. Also onConfigure can be called multiple times I think.. so we can't really set is_initialized_ in constructor of streamcontext afaict...

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: As discussed offline.. keeping it for now and will remove it once we have schema ready in stackdriver.

extensions/stackdriver/metric/record.cc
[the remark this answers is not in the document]
gargnupur: Refactored the whole flow here , and thus both these booleans are gone.. PTAL..

extensions/stackdriver/metric/record.cc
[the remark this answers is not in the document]
gargnupur: Thanks.. Removed these labels for TCP metrics... PTAL

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
gargnupur: DOne

extensions/stackdriver/stackdriver.cc
```diff
@@ -311,34 +343,82 @@ void StackdriverRootContext::record() {
...
+  bool peer_found = getValue(
+      {outbound ? kUpstreamMetadataKey : kDownstreamMetadataKey}, &peer);
+  const ::Wasm::Common::FlatNode& peer_node =
+      *flatbuffers::GetRoot<::Wasm::Common::FlatNode>(
+          peer_found ? peer.data() : empty_node_info_.data());
+  const ::Wasm::Common::FlatNode& local_node = getLocalNode();
+  const ::Wasm::Common::FlatNode& destination_node_info =
+      outbound ? peer_node : local_node;
+
+  if (tcp_request_queue_[id] == nullptr) {
```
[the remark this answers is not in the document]
gargnupur: Thanks..Fixed

extensions/stackdriver/stackdriver.h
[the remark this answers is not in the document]
bianpengyuan: Stream context should always be constructed after root context finishes onConfiguration and onStart. Not sure I follow, why onConfigure called multiple times would be a problem here?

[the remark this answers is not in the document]
gargnupur: Thanks wanted to confirm the ordering.. will change it!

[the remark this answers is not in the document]
gargnupur: Thanks Fixed.. PTAL

extensions/stackdriver/metric/record.cc
```diff
@@ -60,54 +56,185 @@ void record(bool is_outbound, const ::Wasm::Common::FlatNode& local_node_info,
...
+  return inboundMap;
+}
+
+void addHttpSpecificTags(const ::Wasm::Common::RequestInfo& request_info,
+                         TagKeyValueList& tag_map) {
+  const auto& operation =
+      request_info.request_protocol == ::Wasm::Common::kProtocolGRPC
+          ? request_info.request_url_path
+          : request_info.request_operation;
+  tag_map.push_back(std::make_pair(Metric::requestOperationKey(), operation));
```
bianpengyuan, line 222: use emplace_back instead to avoid extra copy.

extensions/stackdriver/metric/record.cc
```diff
@@ -60,54 +56,185 @@ void record(bool is_outbound, const ::Wasm::Common::FlatNode& local_node_info,
...
-         flatbuffers::GetString(local_node_info.namespace_())},
-        {sourceCanonicalRevisionKey(), local_canonical_rev
-                                           ? local_canonical_rev->str()
-                                           : ::Wasm::Common::kLatest.data()}};
+TagKeyValueList getOutboundTagMap(
+    const ::Wasm::Common::FlatNode& local_node_info,
+    const ::Wasm::Common::FlatNode& peer_node_info,
+    const ::Wasm::Common::RequestInfo& request_info) {
+  TagKeyValueList outboundMap = {
+      {std::make_pair<opencensus::tags::TagKey, std::string>(
```
bianpengyuan, line 96: I am wondering do you need to add `std::make_pair<opencensus::tags::TagKey, std::string>`, or just the the pair of strings would also compile, like the original code.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
bianpengyuan: sorry to reopen this, but I am still hesitate to add this for temporary state. Can we just comment out invocation of recordTCP() (or any function that triggers connection stats recording) when cherry-pick to 1.6 and make a new build when TCP metric is ready?

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
gargnupur: If the other PR doesn't get in in time, we will loose the access log we were getting for TCP before...

extensions/stackdriver/stackdriver.h
```diff
@@ -71,8 +71,26 @@ class StackdriverRootContext : public RootContext {
 
   bool useHostHeaderFallback() const { return use_host_header_fallback_; };
 
+  // Records telemetry for the current active stream/connection.
+  bool recordTCP(uint32_t id);
```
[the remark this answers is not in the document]
gargnupur: Fixed

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: Done..

extensions/stackdriver/metric/record.cc
gargnupur: pair of strings worked... it didn't work originally because I think there were some stringview in there and I had to convert them to std::string.. fixed

extensions/stackdriver/metric/record.cc
gargnupur: Thanks..Fixed

extensions/stackdriver/metric/registry.cc
```diff
@@ -219,15 +247,18 @@ StackdriverOptions getStackdriverOptions(
     view_descriptor.RegisterForExport();                            \
   }
 
-#define ADD_TAGS                                             \
-  .add_column(requestOperationKey())                         \
-      .add_column(requestProtocolKey())                      \
+#define ADD_TAGS ADD_COMMON_TAGS ADD_HTTP_HRPC_TAGS
+
+#define ADD_HTTP_HRPC_TAGS \
```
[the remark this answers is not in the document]
gargnupur: Thanks..Fixed

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
gargnupur: Removed.. yeah this wont happen unless this filter is added in TCP chain...

PR #2852, opened 2020-05-14
Add TCP Access Logging

gargnupur:
This is built on top of https://github.com/istio/proxy/pull/2843

Ref: https://github.com/istio/istio/issues/23134

### DIFF
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -26,13 +26,13 @@ package stackdriver.config.v1alpha1;
 
 import "google/protobuf/duration.proto";
 
-// next id: 11
+// next id: 10
 message PluginConfig {
   // Optional. Controls whether to export server access log.
   bool disable_server_access_logging = 1;
 
   // Optional. Controls whether to export TCP server access log.
-  bool enable_tcp_server_access_logging = 10;
+  bool disable_tcp_server_access_logging = 9;
 
   // Optional. FQDN of destination service that the request routed to, e.g.
   // productpage.default.svc.cluster.local. If not provided, request host header
@@ -70,7 +70,4 @@ message PluginConfig {
   // metrics for HTTP traffic. Defaults to false (request and response size
   // metrics are enabled).
   bool disable_http_size_metrics = 8;
-
-  // Optional. Allows enabling of reporting of the metrics for TCP traffic.
-  bool enable_tcp_metrics = 9;
 }
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -26,13 +26,10 @@ package stackdriver.config.v1alpha1;
 
 import "google/protobuf/duration.proto";
 
-// next id: 10
+// next id: 9
 message PluginConfig {
   // Optional. Controls whether to export server access log.
   bool disable_server_access_logging = 1;
-
-  // Optional. Controls whether to export TCP server access log.
-  bool disable_tcp_server_access_logging = 9;
 
   // Optional. FQDN of destination service that the request routed to, e.g.
   // productpage.default.svc.cluster.local. If not provided, request host header
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -26,11 +26,13 @@ package stackdriver.config.v1alpha1;
 
 import "google/protobuf/duration.proto";
 
-// next id: 9
-
+// next id: 11
 message PluginConfig {
   // Optional. Controls whether to export server access log.
   bool disable_server_access_logging = 1;
+
+  // Optional. Controls whether to export TCP server access log.
+  bool enable_tcp_server_access_logging = 10;
 
   // Optional. FQDN of destination service that the request routed to, e.g.
   // productpage.default.svc.cluster.local. If not provided, request host header
@@ -68,4 +70,7 @@ message PluginConfig {
   // metrics for HTTP traffic. Defaults to false (request and response size
   // metrics are enabled).
   bool disable_http_size_metrics = 8;
+
+  // Optional. Allows enabling of reporting of the metrics for TCP traffic.
+  bool enable_tcp_metrics = 9;
 }
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -221,13 +221,13 @@ void Logger::addTCPLabelsToLogEntry(
     const ::Wasm::Common::RequestInfo& request_info,
     google::logging::v2::LogEntry* log_entry) {
   auto label_map = log_entry->mutable_labels();
-  (*label_map)["tcp_remote_ip"] = request_info.source_address;
-  (*label_map)["tcp_server_ip"] = request_info.destination_address;
-  (*label_map)["tcp_remote_port"] = request_info.source_port;
-  (*label_map)["tcp_server_port"] = request_info.destination_port;
-  (*label_map)["tcp_sent_bytes"] = request_info.tcp_total_sent_bytes;
-  (*label_map)["tcp_received_bytes"] = request_info.tcp_total_received_bytes;
-  (*label_map)["tcp_connection_state"] =
+  (*label_map)["source_ip"] = request_info.source_address;
+  (*label_map)["destination_ip"] = request_info.destination_address;
+  (*label_map)["source_port"] = request_info.source_port;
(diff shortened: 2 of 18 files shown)


gargnupur:
@bianpengyuan , @mandarjog , @douglas-reid : This is ready for review..

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -26,12 +26,14 @@ package stackdriver.config.v1alpha1;
...
 
-// next id: 9
-
+// next id: 10
 message PluginConfig {
   // Optional. Controls whether to export server access log.
   bool disable_server_access_logging = 1;
 
+  // Optional. Controls whether to export TCP server access log.
+  bool disable_tcp_server_access_logging = 9;
```
bianpengyuan, line 35: Shall we just use `disable_server_access_logging` to control both tcp and http logging?

extensions/stackdriver/log/logger.cc
```diff
@@ -226,6 +217,41 @@ bool Logger::exportLogEntry(bool is_on_done) {
   return true;
 }
 
+void Logger::addTCPLabelsToLogEntry(
+    const ::Wasm::Common::RequestInfo& request_info,
+    google::logging::v2::LogEntry* log_entry) {
+  auto label_map = log_entry->mutable_labels();
+  (*label_map)["tcp_remote_ip"] = request_info.source_address;
```
bianpengyuan, line 224: Do we need `tcp` prefix? What did we use in mixer v1?

extensions/stackdriver/stackdriver.cc
```diff
@@ -403,6 +404,12 @@ bool StackdriverRootContext::recordTCP(uint32_t id) {
   // Record TCP Metrics.
   ::Extensions::Stackdriver::Metric::recordTCP(outbound, local_node, peer_node,
                                                request_info);
+  // Add LogEntry to Logger. Log Entries are batched and sent on timer
+  // to Stackdriver Logging Service.
+  if (enableTCPServerAccessLog() && shouldLogThisRequest()) {
```
bianpengyuan, line 409: Does `shouldLogThisRequest` apply here? It only bases on http or grpc status right?

extensions/stackdriver/stackdriver.cc
```diff
@@ -403,6 +404,12 @@ bool StackdriverRootContext::recordTCP(uint32_t id) {
...
+    ::Wasm::Common::populateExtendedRequestInfo(&request_info);
+    logger_->addLogEntry(request_info, peer_node, true);
```
bianpengyuan, line 411:
```suggestion
    logger_->addLogEntry(request_info, peer_node, /* is_tcp = */ true);
```

extensions/stackdriver/stackdriver.cc
```diff
@@ -351,7 +352,7 @@ void StackdriverRootContext::record() {
       !config_.disable_http_size_metrics());
   if (enableServerAccessLog() && shouldLogThisRequest()) {
     ::Wasm::Common::populateExtendedHTTPRequestInfo(&request_info);
-    logger_->addLogEntry(request_info, peer_node);
+    logger_->addLogEntry(request_info, peer_node, false);
```
bianpengyuan, line 355:
```suggestion
    logger_->addLogEntry(request_info, peer_node, /* is_tcp = */ false);
```

extensions/stackdriver/stackdriver.h
```diff
@@ -80,7 +80,11 @@ class StackdriverRootContext : public RootContext {
   void deleteFromTCPRequestQueue(uint32_t id);
   void incrementReceivedBytes(uint32_t id, size_t size);
   void incrementSentBytes(uint32_t id, size_t size);
+  void incrementTotalReceivedBytes(uint32_t id, size_t size);
```
bianpengyuan, line 83: Why not reuse the existing functions?

extensions/stackdriver/stackdriver.h
```diff
@@ -155,6 +162,8 @@ class StackdriverContext : public Context {
 
     is_tcp_ = true;
     getRootContext()->addToTCPRequestQueue(context_id_);
+    getRootContext()->setConnectionState(
```
bianpengyuan, line 165: I think you will need to trigger log record in connection open, otherwise it will not be recorded. Also I think it is necessary to record log entry in open and close event, so that they could have correct time stamp.

bianpengyuan: And test make sure that all three types of connection states are covered.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: Good idea.. removing

extensions/stackdriver/log/logger.cc
gargnupur: Changed them to match mixer v1

extensions/stackdriver/stackdriver.cc
gargnupur: Fixed

extensions/stackdriver/stackdriver.cc
gargnupur: Removed for now.. but I was thinking more in terms of sampling successful connections too but will change accesslog sampling filter to select failed request based on presence of response flag.. will do in a separate PR

extensions/stackdriver/stackdriver.cc
gargnupur: Fixed

extensions/stackdriver/stackdriver.h
gargnupur: Because the metrics are different... this is incrementing tcp_total_received_bytes...

extensions/stackdriver/stackdriver.h
gargnupur: Yes.. I was going to ping you offline about this :) I had exact same question in my other PR.. https://github.com/istio/proxy/pull/2843#discussion_r424828627 but reading it now, I think you were referring to Metrics and I was referring to calling Record method in general that does everything.

One of the things for TCP would be that metadata exchange doesn't happen until data comes so, connection open log won't have peer data. But it would be good to still get connection open logged.

gargnupur: Will add this change in a separate PR because of metadata exchange stuff..

extensions/stackdriver/stackdriver.h
bianpengyuan: You can increment total_x_bytes inside the existing function.

extensions/stackdriver/stackdriver.h
bianpengyuan: Ok, I think time stamp is the only thing special with connection open. All other labels should have no difference from other state. You can keep an connection open time stamp somewhere and use that as a signal to record connection open event.

bianpengyuan: Are you planning to fix this in separated PR?

extensions/stackdriver/stackdriver.h
bianpengyuan: This is not fixed yet.

extensions/stackdriver/stackdriver.cc
```diff
@@ -203,7 +204,7 @@ bool StackdriverRootContext::configure(size_t configuration_size) {
     }
   }
 
-  if (!logger_ && enableServerAccessLog()) {
+  if (!logger_ && (enableServerAccessLog())) {
```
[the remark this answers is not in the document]
gargnupur: Thanks..Fixed

extensions/stackdriver/stackdriver.cc
```diff
@@ -313,7 +314,7 @@ bool StackdriverRootContext::onDone() {
   // called, but onConfigure is not triggered. onConfigure is only triggered in
   // thread local VM, which makes it possible that logger_ is empty ptr even
   // when logging is enabled.
-  if (logger_ && enableServerAccessLog() &&
+  if (logger_ && (enableServerAccessLog()) &&
```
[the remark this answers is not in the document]
gargnupur: Thanks..Fixed

extensions/stackdriver/stackdriver.h
gargnupur: yup. thanks.. fixed

extensions/stackdriver/stackdriver.h
gargnupur: Yes, will do in a separate PR.. started a conversation about this with you, Mandar and Doug..

extensions/stackdriver/stackdriver.h
bianpengyuan: I don't see a new commit, did you push?

gargnupur: weird.. change was there in the branch bot not in PR.. I pushed again with a comment.. PTAL

PR #2853, opened 2020-05-15
Clean up test framework

bianpengyuan:
Efforts for https://github.com/istio/istio/issues/23168.

Remove unnecessary files in env package. Also remove SD related code out from driver package to stackdriver plugin test package.


test/envoye2e/env/envoy.go
```diff
@@ -1,168 +0,0 @@
...
-		path := "/dev/shm/envoy_shared_memory_" + s.baseID + "0"
-		if err := os.Remove(path); err != nil {
-			log.Printf("failed to %s\n", err)
-		} else {
-			log.Printf("removed Envoy's shared memory\n")
-		}
-	}
-}
-
-func copyYamlFiles(src, dst string) {
```
gargnupur, line 103: nit: are we loosing this functionality?

bianpengyuan: Yeah, I don't think we we use the `newEnvoy` below anymore. We could add a `dump` step to achieve the same thing.

gargnupur: Cool.. let's do that! approving this for now then...

PR #2854, opened 2020-05-18
fix(stackdriver): improve uids for GCE VMs

douglas-reid:
This PR corrects a number of small data issues observed on GCE VMs with Stackdriver telemetry (in particular, meshtelemetry service calls):

- prefer instance names to numbers in UIDs
- use GCE UIDs for workload instances (instead of `kubernetes://...` uids).

It also adds tests to validate traffic assertions for GCE.

### DIFF
--- a/extensions/stackdriver/edges/edge_reporter.cc
+++ b/extensions/stackdriver/edges/edge_reporter.cc
@@ -49,10 +49,16 @@ void instanceFromMetadata(const ::Wasm::Common::FlatNode& node_info,
   auto namespace_ = node_info.namespace_()
                         ? node_info.namespace_()->string_view()
                         : absl::string_view();
-  if (name.size() > 0 && namespace_.size() > 0) {
-    absl::StrAppend(instance->mutable_uid(), "kubernetes://", name, ".",
-                    namespace_);
+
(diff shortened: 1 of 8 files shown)


extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -49,10 +49,16 @@ void instanceFromMetadata(const ::Wasm::Common::FlatNode& node_info,
   auto namespace_ = node_info.namespace_()
                         ? node_info.namespace_()->string_view()
                         : absl::string_view();
-  if (name.size() > 0 && namespace_.size() > 0) {
-    absl::StrAppend(instance->mutable_uid(), "kubernetes://", name, ".",
-                    namespace_);
+
+  if (Common::isRawGCEInstance(node_info)) {
+    instance->set_uid(Common::getGCEInstanceUID(node_info));
+  } else {
```
bianpengyuan, line 55: You can do `else if () {..}` instead of `else { if () {..} }` here.

douglas-reid: yup. a holdover from more complex logic there. done.

douglas-reid:
Thanks for the reviews!  PTAL.

PR #2855, opened 2020-05-19
[1.6.1] Update envoy sha to pickup 2 crash fixes

mandarjog:
Updates proxy sha to pick up 2 crash fixes in envoy.


mandarjog:
@brian-avery ping ! we need this for 1.6.1

brian-avery:
Thanks @mandarjog I've added this to the list of issues we're tracking for 1.6.1. I will merge it after the 1.6 release.

PR #2857, opened 2020-05-19
ensure that telemetry plugins do not return config errors

mandarjog:
fixes https://github.com/istio/istio/issues/23865

### DIFF
--- a/extensions/access_log_policy/plugin.h
+++ b/extensions/access_log_policy/plugin.h
@@ -82,7 +82,7 @@ class PluginRootContext : public RootContext {
   int32_t max_client_cache_size_ = DefaultClientCacheMaxSize;
   long long log_time_duration_nanos_;
 
-  bool initialized_;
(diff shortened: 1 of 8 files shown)


extensions/access_log_policy/plugin.h
```diff
@@ -72,13 +73,16 @@ class PluginRootContext : public RootContext {
...
+  bool initialized() const { return initialized_; };
 
  private:
   accesslogpolicy::config::v1alpha1::AccessLogPolicyConfig config_;
   // Cache storing last log time by a client.
   absl::flat_hash_map<IstioDimensions, long long> cache_;
   int32_t max_client_cache_size_ = DefaultClientCacheMaxSize;
   long long log_time_duration_nanos_;
+
+  bool initialized_;
```
gargnupur, line 85: nit: would be good to initialize this to empty {} or false...

PR #2858, opened 2020-05-19
MX should continue with sane defaults

mandarjog:
Update MX to continue with sane defaults.
1. A config error is logged if config parsing fails.
2. The filter will still be active, since the only config param is cache size, and the default value is already sane.
3. Update log statements to use macro to be consistent with other plugins and correct line numbers.

fixes:  istio/istio#23865

### DIFF
--- a/extensions/metadata_exchange/plugin.cc
+++ b/extensions/metadata_exchange/plugin.cc
@@ -100,7 +100,7 @@ bool PluginRootContext::onConfigure(size_t size) {
                          " value:", metadata_value_, " node:", node_id_));
 
   // Parse configuration JSON string.
-  if (!configure(size)) {
+  if (size > 0 && !configure(size)) {
     LOG_WARN("configuration has errrors, but initialzation can continue.");
   }
 
--- a/extensions/metadata_exchange/plugin.cc
+++ b/extensions/metadata_exchange/plugin.cc
@@ -55,7 +55,7 @@ bool serializeToStringDeterministic(const google::protobuf::Message& metadata,
 
   mcs.SetSerializationDeterministic(true);
   if (!metadata.SerializeToCodedStream(&mcs)) {
-    logWarn("unable to serialize metadata");
+    LOG_WARN("unable to serialize metadata");
     return false;
   }
   return true;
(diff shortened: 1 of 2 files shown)


mandarjog:
/test test-tsan_proxy
flake?

@gargnupur  ?
```
2020/05/19 23:35:05 stopping XDS server
--- FAIL: TestTCPMetadataExchange (34.80s)
    tcp_metadata_exchange_test.go:79: failed to read bytes from conn read tcp 127.0.0.1:43162->127.0.0.1:20502: read: connection reset by peer
2020/05/19 23:35:05 XDS server starting on 20525
2020/05/19 23:35:05 update config for "client" with version "0"
name: outbound|9080|tcp|server.default.svc.cluster.local
connect_timeout: 1s
skipped 817 lines unfold_more
[2020-05-19 23:35:37.604][61354][warning][config] [bazel-out/k8-fastbuild/bin/external/envoy/source/common/config/_virtual_includes/grpc_stream_lib/common/config/grpc_stream.h:92] StreamAggregatedResources gRPC config stream closed: 13, 
[2020-05-19 23:35:37.635][61354][info][main] [external/envoy/source/server/server.cc:630] exiting
2020/05/19 23:35:37 stop envoy ... done
2020/05/19 23:35:37 removing temp config file
2020/05/19 23:35:37 stopping XDS server
FAIL
FAIL	istio.io/proxy/test/envoye2e/tcp_metadata_exchange	67.155s
FAIL
```

extensions/metadata_exchange/plugin.cc
```diff
@@ -88,34 +88,20 @@ void PluginRootContext::updateMetadataValue() {
...
-    logWarn(absl::StrCat("cannot parse plugin configuration JSON string: ",
-                         configuration->view(), j.dump()));
-    return false;
-  }
-
-  auto max_peer_cache_size =
-      ::Wasm::Common::JsonGetField<int64_t>(j, "max_peer_cache_size");
-  if (max_peer_cache_size.has_value()) {
-    max_peer_cache_size_ = max_peer_cache_size.value();
+  if (!configure(size)) {
```
bianpengyuan, line 103: We can make it `if (size > 0 && configure(size))`, so that this log won't even show when user does not specify anything.

bianpengyuan:
That test has been flaky before. Not seeing it recently though. It should not relate to your change.

extensions/metadata_exchange/plugin.cc
mandarjog: updated. PTAL.

mandarjog:
cherry-pick release-1.6

PR #2861, opened 2020-05-22
bazel_get_workspace_status: Allow users to override variables

jwendell:


bianpengyuan:
I am wondering why do you want to override this?

PiotrSikora:
@jwendell I'm not strongly opposing to this, but what's the use case?

jwendell:
Use case is to be able to build on non-git trees, and offer our own (downstream builders) notion of Status. Note that this kind of flexibility is already supported in istio/istio.

PR #2869, opened 2020-06-04
jwt_authn: change json processing strategy to protobuf

Shikugawa:
Change json processing strategy to protobuf from envoy native json loader.
In WebAssembly context, emscripten don't support pthread on standalone wasm mode, so we should eliminate what depend on them. Envoy (and the dependencies of them) have that.

### DIFF
--- a/src/envoy/http/jwt_auth/BUILD
+++ b/src/envoy/http/jwt_auth/BUILD
@@ -33,8 +33,8 @@ envoy_cc_library(
     ],
     repository = "@envoy",
     deps = [
-        "@envoy//source/exe:envoy_common_lib",
         "@com_google_protobuf//:protobuf",
+        "@envoy//source/exe:envoy_common_lib",
     ],
 )
 
--- a/src/envoy/http/jwt_auth/BUILD
+++ b/src/envoy/http/jwt_auth/BUILD
@@ -29,12 +29,13 @@ envoy_cc_library(
     srcs = ["jwt.cc"],
     hdrs = ["jwt.h"],
     external_deps = [
-        "rapidjson",
+        # "rapidjson",
         "ssl",
     ],
     repository = "@envoy",
     deps = [
         "@envoy//source/exe:envoy_common_lib",
+        "@com_google_protobuf//:protobuf",
     ],
 )
 
@@ -110,7 +111,7 @@ envoy_cc_test(
     repository = "@envoy",
     deps = [
         ":jwt_lib",
-        "@envoy//source/exe:envoy_common_lib",
+        "@com_google_protobuf//:protobuf",
         "@envoy//test/test_common:utility_lib",
     ],
 )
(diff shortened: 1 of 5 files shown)


Shikugawa:
@bianpengyuan ok. I will implement that in another PR.

kyessenov:
This is not helping with fixing system dependencies since we use patched protobuf anyways. We could just have patched rapidjson to remove extra sys calls.

Shikugawa:
Refactored inn #2871.

lizan:
@bianpengyuan @kyessenov how are those JSON libraries are battle tested against untrusted inputs? Do we have fuzzers run against it? In the authn filter the JSON library will be parsing untrusted input from request, so this path is more critical than config parsing path.

More context: JSON parsing library has been source of security vulnerabilities in the past and we decided to remove RapidJSON from Envoy (https://github.com/envoyproxy/envoy/issues/4705) as much as possible. So please don't use RapidJSON for authn.

kyessenov:
nlohmann json is approved for use within google which is the reason we chose it.

lizan:
@kyessenov Is it approved in security critical path? IIRC rapidjson is there too, cc @htuch

kyessenov:
@lizan S&P recommends using either json.hpp or proto2. Tensorflow uses json.hpp.

htuch:
Yeah, I don't think we're comfortable with RapidJSON on the data plane, we've seen some overflows/timeouts with this dependency and would like to minimize our data plane dependency footprint. Can JSON parsing be something that is provided by the Wasm API? That way we can do this Envoy-side and provide back a structured object. I think this is how the plugin config works today.

kyessenov:
@htuch We discussed that with @jplevyak to make JSON parsing be part of ABI. I don't remember the details but I see two problems with that: 1) passing data to/from ABI is expensive, 2) what form would you return parsed JSON to Wasm code that works across languages and doesn't force protobuf?

htuch:
Good point on language independent representation; I assume C structs via FFIs are a non-starter?

Maybe we need binary JSON, http://bsonspec.org/spec.html :D Not seriously suggesting that (I haven't done due diligence).

kyessenov:
I'd pick flatbuffers then :P On a serious note, we have a JSON config with one field. The effort to parse it should be minimal. Using protobuf automatically adds 1MB to the binary (it's too bloated), and using some kind of ABI dance to read a single field is ominous.

htuch:
Sure, I think a lightweight JSON library makes sense, we just need to pick one and have it be as secure/simple/lightweight as possible. We will have a more general ongoing issue with Wasm where we need to link into extensions large external libs, but that's outside the scope of this work.

kyessenov:
OK, I vote for nlohmann/json since it's tiny, recommended by Google S&P, and widely used. It's what in this PR. For large external libs, we use FFI in a few places.

lizan:
Sure if that's battle-tested I'm fine with json.hpp for a lightweight option. Thanks for clarification.

(In note of protobuf binary size we should try upb which grpc uses heavily, but a different topic)

PR #2870, opened 2020-06-08
fix(stackdriver): map grpc status codes into http domain for metrics

douglas-reid:
Adds logic for stackdriver metrics reporting to handle gRPC status codes by translating to HTTP response codes. This should allow monitoring of gRPC services to work seamlessly in HTTP-focused dashboards and alerting.

### DIFF
--- a/extensions/stackdriver/metric/record.cc
+++ b/extensions/stackdriver/metric/record.cc
@@ -169,6 +169,49 @@ TagKeyValueList getInboundTagMap(
   return inboundMap;
 }
 
+// See:
+// https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+uint32_t httpCodeFromGrpc(uint32_t grpc_status) {
+  switch (grpc_status) {
+    case 0:
+      return 200;
+    case 1:
+      return 499;
+    case 2:
+      return 500;
+    case 3:
+      return 400;
+    case 4:
+      return 504;
+    case 5:
+      return 404;
+    case 6:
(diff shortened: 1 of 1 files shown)


extensions/stackdriver/metric/record.cc
```diff
@@ -169,15 +169,63 @@ TagKeyValueList getInboundTagMap(
   return inboundMap;
 }
 
+// See:
+// https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+uint32_t httpCodeFromGrpc(uint32_t grpc_status) {
+  switch (grpc_status) {
+    case 0:
```
gargnupur, line 176: Can we add the status code names here?

douglas-reid: do you mean in the code or in additional comments?

all we have is the uint32_t (in RequestInfo) unfortunately. we could convert and then convert again, but that seems a bit wasteful.

gargnupur: sorry meant just in comments....

douglas-reid: Done.

douglas-reid:
@gargnupur PTAL. Updated to add comments as you suggested.

douglas-reid:
@mandarjog do we need to cherry-pick this back to 1.6 ?

mandarjog:
It is a bug fix of sorts, we should put it back in 1.6.

PR #2871, opened 2020-06-09
jwt_authn: replace json processing strategy

Shikugawa:
Replace json processing strategy from envoy native json loader for cleanup.

### DIFF
--- a/extensions/common/json_util.cc
+++ b/extensions/common/json_util.cc
@@ -22,7 +22,7 @@ namespace Common {
 
 absl::optional<JsonObject> JsonParse(absl::string_view str) {
   const auto result = JsonObject::parse(str, nullptr, false);
-  if (result.empty() || result.is_discarded() || !result.is_object()) {
+  if (result.is_discarded() || !result.is_object()) {
     return absl::nullopt;
   }
   return result;
--- a/extensions/common/json_util.cc
+++ b/extensions/common/json_util.cc
@@ -85,7 +85,8 @@ template <>
 std::pair<absl::optional<std::string>, JsonParserResultDetail>
 JsonValueAs<std::string>(const JsonObject& j) {
   if (j.is_string()) {
-    return std::make_pair(j.get<std::string>(), JsonParserResultDetail::OK);
+    return std::make_pair(j.get_ref<std::string const&>(),
+                          JsonParserResultDetail::OK);
   }
   return std::make_pair(absl::nullopt, JsonParserResultDetail::TYPE_ERROR);
 }
--- a/extensions/common/json_util.cc
+++ b/extensions/common/json_util.cc
@@ -20,21 +20,12 @@
 namespace Wasm {
 namespace Common {
 
-void JsonParser::parse(absl::string_view str) {
-  reset();
+std::pair<JsonObject, JsonParserResultDetail> JsonParse(absl::string_view str) {
   const auto result = JsonObject::parse(str, nullptr, false);
   if (result.empty() || result.is_discarded()) {
-    detail_ = JsonParserResultDetail::PARSE_ERROR;
-    return;
+    return std::make_pair(result, JsonParserResultDetail::PARSE_ERROR);
   }
-  detail_ = JsonParserResultDetail::OK;
-  object_ = result;
-  return;
-}
-
-void JsonParser::reset() {
-  object_ = JsonObject{};
-  detail_ = JsonParserResultDetail::EMPTY;
+  return std::make_pair(result, JsonParserResultDetail::OK);
 }
 
 template <>
--- a/extensions/common/json_util.cc
+++ b/extensions/common/json_util.cc
@@ -20,89 +20,87 @@
 namespace Wasm {
 namespace Common {
 
-JsonObject JsonParse(absl::string_view str, const bool allow_exceptions) {
-  return JsonObject::parse(str, nullptr, allow_exceptions);
+absl::optional<JsonObject> JsonParse(absl::string_view str) {
+  const auto result = JsonObject::parse(str, nullptr, false);
+  if (result.empty() || result.is_discarded()) {
+    return absl::nullopt;
+  }
+  return result;
 }
 
 template <>
-absl::optional<int64_t> JsonValueAs<int64_t>(const JsonObject& j,
-                                             const bool allow_exception) {
+std::pair<absl::optional<int64_t>, absl::optional<JsonParseError>>
+JsonValueAs<int64_t>(const JsonObject& j) {
   if (j.is_number()) {
-    return j.get<int64_t>();
+    return std::make_pair(j.get<int64_t>(), absl::nullopt);
   } else if (j.is_string()) {
     int64_t result = 0;
     if (absl::SimpleAtoi(j.get_ref<std::string const&>(), &result)) {
-      return result;
+      return std::make_pair(result, absl::nullopt);
     }
   }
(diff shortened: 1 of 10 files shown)


extensions/common/json_util.cc
```diff
@@ -20,12 +20,13 @@
 namespace Wasm {
 namespace Common {
 
-::nlohmann::json JsonParse(absl::string_view str) {
-  return ::nlohmann::json::parse(str, nullptr, false);
+::nlohmann::json JsonParse(absl::string_view str, const bool allow_exceptions) {
```
Shikugawa, line 23: I think that we should move this utility to a common place. cc. @kyessenov

kyessenov: This is a common place for istio extensions. Everything under extensions is self-contained.

extensions/common/json_util.cc
```diff
@@ -20,12 +20,13 @@
...
+  return ::nlohmann::json::parse(str, nullptr, allow_exceptions);
```
kyessenov, line 24: Exceptions are not supported in Wasm. I think we should always disallow that to avoid confusion.

extensions/common/json_util.h
```diff
@@ -15,6 +15,9 @@
 
 #pragma once
 
+#include <string>
+#include <utility>
```
kyessenov, line 19: Curious, what is this for? Does it substantially increase generated wasm file size?

Shikugawa: It was not needed, destroyed.

Shikugawa:
@kyessenov ping

Shikugawa:
> I think it is great to get error status in addition, but I don't quite get the reason of wrapping the json util into object. It makes code verbose without bringing much value. Would you mind to explain a bit on this?

@bianpengyuan These object wrapper is introduced to remove return value complexity. If not, we should return pair which includes parsed result and status. But, we don't need parsed status in many case. So I think that wrapping json parser makes simpler.

extensions/common/json_util.cc
```diff
@@ -20,60 +20,99 @@
...
   }
   if (j.is_string()) {
     const std::string& v = j.get_ref<std::string const&>();
     if (v == "true") {
-      return true;
+      return std::make_pair(true, JsonParserResultDetail::OK);
     } else if (v == "false") {
-      return false;
+      return std::make_pair(false, JsonParserResultDetail::OK);
+    } else {
```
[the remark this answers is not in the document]
Shikugawa: I think that It is not required right now but will be required in the future :)

bianpengyuan:
@Shikugawa Ok, I think it is good to keep JsonGetField, but not sure about JsonParser. It really just returns binary status (parsing success or fail). Optional return could distinguish them well.

Shikugawa:
@bianpengyuan ping

src/envoy/http/jwt_auth/jwt.cc
```diff
@@ -254,24 +253,27 @@ Jwt::Jwt(const std::string &jwt) {
...
   header_str_ = Base64UrlDecode(header_str_base64url_);
-  try {
-    header_ = Json::Factory::loadFromString(header_str_);
-  } catch (Json::Exception &e) {
+
+  auto result = Wasm::Common::JsonParse(header_str_);
+  if (result.second != Wasm::Common::JsonParserResultDetail::OK) {
     UpdateStatus(Status::JWT_HEADER_PARSE_ERROR);
     return;
+  } else {
```
bianpengyuan, line 261: Nit: no need to else here.

src/envoy/http/jwt_auth/jwt.cc
```diff
@@ -473,29 +490,34 @@ void Pubkeys::CreateFromPemCore(const std::string &pkey_pem) {
...
-  Json::ObjectSharedPtr jwks_json;
-  try {
-    jwks_json = Json::Factory::loadFromString(pkey_jwks);
-  } catch (Json::Exception &e) {
+  Wasm::Common::JsonObject jwks_json;
+  auto result = Wasm::Common::JsonParse(pkey_jwks);
+  if (result.second != Wasm::Common::JsonParserResultDetail::OK) {
     UpdateStatus(Status::JWK_PARSE_ERROR);
     return;
+  } else {
```
bianpengyuan, line 498: ditto

extensions/common/json_util.cc
```diff
@@ -20,60 +20,91 @@
 namespace Wasm {
 namespace Common {
 
-::nlohmann::json JsonParse(absl::string_view str) {
-  return ::nlohmann::json::parse(str, nullptr, false);
+std::pair<JsonObject, JsonParserResultDetail> JsonParse(absl::string_view str) {
```
bianpengyuan, line 23: Can we return just the JsonObject as the original logic instead of a pair? I don't see the needs of returning detail here. `is_object` on returned object should already represent that.

extensions/common/json_util.h
```diff
@@ -25,46 +25,88 @@
...
+template <>
+std::pair<absl::optional<bool>, JsonParserResultDetail> JsonValueAs<bool>(
+    const JsonObject& j);
+
+template <class T>
+class JsonGetField {
+ public:
+  JsonGetField(const JsonObject& j, absl::string_view field);
+  const JsonParserResultDetail& detail() { return detail_; }
+  T fetch() { return object_; }
```
bianpengyuan, line 72: Can we stick with `value` and `value_or`?

extensions/common/json_util.cc
Shikugawa: I don't agree with this. It seems that `is_object()` can't handle if that is empty or discarded. So we should use this function with basic `PARSE_ERROR` and `is_object()` to validate json strictly.

Shikugawa: But I agree that the usage of this function is pretty redundant. We should handle that is object or not with `JsonParserResultDetail`.

bianpengyuan: Ok, yeah, that's do that then.

src/envoy/http/jwt_auth/jwt.cc
Shikugawa: done

src/envoy/http/jwt_auth/jwt.cc
Shikugawa: done

extensions/common/json_util.h
Shikugawa: done

extensions/common/json_util.cc
```diff
@@ -20,60 +20,91 @@
 namespace Wasm {
 namespace Common {
 
-::nlohmann::json JsonParse(absl::string_view str) {
-  return ::nlohmann::json::parse(str, nullptr, false);
+absl::optional<JsonObject> JsonParse(absl::string_view str) {
+  const auto result = JsonObject::parse(str, nullptr, false);
+  if (result.empty() || result.is_discarded() || !result.is_object()) {
```
bianpengyuan, line 25: Is empty an invalid case?

Shikugawa: Sure. I think that we shouldn't treat empty as a valid value because we can't handle discarded results which are considered as parse error won't be processed when empty value.

Shikugawa: I think that there is no problem to treat empty values as parse failed, but It may cause confusion in the future. I changed this as not valid value.

PR #2876, opened 2020-06-19
authn: relpace json processing strategy

Shikugawa:
Replace json processing strategy from envoy native json loader for cleanup.

This PR depends on https://github.com/istio/proxy/pull/2871.


lizan:
@bianpengyuan can you take a look?

PR #2883, opened 2020-06-26
Update envoy wasm SHA

rshriram:


rshriram:
pipe dream. closing. code needs to be fixed..

PR #2892, opened 2020-07-01
[release-1.5] Update envoy sha rel15

jacob-delgado:

Update proxy sha based on Istio 1.5.7 security patch fix

**Release note**:

```release-note
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -38,7 +38,7 @@ bind(
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
 # istio/envoy commit date: Jun 02 2020
-ENVOY_PROJECT = "istio"
+ENVOY_ORG = "istio"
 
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -38,21 +38,21 @@ bind(
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
 # istio/envoy commit date: Jun 02 2020
```
shamsher31, line 40: nit: Is it require to update this date also if the SHA is updated?

Ref: Last PR https://github.com/istio/proxy/pull/2867/files#diff-09498dbadf45966909850dc8a47ebb13R40

jacob-delgado: It was an oversight, but given that it's just a comment and it already got merged it's fine. Making sure the correct sha is updated is more important. Didn't notice that.

PR #2862, opened 2020-05-22
Switch back envoy upstream to envoy-wasm

bianpengyuan:
Also use bazel to build Wasm modules. Fix https://github.com/istio/istio/issues/23215

PR #2893, opened 2020-07-02
AttributeGen Template fails test

vicentefb:
This PR adds a file called attributegen.yaml.tmpl and adds a unit test to stats_test.go. 

It doesn't fix any issue yet. It yields an error: 

The attributegen.yaml.tmpl was created to be able to test AttributeGen plugin through the function TestAttributeGen in stats_test.go. After running go test -v -run TestAttributeGen the test output yields the following message: 

Error adding/updating listener(s) server: Unable to parse JSON as proto (INVALID_ARGUMENT:(config.configuration): invalid value Missing. 

error code for :20002: 503

**Release note**:

```release-note
```


testdata/filters/attributegen.yaml.tmpl
```diff
@@ -0,0 +1,29 @@
...
+    type_url: type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm
+    value:
+      config:
+        vm_config:
+          runtime: {{ .Vars.WasmRuntime }}
+          vm_id: stats_outbound{{ .N }}
+          code:
+            local: { inline_string: "envoy.wasm.attributegen" }
+        configuration:
+          attributes:
```
bianpengyuan, line 13: You will need to add `@type` url to the configurations, like: https://github.com/istio/proxy/blob/9bbf80d5397880a8a1a4c3c8e4c36c2ff2575e30/testdata/filters/stats_inbound.yaml.tmpl#L14

This is something new and does not need in 1.6.

bianpengyuan: In case you did not notice it, you will also need to move the configurations under a value block as well:
https://github.com/istio/proxy/blob/9bbf80d5397880a8a1a4c3c8e4c36c2ff2575e30/testdata/filters/stats_inbound.yaml.tmpl#L15

Like this change: https://github.com/istio/proxy/pull/2862/files#diff-9aa8081ed5cd5469d42889f40a0ba32e

PR #2894, opened 2020-07-02
Add protocol label to access log only if this is tcp

bianpengyuan:
HTTP protocol is already specified in HTTPRequest message.

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -161,7 +161,6 @@ void Logger::addLogEntry(const ::Wasm::Common::RequestInfo& request_info,
   (*label_map)["service_authentication_policy"] =
       std::string(::Wasm::Common::AuthenticationPolicyString(
           request_info.service_auth_policy));
-  (*label_map)["protocol"] = request_info.request_protocol;
 
   if (is_tcp) {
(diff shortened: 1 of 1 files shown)


extensions/stackdriver/log/logger.cc
```diff
@@ -230,6 +229,7 @@ void Logger::addTCPLabelsToLogEntry(
   (*label_map)["connection_state"] =
       std::string(::Wasm::Common::TCPConnectionStateString(
           request_info.tcp_connection_state));
+  (*label_map)["protocol"] = request_info.request_protocol;
```
mandarjog, line 232: Doesn’t this make an sd query non uniform?

Entry.protocol == tcp and entry.protocol == http is uniform.

If we feel that presence of httpReuestMessage is enough. We can remove protocol completely.

bianpengyuan: Yeah I thought about this as well, and I figured that tcp log name and http log name are different, so we already have enough signals to distinguish them in the query, and we can avoid this duplicated protocol field in http log entry.

But turns out they are using the same log name, so yeah maybe we should just keep this duplication for simplicity.

PR #2898, opened 2020-07-07
Change not to impose a case sensitivity requirement on the word "Bearer"

taisho6339:
- Change not to impose  a case sensitivity requirement on the word "Bearer" in the Authorization header.

**Which issue this PR fixes** : fixes https://github.com/istio/istio/issues/22608

**Release note**:

```release-note
```

### DIFF
--- a/src/envoy/http/jwt_auth/token_extractor.cc
+++ b/src/envoy/http/jwt_auth/token_extractor.cc
@@ -28,7 +28,7 @@ namespace JwtAuth {
 namespace {
 
 // The autorization bearer prefix.
-const std::string kBearerPrefix = "Bearer ";
+const std::string kBearerPrefix = "bearer ";
 
 // The query parameter name to get JWT token.
 const std::string kParamAccessToken = "access_token";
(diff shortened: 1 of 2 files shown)


taisho6339:
OK, I'll fix it to Envoy!
Thanks!

thorhs:
FYI for others looking for a solution, the Bearer keyword is case sensitive according to RFC-6750.

From the RFC:

> [1.1](https://www.rfc-editor.org/rfc/rfc6750.html#section-1.1).  Notational Conventions
> 
>    The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
>    "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
>    document are to be interpreted as described in "Key words for use in
>    RFCs to Indicate Requirement Levels" [[RFC2119](https://www.rfc-editor.org/rfc/rfc2119)].
> 
>    This document uses the Augmented Backus-Naur Form (ABNF) notation of
>    [[RFC5234](https://www.rfc-editor.org/rfc/rfc5234)].  Additionally, the following rules are included from
>    HTTP/1.1 [[RFC2617](https://www.rfc-editor.org/rfc/rfc2617)]: auth-param and auth-scheme; and from "Uniform
>    Resource Identifier (URI): Generic Syntax" [[RFC3986](https://www.rfc-editor.org/rfc/rfc3986)]: URI-reference.
> 
>    Unless otherwise noted, all the protocol parameter names and values
>    are case sensitive.

Please note the bottom paragraph, and:

> [2.1](https://www.rfc-editor.org/rfc/rfc6750.html#section-2.1).  Authorization Request Header Field
> 
>    When sending the access token in the "Authorization" request header
>    field defined by HTTP/1.1 [[RFC2617](https://www.rfc-editor.org/rfc/rfc2617)], the client uses the "Bearer"
>    authentication scheme to transmit the access token.

There is no metion of the bearer keyword being case insensitive that I can find which means it is case sensitive.

PR #2900, opened 2020-07-08
Update Envoy to pull in https://github.com/envoyproxy/envoy/pull/11921

duderino:


duderino:
All tests passed other than the optional macos one which looks like it's been failing for a while.  @fpesce @ericvn and @dgn, sorry to self merge but I want to kick off the 2.5-3 hour post submit build so it's ready for us in the morning

PR #2901, opened 2020-07-09
Bind lifetime of http metadata to DownstreamRequest

mandarjog:
Fixes https://github.com/istio/istio/issues/25274

1. Change lifetime of metadata to DownstreamRequest.
2. Do not read metadata if peer_id not found.

### DIFF
--- a/extensions/metadata_exchange/plugin.cc
+++ b/extensions/metadata_exchange/plugin.cc
@@ -105,7 +105,8 @@ bool PluginRootContext::onConfigure(size_t size) {
   const std::string function = "declare_property";
   envoy::source::extensions::common::wasm::DeclarePropertyArguments args;
   args.set_type(envoy::source::extensions::common::wasm::WasmType::FlatBuffers);
-  args.set_span(envoy::source::extensions::common::wasm::LifeSpan::FilterChain);
+  args.set_span(
+      envoy::source::extensions::common::wasm::LifeSpan::DownstreamRequest);
   args.set_schema(::Wasm::Common::nodeInfoSchema().data(),
                   ::Wasm::Common::nodeInfoSchema().size());
   std::string in;
@@ -113,8 +114,7 @@ bool PluginRootContext::onConfigure(size_t size) {
   args.SerializeToString(&in);
   proxy_call_foreign_function(function.data(), function.size(), in.data(),
                               in.size(), nullptr, nullptr);
-  args.set_span(
-      envoy::source::extensions::common::wasm::LifeSpan::DownstreamConnection);
+
   args.set_name(std::string(::Wasm::Common::kDownstreamMetadataKey));
   args.SerializeToString(&in);
   proxy_call_foreign_function(function.data(), function.size(), in.data(),
--- a/extensions/metadata_exchange/plugin.cc
+++ b/extensions/metadata_exchange/plugin.cc
@@ -105,8 +105,7 @@ bool PluginRootContext::onConfigure(size_t size) {
   const std::string function = "declare_property";
   envoy::source::extensions::common::wasm::DeclarePropertyArguments args;
   args.set_type(envoy::source::extensions::common::wasm::WasmType::FlatBuffers);
-  args.set_span(
-      envoy::source::extensions::common::wasm::LifeSpan::DownstreamConnection);
+  args.set_span(envoy::source::extensions::common::wasm::LifeSpan::FilterChain);
   args.set_schema(::Wasm::Common::nodeInfoSchema().data(),
                   ::Wasm::Common::nodeInfoSchema().size());
   std::string in;
@@ -114,6 +113,8 @@ bool PluginRootContext::onConfigure(size_t size) {
   args.SerializeToString(&in);
   proxy_call_foreign_function(function.data(), function.size(), in.data(),
                               in.size(), nullptr, nullptr);
+  args.set_span(
+      envoy::source::extensions::common::wasm::LifeSpan::DownstreamConnection);
   args.set_name(std::string(::Wasm::Common::kDownstreamMetadataKey));
   args.SerializeToString(&in);
   proxy_call_foreign_function(function.data(), function.size(), in.data(),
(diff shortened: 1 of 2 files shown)


extensions/metadata_exchange/plugin.cc
```diff
@@ -105,15 +105,16 @@ bool PluginRootContext::onConfigure(size_t size) {
...
-      envoy::source::extensions::common::wasm::LifeSpan::DownstreamConnection);
+  args.set_span(envoy::source::extensions::common::wasm::LifeSpan::FilterChain);
   args.set_schema(::Wasm::Common::nodeInfoSchema().data(),
                   ::Wasm::Common::nodeInfoSchema().size());
   std::string in;
   args.set_name(std::string(::Wasm::Common::kUpstreamMetadataKey));
   args.SerializeToString(&in);
   proxy_call_foreign_function(function.data(), function.size(), in.data(),
                               in.size(), nullptr, nullptr);
+  args.set_span(
```
nrjpoddar, line 116: Did you mean to have both bindings or just Filter Chain binding at the top?

mandarjog: Just the upstream key to filterchain.

extensions/metadata_exchange/plugin.cc
```diff
@@ -105,15 +105,16 @@ bool PluginRootContext::onConfigure(size_t size) {
   const std::string function = "declare_property";
   envoy::source::extensions::common::wasm::DeclarePropertyArguments args;
   args.set_type(envoy::source::extensions::common::wasm::WasmType::FlatBuffers);
-  args.set_span(
-      envoy::source::extensions::common::wasm::LifeSpan::DownstreamConnection);
+  args.set_span(envoy::source::extensions::common::wasm::LifeSpan::FilterChain);
```
douglas-reid, line 108: should we have a test to verify the appropriate behavior?  This feels like something we should have validation on.

mandarjog: Absolutely, and also discussion about lifetimes.
@PiotrSikora @jplevyak @lambdai can you take a look at this issue?

bianpengyuan:
Is tcp metadata exchange also going to be problematic because of downstream connection lifespan? https://github.com/istio/proxy/blob/3aaff15a85a5c0cc5c23635d4b858eee001df836/src/envoy/tcp/metadata_exchange/metadata_exchange.cc#L302

kyessenov:
Yes, probably. The issue is having two upstreams, one with sidecar and one without, that share the same downstream connection. This can happen with sidecars, not just ingress (e.g. a pod opening a connection to service A and sending two requests /a and /b where /b is redirected to service B which has no sidecar).

kyessenov:
Alternative solution would be to keep the lifespan, but make MX to write a token value when metadata is not received. Not sure if that works with overlapping requests (multiple h2 streams on the same connection).

mandarjog:
I think we need to rationalize and fix lifespan. TCP mx writes tokens, but for a different reason.

With and without sidecar, this issue can happen, it just very visible in this case.
`setFilterState()` call does not suffer from the same lifetime issue, @kyessenov  why is that?

kyessenov:
MX uses setFilterState (that's the only way to set filter state). The difference is in the declaration since filter state has a schema.

mandarjog:
What is the default lifetime if schema is not specified?

kyessenov:
It's [filter chain](https://github.com/envoyproxy/envoy-wasm/blob/master/source/extensions/common/wasm/wasm_state.h#L44).

extensions/metadata_exchange/plugin.cc
lambdai: Damn, my hunch is that FilterChain has longer life than connection until I read the document. It seems FilterChain here means the http filters in HCM.

mandarjog:
@kyessenov @lambdai 
Looks like filterchain for upstream and connection for downstream is correct. 

```
// - FilterChain has the shortest life span, which is as long as the filter chain lives.
--
  | //
  | // - Request is longer than FilterChain. When internal redirect is enabled, one
  | //   downstream request may create multiple filter chains. Request allows an object to
  | //   survive across filter chains for bookkeeping needs. This is not used for the upstream case.
  | //
  | // - Connection makes an object survive the entire duration of a connection.
  | //   Any stream within this connection can see the same object.
  | //
```

kyessenov:
I disagree, I think it should be `DownstreamRequest`.
For downstream, internal directs should not be flushing metadata since headers are stripped on the first iteration of MX filter.
For upstream, a client reusing connections to send requests to two destinations should not cause aliasing.

extensions/stats/plugin.cc
```diff
@@ -576,9 +576,7 @@ bool PluginRootContext::report(::Wasm::Common::RequestInfo& request_info,
     // been no error in connection.
     uint64_t response_flags = 0;
     getValue({"response", "flags"}, &response_flags);
-    if (peer_node == nullptr &&
-        peer_id != ::Wasm::Common::kMetadataNotFoundValue &&
```
kyessenov, line 580: @gargnupur I think this was added by you: can you take a look?

gargnupur: Speaking it out loud, there are 4 scenarios here:
1) Metadata Exchange has not happened yet -> so return false, so that you come back to report again -> peer_found == false && response_flags => if condition works
2) There is an error(response_flags!=0),  so don't worry about MX and just log -> response_flags != 0 => if condition works as it will go ahead and log
3) Metadata Exchange was attempted but peer not found -> peer_id = kMetadataNotFoundValue -> peer_found = true => if condition works
4) MX has happened, peer is there, no error and thus just log -> peer_found = true -> if condition works

Thus, this will work...

kyessenov:
/test test-asan_proxy
2h timeout is a bit optimistic...

douglas-reid:
Did we ever add the tests we said were needed for this behavior?

kyessenov:
Sure, I can do it in a follow-up. Just wanted to make it the cut since tests can go on master.

PR #2902, opened 2020-07-10
Fix container name as app_container in telemetry v2

gargnupur:

fix container name as app_container in telemetry v2. APP_CONTAINER is an environment variable and thus it come in as a string(https://sourcegraph.com/github.com/kubernetes/kubernetes@4efed0327654afd66346b58ae33f1f6432a274e2/-/blob/staging/src/k8s.io/api/core/v1/types.go#L1837:1). Because of which fbb can't parse it as a list value..

**Release note**:

```release-note
```


extensions/common/proto_util_test.cc
```diff
@@ -46,7 +46,7 @@ constexpr absl::string_view node_metadata_json = R"###(
    "WORKLOAD_NAME":"test_workload",
    "OWNER":"test_owner",
    "NAME":"test_pod",
-   "APP_CONTAINERS": [ "test", "hello" ]
+   "APP_CONTAINERS": "test,hello"
```
kyessenov, line 49: Is this a backwards compatible change? Can you link istio PR that did this?

gargnupur: it should be because it was previously coming in as something like "[\n  test \n, \n hello \n ]".. so it was still a string? 
proxy change:
https://github.com/istio/proxy/pull/2812/
istio change:
https://github.com/istio/istio/pull/22721

douglas-reid: @kyessenov i have some concerns as well, but it looks like this bit of functionality is (no longer?) working as-is, despite the testing in both repos. we need a path forward here, and I think making the change in proxy and then in istio is the right choice.  

i don't know how to back port this gracefully though.

PR #2904, opened 2020-07-11
Remove Istio JWT filter

bianpengyuan:
Clean up Istio Jwt filter. It is not used anymore and pilot is no longer configuring it. This has been replaced by upstream envoy jwt filter.

**Release note**:

```release-note
```


bianpengyuan:
@howardjohn fyi, I am hoping to merge this ibefore [updating envoy SHA](https://github.com/istio/proxy/compare/master...bianpengyuan:update-sha?expand=1) if this could be be turned around quickly, so that we don't pay a toll every time updating envoy sha to make this deprecated code build.

nrjpoddar:
@bianpengyuan I have the same question as Yuechen, what's replacing this implemntation?

yangminzhu:
> @bianpengyuan I have the same question as Yuechen, what's replacing this implemntation?

We have changed to use the Envoy JWT filter by default since 1.3: https://istio.io/latest/news/releases/1.3.x/announcing-1.3/change-notes/#security

bianpengyuan:
@yangminzhu I tried to remove that constant actually, but stuck on making this test pass: https://github.com/istio/proxy/blob/a561c592c858da760649561fea0f6e6bb1ee1abe/src/envoy/http/authn/http_filter_integration_test.cc#L155. It uses a header to metadata filter to simulate istio jwt filter, but I could not find a way to inject protobuf struct using this header to metadata filter to simulate envoy jwt filter. I think this would need larger code change on the test setup to make that test pass, which I don't really want to take on. Would you mind if I add a todo there about removing that constant and assign it to you?

yangminzhu:
> @yangminzhu I tried to remove that constant actually, but stuck on making this test pass:
> 
> https://github.com/istio/proxy/blob/a561c592c858da760649561fea0f6e6bb1ee1abe/src/envoy/http/authn/http_filter_integration_test.cc#L155
> 
> . It uses a header to metadata filter to simulate istio jwt filter, but I could not find a way to inject protobuf struct using this header to metadata filter to simulate envoy jwt filter. I think this would need larger code change on the test setup to make that test pass, which I don't really want to take on. Would you mind if I add a todo there about removing that constant and assign it to you?

Yeah, I'm fine with a TODO. You probably don't need to use the header-to-metadata filter and can just use the Envoy JWT filter to generate the real metadata in the test.

bianpengyuan:
> You probably don't need to use the header-to-metadata filter and can just use the Envoy JWT filter to generate the real metadata in the test.

@yangminzhu yes, that is the change I was talking about :). Added a todo, thanks!

bianpengyuan:
Test failure is because of port conflict which should be resolved by https://github.com/istio/proxy/pull/2906

PR #2910, opened 2020-07-14
Fill in unknown for stackdriver metric empty labels

bianpengyuan:

**Release note**:

```release-note
```


extensions/stackdriver/metric/record.cc
```diff
@@ -93,38 +95,45 @@ TagKeyValueList getOutboundTagMap(
     const ::Wasm::Common::FlatNode& peer_node_info,
     const ::Wasm::Common::RequestInfo& request_info) {
   TagKeyValueList outboundMap = {
-      {meshUIDKey(), flatbuffers::GetString(local_node_info.mesh_id())},
-      {requestProtocolKey(), std::string(request_info.request_protocol)},
+      {meshUIDKey(),
+       unknownIfEmpty(flatbuffers::GetString(local_node_info.mesh_id()))},
+      {requestProtocolKey(),
+       unknownIfEmpty(std::string(request_info.request_protocol))},
```
gargnupur, line 101: for things we get directly from request_info.. shouldn't we set them to "unknown" if it's empty for all telemetry and not just for metrics?

bianpengyuan: It is going to make the code in context quite verbose. All wasm ABI/API set strings to empty if value is not found. If we want to set unknown in context, that would mean each getValue call would need to be wrapped by an `if (value.empty) then unknown` block. I think it is fine to make code in context follow wasm ABI/API convention and let each caller decide what to do with empty values.

gargnupur: Sounds good!

PR #2912, opened 2020-07-14
Add TCP Access Log on Open

gargnupur:

Add TCP Access Log on Open. This PR adds TCP access log onNewConnection too.
As TCP metadata exchange doesn't happen till OnData call happens, we wait for any access log entries to be logged till peer info is available until a timeout of 10 mins.

Adding more tests to capture no mx scenario too..

**Release note**:

```release-note
```


bianpengyuan:
I am a bit lost on this. This seems more complicated than it needs to be. I am thinking For TCP open event or metric, can we just add the connection to the queue with a start time, and whenever recordTCP is triggered, check the elapse between current time and connection open time, and decide whether we still want to wait for metadata? Seems like there are much more change in this PR, can you elaborate a bit more or add more comment on the context and reason?

gargnupur:
> I am a bit lost on this. This seems more complicated than it needs to be. I am thinking For TCP open event or metric, can we just add the connection to the queue with a start time, and whenever recordTCP is triggered, check the elapse between current time and connection open time, and decide whether we still want to wait for metadata? Seems like there are much more change in this PR, can you elaborate a bit more or add more comment on the context and reason?

@bianpengyuan : ptal now.. it's based on the comment above and what we discussed offline...

extensions/stackdriver/stackdriver.cc
```diff
@@ -472,6 +497,48 @@ StackdriverRootContext* StackdriverContext::getRootContext() {
...
+  if (enableServerAccessLog() && !record_info.tcp_open_log_entry_logged &&
+      ((cur - request_info.start_time) > tcp_log_entry_timeout_)) {
+    // Populate RequestInfo with local node information.
+    ::Wasm::Common::populateTCPRequestInfo(
+        outbound, record_info.request_info.get(), destination_namespace);
+    ::Wasm::Common::populateExtendedRequestInfo(record_info.request_info.get());
+    logTCPOpen(record_info, peer_node);
+    // Set is_populated to false, so that populateTCPRequestInfo can be called
+    // again when peer metadata is available by RecordTCP.
+    record_info.request_info->is_populated = false;
```
bianpengyuan, line 518: Once peer metadata arrives, the only difference will be peer metadata, right? All "request" info (we should really rename it to stream info, but let's do it in the following PRs) should stay the same. So I don't think you need to populate it again.

extensions/stackdriver/stackdriver.cc
```diff
@@ -394,6 +408,9 @@ bool StackdriverRootContext::recordTCP(uint32_t id) {
   getValue({"response", "flags"}, &response_flags);
   if (!peer_found && peer_id != ::Wasm::Common::kMetadataNotFoundValue &&
       response_flags == 0) {
+    logTCPOpenOnTimeout(
```
bianpengyuan, line 411: Do we need to add this function call here? Can we extend the condition above to：
```
if (!peer_found && peer_id != ::Wasm::Common::kMetadataNotFoundValue &&
      response_flags == 0 && cur_time - start_time <= timeout)
```
and if it times out, just running through the following logic to record metrics and add logs.

extensions/stackdriver/stackdriver.cc
```diff
@@ -472,6 +497,48 @@ StackdriverRootContext* StackdriverContext::getRootContext() {
...
+void StackdriverRootContext::logTCPOpen(
+    StackdriverRootContext::TcpRecordInfo& record_info,
+    const ::Wasm::Common::FlatNode& peer_node) {
+  // Log TCP Open event log, if it has not been logged before.
+  if (!record_info.tcp_open_log_entry_logged) {
+    auto state = record_info.request_info->tcp_connection_state;
+    // It's possible that for a short lived TCP connection, we log TCP
+    // Connection Open log entry on connection close. To make a distinction
+    // between open and close log entries, we change the state to Open here.
+    record_info.request_info->tcp_connection_state =
```
bianpengyuan, line 531: Does it make sense to always just set `tcp_connection_state` to Open in this function?

extensions/stackdriver/stackdriver.cc
```diff
@@ -472,6 +497,48 @@ StackdriverRootContext* StackdriverContext::getRootContext() {
...
+    // again when peer metadata is available by RecordTCP.
+    record_info.request_info->is_populated = false;
+  }
+}
+
+void StackdriverRootContext::logTCPOpen(
+    StackdriverRootContext::TcpRecordInfo& record_info,
+    const ::Wasm::Common::FlatNode& peer_node) {
+  // Log TCP Open event log, if it has not been logged before.
+  if (!record_info.tcp_open_log_entry_logged) {
```
bianpengyuan, line 526: Nit: early return here and move the logic out of the block:
```
if (record_info.tcp_open_log_entry_logged) {
  return;
}
```

extensions/stackdriver/stackdriver.cc
gargnupur: we don't plan to record metrics at that time.. just logs... hence having a separate function call here makes it easy to figure out the logic...

extensions/stackdriver/stackdriver.cc
gargnupur: Agreed for inbound case it will remain same...  removed it for now.. and will figure out a better way to refactor populateRequestInfo when we add client side log..

extensions/stackdriver/stackdriver.cc
gargnupur: That is so that we can capture `Connected` state too at timeout as this would tell if there was data sent on connection.. and not just `Open`...

extensions/stackdriver/stackdriver.cc
bianpengyuan: Should we keep the behavior consistent on logging and metric? I don't see downside on recording connection open on timeout here.

gargnupur: @bianpengyuan : thanks fixed.. recording metric on connection open on timeout too.. please review

extensions/stackdriver/stackdriver.cc
gargnupur: Removed now

extensions/stackdriver/stackdriver.cc
```diff
@@ -475,6 +518,17 @@ StackdriverRootContext* StackdriverContext::getRootContext() {
   return dynamic_cast<StackdriverRootContext*>(root);
 }
 
+void StackdriverRootContext::logTCPOpen(
+    StackdriverRootContext::TcpRecordInfo& record_info,
+    const ::Wasm::Common::FlatNode& peer_node) {
+  record_info.request_info->tcp_connection_state =
```
bianpengyuan, line 524: nit: Seems weird to me to make this a function. It handles a very specific case, instead of a general functionality. How about just remove this function and copy the function body into that if block at line 437?

extensions/stackdriver/stackdriver.h
```diff
@@ -111,6 +118,17 @@ class StackdriverRootContext : public RootContext {
   // Indicates whether or not to report TCP Logs.
   bool enableTCPServerAccessLog();
 
+  // Helper functions to log the tcp request on open. On NewConnection, we don't
+  // have peer metadata available, thus we wait for peer metadata to be
+  // available. If it's available within a timeout, it is logged  via logTCPOpen
+  // function else via logTCPOpenOnTimeout function.
+  void logTCPOpenOnTimeout(StackdriverRootContext::TcpRecordInfo& record_info,
```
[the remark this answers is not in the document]
gargnupur: Thanks..removed

extensions/stackdriver/stackdriver.cc
gargnupur: moved it for now.. but will refactor RecordTcp function in a different PR later.. feel it's become too big..

PR #2914, opened 2020-07-15
Pre-compile the .wasm files for V8.

jplevyak:

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -80,6 +80,10 @@ build_wasm:
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //extensions:stats.wasm
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //extensions:metadata_exchange.wasm
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //extensions:attributegen.wasm
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //src/envoy:envoy @envoy//test/tools/wee8_compile:wee8_compile_tool
+	bazel-bin/external/envoy/test/tools/wee8_compile/wee8_compile_tool bazel-bin/extensions/stats.wasm bazel-bin/extensions/stats.compiled.wasm
+	bazel-bin/external/envoy/test/tools/wee8_compile/wee8_compile_tool bazel-bin/extensions/metadata_exchange.wasm bazel-bin/extensions/metadata_exchange.compiled.wasm
+	bazel-bin/external/envoy/test/tools/wee8_compile/wee8_compile_tool bazel-bin/extensions/attributegen.wasm bazel-bin/extensions/attributegen.compiled.wasm
 
 # NOTE: build_wasm has to happen before build_envoy, since the integration test references bazel-bin symbol link for envoy binary,
 # which will be overwritten if wasm build happens after envoy.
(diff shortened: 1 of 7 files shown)


jplevyak:
I presume there will be packaging and config changes unless we want to just replace the existing files with the ones with the compiled v8 code.

Makefile.core.mk
```diff
@@ -80,6 +80,10 @@ build_wasm:
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //extensions:stats.wasm
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //extensions:metadata_exchange.wasm
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //extensions:attributegen.wasm
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //src/envoy:envoy @envoy//test/tools/wee8_compile:wee8_compile_tool
+	bazel-bin/external/envoy/test/tools/wee8_compile/wee8_compile_tool bazel-bin/extensions/stats.wasm bazel-bin/extensions/stats.compiled.wasm
+	bazel-bin/external/envoy/test/tools/wee8_compile/wee8_compile_tool bazel-bin/extensions/metadata_exchange.wasm bazel-bin/extensions/metadata_exchange.compiled.wasm
+	bazel-bin/external/envoy/test/tools/wee8_compile/wee8_compile_tool bazel-bin/extensions/attributegen.wasm bazel-bin/extensions/attributegen.compiled.wasm
```
mandarjog, line 86: Can vm load precompiled and non precompiled versions of files without any other config change?

Makefile.core.mk
```diff
@@ -80,6 +80,10 @@ build_wasm:
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //extensions:stats.wasm
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //extensions:metadata_exchange.wasm
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //extensions:attributegen.wasm
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) && bazel $(BAZEL_STARTUP_ARGS) build $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_REL) //src/envoy:envoy @envoy//test/tools/wee8_compile:wee8_compile_tool
```
mandarjog, line 83: we need to upload this to the gcs bucket like the other modules.

@bianpengyuan

bianpengyuan:
Do you also want to update the release-binary script to push these wasm files? https://github.com/istio/proxy/blob/8e608eafcee3a5115a5af1e1c1537833e8f52cb4/scripts/release-binary.sh#L190

Also could you extend the current test to also test the pre compiled version? https://github.com/istio/proxy/blob/8e608eafcee3a5115a5af1e1c1537833e8f52cb4/test/envoye2e/stats_plugin/stats_test.go#L56-L71

mandarjog:
Let's not replace right now, we can have that as a configuration option.

Makefile.core.mk
PiotrSikora, line 83: You don't need `//src/envoy:envoy` here, only `@envoy//test/tools/wee8_compile:wee8_compile_tool`.

Makefile.core.mk
PiotrSikora: Compiled version is embedded in a custom section of the `.wasm` module, so `.compiled.wasm` contains both.

When `allow_precompiled` is set to `true`, then compiled version will be loaded, otherwise only Wasm bytecode will be loaded, parsed and compiled.

jplevyak: Done.

Makefile.core.mk
jplevyak: Done

jplevyak: Done.

PR #2916, opened 2020-07-15
Add hooks for custom Istio Wasm extensions.

jplevyak:


jplevyak:
This is the companion to https://github.com/envoyproxy/envoy-wasm/pull/576

src/envoy/extensions/wasm/wasm.cc
```diff
@@ -0,0 +1,137 @@
+#include "extensions/common/wasm/wasm.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace Common {
+namespace Wasm {
+namespace Istio {
+
+class IstioContext : public Context {
```
bianpengyuan, line 9: Can you extract IstioContext to a separated `context.h` file? Extended ABI implemented in other files will need to include and reference this context.

jplevyak: Done.

PR #2918, opened 2020-07-16
Add filtered label for sampled logs

gargnupur:

Add filtered label for sampled logs

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -162,6 +162,7 @@ void Logger::addLogEntry(const ::Wasm::Common::RequestInfo& request_info,
       std::string(::Wasm::Common::AuthenticationPolicyString(
           request_info.service_auth_policy));
   (*label_map)["protocol"] = request_info.request_protocol;
+  (*label_map)["filtered"] = request_info.sampled ? "True" : "False";
 
   if (is_tcp) {
     addTCPLabelsToLogEntry(request_info, new_entry);
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -162,7 +162,7 @@ void Logger::addLogEntry(const ::Wasm::Common::RequestInfo& request_info,
       std::string(::Wasm::Common::AuthenticationPolicyString(
           request_info.service_auth_policy));
   (*label_map)["protocol"] = request_info.request_protocol;
-  (*label_map)["filtered"] = request_info.sampled ? "True" : "False";
+  (*label_map)["log_sampled"] = request_info.log_sampled ? "true" : "false";
 
   if (is_tcp) {
     addTCPLabelsToLogEntry(request_info, new_entry);
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -432,7 +432,8 @@ bool StackdriverRootContext::shouldLogThisRequest(
     LOG_DEBUG("cannot get envoy access log info from filter state.");
     return true;
   }
-  request_info.sampled = (shouldLog != "no");
+  // Add label log_sampled if Access Log Policy sampling was applied to logs.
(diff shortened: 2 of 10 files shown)


extensions/stackdriver/log/logger.cc
```diff
@@ -162,6 +162,7 @@ void Logger::addLogEntry(const ::Wasm::Common::RequestInfo& request_info,
       std::string(::Wasm::Common::AuthenticationPolicyString(
           request_info.service_auth_policy));
   (*label_map)["protocol"] = request_info.request_protocol;
+  (*label_map)["filtered"] = request_info.sampled ? "True" : "False";
```
bianpengyuan, line 165: nit: could you make it lower case? All other labels are lower cased except enum ones.

extensions/stackdriver/stackdriver.cc
```diff
@@ -350,7 +350,7 @@ void StackdriverRootContext::record() {
   ::Extensions::Stackdriver::Metric::record(
       outbound, local_node, peer_node, request_info,
       !config_.disable_http_size_metrics());
-  if (enableServerAccessLog() && shouldLogThisRequest()) {
+  if (enableServerAccessLog() && shouldLogThisRequest(request_info)) {
```
bianpengyuan, line 353: With this code, Isn't it that all entries exported will just have `filtered: true` label? So I am confused on the meaning of this label? Could you add some comment about this?

extensions/stackdriver/log/logger.cc
douglas-reid: should we be consistent with terminology here?  if the request is selected for logs-sampling, should this just be `sampled` ?

extensions/stackdriver/stackdriver.cc
gargnupur: Only the logs that were sampled will have `filtered: true` label

extensions/stackdriver/log/logger.cc
gargnupur: I myself was thinking the same... as PAY as you Go(https://cloud.google.com/pricing) team has not implemented "filtered" label yet.. changing it to sampled...

extensions/stackdriver/stackdriver.cc
bianpengyuan: Yeah, but every log entry reaches the backend has `filtered: true` label, doesn't it? If this label presents on all log entries with the same value, it essentially brings no value, no?

gargnupur: As talked offline... this will only mark the label "filtered":"true" changed to "log_sampled":"true" for whom kAccessLogPolicyKey is set. Which is set by Access Log Policy sampling filter when places in front of stackdriver.

extensions/stackdriver/stackdriver.cc
```diff
@@ -425,12 +425,15 @@ inline bool StackdriverRootContext::enableEdgeReporting() {
...
-bool StackdriverRootContext::shouldLogThisRequest() {
+bool StackdriverRootContext::shouldLogThisRequest(
+    ::Wasm::Common::RequestInfo& request_info) {
   std::string shouldLog = "";
   if (!getValue({::Wasm::Common::kAccessLogPolicyKey}, &shouldLog)) {
     LOG_DEBUG("cannot get envoy access log info from filter state.");
     return true;
   }
+  // Add label log_sampled if Access Log Policy sampling was applied to logs.
+  request_info.log_sampled = (shouldLog != "no");
```
bianpengyuan, line 436: super nit: could we just do string comparison for one time? Just return request_info.log_sampled in the next line.

bianpengyuan:
just one nit, please unblock once fixed it.

extensions/stackdriver/stackdriver.cc
gargnupur: thanks.. done

PR #2919, opened 2020-07-17
Add option to enable Log Compression for stackdriver access logs

gargnupur:
Add option to enable Log Compression for stackdriver access logs

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -25,8 +25,9 @@ syntax = "proto3";
 package stackdriver.config.v1alpha1;
 
 import "google/protobuf/duration.proto";
+import "google/protobuf/wrappers.proto";
 
-// next id: 9
+// next id: 10
 message PluginConfig {
   // Optional. Controls whether to export server access log.
   bool disable_server_access_logging = 1;
@@ -67,4 +68,7 @@ message PluginConfig {
   // metrics for HTTP traffic. Defaults to false (request and response size
   // metrics are enabled).
   bool disable_http_size_metrics = 8;
+
+  // Optional. Allows enabling log compression for stackdriver access logs.
+  google.protobuf.BoolValue enable_log_compression = 9;
 }
--- a/extensions/stackdriver/common/utils.h
+++ b/extensions/stackdriver/common/utils.h
@@ -42,6 +42,7 @@ struct StackdriverStubOption {
   std::string insecure_endpoint;
   std::string monitoring_endpoint;
   std::string project_id;
+  bool enable_log_compression;
 };
 
 // Build Envoy GrpcService proto based on the given stub option.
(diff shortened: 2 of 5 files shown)


gargnupur:
@bianpengyuan : I wanted to give easiest option for users to turn on this feature and change via istio/istio through node metadata seems easier than envoyfilter config change... I am open to both options.. just feel node metadata changes are easier for users to change in case of an error...

bianpengyuan:
Node metadata is not easy to change for user. It requires customized bootstrap like https://github.com/istio/istio/blob/master/tests/integration/telemetry/stackdriver/testdata/custom_bootstrap.yaml.tmpl. User needs to annotate their pod to point to that config map and patch node metadata.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -67,4 +67,7 @@ message PluginConfig {
   // metrics for HTTP traffic. Defaults to false (request and response size
   // metrics are enabled).
   bool disable_http_size_metrics = 8;
+
+  // Optional. Allows enabling log compression for stackdriver access logs.
+  bool enable_log_compression = 9;
```
bianpengyuan, line 72: Can you make it a [BoolValue](https://github.com/protocolbuffers/protobuf/blob/d4ec70fdd3380e7113e9e3c23cef2bea50dbf2e7/src/google/protobuf/wrappers.proto#L104), so that we can flip the default behavior of unset easily?

out/.env
```diff
@@ -0,0 +1,8 @@
+TARGET_OUT_LINUX=/usr/local/google/home/gargnupur/go/src/istio.io/proxy/out/linux_amd64
```
bianpengyuan, line 1: accident commit?

gargnupur: hehe yes.. working on too many pr's at the same time :(

mandarjog:
Nupur did you try it with cloud logging? Is it working?

gargnupur:
@mandarjog : I was getting this error when I tried e2e with just proxy built out of this:
Envoy command: [-c etc/istio/proxy/envoy-rev0.json --restart-epoch 0 --drain-time-s 45 --parent-shutdown-time-s 60 --service-cluster istio-ingressgateway --service-node router~10.4.2.67~istio-ingressgateway-686445f844-2br6g.istio-system~istio-system.svc.cluster.local --max-obj-name-len 189 --local-address-ip-version v4 --log-format-prefix-with-location 0 --log-format %Y-%m-%dT%T.%fZ	%l	envoy %n	%v -l warning --component-log-level misc:error]
PARSE ERROR: Argument: --max-obj-name-len
             Couldn't find match for argument

I don't think it's related to this pr... so will debug and try again before submitting the pr..

mandarjog:
: --max-obj-name-len was removed as an arg in envoy 1.15.

In a proxy sha update pr, it has been fixed so make sure you rebase from latest master

gargnupur:
@mandarjog , @bianpengyuan : Verified logs are flowing e2e without any issues :)

bianpengyuan:
Would be great to know how much does it help by looking at pod level network bytes in/out.

bianpengyuan:
Also does that mean we can just make it not configurable? What is the drawback of always enabling it?

gargnupur:
@bianpengyuan , @mandarjog : so here's the summary:
1) PR as is works fine.
2) When compression is enabled, we currently get the InValidArgument error with message "Error skipping unselected field of a message". I talked to person working on this, and they were same error which was just recently fixed and thus only available in very recent version of test proxy.

PR #2922, opened 2020-07-21
Add connection_id in stackdriver access log

gargnupur:

Add connection_id in stackdriver access log
Added for both http and tcp, so that for http we can get underlying connection's id too for debugging..

**Release note**:

```release-note
```


extensions/common/context.cc
```diff
@@ -170,6 +170,8 @@ void populateRequestInfo(bool outbound, bool use_host_header_fallback,
   uint64_t response_flags = 0;
   getValue({"response", "flags"}, &response_flags);
   request_info->response_flag = parseResponseFlag(response_flags);
+
+  getValue({"connection_id"}, &request_info->connection_id);
```
bianpengyuan, line 174: Shall we put this in extendedRequestInfo?

gargnupur: Thanks..done

PR #2924, opened 2020-07-21
This PR outputs binaries with BUILD_W_C=1

sdake:
When BUILD_WITH_CONTAINER=1 is set, this PR outputs all relevant binaries
for envoy. This target can be used to export the build cache, or
alternatively it may be viable to use the build cache directly in the
Istio build, linking these two repositories via a volume mount.

**Release note**:

```release-note
```

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -165,6 +165,13 @@ test_release:
 push_release: build
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/release-binary.sh -d "$(RELEASE_GCS_PATH)" -p
 
+# Used by build container to export the build output from the docker volume cache
+exportcache:
+	mkdir -p /work/out/linux_amd64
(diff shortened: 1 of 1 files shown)


sdake:
Part of: https://github.com/istio/istio/issues/25572

sdake:
@bianpengyuan can you please review? This PR doesn't change anything of significance that would affect those that don't use BUILD_WITH_CONTAINER=1 as an environment variable.

Cheers,
-steve

bianpengyuan:
Trying to understand. Assuming BUILD_W_C=1 is set, how is this target going to be used? I think `make export-cache` after `make build` won't work?

sdake:
@bianpengyuan, 

EXPORT BUILD_WITH_CONTAINER=1 make build
EXPORT BUILD_WITH_CONTAINER=1 make export-cache

the first operation will build envoy and store the results in a docker volume in /home/.cache.
The second operation will copy these files from the docker volume which is persistent to out/linux_amd64/(the 4 files)

Should be fine, unless someone has not built envoy in the first place.

Cheers,
-steve

PR #2926, opened 2020-07-21
Use root context to differentiate stream direction in stats filter

bianpengyuan:

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stats/plugin.cc
+++ b/extensions/stats/plugin.cc
@@ -240,7 +240,7 @@ const std::vector<MetricFactory>& PluginRootContext::defaultMetrics() {
   return default_metrics;
 }
 
-bool PluginRootContext::initializeDimensions(const json& j) {
+bool PluginRootContext::initializeDimensions(const json& j, bool is_outbound) {
   // Clean-up existing expressions.
   cleanupExpressions();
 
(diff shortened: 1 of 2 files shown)


extensions/stats/plugin.cc
```diff
@@ -240,7 +240,7 @@ const std::vector<MetricFactory>& PluginRootContext::defaultMetrics() {
   return default_metrics;
 }
 
-bool PluginRootContext::initializeDimensions(const json& j) {
+bool PluginRootContext::initializeDimensions(const json& j, bool is_outbound) {
```
kyessenov, line 243: So this is going to be called once in a shared VM? Don't we need to call it twice for anticipate both inbound and outbound? I think we have to build an internal dispatch in the root context based on VM type: `map[vm_type]RootContext`.

bianpengyuan: Yeah, good point. Rework this to use different root context to distinguish outbound/inbound.

kyessenov: Ok, but this is not saving us on VMs, right? A shared VM would have a single root context that is shared between outbound and inbound stat records. Or maybe I am once again confused by Wasm.

mandarjog: Root context is orthogonal to vm.
@jplevyak ?

bianpengyuan: Yeah, both root contexts get created at VM bootstrap, then later on stream context find its corresponding root context based on root_id in filter configuration.

kyessenov: What does this change then achieve? We could have just kept the field in the root context since we end up having two root contexts.

bianpengyuan: We always have two root context in the code base. I am not adding a new one. Without this change, we need to rely on two VM bootstraps to distinguish direction of root context: one triggered by inbound listener, one triggered by outbound listener, and each VM has one root context. With this change, two root contexts will be created on one shared VM, and each root context holds state for inbound or outbound.

I think it is fine to have multiple root context. It is designed to work that way. Later on we could even merge mx with stats vm to minimize VM overhead, which means one more root context in the VM.

kyessenov: What I'm saying is that I think it would be fine to leave code as-is and simply share the VM. Since streams associate with two root context, and traffic direction returns consistently the same value for each root/stream contexts. What am I missing?

bianpengyuan: Without this change, if we remove vm_ids, then only one VM will be bootstrapped, triggered by either outbound listener, or outbound listener. At that moment, both root contexts are initialized, and getDirection call in onConfigure will always either inbound or outbound. So we end up with two root contexts with same directional state.

kyessenov: I see. onConfigure is called per VM (or the last one).

PR #2949, opened 2020-07-27
Add extra fields in stackdriver access logs to match envoy access logs

gargnupur:

Add extra fields in stackdriver access logs to match envoy access logs

**Release note**:

```release-note
```

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -96,9 +96,8 @@ void extractServiceName(const std::string& host,
 //   host for destination service name.
 void getDestinationService(const std::string& dest_namespace,
                            bool use_host_header, std::string* dest_svc_host,
-                           std::string* dest_svc_name) {
-  std::string cluster_name;
-  getValue({"cluster_name"}, &cluster_name);
+                           std::string* dest_svc_name,
+                           std::string cluster_name) {
   *dest_svc_host = use_host_header
                        ? getHeaderMapValue(WasmHeaderMapType::RequestHeaders,
                                            kAuthorityHeaderKey)
@@ -135,12 +134,14 @@ void populateRequestInfo(bool outbound, bool use_host_header_fallback,
                          RequestInfo* request_info,
                          const std::string& destination_namespace) {
   request_info->is_populated = true;
+  getValue({"cluster_name"}, &request_info->upstream_cluster);
   // Fill in request info.
   // Get destination service name and host based on cluster name and host
   // header.
   getDestinationService(destination_namespace, use_host_header_fallback,
                         &request_info->destination_service_host,
-                        &request_info->destination_service_name);
+                        &request_info->destination_service_name,
+                        request_info->upstream_cluster);
 
   getValue({"request", "url_path"}, &request_info->request_url_path);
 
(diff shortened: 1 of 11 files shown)


extensions/common/context.cc
```diff
@@ -96,9 +96,8 @@ void extractServiceName(const std::string& host,
 //   host for destination service name.
 void getDestinationService(const std::string& dest_namespace,
                            bool use_host_header, std::string* dest_svc_host,
-                           std::string* dest_svc_name) {
-  std::string cluster_name;
-  getValue({"cluster_name"}, &cluster_name);
+                           std::string* dest_svc_name,
+                           std::string cluster_name) {
```
bianpengyuan, line 100:
```suggestion
                           const std::string& cluster_name) {
```

extensions/common/context.cc
```diff
@@ -338,6 +339,19 @@ void populateExtendedRequestInfo(RequestInfo* request_info) {
   getValue({"destination", "address"}, &request_info->destination_address);
   getValue({"source", "port"}, &request_info->source_port);
   getValue({"connection_id"}, &request_info->connection_id);
+  getValue({"route_name"}, &request_info->route_name);
+  getValue({"upstream", "address"}, &request_info->upstream_host);
+  getValue({"upstream", "cluster"}, &request_info->upstream_cluster);
```
bianpengyuan, line 344: This is fetched twice.

extensions/stackdriver/log/logger.cc
```diff
@@ -185,6 +185,13 @@ void Logger::fillAndFlushLogEntry(
   (*label_map)["protocol"] = request_info.request_protocol;
   (*label_map)["log_sampled"] = request_info.log_sampled ? "true" : "false";
   (*label_map)["connection_id"] = std::to_string(request_info.connection_id);
+  (*label_map)["route_name"] = request_info.route_name;
+  (*label_map)["upstream_host"] = request_info.upstream_host;
+  (*label_map)["upstream_cluster"] = request_info.upstream_cluster;
+  (*label_map)["requested_server_name"] = request_info.request_serever_name;
+  (*label_map)["x-envoy-original-path"] = request_info.x_envoy_original_path;
+  (*label_map)["x-envoy-original-dst-path"] =
```
bianpengyuan, line 193: s/path/host

extensions/common/context.cc
gargnupur: we are modifying cluster_name.. hence didn't make it a const ref.. and passed it by value rather than copying it to a temp variable... but looks like it's not a good idea from here: https://www.cplusplus.com/articles/z6vU7k9E/

Pass by value when the function does not want to modify the parameter and the
value is easy to copy (ints, doubles, char, bool, etc... simple types. std::string,
std::vector, and all other STL containers are NOT simple types.)

extensions/common/context.cc
gargnupur: Thanks.. removing

extensions/stackdriver/log/logger.cc
gargnupur: Thanks.. fixing

extensions/common/context.cc
bianpengyuan: Now look at the logic, it seems to be unnecessary to change value of cluster_name. We can just update the logic to:
```
  if (route_name == kBlackHoleRouteName) {
    *dest_svc_name = kBlackHoleCluster;
    return;
  }
  if (route_name == kPassThroughRouteName) {
    *dest_svc_name = kPassThroughCluster;
    return;
  }
  if (cluster_name == kInboundPassthroughClusterIpv4 ||
      cluster_name == kInboundPassthroughClusterIpv6) {
    *dest_svc_name = cluster_name;
    return;
  }
```

Also seems like we are getting route name here as well, so we are fetching it twice as well. Can we pass it in instead?

extensions/common/context.cc
mandarjog, line 344: When we have a 404 NR or in the other case 503 UC or UH it will not leave garbage value, right?

extensions/stackdriver/log/logger.cc
```diff
@@ -185,6 +185,13 @@ void Logger::fillAndFlushLogEntry(
   (*label_map)["protocol"] = request_info.request_protocol;
   (*label_map)["log_sampled"] = request_info.log_sampled ? "true" : "false";
   (*label_map)["connection_id"] = std::to_string(request_info.connection_id);
+  (*label_map)["route_name"] = request_info.route_name;
+  (*label_map)["upstream_host"] = request_info.upstream_host;
+  (*label_map)["upstream_cluster"] = request_info.upstream_cluster;
+  (*label_map)["requested_server_name"] = request_info.request_serever_name;
+  (*label_map)["x-envoy-original-path"] = request_info.x_envoy_original_path;
```
mandarjog, line 192: Are sd log entries searchable by these labels ?

extensions/common/context.cc
gargnupur: it should be empty in that case

extensions/stackdriver/log/logger.cc
gargnupur: yes.. it would be something like: labels.x-envoy-original-path

PR #2952, opened 2020-07-28
authn: revised request authenticator

Shikugawa:
This is a part of replacement of authn wasm implementation. It is using revised request authentication API. https://github.com/istio/api/pull/1536
#2800 is too large and hard to review all codes so that I split all of authn wasm implementation.

### DIFF
--- a/extensions/authn/filter_context.cc
+++ b/extensions/authn/filter_context.cc
@@ -21,44 +21,23 @@
 using istio::authn::Payload;
 using istio::authn::Result;
 
-// WASM_PROLOG
-#ifndef NULL_PLUGIN
-
-#include "proxy_wasm_intrinsics.h"
-
-#else  // NULL_PLUGIN
-
-#include "include/proxy-wasm/null_plugin.h"
-
-namespace proxy_wasm {
-namespace null_plugin {
+namespace Extensions {
 namespace AuthN {
-
-#endif  // NULL_PLUGIN
 
 using Envoy::Extensions::HttpFilters::HttpFilterNames;
 using Envoy::Utils::IstioFilterName;
 using google::protobuf::util::MessageToJsonString;
 
-void FilterContext::setPeerResult(const Payload* payload) {
-  if (payload != nullptr) {
-    switch (payload->payload_case()) {
-      case Payload::kX509:
-        // logDebug(
-        //     absl::StrCat("Set peer from X509: {}", payload->x509().user()));
-        result_.set_peer_user(payload->x509().user());
-        break;
-      case Payload::kJwt:
-        // logDebug(absl::StrCat("Set peer from JWT: {}",
-        // payload->jwt().user()));
-        result_.set_peer_user(payload->jwt().user());
-        break;
-      default:
-        // logDebug("Payload has not peer authentication data");
-        break;
-    }
-  }
-}
+FilterContext::FilterContext(
+    const envoy::config::core::v3::Metadata& dynamic_metadata,
+    const RequestHeaderMap& header_map, const Connection* connection,
+    const istio::envoy::config::filter::http::authn::v2alpha2::FilterConfig&
+        filter_config)
+    : dynamic_metadata_(dynamic_metadata),
+      header_map_(header_map),
+      connection_(connection),
+      filter_config_(filter_config) {}
+
 void FilterContext::setOriginResult(const Payload* payload) {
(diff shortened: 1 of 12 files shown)


extensions/authn/request_authenticator.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include "extensions/authn/authn_utils.h"
+#include "extensions/authn/filter_context.h"
+#include "security/v1beta1/request_authentication.pb.h"
+#include "src/istio/authn/context.pb.h"
+
+namespace Extensions {
+namespace AuthN {
+
+class IRequestAuthenticator {
```
yangminzhu, line 26: nit: I feel it's more common to name the abstract class as RequestAuthentication, and the implementation as RequestAuthenticationImpl in Envoy?

extensions/authn/request_authenticator.cc
```diff
@@ -0,0 +1,122 @@
...
+      filter_context_->setOriginResult(payload);
+      return true;
+    }
+  }
+
+  return false;
+}
+
+bool RequestAuthenticator::validateJwt(istio::authn::JwtPayload* jwt) {
+  for (const auto& jwt_rule : request_authentication_policy_.jwt_rules()) {
```
yangminzhu, line 93: I feel this could be greatly simplified without depending on the request_authentication policy at all.

Istiod already generates the Envoy JWT filter config from consolidated RequestAuthN policies, Envoy JWT filter should validate and output the JWT token to the metadata (via PayloadInMetadata)

can we just pass a list of issuers in the authn filter's config instead of passing the whole RequestAuthN policy? (I want to avoid dependency on the user-facing API in proxy)

extensions/authn/filter_context.cc
```diff
@@ -0,0 +1,113 @@
...
+  if (entry_it->second.struct_value().fields().empty()) {
+    return false;
+  }
+
+  // Serialize the payload from Envoy jwt filter first before writing it to
+  // |payload|.
+  // TODO (pitlv2109): Return protobuf Struct instead of string, once Istio jwt
+  // filter is removed. Also need to change how Istio authn filter processes the
+  // jwt payload.
+  MessageToJsonString(entry_it->second.struct_value(), payload);
```
yangminzhu, line 84: As you can see there is a TODO here, here we parse the struct to string, and later in ExtractOriginalPayload we parse the string into JSON object via JsonParse(), is it possible to avoid this and use the struct directly?

extensions/authn/filter_context.h
```diff
@@ -0,0 +1,97 @@
...
+using Envoy::Http::RequestHeaderMap;
+using Envoy::Network::Connection;
+
+// FilterContext holds inputs, such as request dynamic metadata and connection
+// and result data for authentication process.
+class FilterContext {
+ public:
+  FilterContext(
+      const envoy::config::core::v3::Metadata& dynamic_metadata,
+      const RequestHeaderMap& header_map, const Connection* connection,
```
kyessenov, line 36: This seems to be using Envoy interfaces. How are you going to compile this to Wasm?

Shikugawa: I'm not considering about wasm compilation in this PR. I will submit a PR which will include plugin.h/.cc using proxy-wasm interface. I will fix them in that.

kyessenov: I'm just wondering that if we're revising the implementation anyways, why not write it once with Wasm ABI.

Shikugawa: This PR doesn't include the build components which will be required to build AuthN filter. I'd like to split AuthN wasm implementations described in #2800. If we consider about wasm build in this PR. It will be complex to review all of codes, and out of request authenticator context. We have had already the implementation to be available not NullVM in #2800.

Shikugawa:
@kyessenov Discussed with @istio/wg-security-maintainers, we should revise authn filter with some of components, not monorith authn filter. I shared design doc of that. But there is no consensus with googlers. Can you review it and add some comments? https://docs.google.com/document/d/1JCVR3IoHzuvIvl2Qrep785eNLGOVdwZ3Ha2rrsM-K5w

kyessenov:
Splitting the filter into multiple is fine with me.

One thing to note is that we were trying to combine multiple Wasm modules
into one big fat module for performance reasons. So it might end up being
one module in the end.

zirain:
outdated

PR #2954, opened 2020-07-30
Basic Auth extension implementation and integration test

vicentefb:
**What this PR does / why we need it**: This PR adds Basic Auth plugin implementation and integration test.


bianpengyuan:
you can run `./scripts/check-style.sh` to fix lint issue.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,135 @@
...
+}
+
+FilterHeadersStatus PluginContext::onRequestHeaders(uint32_t, bool) {
+  auto result = getRequestHeaderPairs();
+  auto pairs = result->pairs();
+  LOG_WARN(std::string("headers: ") + std::to_string(pairs.size()));
+  for (auto& p : pairs) {
+    LOG_INFO(std::string(p.first) + std::string(" -> ") + std::string(p.second));
+    if(std::string(p.first).compare(std::string("authorization")) ==  0) {
+        // This will return Basic YWRtaW46YWRtaW4= which equals to admin:admin
```
richardwxn, line 100: This comment is for your testing case only, remove

testdata/filters/credentials.txt
```diff
@@ -0,0 +1 @@
+{ "credentials" : ["admin:admin", "admin2:admin2", "admin3:admin3", "admin4:admin4"]}
```
richardwxn, line 1: It might be ok to use a file for testing purpose, but is it possible to load credentials from k8s secret in cluster?

vicentefb: okok I'm going to research on that!

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,118 @@
...
+
+FilterHeadersStatus PluginContext::onRequestHeaders(uint32_t, bool) {
+  auto result = getRequestHeaderPairs();
+  auto pairs = result->pairs();
+  LOG_WARN(std::string("headers: ") + std::to_string(pairs.size()));
+  for (auto& p : pairs) {
+    if (std::string(p.first).compare(std::string("authorization")) == 0) {
+      basicAuthorizationHeader = std::string(p.second);
+      // obtain the Base 64 encoded username:passwd
+      basicAuthorizationHeader = basicAuthorizationHeader.substr(6);
```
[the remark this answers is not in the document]
vicentefb: okok true, will add that!

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,118 @@
...
+  // Parse configuration JSON string.
+  auto result = ::Wasm::Common::JsonParse(configuration_data->view());
+  if (!result.has_value()) {
+    LOG_WARN(absl::StrCat("cannot parse plugin configuration JSON string: ",
+                          configuration_data->view()));
+    return false;
+  }
+  // j is a JsonObject holds configuration data
+  auto j = result.value();
+  credentials = ::Wasm::Common::JsonGetField<std::string>(j, "credentials")
```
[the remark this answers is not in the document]
vicentefb: I can use JsonArrayIterate to verify if there's an array inside my JSON object but not to retrieve its contents.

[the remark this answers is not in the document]
bianpengyuan: You can pass in a lambda and retrieve string_views from the json object into a captured container. Example usage: https://github.com/istio/proxy/blob/77466b11af0aa9a5a8b1817ed02360ea90906113/extensions/stats/plugin.cc#L266

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,118 @@
...
+                    .JsonGetArray(j, "credentials");
+  if (!credentials.empty()) {
+    for (auto credential : credentials) {
```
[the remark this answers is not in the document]
vicentefb: okok, just waiting on any comments/suggestions on the config doc

vicentefb:
> @vicentefb: The following tests **failed**, say `/retest` to rerun all failed tests:
> 
> Test name	Commit	Details	Rerun command
> test-tsan_proxy	[203ed63](https://github.com/istio/proxy/commit/203ed63edeeb2ea16ead1d6821a8a9400f1187d1)	[link](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_proxy/2954/test-tsan_proxy/1057)	`/test test-tsan_proxy`
> test-asan_proxy	[203ed63](https://github.com/istio/proxy/commit/203ed63edeeb2ea16ead1d6821a8a9400f1187d1)	[link](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_proxy/2954/test-asan_proxy/998)	`/test test-asan_proxy`
> test_proxy	[4fa7c43](https://github.com/istio/proxy/commit/4fa7c43704117c7e502aa4f140d1fc3838c99e12)	[link](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_proxy/2954/test_proxy/1006)	`/test test_proxy`

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+  auto result = ::Wasm::Common::JsonParse(configuration_data->view());
+  if (!result.has_value()) {
+    LOG_WARN(absl::StrCat("cannot parse plugin configuration JSON string: ",
+                          configuration_data->view()));
+    return false;
+  }
+  // j is a JsonObject holds configuration data
+  auto j = result.value();
+
+  if (!JsonArrayIterate(j, "pathPrefixes", [&](const json& prefixes) -> bool {
```
bianpengyuan, line 77: After moving `configurationData` into RootContext, you should make it capture `this` instead.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+        auto pathPrefix =
+            JsonGetField<std::string>(prefixes, "pathPrefix").value_or("");
+        if (pathPrefix == "") return false;
```
bianpengyuan, line 80:
```suggestion
        if (pathPrefix.empty()) {
          return false;
        }
```

bianpengyuan: Please make sure every if else block is wrapped.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+        struct S s1;
+        if (!JsonArrayIterate(prefixes, "requestMethods",
+                              [&](const json& methods) -> bool {
+                                s1.requestMethods.insert(std::string(methods));
```
bianpengyuan, line 84: Can you use jsonValueAs to prevent undesired data type? https://github.com/istio/proxy/blob/422ba7d52af4e9e12ae75497f0d7483a565908b1/extensions/common/json_util.cc#L77

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+                                return true;
+                              })) {
+          LOG_WARN("Failed to parse configuration for request methods.");
+          return false;
+        }
+
+        if (!JsonArrayIterate(prefixes, "credentials",
+                              [&](const json& credentials) -> bool {
+                                auto cred = std::string(credentials);
```
bianpengyuan, line 93: Ditto

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+                                s1.encodedCredentials.insert(
+                                    Base64::encode(cred.data(), cred.size()));
+                                return true;
+                              })) {
+          LOG_WARN("Failed to parse configuration for credentials.");
+          return false;
+        }
+        auto checkPath = configurationData.find(pathPrefix);
```
bianpengyuan, line 101: c++ variable names should be snake case. Please update all variables in this change to follow the pattern.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+        if (checkPath != configurationData.end()) {
+          for (auto& methods : s1.requestMethods) {
```
bianpengyuan, line 103:
```suggestion
          for (const auto& methods : s1.requestMethods) {
```

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+            configurationData[pathPrefix].requestMethods.insert(methods);
+          }
+
+          for (auto& creds : s1.encodedCredentials) {
```
bianpengyuan, line 107:
```suggestion
          for (const auto& creds : s1.encodedCredentials) {
```

extensions/basic_auth/plugin.cc
bianpengyuan: Also please rename this to something ended with `_iter`.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+      })) {
+    LOG_WARN("Failed to parse configuration information.");
+  }
+
+  return true;
+}
+
+FilterHeadersStatus PluginContext::onRequestHeaders(uint32_t, bool) {
+  auto authorizationHeader = getRequestHeader("authorization")->toString();
+  auto pathPrefix = getRequestHeader(":path")->toString();
```
bianpengyuan, line 123: You don't need to get these two headers unless there is a path match from config, which would make this cheaper.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+  auto method = getRequestHeader(":method")->toString();
+  auto checkPathPrefix = configurationData.find(pathPrefix);
+  if (checkPathPrefix != configurationData.end()) {
+    auto checkMethod =
```
bianpengyuan, line 127: const auto& rule = configurationData[pathPrefix]; and then use rule afterwards instead of `configurationData[pathPrefix]`.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  return true;
+}
+
+FilterHeadersStatus PluginContext::onRequestHeaders(uint32_t, bool) {
+  auto authorizationHeader = getRequestHeader("authorization")->toString();
+  auto pathPrefix = getRequestHeader(":path")->toString();
+  auto method = getRequestHeader(":method")->toString();
+  auto checkPathPrefix = configurationData.find(pathPrefix);
```
bianpengyuan, line 125: This checks the whole path against prefix in configuration... Please make sure the logic here is right.. this is going to be a very sensitive code path since it relates to security.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+  if (checkPathPrefix != configurationData.end()) {
+    auto checkMethod =
+        configurationData[pathPrefix].requestMethods.find(method);
+    auto checkEmptyMethod =
```
bianpengyuan, line 129: It is hard for me to understand code in this function. Can you either break it down into sub function, or at least add more comments?

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+
+struct S {
+  std::unordered_set<std::string> requestMethods;
+  std::unordered_set<std::string> encodedCredentials;
+};
+
+using ::nlohmann::json;
+using ::Wasm::Common::JsonArrayIterate;
+using ::Wasm::Common::JsonGetField;
+std::unordered_map<std::string, S> configurationData;
```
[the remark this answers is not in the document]
vicentefb: If I'm using that variable in two functions from two different classes (RootContext and PluginContext), what would be the best C++ approach to this? Should I make it static variable inside RootContext class and make that class a friend of PluginContext class?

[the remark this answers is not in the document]
bianpengyuan: No you should create a method in root context class. and call it in stream context. There are plenty examples about how to access root context in stream context in other extensions.

extensions/basic_auth/config.proto
```diff
@@ -0,0 +1,62 @@
...
+import "google/protobuf/duration.proto";
+
+message PluginConfig {
+  // Multiple independent basic auth configurations
+  repeated BasicAuth pathPrefixes = 1;
+}
+
+message BasicAuth {
+  // HTTP path prefix to restrict access to.
+  string pathPrefix = 1;
```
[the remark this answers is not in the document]
kyessenov: I think we might need to delegate to attributegen adapter instead. Perhaps, add a condition based on what attribute generation adapter produces: see https://github.com/istio/proxy/blob/master/extensions/attributegen/config.proto#L31. Matchers can get pretty complicated (https://github.com/envoyproxy/envoy/pull/12317/files#diff-dc10a97bd5c236d9ce6e7e0d1c1c4a3c), so we need to factor it out from each module.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,118 @@
...
+      return true;
+    }
+  }
+  return false;
+}
+
+bool PluginRootContext::onConfigure(size_t size) {
+  // Parse configuration JSON string.
+  if (size > 0 && !configure(size)) {
+    LOG_WARN("configuration has errors, but initialzation can continue.");
```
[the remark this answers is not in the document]
kyessenov: Yes, it should fail.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+PROXY_WASM_NULL_PLUGIN_REGISTRY;
+
+using Base64 = Envoy::Base64;
+
+#endif
+
+static RegisterContextFactory register_BasicAuth(
+    CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContext));
+
+struct S {
```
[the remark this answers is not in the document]
kyessenov: You can also just make a tuple for something like this. C++17 has nice tuple syntax https://en.cppreference.com/w/cpp/language/structured_binding

extensions/basic_auth/plugin.cc
kyessenov: Yes, you need to iterate over match conditions and compare for a prefix. This would be good enough for now. TRIE would be faster.

extensions/basic_auth/config.proto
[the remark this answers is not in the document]
vicentefb: okok changed

extensions/basic_auth/config.proto
```diff
@@ -0,0 +1,62 @@
...
+//  ]
+// }
+
+package basic_auth;
+
+import "google/protobuf/duration.proto";
+
+message PluginConfig {
+  // Multiple independent basic auth configurations
+  repeated BasicAuth pathPrefixes = 1;
```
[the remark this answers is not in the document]
vicentefb: okok done

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,192 @@
...
+  std::unordered_set<std::string> requestMethods;
+  std::unordered_set<std::string> encodedCredentials;
+};
+
+using ::nlohmann::json;
```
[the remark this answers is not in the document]
vicentefb: okok done.

extensions/basic_auth/plugin.cc
vicentefb: okok updated all variables

extensions/basic_auth/plugin.cc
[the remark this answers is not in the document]
vicentefb: okok done

extensions/basic_auth/plugin.cc
vicentefb: Okok sound good, changed.

extensions/basic_auth/plugin.cc
vicentefb: okok sure!

extensions/basic_auth/plugin.cc
vicentefb: oh that's true and more readable. Thanks!

extensions/basic_auth/plugin.cc
vicentefb: True okok changed that.

extensions/basic_auth/plugin.cc
vicentefb: I've already moved `configuration_data` into RootContext. By "capture `this` " you mean to store the values into the container, right?

extensions/basic_auth/plugin.cc
vicentefb: sure, i splitted it into subfunctions and added comments.

extensions/basic_auth/plugin.cc
vicentefb: Oh i see, i just realize that the naming is misleading. In fact, this logic covers the case where the `match_pattern` is exact.

extensions/common/json_util.h
```diff
@@ -71,7 +71,15 @@ JsonValueAs<std::vector<std::string_view>>(const JsonObject& j);
 template <class T>
 class JsonGetField {
  public:
+<<<<<<< HEAD
```
richardwxn: Resolve the conflicts in this file and clean up

testdata/stats/stats_filter_config.yaml
```diff
@@ -0,0 +1,8 @@
+debug: "true"
```
richardwxn, line 1: is this change related?

richardwxn:
Can you exclude unrelated commits? Seems like you you are bringing them back

extensions/basic_auth/config.proto
```diff
@@ -0,0 +1,72 @@
...
+// {
+//   "basic_auth_rules": [
+//     {
+//       "request_path":"/api/reviews/pay",
+//       "match": "exact",
+//       "request_methods":[ "GET", "POST" ],
+//       "credentials":[ "admin:admin", "admin2:admin2" ]
+//     },
+//     {
+//       "request_path":"/api",
```
kyessenov, line 33: The example does not seem to match the schema.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,330 @@
...
+using ::Wasm::Common::JsonValueAs;
+
+static RegisterContextFactory register_BasicAuth(
+    CONTEXT_FACTORY(PluginContext), ROOT_FACTORY(PluginRootContext));
+
+namespace {
+void deniedNoAccessToCredentials() {
+  sendLocalResponse(401,
+                    "Request denied by Basic Auth check. No credential "
+                    "has accessed to specified prefix path.",
```
kyessenov, line 55: has access

kyessenov: also, why is it prefix always?

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,330 @@
...
+  sendLocalResponse(401,
+                    "Request denied by Basic Auth check. Invalid "
+                    "username and/or password",
+                    "", {});
+}
+
+void deniedNoAcessToMethod() {
+  sendLocalResponse(401,
+                    "Request denied by Basic Auth check. No method "
+                    "has access to specified prefix path.",
```
kyessenov, line 83: Consider rephrasing: Request method not authorized.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,330 @@
...
+}
+
+void deniedInvalidCredentials() {
+  sendLocalResponse(401,
+                    "Request denied by Basic Auth check. Invalid "
+                    "username and/or password",
+                    "", {});
+}
+
+void deniedNoAcessToMethod() {
```
kyessenov, line 80: "access"

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,330 @@
...
+                    "Request denied by Basic Auth check. No method "
+                    "has access to specified prefix path.",
+                    "", {});
+}
+}  // namespace
+
+bool PluginRootContext::onConfigure(size_t size) {
+  // Parse configuration JSON string.
+  if (size > 0 && !configure(size)) {
+    LOG_WARN("configuration has errors, but initialzation can continue.");
```
kyessenov, line 91: I think the comment is wrong now. Returning false will not make it continue.

extensions/basic_auth/config.proto
vicentefb: I added the submessage "match" to be able to know which match pattern to apply. What part doesn't match the schema?

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,330 @@
...
+                                         credential.first.value().size()));
+                      return true;
+                    })) {
+              LOG_WARN("Failed to parse configuration for credentials.");
+              return false;
+            }
+            auto path_iter = basic_auth_configuration_.find(match);
+            if (path_iter != basic_auth_configuration_.end() &&
+                path_iter->second.find(request_path) !=
+                    basic_auth_configuration_[match].end()) {
```
kyessenov, line 178: This looks like a second lookup. You can just path_iter->second.end().

extensions/basic_auth/plugin.cc
vicentefb: oh, i missed that one.

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,100 @@
...
+      : RootContext(id, root_id) {}
+  ~PluginRootContext() {}
+  bool onConfigure(size_t) override;
+  PluginRootContext();
+
+  struct headerData {
+    std::unordered_set<std::string> request_methods;
+    std::unordered_set<std::string> encoded_credentials;
+  };
+  std::unordered_map<
```
kyessenov, line 58: This incurs a copy. Return it as const reference.

extensions/basic_auth/plugin.cc
vicentefb: sure!

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,100 @@
...
+      basic_auth_configuration_;
+};
+
+// Per-stream context.
+class PluginContext : public Context {
+ public:
+  explicit PluginContext(uint32_t id, RootContext* root) : Context(id, root) {}
+  FilterHeadersStatus onRequestHeaders(uint32_t, bool) override;
+  FilterHeadersStatus credentialsCheck(
+      std::unordered_map<std::string,
```
kyessenov, line 79: Pass as const reference.

extensions/basic_auth/plugin.cc
vicentefb: ok, changed!

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,330 @@
...
+              for (auto& methods : header.request_methods) {
+                basic_auth_configuration_[match][request_path]
```
kyessenov, line 180: This is a bit strange layout. I would expect it to be keyed by request method since there is a fixed limited set of them. Paths are usually prefix-matched, which is difficult to do without iterating over the map.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,330 @@
...
+            }
+            return true;
+          })) {
+    LOG_WARN("Failed to parse configuration information.");
+  }
+  return true;
+}
+
+FilterHeadersStatus PluginContext::credentialsCheck(
+    std::unordered_map<std::string, PluginRootContext::headerData>::mapped_type
```
kyessenov, line 199: mapped_type just resolves to headerData, right?

extensions/basic_auth/plugin.cc
vicentefb: okok that's true!

extensions/basic_auth/plugin.cc
vicentefb: yes

extensions/basic_auth/plugin.cc
vicentefb: I chose `match` to be the key because first I try to find an exact match for the requested path, if that's not the case I proceed to iterate over the request_paths that are mapped to prefix and suffix match patterns in order to find a prefix or suffix that matches the HTTP request path. I don't know if this makes sense. I am basing my logic off of this example: https://swagger.io/docs/specification/2-0/authentication/basic-authentication/ .If I change the key to be `request_method` instead of `match` I would still need to have as key of my second map `match`, right?

extensions/basic_auth/config.proto
vicentefb: oh i see, you mean that instead of being `"request_path": "/api"` it should be `"prefix": "/api"` , right?

bianpengyuan:
@vicentefb did you forget to push new change?

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,100 @@
...
+// PluginRootContext is the root context for all streams processed by the
+// thread. It has the same lifetime as the worker thread and acts as target for
+// interactions that outlives individual stream, e.g. timer, async calls.
+class PluginRootContext : public RootContext {
+ public:
+  PluginRootContext(uint32_t id, std::string_view root_id)
+      : RootContext(id, root_id) {}
+  ~PluginRootContext() {}
+  bool onConfigure(size_t) override;
+  PluginRootContext();
```
bianpengyuan, line 52: I don't think this is needed?

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,100 @@
...
+      std::unordered_map<std::string, PluginRootContext::headerData>>
+      basic_auth_configuration_;
+};
+
+// Per-stream context.
+class PluginContext : public Context {
+ public:
+  explicit PluginContext(uint32_t id, RootContext* root) : Context(id, root) {}
+  FilterHeadersStatus onRequestHeaders(uint32_t, bool) override;
+  FilterHeadersStatus credentialsCheck(
```
bianpengyuan, line 78: This can be moved to private.

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,100 @@
...
+      const std::unordered_map<std::string,
+                               PluginRootContext::headerData>::mapped_type,
+      std::string);
```
bianpengyuan, line 81: const std::string&

extensions/basic_auth/plugin.cc
bianpengyuan: Why not just use PluginRootContext::headerData then? `const PluginRootContext::headerData&`

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,100 @@
...
+// interactions that outlives individual stream, e.g. timer, async calls.
+class PluginRootContext : public RootContext {
+ public:
+  PluginRootContext(uint32_t id, std::string_view root_id)
+      : RootContext(id, root_id) {}
+  ~PluginRootContext() {}
+  bool onConfigure(size_t) override;
+  PluginRootContext();
+
+  struct headerData {
```
bianpengyuan, line 54: Type def should start with capital letter.

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,100 @@
...
+      std::string);
+
+ private:
+  inline PluginRootContext* rootContext() {
+    return dynamic_cast<PluginRootContext*>(this->root());
+  };
+  inline std::unordered_map<
+      std::string,
+      std::unordered_map<std::string, PluginRootContext::headerData>>
+  basicAuthConfigurationValue() {
```
bianpengyuan, line 90: This is just used at one place. And wrap this call with a function complicate the code. You don't need to wrap it. Just replace this with `rootContext()->basicAuthConfigurationValue()` at the place that this function is invoked.

extensions/basic_auth/plugin.cc
bianpengyuan: I agree with Kuat on this. Using pattern as key is not saving you any runtime on searching. It makes more sense to use method as key and value could just be an array of struct, which has a path matcher{prefix, exact, suffix} and allowed credentials on each matcher.

extensions/basic_auth/plugin.h
bianpengyuan: Also headerData is not a informative name to me, It should follow the name used in configuration proto, such as rule etc.

extensions/basic_auth/config.proto
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+syntax = "proto3";
+
+// $title: Basic Auth Config
+// $description: Configuration for Basic Auth Filter.
+// $weight: 20
+
+// The following is an example of a configuration that restricts access to
+// given path prefixes, request methods and credentials.
```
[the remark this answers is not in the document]
vicentefb: okok true, changed!

extensions/basic_auth/plugin.h
vicentefb: true it is not needed

extensions/basic_auth/plugin.cc
vicentefb: done!

extensions/basic_auth/plugin.h
vicentefb: okok sure!

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+                    })) {
+              LOG_WARN("Failed to parse configuration for request path.");
+              return false;
+            }
+            if (request_path == "" || match == "") {
+              return false;
+            }
+            if (!JsonArrayIterate(
+                    configuration, "request_methods",
+                    [&](const json& methods) -> bool {
```
bianpengyuan, line 157: should not be plural. Should be method and next line should use method_string variable.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+                        return false;
+                      }
+                      request_methods.push_back(
+                          std::string(method.first.value()));
+                      return true;
+                    })) {
+              LOG_WARN("Failed to parse configuration for request methods.");
+              return false;
+            }
+            struct BasicAuthConfigRule rules;
```
bianpengyuan, line 171: This is just one rule, Should not be plural.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+            if (!JsonArrayIterate(
+                    configuration, "credentials",
+                    [&](const json& credentials) -> bool {
+                      auto credential = JsonValueAs<std::string>(credentials);
+                      if (credential.second !=
+                          Wasm::Common::JsonParserResultDetail::OK) {
+                        LOG_WARN("unexpected credential");
```
bianpengyuan, line 178: Same here.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+
+            if (match == "prefix") {
+              rules.pattern = Prefix;
+            } else if (match == "exact") {
+              rules.pattern = Exact;
+            } else if (match == "suffix") {
+              rules.pattern = Suffix;
+            }
+            rules.request_path = request_path;
+            for (auto& methods : request_methods) {
```
bianpengyuan, line 198:
```suggestion
            for (auto& method : request_methods) {
```

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+              basic_auth_configuration_[methods].push_back(rules);
+            }
+            return true;
+          })) {
+    LOG_WARN("Failed to parse configuration information.");
```
bianpengyuan, line 203: Update this error message to something more informative, like `absl::StrCat("cannot parse plugin configuration JSON string: ", configuration_data->view())`

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,92 @@
...
+};
+
+// Per-stream context.
+class PluginContext : public Context {
+ public:
+  explicit PluginContext(uint32_t id, RootContext* root) : Context(id, root) {}
+  FilterHeadersStatus onRequestHeaders(uint32_t, bool) override;
+
+ private:
+  FilterHeadersStatus credentialsCheck(
```
bianpengyuan, line 80: Instead of implementing this in PluginContext, you can just implement it in RootContext, so that you can avoid copying configuration from it. When PluginContext needs to do credential check, it can just call `rootContext()->checkCredential(...)`.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+  auto method_iter = basic_auth_configuration.find(method);
+  // First we check if the request method is present in our container
+  if (method_iter != basic_auth_configuration.end()) {
+    // We iterate through our vector of struct in order to find if the
+    // request_path according to given match patterns, is part of the plugin's
+    // configuration data. If that's the case we check the credentials
+    for (auto& rules : basic_auth_configuration[method]) {
+      if (rules.pattern == PluginRootContext::MATCH_TYPE::Prefix) {
+        absl::string_view path = request_path;
+        if (absl::ConsumePrefix(&path, rules.request_path)) {
```
bianpengyuan, line 256: ConsumePrefix/Suffix would strip prefix/suffix from the string right? You just need to check the prefix/suffix. absl has StartWith and EndWith: https://github.com/abseil/abseil-cpp/blob/d0c433455801e1c1fb6f486f0b447e22f946ab52/absl/strings/match.h#L53-L62 which are better fit here.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+    return FilterHeadersStatus::StopIteration;
+  } else {
+    return FilterHeadersStatus::Continue;
+  }
+  return FilterHeadersStatus::StopIteration;
+}
+
+FilterHeadersStatus PluginContext::onRequestHeaders(uint32_t, bool) {
+  auto basic_auth_configuration = rootContext()->basicAuthConfigurationValue();
+  auto request_path = getRequestHeader(":path")->toString();
```
bianpengyuan, line 245: I don't think you will need to get them as string. This will make a copy of the path, which is inefficient. `auto request_path = getRequestHeader(":path")` returns `WasmData` and I think you can use string_view for most of the code beneath: `request_path->view()`

bianpengyuan: Same comment to method and authorization header.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+    // We iterate through our vector of struct in order to find if the
+    // request_path according to given match patterns, is part of the plugin's
+    // configuration data. If that's the case we check the credentials
+    for (auto& rules : basic_auth_configuration[method]) {
+      if (rules.pattern == PluginRootContext::MATCH_TYPE::Prefix) {
+        absl::string_view path = request_path;
+        if (absl::ConsumePrefix(&path, rules.request_path)) {
+          auto authorization_header =
+              getRequestHeader("authorization")->toString();
+          return credentialsCheck(rules, authorization_header);
```
bianpengyuan, line 259: You probably need to go through all the rules, instead of just find one match and claim done. Consider case where several rules match the same path, this needs to check the authorization header against all of them.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+  return true;
+}
+
+FilterHeadersStatus PluginContext::credentialsCheck(
+    const PluginRootContext::BasicAuthConfigRule& rule,
+    const std::string& authorization_header) {
+  auto credential_iter = rule.encoded_credentials.find("");
+  auto credential_size = rule.encoded_credentials.size();
+  // Check if credential set is of of size 1 and its an empty string.
+  if (credential_size == 1 &&
```
bianpengyuan, line 214: What is the purpose of this condition? Empty credential should be considered as an invalid configuration.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+      credential_iter != rule.encoded_credentials.end()) {
+    deniedNoAccessToCredentials();
+    return FilterHeadersStatus::StopIteration;
+  }
+  // Check if there is an authorization header
+  if (authorization_header.compare("") == 0) {
+    deniedNoAuthorizationHeader();
+    return FilterHeadersStatus::StopIteration;
+  }
+  absl::string_view basic_authorization_header = authorization_header;
```
bianpengyuan, line 224: You can just pass in string_view.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+  auto credential_iter = rule.encoded_credentials.find("");
+  auto credential_size = rule.encoded_credentials.size();
+  // Check if credential set is of of size 1 and its an empty string.
+  if (credential_size == 1 &&
+      credential_iter != rule.encoded_credentials.end()) {
+    deniedNoAccessToCredentials();
+    return FilterHeadersStatus::StopIteration;
+  }
+  // Check if there is an authorization header
+  if (authorization_header.compare("") == 0) {
```
bianpengyuan, line 220: I don't think you'd need to check if it empty or not. The following check will handle empty authorization header.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+      rule.encoded_credentials.find(std::string(basic_authorization_header));
+  // Check if encoded credential is part of the encoded_credentials
+  // set from our container to grant or deny access.
+  if (auth_credential_iter == rule.encoded_credentials.end()) {
+    deniedInvalidCredentials();
+    return FilterHeadersStatus::StopIteration;
+  } else {
+    return FilterHeadersStatus::Continue;
+  }
+  return FilterHeadersStatus::StopIteration;
```
bianpengyuan, line 240: This line will never be reached right? Should we just move `return FilterHeadersStatus::Continue;` out from else and use that as the final return.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+            std::string request_path;
+            std::vector<std::string> request_methods;
+            if (!JsonObjectIterate(
+                    configuration, "request_path",
+                    [&](std::string pattern) -> bool {
+                      match = pattern;
+                      auto path = JsonGetField<std::string>(
+                                      configuration["request_path"], pattern)
+                                      .value_or("");
+                      auto path_string = JsonValueAs<std::string>(path);
```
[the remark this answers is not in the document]
vicentefb: yes, you are right, removed.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,287 @@
...
+                        return false;
+                      }
+                      request_path = path;
+                      return true;
+                    })) {
+              LOG_WARN("Failed to parse configuration for request path.");
+              return false;
+            }
+            if (request_path == "" || match == "") {
+              return false;
```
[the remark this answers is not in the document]
vicentefb: okok, sure, implemented.

extensions/basic_auth/plugin.cc
vicentefb: sure!

extensions/basic_auth/plugin.h
vicentefb: oh i see! done.

test/envoye2e/basic_auth/basic_auth_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"fmt"
+	"istio.io/proxy/test/envoye2e"
+	"istio.io/proxy/test/envoye2e/driver"
+	"os"
+	"path/filepath"
+	"strconv"
+	"testing"
+	"time"
+
+	"istio.io/proxy/test/envoye2e/env"
```
bianpengyuan, line 27: The import here does not look right to me, should be something like:
```suggestion
	"fmt"
	"os"
	"path/filepath"
	"strconv"
	"testing"
	"time"

	"istio.io/proxy/test/envoye2e"
	"istio.io/proxy/test/envoye2e/driver"
	"istio.io/proxy/test/envoye2e/env"
```

Wondering why CI is not catching it.

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,92 @@
...
+// Per-stream context.
+class PluginContext : public Context {
+ public:
+  explicit PluginContext(uint32_t id, RootContext* root) : Context(id, root) {}
+  FilterHeadersStatus onRequestHeaders(uint32_t, bool) override;
+
+ private:
+  inline PluginRootContext* rootContext() {
+    return dynamic_cast<PluginRootContext*>(this->root());
+  };
```
bianpengyuan, line 84:
```suggestion
  }
```

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,92 @@
...
+
+#pragma once
+
+#include "extensions/common/context.h"
+#include "google/protobuf/util/json_util.h"
+
+#ifndef NULL_PLUGIN
+
+#include <assert.h>
+#define ASSERT(_X) assert(_X)
```
bianpengyuan, line 24: I think you are not using assert here right? Could you clean this up? So does line 28.

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,92 @@
...
+
+namespace proxy_wasm {
+namespace null_plugin {
+namespace BasicAuth {
+namespace Plugin {
+
+#endif
+
+using google::protobuf::util::JsonParseOptions;
+using google::protobuf::util::Status;
```
bianpengyuan, line 42: These are also left over of copy paste?

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,92 @@
...
+                           std::vector<PluginRootContext::BasicAuthConfigRule>>
+  basicAuthConfigurationValue() {
+    return basic_auth_configuration_;
+  };
+  FilterHeadersStatus credentialsCheck(const BasicAuthConfigRule&,
+                                       std::string_view);
+
+ private:
+  bool configure(size_t);
+  std::unordered_map<std::string,
```
bianpengyuan, line 70: Can you add some comment here about this map? What is the key and corresponding value?

extensions/basic_auth/plugin.h
bianpengyuan: Hmm I might not be clearly here. I think the whole logic inside `onRequestHeaders` could just be moved into a method defined in RootContext (you can call it check or something). Then you don't need to exchange data structure between root context and plugin context. Check logic could just contained inside root context.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,258 @@
...
+  if (method_iter != basic_auth_configuration.end()) {
+    // We iterate through our vector of struct in order to find if the
+    // request_path according to given match patterns, is part of the plugin's
+    // configuration data. If that's the case we check the credentials
+    FilterHeadersStatus header_status = FilterHeadersStatus::Continue;
+    for (auto& rules : basic_auth_configuration[method]) {
+      if (rules.pattern == PluginRootContext::MATCH_TYPE::Prefix) {
+        if (absl::StartsWith(request_path, rules.request_path)) {
+          auto authorization_header = getRequestHeader("authorization");
+          std::string_view authorization = authorization_header->view();
```
bianpengyuan, line 227: This will fetch authorization header multiple times and I guess you are doing it because consumePrefix/Suffix will change the string?

Absl has a function called stripPrefix/Suffix and I think it will leave the string intact, so that you can pass in authorization string_view without fetching it repeatedly.

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,258 @@
...
+            }
+            return true;
+          })) {
+    LOG_WARN(absl::StrCat("cannot parse plugin configuration JSON string: ",
+                          configuration_data->view()));
+  }
+  return true;
+}
+
+FilterHeadersStatus PluginRootContext::credentialsCheck(
```
bianpengyuan, line 192: This does not need to be method of RootContext right? It can just be an util function in anonymous namespace.

test/envoye2e/basic_auth/basic_auth_test.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+func skipWasm(t *testing.T, runtime string) {
```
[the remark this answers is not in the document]
vicentefb: okok going to move it to `test/envoye2e/env/utils.go` file and will modify `stats_test.go` as well which is the only file that uses skipWasm function. I'm going to also change the package name in this file.

test/envoye2e/basic_auth/basic_auth_test.go
vicentefb: okok changed.

extensions/basic_auth/plugin.h
vicentefb: yes.

extensions/basic_auth/plugin.h
vicentefb: done.

extensions/basic_auth/plugin.cc
vicentefb: okok sure!

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,93 @@
...
+  ~PluginRootContext() {}
+  bool onConfigure(size_t) override;
+
+  enum MATCH_TYPE { Prefix, Exact, Suffix };
+  struct BasicAuthConfigRule {
+    std::string request_path;
+    MATCH_TYPE pattern;
+    std::unordered_set<std::string> encoded_credentials;
+  };
+  FilterHeadersStatus check() { return checkRequestHeaders(); };
```
bianpengyuan, line 57: I don't think you need to do this redirect. You can just copy all logic from checkRequestHeaders() to check().

extensions/basic_auth/plugin.h
```diff
@@ -0,0 +1,93 @@
...
+      : RootContext(id, root_id) {}
+  ~PluginRootContext() {}
+  bool onConfigure(size_t) override;
+
+  enum MATCH_TYPE { Prefix, Exact, Suffix };
+  struct BasicAuthConfigRule {
+    std::string request_path;
+    MATCH_TYPE pattern;
+    std::unordered_set<std::string> encoded_credentials;
+  };
```
bianpengyuan, line 56: nit: These no longer need to be public right?

extensions/basic_auth/plugin.cc
```diff
@@ -0,0 +1,262 @@
...
+
+FilterHeadersStatus credentialsCheck(
+    const PluginRootContext::BasicAuthConfigRule& rule,
+    std::string_view authorization_header) {
+  // Check if the Basic auth header starts with "Basic "
+  std::string_view authorization_header_strip =
+      absl::StripPrefix(authorization_header, "Basic ");
+  // If authorization_header_strip is the string as the original string
+  // it means there was no "Basic " prefix found.
+  if (authorization_header_strip == authorization_header) {
```
bianpengyuan, line 75: nit: instead of checking equality here, I would suggest at the beginning of the function, Use StartWith check whether authorization header start with `Basic `

extensions/basic_auth/plugin.h
bianpengyuan: By copying I mean just remove `checkRequestHeaders` and make `check` includes all the logic in .cc file.

extensions/basic_auth/plugin.h
vicentefb: okok they were public because `checkCredentials()` function is inside anonymous namespace. I'm going to place the struct and enum private and move the function outside the namespace and include it as a private method in PluginRootContext.

bianpengyuan: ok, sgtm.

PR #2955, opened 2020-07-31
Enable Client Side Access Logs for SD

gargnupur:
Enable Client Side Access Logs for SD

Will need sampling for client side logs too.. will add in a follow-up PR.

**Release note**:

```release-note
```


douglas-reid:
Do we want a way to control whether or not we want both server and client access logs in config?

extensions/stackdriver/stackdriver.cc
```diff
@@ -461,8 +461,8 @@ inline bool StackdriverRootContext::isOutbound() {
   return direction_ == ::Wasm::Common::TrafficDirection::Outbound;
 }
 
-inline bool StackdriverRootContext::enableServerAccessLog() {
-  return !config_.disable_server_access_logging() && !isOutbound();
+inline bool StackdriverRootContext::enableAccessLog() {
+  return !config_.disable_server_access_logging();
```
douglas-reid, line 465: I feel like we want more config around enabling client access logging and, relatedly, we need to assess whether or not `disable_server_access_logging` should be renamed.

testdata/stackdriver/client_access_log_entry_sampled.yaml.tmpl
```diff
@@ -0,0 +1,27 @@
...
+  destination_namespace: default
+  source_principal: "{{ .Vars.SourcePrincipal }}"
+  destination_workload: productpage-v1
+  destination_app: productpage
+  destination_canonical_service: productpage-v1
+  destination_canonical_revision: version-1
+  destination_version: v1
+  protocol: http
+  log_sampled: "true"
+  upstream_cluster: "outbound|9080|http|server.default.svc.cluster.local"
```
mandarjog, line 22: why does this entry not mention any trace related information?

extensions/stackdriver/stackdriver.cc
gargnupur: should we just  add `disable_client_access_logging` ? changing `disable_server_access_logging` might cause compatibility issues?

gargnupur:
> Do we want a way to control whether or not we want both server and client access logs in config?

@douglas-reid : I think it we be good to have a separate configs for both? @bianpengyuan , @mandarjog : wdyt?

mandarjog:
Per discussion
1.  separate config options
2. `bool always_log: default=false`
3.  `bool skip_if_server_logs default=true`

or just one bool?

gargnupur:
> Per discussion
> 
> 1. separate config options
> 2. `bool always_log: default=false`
> 3. `bool skip_if_server_logs default=true`
> 
> or just one bool?

Need 2 in my mind.. 
so we will have following access logs configs:
1. disable_server_access_logging -> default = true
2. enable_client_access_logging -> default = false
3. skip_client_log_if_server_logs -> default = true (=> logs when MX fails)
/cc @mandarjog , @bianpengyuan , @douglas-reid , @kyessenov

extensions/stackdriver/log/logger.cc
```diff
@@ -33,53 +33,22 @@ namespace Log {
...
-  Common::getMonitoredResource(resource_type, local_node_info,
-                               &monitored_resource);
-  log_entries_request_->mutable_resource()->CopyFrom(monitored_resource);
-
-  // Set common labels shared by all entries.
-  auto label_map = log_entries_request_->mutable_labels();
-  (*label_map)["destination_name"] =
-      flatbuffers::GetString(local_node_info.name());
+void Logger::fillDestinationLabels(
+    const ::Wasm::Common::FlatNode& node_info,
```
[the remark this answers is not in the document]
gargnupur: Fixed

extensions/stackdriver/log/logger.cc
```diff
@@ -105,73 +74,179 @@ Logger::Logger(const ::Wasm::Common::FlatNode& local_node_info,
           flatbuffers::GetString(rev_iter->value());
     }
   }
+}
+
+void Logger::fillSourceLabels(
+    const ::Wasm::Common::FlatNode& node_info,
```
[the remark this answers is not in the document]
gargnupur: Fixed

extensions/stackdriver/stackdriver.cc
gargnupur: Added `disable_client_access_logging`

testdata/stackdriver/client_access_log_entry_sampled.yaml.tmpl
gargnupur: We are not sending any request with trace headers... will add a test for that in a separate pr

gargnupur:
@bianpengyuan, @douglas-reid , @kyessenov, @mandarjog   : This is ready for review..

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -71,4 +71,11 @@ message PluginConfig {
 
   // Optional. Allows enabling log compression for stackdriver access logs.
   google.protobuf.BoolValue enable_log_compression = 9;
+
+  // Optional. Controls whether to export client access log.
+  bool disable_client_access_logging = 10;
```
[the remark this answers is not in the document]
gargnupur: I like the idea as it saves from problems of multiple combinations of booleans... we would just have to match with configuring it in istio via values properly..

how about this?
`
enum AccessLogging {
  NONE = 0;
  FULL_SERVER_ONLY = 1;
  FULL_SERVER_AND_CLIENT_ERRORS_ONLY = 2;
  FULL_SERVER_AND_FULL_CLIENT = 3;
  CLIENT_ERRORS_ONLY = 4;
  FULL_CLIENT_ONLY = 5;
}

AccessLogging access_logging = 10;
`

just to clarify enum option 2 above?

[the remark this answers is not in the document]
bianpengyuan: FULL seems to be confusing to me. It makes sense to me that SERVER_ONLY means log all request.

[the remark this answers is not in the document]
gargnupur: PTAL now..

gargnupur:
@bianpengyuan, @douglas-reid , @kyessenov, @mandarjog : This is ready for review..

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -27,10 +27,22 @@ package stackdriver.config.v1alpha1;
...
 message PluginConfig {
+  // Types of Access logs to export.
+  enum AccessLogging {
+    // No Logs.
+    None = 0;
+    // All logs only including both success and error logs.
+    FULL = 1;
+    // All error logs when no metadata is available of the peer. This is
+    // currently only available for outbound/client side logs.
+    ERR_ONLY_ON_NO_MX = 2;
```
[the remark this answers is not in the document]
bianpengyuan: If offloading to log sample filter is the agreed approach, do we need to make this an enum? Shall we keep it as boolean (or a BoolValue if `disable` field name is undesired) and rename it to something without `server` in it, like `enable_access_log`?

extensions/stackdriver/log/logger.h
```diff
@@ -74,18 +88,55 @@ class Logger {
...
 
-  // Request that the new log entry should be written into.
-  std::unique_ptr<google::logging::v2::WriteLogEntriesRequest>
-      log_entries_request_;
-
-  // Estimated size of the current WriteLogEntriesRequest.
-  int size_ = 0;
+  // Stores client/server requests that the new log entry should be written
+  // into.
+  std::unordered_map<Logger::LogEntryType,
```
bianpengyuan, line 137: Is it necessary to make a map for only two possible keys? Feel like this is simpler to just have two member variables.

extensions/stackdriver/stackdriver.cc
```diff
@@ -461,8 +464,28 @@ inline bool StackdriverRootContext::isOutbound() {
...
 }
 
+inline bool StackdriverRootContext::enableAccessLog() {
+  return enableServerAccessLog() || enableClientAccessLog() ||
+         enableClientAccessLogOnNoMxAndError();
+}
+
 inline bool StackdriverRootContext::enableServerAccessLog() {
-  return !config_.disable_server_access_logging() && !isOutbound();
+  return (!config_.disable_server_access_logging() ||
```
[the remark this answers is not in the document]
gargnupur: The direction will matter if suppose we want to enable only server logs.. but stackdriver filter will be added in both directions for metrics...

extensions/stackdriver/log/logger.h
gargnupur: it's not just 2 variables.. it will be 4.. request and size.. and then if we add audit logs to the mix it would become 6...

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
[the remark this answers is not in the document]
gargnupur: Offloading to sample filter would involve change in sampling filter and this option was added to fasttrack client side logs.. so that we don't enable client side and server logs by default causing increase in logs cost.

I totally agree with offloading the decision to log to sampling filter rather than stackdriver filter. In case of http,  we can do on presence of header and for tcp on presence of filter state..

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
bianpengyuan: But we are going to have different configurations for inbound and outbound filter, right? Each logger/root context should be either inbound or outbound, so in terms of enablement, I think it does not need to know about directions.

extensions/stackdriver/log/logger.h
bianpengyuan: Yeah, it makes sense to me to wrap the request and size, but I mean the possible entries for the map is just inbound or outbound. Seems unnecessary to make a map just for two known keys, two member variables would be a better fit here I think. Even it grows to 3 types later, map still seems to be an unnecessary overhead.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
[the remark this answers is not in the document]
bianpengyuan: Judging whether to log based on presence of mx data seems to be the ultimate way of reducing log volume and not sure if that is really necessary for initial release. I think only logging errors (which log sample filter already supports) already seems to be a good budget choice. IMHO we should start with this simple option and avoid complexity. @mandarjog @kyessenov @douglas-reid wdyt?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -27,10 +27,22 @@ package stackdriver.config.v1alpha1;
...
 import "google/protobuf/wrappers.proto";
 
-// next id: 10
+// next id: 12
 message PluginConfig {
+  // Types of Access logs to export.
+  enum AccessLogging {
+    // No Logs.
+    None = 0;
+    // All logs only including both success and error logs.
```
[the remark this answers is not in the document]
gargnupur: Fixed

extensions/stackdriver/log/logger.h
gargnupur: Well it makes code cleaner, otherwise we will just have too many if else's...checkout the flush methods itself..

This is the current state:
void Logger::flush(Logger::LogEntryType log_entry_type) {
  std::unique_ptr<Logger::WriteLogEntryRequest> request;
  if(log_entry_type == LogEntryType::Client){
    request = client_log_entry_request_;
  } else if(log_entry_type == LogEntryType::Server){
    request = server_log_entry_request_;
  } else {
    request = audit_log_entry_request_;
  }
  std::unique_ptr<google::logging::v2::WriteLogEntriesRequest> cur =
      std::make_unique<google::logging::v2::WriteLogEntriesRequest>();
  cur->set_log_name(request->log_name());
  cur->mutable_resource()->CopyFrom(request->resource());
  *cur->mutable_labels() = request->labels();

  // Swap the new request with the old one and export it.
  request.swap(cur);
  request_queue_.emplace_back(std::move(cur));

  // Reset size counter.
  if(log_entry_type == LogEntryType::Client){
    client_log_entry_request_size_ = 0;
  } else if(log_entry_type == LogEntryType::Server){
    server_log_entry_request_size_ = 0;
  } else {
    audit_log_entry_request_size_ = 0;
  }
}

without map this would look like this:
void Logger::flush(Logger::LogEntryType log_entry_type) {
  std::unique_ptr<Logger::WriteLogEntryRequest> request;
  if(log_entry_type == LogEntryType::Client){
    request = client_log_entry_request_;
  } else if(log_entry_type == LogEntryType::Server){
    request = server_log_entry_request_;
  } else {
    request = audit_log_entry_request_;
  }
  std::unique_ptr<google::logging::v2::WriteLogEntriesRequest> cur =
      std::make_unique<google::logging::v2::WriteLogEntriesRequest>();
  cur->set_log_name(request->log_name());
  cur->mutable_resource()->CopyFrom(request->resource());
  *cur->mutable_labels() = request->labels();

  // Swap the new request with the old one and export it.
  request.swap(cur);
  request_queue_.emplace_back(std::move(cur));

  // Reset size counter.
  if(log_entry_type == LogEntryType::Client){
    client_log_entry_request_size_ = 0;
  } else if(log_entry_type == LogEntryType::Server){
    server_log_entry_request_size_ = 0;
  } else {
    audit_log_entry_request_size_ = 0;
  }
}

The logic looks more clearer as there are going to be more than 2 types of logs and it's less prone to error...

bianpengyuan: These two blocks seem to be identical, but yeah if this makes code simple let's do this way.

extensions/stackdriver/log/logger.h
```diff
@@ -74,18 +88,55 @@ class Logger {
...
+      const ::Wasm::Common::FlatNode& node_info,
+      google::protobuf::Map<std::string, std::string>* label_map);
+
+  // Helper methods to fill source Labels.
+  void fillSourceLabels(
+      const ::Wasm::Common::FlatNode& node_info,
+      google::protobuf::Map<std::string, std::string>* label_map);
+
+  // Helper method to set monitored resource.
+  void setMonitoredResource(
```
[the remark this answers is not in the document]
gargnupur: You know started with anonymous but somehow moved to logger.. thanks, changed back to anonymous!
`GetLogEntryType` can't be moved as it's accessing private member of Logger class

extensions/stackdriver/log/logger.cc
```diff
@@ -120,70 +101,176 @@ Logger::Logger(const ::Wasm::Common::FlatNode& local_node_info,
...
+    const ::Wasm::Common::FlatNode& local_node_info,
+    const std::string& resource_type,
+    google::logging::v2::WriteLogEntriesRequest* log_entries_request) {
+  google::api::MonitoredResource monitored_resource;
+  Common::getMonitoredResource(resource_type, local_node_info,
+                               &monitored_resource);
+  log_entries_request->mutable_resource()->CopyFrom(monitored_resource);
+}
+
+void Logger::initializeInboundLogEntryRequest(
```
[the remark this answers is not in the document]
gargnupur: Done..

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
gargnupur: Agreed.. that's why I didn't have it in my initial pr... currently it's needed for errors only case as we are enabling it for client only.. I will refactor this part once we make a decision on configs...

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -27,10 +27,22 @@ package stackdriver.config.v1alpha1;
 import "google/protobuf/duration.proto";
 import "google/protobuf/wrappers.proto";
 
-// next id: 10
+// next id: 12
 message PluginConfig {
+  // Types of Access logs to export.
+  enum AccessLogging {
+    // No Logs.
+    None = 0;
```
douglas-reid, line 35: nit: caps ?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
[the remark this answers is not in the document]
douglas-reid: so:  { `NONE`, `ALL`, `ERR_ONLY`} ?  Seems fine to me...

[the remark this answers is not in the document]
gargnupur: Updated based on offline conversation. Changed to Err_Only to log on errors.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: Thanks..Fixed

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
gargnupur: Fixed...

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -27,10 +27,22 @@ package stackdriver.config.v1alpha1;
...
 message PluginConfig {
+  // Types of Access logs to export.
+  enum AccessLogging {
+    // No Logs.
+    NONE = 0;
+    // All logs including both success and error logs.
+    FULL = 1;
+    // All error logs. This is
+    // currently only available for outbound/client side logs.
+    ERR_ONLY = 2;
```
douglas-reid, line 40: super nit: slight preference for fully spelling this out `ERRORS_ONLY`.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -27,10 +27,22 @@ package stackdriver.config.v1alpha1;
...
+// next id: 12
 message PluginConfig {
+  // Types of Access logs to export.
+  enum AccessLogging {
+    // No Logs.
+    NONE = 0;
+    // All logs including both success and error logs.
+    FULL = 1;
+    // All error logs. This is
+    // currently only available for outbound/client side logs.
```
douglas-reid, line 39: nit: can we add the detail from the discussion around response flags and codes here too?

gargnupur: Done

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: Sure..fixed!

extensions/stackdriver/log/logger.cc
```diff
@@ -120,70 +105,149 @@ Logger::Logger(const ::Wasm::Common::FlatNode& local_node_info,
...
+  log_entries_request_map_[log_entry_type]->request =
+      std::make_unique<google::logging::v2::WriteLogEntriesRequest>();
+  log_entries_request_map_[log_entry_type]->size = 0;
+  auto log_entries_request =
+      log_entries_request_map_[log_entry_type]->request.get();
+  std::string log_name = outbound ? kClientAccessLogName : kServerAccessLogName;
+  log_entries_request->set_log_name("projects/" + project_id_ + "/logs/" +
+                                    log_name);
+
+  std::string resource_type = outbound ? Common::kPodMonitoredResource
```
bianpengyuan, line 167: nit: const std::string&

extensions/stackdriver/log/logger.cc
```diff
@@ -120,70 +105,149 @@ Logger::Logger(const ::Wasm::Common::FlatNode& local_node_info,
...
+    const flatbuffers::Vector<flatbuffers::Offset<Wasm::Common::KeyVal>>*
+        platform_metadata,
+    const ::Wasm::Common::FlatNode& local_node_info, bool outbound) {
+  auto log_entry_type = GetLogEntryType(outbound);
+  log_entries_request_map_[log_entry_type]->request =
+      std::make_unique<google::logging::v2::WriteLogEntriesRequest>();
+  log_entries_request_map_[log_entry_type]->size = 0;
+  auto log_entries_request =
+      log_entries_request_map_[log_entry_type]->request.get();
+  std::string log_name = outbound ? kClientAccessLogName : kServerAccessLogName;
```
bianpengyuan, line 163: nit: const std::string&

extensions/stackdriver/stackdriver.cc
```diff
@@ -364,9 +363,13 @@ void StackdriverRootContext::record() {
   ::Extensions::Stackdriver::Metric::record(
       outbound, local_node, peer_node, request_info,
       !config_.disable_http_size_metrics());
-  if (enableServerAccessLog() && shouldLogThisRequest(request_info)) {
+  if ((enableAllAccessLog() ||
+       (enableClientAccessLogOnError() &&
```
bianpengyuan, line 367: I still don't think we need to distinguish client or not. This can just be enableAccessLogOnError. Direction is already distinguished at filter configuration.

extensions/stackdriver/log/logger.h
```diff
@@ -44,26 +44,40 @@ class Logger {
...
+  };
+
+  // Type of log Entry.
+  enum LogEntryType { Client, Server };
+
   // Flush rotates the current WriteLogEntriesRequest. This will be triggered
   // either by a timer or by request size limit. Returns false if there is no
   // log entry to be exported.
   bool flush();
+  void flush(Logger::LogEntryType log_entry_type);
```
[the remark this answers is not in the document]
gargnupur: There is no status to return there...it's more of a helper method..

extensions/stackdriver/log/logger.cc
gargnupur: Thanks fixed!

extensions/stackdriver/log/logger.cc
gargnupur: Thanks Fixed!

extensions/stackdriver/stackdriver.cc
gargnupur: We are saying that we enable access log on error on client side only... thus the distinction...
I think the code will work for server side too if we remove the distinction.. would be a good add to server side logging... @mandarjog, @douglas-reid  wdyt?

extensions/stackdriver/log/logger.cc
gargnupur: Cant change to const ref as we assigning resource_type based on cluster name key below...

extensions/stackdriver/stackdriver.cc
mandarjog: I think “logOnError” will suffice. 
We will not set it in server side logs config for now, but will work just as well.

gargnupur: Done!

PR #2956, opened 2020-08-02
Update Envoy-WASM SHA to latest.

PiotrSikora:


PiotrSikora:
@lizan @yangminzhu Any ideas what might be broken? Both `//src/envoy/http/authn:authn_utils_test` and `//src/envoy/tcp/sni_verifier:sni_verifier_test` are crashing on `nullptr` in `string_view`.

src/envoy/tcp/sni_verifier/sni_verifier.cc
```diff
@@ -47,8 +47,8 @@ Config::Config(Stats::Scope& scope, size_t max_client_hello_size)
         Filter* filter = static_cast<Filter*>(SSL_get_app_data(ssl));
 
         if (filter != nullptr) {
-          filter->onServername(
-              SSL_get_servername(ssl, TLSEXT_NAMETYPE_host_name));
+          filter->onServername(absl::NullSafeStringView(
```
kyessenov, line 50: What is this for?

PiotrSikora: `{absl,std}::string_view(nullptr)` is not a valid thing in C++17 and results in segfault, and `SSL_get_servername()` returns `nullptr` when SNI was not present in the TLS handshake.

kyessenov: Hm, this makes me concerned that changing from absl to std is safe.

PiotrSikora: The change to `std::string_view` only happened in `//extensions`, to match Proxy-Wasm C++ SDK.

This code here is still using `absl::string_view`, so what you're concerned with is update from C++14 to C++17, but that already happened in Envoy. Basically, `absl::string_view` uses internal implementation when compiled with C++14, but it's just an alias for `std::string_view` when compiled with C++17.

lizan: @kyessenov `std::string_view` isn't safe with nullptr in the C++ standard, some implementation (libstdc++ or absl in c++14 compat mode) allows them while we shouldn't rely on them.

PiotrSikora: @kyessenov could you approve? It's not like we can force Envoy back to C++14.

PiotrSikora: Ping.

PR #2961, opened 2020-08-04
Set TextPayload for TCP Access Log 

gargnupur:
Set TextPayload for TCP Access Log  and fix source_port showing up as 0 in TCP logs

**Release note**:

```release-note
```

### DIFF
--- a/extensions/common/context.h
+++ b/extensions/common/context.h
@@ -105,10 +105,10 @@ struct RequestInfo {
   int64_t response_size = 0;
 
   // Destination port that the request targets.
-  uint32_t destination_port = 0;
+  uint64_t destination_port = 0;
 
   // Source port of the client.
-  uint32_t source_port = 0;
+  uint64_t source_port = 0;
 
   // Protocol used the request (HTTP/1.1, gRPC, etc).
   std::string request_protocol;
--- a/extensions/common/context.h
+++ b/extensions/common/context.h
@@ -105,7 +105,7 @@ struct RequestInfo {
   int64_t response_size = 0;
 
   // Destination port that the request targets.
-  uint64_t destination_port = 0;
+  uint32_t destination_port = 0;
 
   // Source port of the client.
   uint64_t source_port = 0;
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -30,6 +30,21 @@
 namespace Extensions {
 namespace Stackdriver {
 namespace Log {
+namespace {
+void setSourceCanonicalService(
+    const ::Wasm::Common::FlatNode& peer_node_info,
+    google::protobuf::Map<std::string, std::string>* label_map) {
+  const auto peer_labels = peer_node_info.labels();
+  if (peer_labels) {
+    auto ics_iter = peer_labels->LookupByKey(
+        Wasm::Common::kCanonicalServiceLabelName.data());
+    if (ics_iter) {
+      (*label_map)["source_canonical_service"] =
+          flatbuffers::GetString(ics_iter->value());
+    }
+  }
+}
+}  // namespace
 
 using google::protobuf::util::TimeUtil;
 
@@ -119,7 +134,7 @@ void Logger::addTcpLogEntry(const ::Wasm::Common::RequestInfo& request_info,
   *new_entry->mutable_timestamp() =
       google::protobuf::util::TimeUtil::NanosecondsToTimestamp(log_time);
 
-  addTCPLabelsToLogEntry(request_info, new_entry);
+  addTCPLabelsToLogEntry(request_info, peer_node_info, new_entry);
   fillAndFlushLogEntry(request_info, peer_node_info, new_entry);
 }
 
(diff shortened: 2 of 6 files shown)


extensions/stackdriver/log/logger.cc
```diff
@@ -160,11 +175,13 @@ void Logger::fillAndFlushLogEntry(
     if (app_iter) {
       (*label_map)["source_app"] = flatbuffers::GetString(app_iter->value());
     }
-    auto ics_iter = peer_labels->LookupByKey(
-        Wasm::Common::kCanonicalServiceLabelName.data());
-    if (ics_iter) {
-      (*label_map)["source_canonical_service"] =
-          flatbuffers::GetString(ics_iter->value());
+    if (label_map->find("source_canonical_service") == label_map->end()) {
```
douglas-reid, line 178: why not `setSourceCanonicalService()` ?

extensions/stackdriver/log/logger.cc
```diff
@@ -243,8 +260,16 @@ bool Logger::exportLogEntry(bool is_on_done) {
...
     const ::Wasm::Common::RequestInfo& request_info,
+    const ::Wasm::Common::FlatNode& peer_node_info,
     google::logging::v2::LogEntry* log_entry) {
   auto label_map = log_entry->mutable_labels();
+  setSourceCanonicalService(peer_node_info, label_map);
+  log_entry->set_text_payload(absl::StrCat(
+      label_map->find("source_canonical_service") != label_map->end()
+          ? (*label_map)["source_canonical_service"]
+          : flatbuffers::GetString(peer_node_info.workload_name()),
+      "-->",
```
douglas-reid, line 271: curious: was this specific text format requested by any of the users?  should we include spaces between the names and the `-->` ?

extensions/stackdriver/log/logger.cc
gargnupur: Thanks,, will fix..

extensions/stackdriver/log/logger.cc
gargnupur: no, we are adding this because otherwise tcp connection logs looks empty in the first look in stackdriver log viewer.

sure, will include spaces..

gargnupur: Also, added a condition for destination in payload that if destination_canonical_service is not found, change to destination_service_name.

extensions/stackdriver/log/logger.cc
```diff
@@ -243,8 +255,19 @@ bool Logger::exportLogEntry(bool is_on_done) {
 
 void Logger::addTCPLabelsToLogEntry(
     const ::Wasm::Common::RequestInfo& request_info,
+    const ::Wasm::Common::FlatNode& peer_node_info,
     google::logging::v2::LogEntry* log_entry) {
   auto label_map = log_entry->mutable_labels();
+  setSourceCanonicalService(peer_node_info, label_map);
+  log_entry->set_text_payload(absl::StrCat(
+      label_map->find("source_canonical_service") != label_map->end()
```
bianpengyuan, line 263: This looks up map two times. Better get iter beforehand and look up map for only one time.

extensions/common/context.h
```diff
@@ -105,10 +105,10 @@ struct RequestInfo {
   int64_t response_size = 0;
 
   // Destination port that the request targets.
-  uint32_t destination_port = 0;
+  uint64_t destination_port = 0;
```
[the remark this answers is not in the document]
gargnupur: just keeping it consistent with what is it's exact type from ABI...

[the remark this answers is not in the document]
gargnupur: actually changed it back.. don't want any surprises late in the cycle..

extensions/stackdriver/log/logger.cc
gargnupur: Thanks..Fixed

extensions/common/context.h
```diff
@@ -108,7 +108,7 @@ struct RequestInfo {
   uint32_t destination_port = 0;
 
   // Source port of the client.
-  uint32_t source_port = 0;
+  uint64_t source_port = 0;
```
bianpengyuan, line 111:
```suggestion
  uint32_t source_port = 0;
```

bianpengyuan: This is just to keep it consistent with destination_port

bianpengyuan:
Feel free to unblock after taking a look at the one last suggestion.

extensions/common/context.h
gargnupur: No this is needed as we get back uint64 from ABI and if we cast to uint32.. it comes up as 0..

bianpengyuan: Not sure why it comes up as 0, but if that is the case, could you also change destination_port to use uint64_t? If it is going to cause problem to source_port, it would cause the same problem to destination port, wouldn't it?

gargnupur: It doesn't cause problem for destination port because when we get it, we get it as uint64.. https://github.com/istio/proxy/blob/master/extensions/common/context.cc#L151:12 and then cast it as uint32.. thus I had changed it to uint64 too.. but hadn't tested it so changed it back.. 
we should make it consistent and will change to uint64 in a different pr..

PR #2965, opened 2020-08-05
Use smaller Wasm heap and stack

mandarjog:
This is working fine now. Ok to proceed.

Reduced observed size in a full namespace servicegraph experiment from 150MB to 125MB per proxy.
So the reduction was not on par with 16MB --> 2BM x number of threads.

### DIFF
--- a/extensions/BUILD
+++ b/extensions/BUILD
@@ -26,6 +26,12 @@ load(
 
 envoy_package()
 
+WASM_LINKOPTS = [
+    "--js-library external/proxy_wasm_cpp_sdk/proxy_wasm_intrinsics.js",
+    "-s TOTAL_MEMORY=2MB",
+    "-s TOTAL_STACK=1MB",
+]
+
 wasm_cc_binary(
     name = "stats.wasm",
     srcs = [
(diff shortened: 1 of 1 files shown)


PiotrSikora:
Any reason why this wasn't merged, @mandarjog?

kyessenov:
We do have a recursive descent parser. Probably need to check if stack is big enough.

mandarjog:
Not sure why it was not merged.
@kyessenov with recursive descent parser, what question are you trying to answer ? Just stack depth or how much each stack frame uses?

kyessenov:
I just don't know where these numbers came from? We don't have any tests with large configs, so it might be that stack/heap overflow for them.

PiotrSikora:
Could we move this forward? Using 16MB is quite expensive.

By default, Emscripten uses `INITIAL_MEMORY=16MB` (5MB for stack + 11MB for heap).

At the very least, we could shrink it to 8MB (5MB for stack + 3MB for heap) and allow the memory to grow on demand. This shouldn't have any downsides (stack is the same), other than possible reallocation if the heap grows above the initial 3MB.

mandarjog:
closing in favor of https://github.com/istio/proxy/pull/3222

PR #2970, opened 2020-08-06
Add audit log stream to stackdriver filter

davidraskin:
This adds an audit log stream to stackdriver. The goal would be to have requests to be auditted will be filtered through the audit action on Istio Authorization Policy. 
See: https://github.com/istio/api/pull/1552, [Audit log stream proposal](https://docs.google.com/document/d/1I3Zu7qImw-CK1_BzxWcDFpB9rAbuy6DBQhejpbAVUGo/edit#)

cc @liminw @gargnupur @kyessenov @mandarjog 

This is WIP. It should only be merged after https://github.com/envoyproxy/envoy/pull/11705


extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -27,11 +27,15 @@ package stackdriver.config.v1alpha1;
...
-// next id: 10
+// next id: 11
 message PluginConfig {
-  // Optional. Controls whether to export server access log.
+  // Optional. Controls whether to export server access log. This does not
+  // affect audit logging
   bool disable_server_access_logging = 1;
 
+  // Optional. Controls whether to export audit log.
+  bool enable_server_access_auditting = 10;
```
douglas-reid, line 37: nit: can we name this `audit_server_access` ?

extensions/stackdriver/log/logger.cc
```diff
@@ -50,13 +50,17 @@ using google::protobuf::util::TimeUtil;
 
 // Name of the HTTP server access log.
 constexpr char kServerAccessLogName[] = "server-accesslog-stackdriver";
+constexpr char kAuditAccessLogName[] = "server-accesslog-stackdriver-audit";
```
douglas-reid, line 53: is this naming convention important?  does it need to match the historic name for the other access logs, or should we consider calling this `server-audit-log` (or something else less verbose)?

davidraskin: I don't think that's important. I just went the simplest route. I'll rename it.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
davidraskin: Sure.

bianpengyuan:
This might be a dumb question, but isn't it that request should not proceed until audit log succeed? Looks like the current implementation is still logging at Log callback which happens after request finishes.

davidraskin:
The request is not exported if there are no audit log entries added to it. This works the same way as with regular log entries currently.

extensions/stackdriver/log/logger.cc
```diff
@@ -50,13 +50,17 @@ using google::protobuf::util::TimeUtil;
...
 
 Logger::Logger(const ::Wasm::Common::FlatNode& local_node_info,
                std::unique_ptr<Exporter> exporter, int log_request_size_limit) {
   // Initalize the current WriteLogEntriesRequest.
   log_entries_request_ =
       std::make_unique<google::logging::v2::WriteLogEntriesRequest>();
 
+  audit_entries_request_ =
```
gargnupur, line 61: check this out... I have made it a map: https://github.com/istio/proxy/pull/2955/files#diff-d116ca95f1292dece4d352210d220a13 so would be easier to add many types of entries now...

bianpengyuan:
@davidraskin Yeah, but current access log implementation is for telemetry and at best effort. The exporting could have transient failure which left some requests out of log. Is this fine for audit log? I thought the assumption of audit log is that every request reaches application must be logged.

davidraskin:
@bianpengyuan I see what your saying. I think it's still important to see the result of the request, and whether it succeeds or fails. But maybe it should fail closed rather than open.

extensions/stackdriver/log/logger.cc
bianpengyuan, line 53: Out of curiosity, have you thought about fill in `protoPayload` in logEntry, which supports `google.cloud.audit.AuditLog`: https://cloud.google.com/logging/docs/reference/audit/auditlog/rest/Shared.Types/AuditLog. Sound like that is a more standard way to do audit log at Google?

davidraskin: Yes, the AuditLog payload is used primarily for auditting GCP service calls, it doesn't apply to application logging, afaik.

mandarjog: @bianpengyuan @davidraskin can you find out if `google.cloud.audit.AuditLog` **cannot** be used to audit other services?

davidraskin: @mandarjog Maybe it can theoretically be used, because I think the only difference is the proto payload, but the [Audit Log docs](https://cloud.google.com/logging/docs/reference/audit/auditlog/rest/Shared.Types/AuditLog#AuthenticationInfo) state that it is for GCP API calls. Additionally, [configuring the Data Access Audit Logs](https://cloud.google.com/logging/docs/audit/configure-data-access) is done through the Cloud Console, which could be confusing since that wouldn't be the case here.

gargnupur: Should these logs be actually called audit logs? https://cloud.google.com/logging/docs/audit cloud audit logs looks like are logs for telling who or what accessed a GCP resource... This is more about capturing auth info for application?

davidraskin: They're still audit logs I think. Just ones at the application level, separate from GCP audit logs. We can call them IstioAudit logs to reduce confusion.

extensions/common/context.h
```diff
@@ -235,5 +235,9 @@ void populateExtendedRequestInfo(RequestInfo* request_info);
 void populateTCPRequestInfo(bool outbound, RequestInfo* request_info,
                             const std::string& destination_namespace);
 
+// Read value of 'access_log_hint' key in envoy dynamic metadata which
```
gargnupur, line 238:
```suggestion
// Read value of 'access_log_hint' key from envoy dynamic metadata which
```

extensions/common/context.h
```diff
@@ -235,5 +235,9 @@ void populateExtendedRequestInfo(RequestInfo* request_info);
...
+// determines whether to audit a request
```
gargnupur, line 239:
```suggestion
// determines whether to audit a request or not.
```

mandarjog:
lgtm.

@bianpengyuan @gargnupur please approve if your comments are addressed.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -45,6 +45,9 @@ message PluginConfig {
   // This is deprecated in favor of AccessLogging enum.
   bool disable_server_access_logging = 1 [deprecated = true];
 
+  // Optional. Controls whether to export audit log.
+  bool enable_audit_log = 11;
```
gargnupur, line 49: Would we have both Full access logs and audit logs at the same time? Because otherwise we can put this in accesslogging enum?

@mandarjog , @bianpengyuan , @douglas-reid

extensions/stackdriver/log/logger.cc
```diff
@@ -195,80 +214,96 @@ Logger::Logger(const ::Wasm::Common::FlatNode& local_node_info,
...
 
   log_request_size_limit_ = log_request_size_limit;
   exporter_ = std::move(exporter);
 }
 
 void Logger::addTcpLogEntry(const ::Wasm::Common::RequestInfo& request_info,
                             const ::Wasm::Common::FlatNode& peer_node_info,
-                            long int log_time, bool outbound) {
+                            long int log_time, bool outbound, bool audit) {
+  LogEntryType log_entry_type = GetLogEntryType(outbound, audit);
```
gargnupur, line 241: why create a new variable? I don't think this is getting re-used?

extensions/stackdriver/log/logger.cc
```diff
@@ -331,43 +366,36 @@ bool Logger::exportLogEntry(bool is_on_done) {
 
 void Logger::addTCPLabelsToLogEntry(
     const ::Wasm::Common::RequestInfo& request_info,
-    const ::Wasm::Common::FlatNode& peer_node_info, bool outbound,
-    google::logging::v2::LogEntry* log_entry) {
+    const ::Wasm::Common::FlatNode& peer_node_info,
+    google::logging::v2::LogEntry* log_entry, bool outbound, bool audit) {
+  LogEntryType log_entry_type = GetLogEntryType(outbound, audit);
+  auto& entries_request = log_entries_request_map_[log_entry_type]->request;
```
gargnupur, line 372: const?

extensions/stackdriver/log/logger.cc
```diff
@@ -331,43 +366,36 @@ bool Logger::exportLogEntry(bool is_on_done) {
...
     auto destination_cs_iter = label_map->find("destination_canonical_service");
-    source =
-        source_cs_iter != log_entries_request_map_[log_entry_type]
-                              ->request->labels()
-                              .end()
-            ? source_cs_iter->second
-            : log_entries_request_map_[log_entry_type]->request->labels().at(
-                  "source_workload");
+    source = source_cs_iter != log_entries_request_map_[log_entry_type]
+                                   ->request->labels()
```
gargnupur, line 381: entries_request?

extensions/stackdriver/log/logger.cc
```diff
@@ -403,7 +431,9 @@ void Logger::fillHTTPRequestInLogEntry(
           request_info.duration);
   http_request->set_referer(request_info.referer);
   auto label_map = log_entry->mutable_labels();
-  (*label_map)["request_id"] = request_info.request_id;
+  if (!audit) {
+    (*label_map)["request_id"] = request_info.request_id;
```
gargnupur, line 435: request id is good to have as it helps correlate problems with application logs...

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
davidraskin: This would require four additions to the enum e.g. Audit_Full, Audit_Errors_Only, All_Errors_Only, and All. Or we could also have a separate enum for log type with values Regular, Audit, and All.

extensions/stackdriver/log/logger.cc
davidraskin: Sure we can include it then.

extensions/stackdriver/log/logger.cc
davidraskin: Do you mean const on log_entry_type?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: I think just 2 should be good? 
AUDIT_FULL and AUDIT_ERRORS_ONLY ?

And user will select just one of the 4 options..

extensions/stackdriver/stackdriver.h
```diff
@@ -119,6 +119,11 @@ class StackdriverRootContext : public RootContext {
 
   bool shouldLogThisRequest(::Wasm::Common::RequestInfo& request_info);
 
+  // Indicates whether to export server audit log or not.
+  bool enableAuditLog();
+
+  bool shouldAuditThisRequest();
```
gargnupur, line 125: can you add comment for this function too?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: Talked to David offline.. there is a possibility of having regular and audit logs both based on permissions... so having a bool is better then making this enum complicated I feel..

extensions/stackdriver/stackdriver.h
davidraskin: Yep.

extensions/stackdriver/stackdriver.cc
```diff
@@ -25,6 +25,7 @@
 #include "extensions/stackdriver/edges/mesh_edges_service_client.h"
 #include "extensions/stackdriver/log/exporter.h"
 #include "extensions/stackdriver/metric/registry.h"
+#include "stackdriver.h"
```
bianpengyuan, line 28: not needed change?

extensions/stackdriver/log/logger.cc
```diff
@@ -73,14 +73,18 @@ void setMonitoredResource(
 // Helper methods to fill destination Labels.
 void fillDestinationLabels(
```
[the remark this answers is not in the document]
davidraskin: Done.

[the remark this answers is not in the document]
bianpengyuan: Can you add more details? What labels are qualified to be put in audit log? In the future, If I am going to add a new label, I need some guidance on whether that label should be in audit log or not.

extensions/stackdriver/stackdriver.cc
```diff
@@ -442,13 +452,35 @@ bool StackdriverRootContext::recordTCP(uint32_t id) {
...
       record_info.request_info->tcp_connection_state =
           ::Wasm::Common::TCPConnectionState::Close;
     }
     logger_->addTcpLogEntry(request_info, peer_node,
-                            getCurrentTimeNanoseconds(), outbound);
+                            getCurrentTimeNanoseconds(), outbound,
+                            false /* audit */);
   }
+
+  if (enableAuditLog() && shouldAuditThisRequest()) {
```
gargnupur, line 465: need to call ::Wasm::Common::populateExtendedRequestInfo(&request_info); here too?

extensions/stackdriver/log/logger.cc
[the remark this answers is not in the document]
gargnupur: I agree, I think it would also be good idea to add a separate function like fillAuditAndCommonDestinationLabel and fillDestinationLabels .. something on those lines..

extensions/stackdriver/log/logger.cc
```diff
@@ -150,18 +159,26 @@ constexpr char kServerAccessLogName[] = "server-accesslog-stackdriver";
...
+constexpr char kClientAuditLogName[] = "client-istio-audit-log";
+
 void Logger::initializeLogEntryRequest(
     const flatbuffers::Vector<flatbuffers::Offset<Wasm::Common::KeyVal>>*
         platform_metadata,
-    const ::Wasm::Common::FlatNode& local_node_info, bool outbound) {
-  auto log_entry_type = GetLogEntryType(outbound);
+    const ::Wasm::Common::FlatNode& local_node_info, bool outbound,
+    bool audit) {
+  const LogEntryType log_entry_type = GetLogEntryType(outbound, audit);
```
gargnupur, line 172: no need for const with enum

extensions/stackdriver/log/logger.cc
```diff
@@ -331,43 +366,34 @@ bool Logger::exportLogEntry(bool is_on_done) {
 
 void Logger::addTCPLabelsToLogEntry(
     const ::Wasm::Common::RequestInfo& request_info,
-    const ::Wasm::Common::FlatNode& peer_node_info, bool outbound,
-    google::logging::v2::LogEntry* log_entry) {
+    const ::Wasm::Common::FlatNode& peer_node_info,
+    google::logging::v2::LogEntry* log_entry, bool outbound, bool audit) {
+  auto& entries_request =
```
gargnupur, line 371: this can be const auto& i think...

gargnupur:
code looks ok, with some minor nits.. can you also add an integration test (https://github.com/istio/proxy/blob/master/test/envoye2e/stackdriver_plugin/stackdriver_test.go) ?

liminw:
> @davidraskin Yeah, but current access log implementation is for telemetry and at best effort. The exporting could have transient failure which would leave some requests not logged. Is this fine for audit log? I thought the assumption of audit log is that every request reaches application must be logged.

Peter, you are totally correct. Audit log has strong reliability requirement. This is the first step of enabling configurable audit feature. Going forward, we plan to improve the reliability of audit stream, which requires a separate design.

bianpengyuan:
@liminw Ok, thanks for clarification! Can you elaborate a bit on `the first step of enabling configurable audit feature`? It does not look right to me to check in code that is not going to used for production. It will likely just become tech debt so hope we can avoid that here.

davidraskin:
@bianpengyuan If request completion is the concern, we can just do the logging in the requestHeaders callback, and fail if there is some issue, and/or have an extra verification step to see that the log is received.

bianpengyuan:
> we can just do the logging in the requestHeaders callback

That would be ideal IMO. thanks!

liminw:
> @liminw Ok, thanks for clarification! Can you elaborate a bit on `the first step of enabling configurable audit feature`? It does not look right to me to check in code that is not going to used for production. It will likely just become tech debt so hope we can avoid that here.

I don't think we are close to claim audit feature production ready. As far as I know, Cloud audit team has done a lot of work to improve the reliability of audit stream. We still need to learn from them and design for it. But this can be done incrementally.

test/envoye2e/stackdriver_plugin/stackdriver_test.go
```diff
@@ -640,3 +640,78 @@ func TestStackdriverTCPMetadataExchange(t *testing.T) {
...
 }
+
+func TestStackdriverAuditLog(t *testing.T) {
+	t.Parallel()
+	var TestCases = []struct {
+		name          string
+		respCode      string
+		logEntryCount int
+	}{
+		{"StackdriverAudit", "200", 5},
```
gargnupur, line 651: As there is only one testcase, we don't need this?

PR #2978, opened 2020-08-08
authn: add revised peer authenticator

Shikugawa:
This is a part of replacement of authn wasm implementation. It is using revised peer authentication API. istio/api#1536
istio#2800 is too large and hard to review all codes so that I split all of authn wasm implementation.


zirain:
outdated

PR #2980, opened 2020-08-10
Update Envoy-WASM SHA to latest with C++14 fix.

PiotrSikora:

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -37,10 +37,10 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/istio/envoy/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
-# Commit date: 2020-08-07 - Branch: release-1.7
-ENVOY_SHA = "ac9fe44075bcce73befe9ba62f352b2351f7ce1d"
+# Commit date: 2020-08-10 - Branch: release-1.7
+ENVOY_SHA = "d13b204e4fb382836dd462805b2786638acb1dc6"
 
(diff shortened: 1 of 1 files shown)


kyessenov:
Builds fine locally. As a follow up we should investigate why CI passed the first time.

PiotrSikora:
> Builds fine locally. As a follow up we should investigate why CI passed the first time.

That's because `libc++` allows `std::string_view` in C++14, and we use `libc++`.

```
$ cat main.cc
```
```
#include <string>

int main() {
   std::string_view x;
   return 1;
}
```
```
$ clang++ -std=c++14 main.cc
```
```
main.cc:4:9: error: no type named 'string_view' in namespace 'std'
   std::string_view bla;
   ~~~~~^
1 error generated.
```
```
$ clang++ -std=c++14 -stdlib=libc++ main.cc
```
```
(OK)
```

kyessenov:
@PiotrSikora IMHO libc++ is not conformant to C++ spec, and we should not side on its peculiar behavior.

PiotrSikora:
> @PiotrSikora IMHO libc++ is not conformant to C++ spec, and we should not side on its peculiar behavior.

I'm not saying that we should, I'm just saying why the CI passed.

kyessenov:
OK. Should we build without libc++ in CI then? That will keep us honest to c++ standards.

PiotrSikora:
> OK. Should we build without libc++ in CI then? That will keep us honest to c++ standards.

We're shipping production binaries with `libc++` and that's unlikely to change.

If you want to add another CI target testing C++ conformance using `stdlibc++`, be my guest.

PiotrSikora:
@howardjohn I've been seeing some build timeouts recently. Could you check if we're swapping on the build machines?

PR #2981, opened 2020-08-14
Update altlr to 4.7.2

mandarjog:
Updates antlr to fix https://github.com/envoyproxy/envoy-wasm/issues/497

At that time only cel-cpp was updated, but not antlr.
We do not have a unit test that produces the crash.
I have verified that crash *does* happen with 1.7 build and it *does not* happen with build that includes antlr-4.7.2.


jwendell:
@mandarjog @PiotrSikora You are probably working on this upstream. Should this be considered a release blocker for 1.7.0?

jwendell:
Ah, I just saw you fixed it in istio/envoy. Nervermind. I opened https://github.com/istio/proxy/pull/2982

PR #2985, opened 2020-08-14
Create the VM context as an IstioContext for consistency.

jplevyak:


kyessenov:
/retest
Can you explain what this change actually does?

jplevyak:
In order to allow istio to specialize the Context object e.g. to extend it for istio specific ABI calls we have virtual functions down in proxy-wasm to create these objects.  While the VM context is not used typically for anything other than the _malloc call and so would not normally require any extensions which istio might make, it is inconsistent to have some contexts raw proxy_wasm::ContextBase instead of IstioContext.  This change corrects that inconsistency.

PR #2986, opened 2020-08-14
tests: fix parallel testing

kyessenov:
Templates were filled once instead of repeatedly for every iteration. So it kept the config identical and not exercising VM reloading. Fix by delaying templating.

Also, move out metadata exchange config. This will be used to load it dynamically just like others.


kyessenov:
Filed issue https://github.com/istio/istio/issues/26548 for the test fail.
/retest

PR #2991, opened 2020-08-19
Fix clang-10 compatibility: designated initializers are a C++20 exten…

jplevyak:
…sion.


jplevyak:
Note: envoy upstream is now using clang-10.  This is a new "warning" which we consider an error in clang-10.

jplevyak:
I agree that it is less readable.  I like that syntax and it is a pity that they made it a warning now.

PR #2993, opened 2020-08-28
Update stackdriver filter to fallback to default mesh id format when not specified

bianpengyuan:

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -183,9 +183,7 @@ void getLocalNodeMetadata(google::protobuf::Struct* node_metadata) {
   auto project_number = getProjectNumber();
   auto* mesh_id_field =
       (*node_metadata->mutable_fields())["MESH_ID"].mutable_string_value();
-  if (project_number.empty()) {
-    *mesh_id_field = "unknown";
-  } else {
+  if (!project_number.empty()) {
     *mesh_id_field = absl::StrCat("proj-", project_number);
   }
 }
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -176,7 +176,7 @@ void getLocalNodeMetadata(google::protobuf::Struct* node_metadata) {
   const auto mesh_id_it = node_metadata->fields().find("MESH_ID");
   if (mesh_id_it != node_metadata->fields().end() &&
       !mesh_id_it->second.string_value().empty() &&
-      absl::HasPrefix(mesh_id_it->second.string_value()], "proj-")) {
+      absl::StartsWith(mesh_id_it->second.string_value(), "proj-")) {
     return;
   }
 
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -151,10 +151,43 @@ std::string getMonitoringEndpoint() {
   return monitoring_endpoint;
 }
 
+// Get GCP project number.
+std::string getProjectNumber() {
+  std::string project_number;
+  if (!getValue({"node", "metadata", "PLATFORM_METADATA", kGCPProjectNumberKey},
+                &project_number)) {
+    return "";
+  }
+  return project_number;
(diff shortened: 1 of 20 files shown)


extensions/stackdriver/stackdriver.cc
```diff
@@ -151,12 +151,45 @@ std::string getMonitoringEndpoint() {
...
+      !mesh_id_it->second.string_value().empty()) {
+    return;
+  }
+
+  // Insert or update mesh id to default mesh id if empty.
+  auto project_number = getProjectNumber();
+  auto* mesh_id_field =
+      (*node_metadata->mutable_fields())["MESH_ID"].mutable_string_value();
+  if (project_number.empty()) {
+    *mesh_id_field = "unknown";
```
mandarjog, line 187: What is the difference between meshid = unknown and unset?

bianpengyuan: Actually not much, I think both edge and metric handle this case well (if empty, set it to unknown). For log, it currently leaves unknown fields empty, so this is going to be make a difference there.

bianpengyuan:
@elfinhe fyi.

extensions/stackdriver/stackdriver.cc
```diff
@@ -151,12 +151,43 @@ std::string getMonitoringEndpoint() {
...
+  if (!getMessageValue({"node", "metadata"}, node_metadata)) {
+    return;
+  }
+  const auto mesh_id_it = node_metadata->fields().find("MESH_ID");
+  if (mesh_id_it != node_metadata->fields().end() &&
+      !mesh_id_it->second.string_value().empty()) {
+    return;
+  }
+
+  // Insert or update mesh id to default mesh id if empty.
```
gargnupur, line 182: should the comment says "as it's empty"

with  "if empty", it feels it should check for it being empty below...

bianpengyuan: Done, thanks!

extensions/stackdriver/stackdriver.cc
```diff
@@ -151,12 +151,43 @@ std::string getMonitoringEndpoint() {
...
   request_info.tcp_received_bytes = 0;
 }
 
+// Get local node metadata. If mesh id is not filled or does not exist,
+// fall back to default format `proj-<project-number>`.
+void getLocalNodeMetadata(google::protobuf::Struct* node_metadata) {
+  if (!getMessageValue({"node", "metadata"}, node_metadata)) {
+    return;
+  }
+  const auto mesh_id_it = node_metadata->fields().find("MESH_ID");
```
elfinhe, line 176: Is there any default value for the MESH_ID here?

extensions/stackdriver/stackdriver.cc
```diff
@@ -151,12 +151,43 @@ std::string getMonitoringEndpoint() {
...
+  if (mesh_id_it != node_metadata->fields().end() &&
+      !mesh_id_it->second.string_value().empty()) {
```
elfinhe, line 178: Today the meshid will always fallback to trust domain "<projectname>.svc.id.goog", which cannot be recognized by stackdriver.

we need to add one more line, 

```
// Today GCP Monitoring use "proj-<ProjectNumber>" as mesh ID, reuse input if it is a valid mesh ID.
 if (mesh_id_it != node_metadata->fields().end() &&
      !mesh_id_it->second.string_value().empty() &&
      mesh_id_it->second.string_value().<HAS_PREFIX>("proj-")) {
            return;
      }
```

mandarjog: Great catch.
I question the useless default: svc.id.goog ?

extensions/stackdriver/stackdriver.cc
bianpengyuan: It will just be empty string if not set.

extensions/stackdriver/stackdriver.cc
bianpengyuan: Done.

PR #2994, opened 2020-08-31
Add sampling for Client Side access logs and TCP Connection Access logs

gargnupur:

Add sampling for Client Side access logs and TCP Connection Access logs

For tcp, an error connection is decided based on presence of response_flag

**Release note**:

```release-note
```


bianpengyuan (commented):
A meta question: do we need to sample log for TCP connection? The volume should be very low since it happens periodically and there ought to be only handful connections that is processed by networking filter.

gargnupur:
> A meta question: do we need to sample log for TCP connection? The volume should be very low since it happens periodically and there ought to be only handful connections that is processed by networking filter.

I feel it would be good to have it as an option and users can enable it if they need it.
TCP starts collecting telemetry at the start of connection itself and thus because of logging, extended RequestInfo gets populated at the start of the connection and hangs around till the connection dies. So it could be expensive and thus having sampling can help, not hold around this extra data from handful of connections to 1 connection... wdyt?

bianpengyuan:
The request info is way cheaper than other state that Envoy keeps for a connection, so I don't really think the memory overhead from telemetry filter for TCP connection is a good motivation for this. Also the point here is that I really don't think there will be that many TCP connections per pod, so we might be solving a problem that does not exist. How about we avoid the complexity and add this when it is demanded?

kyessenov:
Is sampling disabled without the access log policy filter? If yes, then it is reasonable to make it support the policy filter but not install access log policy filter initially.

gargnupur:
> Is sampling disabled without the access log policy filter? If yes, then it is reasonable to make it support the policy filter but not install access log policy filter initially.

Yes, it is disabled without the access log policy filter...

gargnupur:
Closing this as we don't need TCP Sampling as of now... Opened a new PR for client side sampling... https://github.com/istio/proxy/pull/3031

PR #2995, opened 2020-09-01
Add prow script for centos postsubmit build

howardjohn:
This adds a postsubmit script for centos build. Will be enabled in https://github.com/istio/test-infra/pull/2896.

Additionally, there are some minor fixes to the centos logic which broke since the original merge (its not yet blocking)

### DIFF
--- a/prow/proxy-common.inc
+++ b/prow/proxy-common.inc
@@ -56,7 +56,7 @@ if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
   gcloud auth activate-service-account --key-file="${GOOGLE_APPLICATION_CREDENTIALS}"
 
   # Use RBE when logged in.
-  BAZEL_BUILD_RBE_INSTANCE="${BAZEL_BUILD_RBE_INSTANCE:-projects/istio-testing/instances/default_instance}"
+  BAZEL_BUILD_RBE_INSTANCE="${BAZEL_BUILD_RBE_INSTANCE-projects/istio-testing/instances/default_instance}"
   BAZEL_BUILD_RBE_CACHE="${BAZEL_BUILD_RBE_CACHE:-grpcs://remotebuildexecution.googleapis.com}"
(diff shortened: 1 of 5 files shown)


prow/proxy-common.inc
```diff
@@ -56,7 +56,7 @@ if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
   gcloud auth activate-service-account --key-file="${GOOGLE_APPLICATION_CREDENTIALS}"
 
   # Use RBE when logged in.
-  BAZEL_BUILD_RBE_INSTANCE="${BAZEL_BUILD_RBE_INSTANCE:-projects/istio-testing/instances/default_instance}"
+  BAZEL_BUILD_RBE_INSTANCE="${BAZEL_BUILD_RBE_INSTANCE-projects/istio-testing/instances/default_instance}"
```
bianpengyuan, line 59: Accidental change?

howardjohn: Intentional, I found we are still using RBE with centos which I think is wrong. The `:-` means it will set it even if `BAZEL_BUILD_RBE_INSTANCE=""` is set, `-` is needed to allow us to unset it

bianpengyuan:
Wondering why centos presubmit test result is not listed here?

howardjohn:
@bianpengyuan its marked as hidden since it was under development. Once this merges  Iwill merge https://github.com/istio/test-infra/pull/2896 which will make it visible. You can see it passes from https://prow.istio.io/pr-history/?org=istio&repo=proxy&pr=2995

bianpengyuan (approved):

prow/proxy-common.inc
PiotrSikora: Do you recall what was the issue here? While the the remote execution engine uses Ubuntu toolchain, I believe that CentOS should be fine using object cache.

PiotrSikora: Let's continue discussion in https://github.com/istio/proxy/pull/3173.

PR #3000, opened 2020-09-03
Minor additions to Access Logs

gargnupur:
Minor additions to Access Logs
1) Made batch size and interval for reporting configurable
2) Don't add labels if they are empty-> this is done for labels that are known to empty sometimes
3) Add upstream_transport_failure_reason label to logs

**Release note**:

```release-note
```

### DIFF
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -27,7 +27,7 @@ package stackdriver.config.v1alpha1;
 import "google/protobuf/duration.proto";
 import "google/protobuf/wrappers.proto";
 
-// next id: 12
+// next id: 14
 message PluginConfig {
   // Types of Access logs to export. Does not affect audit logging.
   enum AccessLogging {
@@ -44,6 +44,17 @@ message PluginConfig {
   // Optional. Controls whether to export server access log.
   // This is deprecated in favor of AccessLogging enum.
   bool disable_server_access_logging = 1 [deprecated = true];
+
+  // Optional. Allows configuration of the size of the LogWrite request. The
+  // size is in bytes, so that it allows for better performance. Default is 4MB.
+  // The size of one log entry within LogWrite request is approx 1Kb.
+  int32 max_log_batch_size_in_bytes = 12;
+
+  // Optional. Allows configuration of the time between calls out to the
+  // stackdriver logging service to report buffered LogWrite request.
+  // Customers can choose to report more aggressively by keeping shorter report
+  // interval if needed. Default is 10s.
+  google.protobuf.Duration log_report_duration = 13;
 
   // Optional. Controls whether to export audit log.
   bool enable_audit_log = 11;
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -298,13 +298,29 @@ void Logger::fillAndFlushLogEntry(
     (*label_map)["protocol"] = request_info.request_protocol;
     (*label_map)["log_sampled"] = request_info.log_sampled ? "true" : "false";
     (*label_map)["connection_id"] = std::to_string(request_info.connection_id);
-    (*label_map)["route_name"] = request_info.route_name;
-    (*label_map)["upstream_host"] = request_info.upstream_host;
+    if (!request_info.route_name.empty()) {
+      (*label_map)["route_name"] = request_info.route_name;
+    }
+    if (!request_info.upstream_host.empty()) {
+      (*label_map)["upstream_host"] = request_info.upstream_host;
+    }
     (*label_map)["upstream_cluster"] = request_info.upstream_cluster;
-    (*label_map)["requested_server_name"] = request_info.request_serever_name;
-    (*label_map)["x-envoy-original-path"] = request_info.x_envoy_original_path;
-    (*label_map)["x-envoy-original-dst-host"] =
-        request_info.x_envoy_original_dst_host;
+    if (!request_info.requested_server_name.empty()) {
+      (*label_map)["requested_server_name"] =
+          request_info.requested_server_name;
+    }
+    if (!request_info.x_envoy_original_path.empty()) {
+      (*label_map)["x-envoy-original-path"] =
+          request_info.x_envoy_original_path;
+    }
+    if (!request_info.x_envoy_original_dst_host.empty()) {
+      (*label_map)["x-envoy-original-dst-host"] =
+          request_info.x_envoy_original_dst_host;
+    }
+    if (!request_info.upstream_transport_failure_reason.empty()) {
+      (*label_map)["upstream_transport_failure_reason"] =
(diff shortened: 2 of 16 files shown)


extensions/stackdriver/stackdriver.cc
```diff
@@ -221,6 +219,17 @@ bool StackdriverRootContext::configure(size_t configuration_size) {
             configuration + ", " + status.message().ToString());
     return false;
   }
+
+  int duration = 0;
```
bianpengyuan, line 223: If I remembered correctly, topology report is also control by this ticker. Could you check if there is any side effect on that?

extensions/stackdriver/log/logger.cc
```diff
@@ -296,13 +296,28 @@ void Logger::fillAndFlushLogEntry(
...
+    }
+    if (!request_info.upstream_host.empty()) {
+      (*label_map)["upstream_host"] = request_info.upstream_host;
+    }
     (*label_map)["upstream_cluster"] = request_info.upstream_cluster;
-    (*label_map)["requested_server_name"] = request_info.request_serever_name;
-    (*label_map)["x-envoy-original-path"] = request_info.x_envoy_original_path;
-    (*label_map)["x-envoy-original-dst-host"] =
-        request_info.x_envoy_original_dst_host;
+    if (!request_info.request_serever_name.empty()) {
```
kyessenov, line 306: server -> server

extensions/stackdriver/stackdriver.cc
```diff
@@ -70,7 +70,7 @@ int getMonitoringExportInterval() {
   return 60;
 }
 
-// Get logging export interval from node metadata in milliseconds. Returns 60
+// Get logging export interval from node metadata in milliseconds. Returns 10
```
kyessenov, line 73: What does "returns 10" mean? Is it the default value? That sounds too low for milliseconds.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -45,6 +45,16 @@ message PluginConfig {
...
 
+  // Optional. Allows configuration of the number of log entries request to
+  // batch into a single request. The size is in bytes, so that it allows for
+  // better performance. Default is 4MB. The size of one log entry is approx
+  // 1Kb.
+  int32 max_log_batch_size_in_bytes = 12;
+
+  // Optional. Allows configuration of the time between calls out to the
+  // stackdriver logging service to report new log entries.
+  google.protobuf.Duration log_report_duration = 13;
```
kyessenov, line 56: These two options sound contradictory. Do we report when the batch size reaches certain limit or when the ticker ticks? Are there many batches in a single request?

extensions/stackdriver/stackdriver.cc
gargnupur: "Returns 10 second".. 10 seconds is the default value..

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: We flush when the batch size reaches certain limit(configured at 4MB) and export(make grpc call to logging service) the flushed entried when ticker ticks. Yeah there are many log entries batched in a single request.

extensions/stackdriver/stackdriver.cc
gargnupur: As we talked offline.. added a default timer for ticker...

extensions/stackdriver/log/logger.cc
gargnupur: Fixed

gargnupur:
@kyessenov , @bianpengyuan : This is ready for review...

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -45,6 +45,16 @@ message PluginConfig {
   // This is deprecated in favor of AccessLogging enum.
   bool disable_server_access_logging = 1 [deprecated = true];
 
+  // Optional. Allows configuration of the number of log entries request to
```
bianpengyuan, line 48: `Allows configuration of the number of log entries request to batch into a single request` This comment is confusing, What it really controls is the size of a LogWrite request.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
bianpengyuan: Sorry, I lost track about the context here a bit. Why do we make report duration configurable? Can you add the reason to the description so that we don't forget? Also would the option still apply if we switch to dedicated thread for log reporting?

Also probably better to clarify the comment a bit. Instead of `to report new log entries`, we can make it `to report buffered LogWrite request`.

extensions/stackdriver/stackdriver.cc
```diff
@@ -352,7 +396,8 @@ void StackdriverRootContext::onTick() {
     }
   }
 
-  if (enableAccessLog()) {
+  if (enableAccessLog() &&
+      (cur - last_log_report_call_nanos_ > log_report_duration_nanos_)) {
```
bianpengyuan, line 400: Where is last_log_report_call_nanos_ updated?

extensions/stackdriver/stackdriver.cc
```diff
@@ -70,14 +70,24 @@ int getMonitoringExportInterval() {
   return 60;
 }
 
-// Get logging export interval from node metadata in milliseconds. Returns 60
+// Get logging export interval from node metadata in nanoseconds. Returns 10
 // seconds if interval is not found in metadata.
-int getLoggingExportIntervalMilliseconds() {
+long int getLoggingExportIntervalNanoseconds() {
```
bianpengyuan, line 75: Should we get rid of this node metadata override if we are going to control it via filter config?

bianpengyuan:
Fixes https://github.com/istio/istio/issues/26485

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: Fixed.

extensions/stackdriver/stackdriver.cc
gargnupur: Thanks..Fixed

extensions/stackdriver/stackdriver.cc
gargnupur: Removed

gargnupur:
Fixes https://github.com/istio/istio/issues/21252 too

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
gargnupur: Added.. Specifically to fix https://github.com/istio/istio/issues/21252

bianpengyuan (approved):

PR #3010, opened 2020-09-10
Remove unneeded check in release build

howardjohn:
This blocks centos, which is intentionally passing -i and -d. I think
its reasonable to remove the check entirely rather than changing the
logic around, since passing `-i` is already explicitly opting out of the
check

**Release note**:

```release-note
```

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -188,7 +188,7 @@ push_release: build
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/release-binary.sh -d "$(RELEASE_GCS_PATH)" -p
 
 push_release_centos:
-	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS) $(CENTOS_BUILD_ARGS)" BUILD_ENVOY_BINARY_ONLY=1 BINARY_NAME=envoy-centos && ./scripts/release-binary.sh -d "$(RELEASE_GCS_PATH)"
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS) $(CENTOS_BUILD_ARGS)" BUILD_ENVOY_BINARY_ONLY=1 BINARY_NAME=envoy-centos && ./scripts/release-binary.sh -c -d "$(RELEASE_GCS_PATH)"
 
 # Used by build container to export the build output from the docker volume cache
 exportcache:
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -182,13 +182,13 @@ test_release:
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/release-binary.sh
 
 test_release_centos:
-	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS) $(CENTOS_BUILD_ARGS)" BUILD_ENVOY_BINARY_ONLY=1 BINARY_NAME=envoy-centos && ./scripts/release-binary.sh -i
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS) $(CENTOS_BUILD_ARGS)" BUILD_ENVOY_BINARY_ONLY=1 BINARY_NAME=envoy-centos && ./scripts/release-binary.sh
 
 push_release: build
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/release-binary.sh -d "$(RELEASE_GCS_PATH)" -p
 
 push_release_centos:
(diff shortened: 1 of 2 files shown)


PiotrSikora (changes requested):
`-i` should not be passed in the release builds. This check was added so that we would never accidentally push releases built on unsupported platforms, and it prevented failures in the past.

Instead, you should extend the script to recognize CentOS as a supported platform.

howardjohn:
@PiotrSikora we previously made the script OS generic and look at BUILD_ENVOY_BINARY_ONLY to determine to skip wasm/docker/etc. I added an explicit flag now but we could also disable it only if BUILD_ENVOY_BINARY_ONLY is set although its not as explicit

scripts/release-binary.sh
```diff
@@ -82,7 +87,7 @@ if [ "${CHECK}" -eq 1 ]; then
     [[ "${UBUNTU_RELEASE}" == 'xenial' ]] || { echo 'Must run on Ubuntu 16.04 (Xenial).'; exit 1; }
   fi
   [[ "$(uname -m)" == 'x86_64' ]] || { echo 'Must run on x86_64.'; exit 1; }
-elif [ -n "${DST}" ]; then
+elif [ -n "${DST}" ] && [ "${BUILD_FOR_CENTOS}" -ne 0 ]; then
```
PiotrSikora, line 90: Since you're explicitly marking builds for CentOS, you should perform the check to make sure it's running on CentOS (i.e. provide CentOS equivalent to the regular check), and then you can remove `-i` from the `Makefile` targets.

PiotrSikora (changes requested):

scripts/release-binary.sh
```diff
@@ -76,12 +81,15 @@ if [ "${DST}" == "none" ]; then
...
 # Make sure the release binaries are built on x86_64 Ubuntu 16.04 (Xenial)
-if [ "${CHECK}" -eq 1 ]; then
+if [ "${CHECK}" -eq 1 ] && [ "${BUILD_FOR_CENTOS}" -eq 0 ]; then
   if [[ "${BAZEL_BUILD_ARGS}" != *"--config=remote-"* ]]; then
     UBUNTU_RELEASE=${UBUNTU_RELEASE:-$(lsb_release -c -s)}
     [[ "${UBUNTU_RELEASE}" == 'xenial' ]] || { echo 'Must run on Ubuntu 16.04 (Xenial).'; exit 1; }
   fi
   [[ "$(uname -m)" == 'x86_64' ]] || { echo 'Must run on x86_64.'; exit 1; }
+elif [ "${CHECK}" -eq 1 ] && [ "${BUILD_FOR_CENTOS}" -eq 1 ]; then
+# Make sure the release binaries are built on CentOS 7
```
PiotrSikora, line 91: Nit: wrong indent?

scripts/release-binary.sh
```diff
@@ -49,12 +49,16 @@ BUILD_ENVOY_BINARY_ONLY="${BUILD_ENVOY_BINARY_ONLY:-0}"
...
+# Support CentOS builds
+BUILD_FOR_CENTOS=0
+
 function usage() {
   echo "$0
     -d  The bucket name to store proxy binary (optional).
         If not provided, both envoy binary push and docker image push are skipped.
     -i  Skip Ubuntu Xenial check. DO NOT USE THIS FOR RELEASED BINARIES.
-        Cannot be used together with -d option.
+        Cannot be used together with -d option unless -c is set.
```
PiotrSikora, line 60: `unless -c is set` is no longer true.

Makefile.core.mk
```diff
@@ -182,13 +182,13 @@ test_release:
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/release-binary.sh
 
 test_release_centos:
-	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS) $(CENTOS_BUILD_ARGS)" BUILD_ENVOY_BINARY_ONLY=1 BINARY_NAME=envoy-centos && ./scripts/release-binary.sh -i
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS) $(CENTOS_BUILD_ARGS)" BUILD_ENVOY_BINARY_ONLY=1 BINARY_NAME=envoy-centos && ./scripts/release-binary.sh
```
PiotrSikora, line 185: You need to add `-c`.

Makefile.core.mk
```diff
@@ -182,13 +182,13 @@ test_release:
...
 
 push_release: build
 	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS)" && ./scripts/release-binary.sh -d "$(RELEASE_GCS_PATH)" -p
 
 push_release_centos:
-	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS) $(CENTOS_BUILD_ARGS)" BUILD_ENVOY_BINARY_ONLY=1 BINARY_NAME=envoy-centos && ./scripts/release-binary.sh -i -d "$(RELEASE_GCS_PATH)"
+	export PATH=$(PATH) CC=$(CC) CXX=$(CXX) BAZEL_BUILD_ARGS="$(BAZEL_BUILD_ARGS) $(CENTOS_BUILD_ARGS)" BUILD_ENVOY_BINARY_ONLY=1 BINARY_NAME=envoy-centos && ./scripts/release-binary.sh -d "$(RELEASE_GCS_PATH)"
```
PiotrSikora, line 191: You need to add `-c`.

Makefile.core.mk
PiotrSikora: Still missing here.

howardjohn: Thanks! I think I got it now

PiotrSikora (approved):
Thanks!

PiotrSikora:
(It might be worth changing commit summary and/or message before bot merges this)

PR #3016, opened 2020-09-17
Import edge proto directly in go integration test

bianpengyuan:
When trying to import the test module from this repo, it fails because the go mod tries to download mesh telemetry proto from cloud.google.com. This change removes the indirection (`cloud.google.com/go/meshtelemetry/v1alpha1 v0.0.0 => ./test/envoye2e/stackdriver_plugin/edges`) and import the generated edge proto directly.


kyessenov:
This is fine as long as edges are not in googleapis. Once they show up there, we are going to get conflicting protobuf definitions that would need to be taken care of.

douglas-reid:
@kyessenov we probably have a _long_ time before they make their way there.

kyessenov (approved):

PR #3018, opened 2020-09-17
fix(fake): preserve trace attributes in stackdriver fake

douglas-reid:
This enables testing that checks on the `custom_tags` functionality for Stackdriver. In particular, it will be used to verify the ability to annotate Cloud Trace spans with canonical service information.


bianpengyuan (approved):

bianpengyuan:
/test test-asan_proxy

We probably should try to reduce some unnecessary build target from asan build

PR #3040, opened 2020-10-06
Add local rate limiting wasm filter

gargnupur:

Adds local rate limiting wasm filter

### DIFF
--- a/extensions/local_ratelimit/BUILD
+++ b/extensions/local_ratelimit/BUILD
@@ -1,0 +1,29 @@
+licenses(["notice"])  # Apache 2
+
+load(
+    "@envoy//bazel:envoy_build_system.bzl",
+    "envoy_cc_library",
+    "envoy_package",
+)
+
+envoy_package()
+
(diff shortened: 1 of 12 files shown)


bianpengyuan (commented):
Can we do it in Wasm ecosystem repo: https://github.com/istio-ecosystem/wasm-extensions if this is going to be Wasm only plugin?

kyessenov:
I think it makes sense to make it Wasm-only. Rate limiting has a wide range of requirements, so this would be a good example. Since I don't think one-size-fits-all solution, it is not useful to compile in.

We can also simplify the build there, and drop the extra cruft with protos.

kyessenov:
It's really important to measure performance hit for this filter. We should not add too much overhead at a very high rate limit.

gargnupur:
@kyessenov , @bianpengyuan : Thanks, will add this to istio-ecosystem repo and add benchmark perf tests too..

kyessenov:
Merged into wasm-extensions.

PR #3045, opened 2020-10-10
jwt: support nested claims

yangminzhu:
https://github.com/istio/istio/issues/21340

### DIFF
--- a/src/envoy/http/authn/authn_utils.cc
+++ b/src/envoy/http/authn/authn_utils.cc
@@ -39,11 +39,12 @@ static const std::string kExchangedTokenOriginalPayload = "original_claims";
 void process(const Wasm::Common::JsonObject& json_obj,
              google::protobuf::Struct& claims) {
   for (const auto& claim : json_obj.items()) {
-    auto json_key = Wasm::Common::JsonValueAs<std::string>(claim.key());
+    auto claim_key = claim.key();
+    auto json_key = Wasm::Common::JsonValueAs<absl::string_view>(claim_key);
     if (json_key.second != Wasm::Common::JsonParserResultDetail::OK) {
       continue;
     }
-    std::string key = json_key.first.value();
+    std::string key(json_key.first.value());
 
     // 1. Try to parse as string.
     auto value_string =
--- a/src/envoy/http/authn/authn_utils.cc
+++ b/src/envoy/http/authn/authn_utils.cc
@@ -36,6 +36,52 @@ static const std::string kExchangedTokenOriginalPayload = "original_claims";
 
 };  // namespace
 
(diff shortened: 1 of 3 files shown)


src/envoy/http/authn/authn_utils.cc
```diff
@@ -36,6 +36,52 @@ static const std::string kExchangedTokenOriginalPayload = "original_claims";
 
 };  // namespace
 
+void process(const Wasm::Common::JsonObject& json_obj,
+             google::protobuf::Struct& claims) {
+  for (const auto& claim : json_obj.items()) {
+    auto json_key = Wasm::Common::JsonValueAs<std::string>(claim.key());
```
bianpengyuan, line 42: As string_view? so that we can save a copy.

bianpengyuan:
Looks like there is memory leak in implementation

yangminzhu:
> Looks like there is memory leak in implementation

@bianpengyuan I think it's due to the change of using `string_view`. I'm not sure why the tsan/asan still failed even after I move the `claim.key()` to be above the `JsonValueAs`, it seems it still get destroyed too early.

src/envoy/http/authn/authn_utils_test.cc
```diff
@@ -58,6 +58,27 @@ const std::string kSecIstioAuthUserInfoHeaderWithAudValueArray =
...
+const std::string kSecIstioAuthUserInfoHeaderWithNestedClaims =
+    R"(
+       {
+         "iss": "issuer@foo.com",
+         "sub": "sub@foo.com",
+         "nested1": {
+           "aud1": "aud1a  aud1b",
+           "list1": ["list1a", "list1b"],
+           "other1": "str1",
+           "non-string-ignored": 111,
```
liminw, line 70: Do we plan to support non-string claims going forward?

liminw (commented):
Looks good.

src/envoy/http/authn/authn_utils_test.cc
yangminzhu: maybe, the Istio API syntax will need update to support non-string match.

src/envoy/http/authn/authn_utils.cc
```diff
@@ -36,6 +36,52 @@ static const std::string kExchangedTokenOriginalPayload = "original_claims";
...
+    if (json_key.second != Wasm::Common::JsonParserResultDetail::OK) {
+      continue;
+    }
+    std::string key = json_key.first.value();
```
bianpengyuan, line 46: const string&?

liminw (approved):

bianpengyuan (approved):

PR #3047, opened 2020-10-12
Add customized access logging support for stackdriver access logs

gargnupur:
Add customized access logging support for stackdriver access logs 

TODO: add support for removal of tags

### DIFF
--- a/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
+++ b/extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
@@ -27,7 +27,21 @@ package stackdriver.config.v1alpha1;
 import "google/protobuf/duration.proto";
 import "google/protobuf/wrappers.proto";
 
-// next id: 14
+// Custom instance configuration overrides.
+// Provides a way to customize metrics/logs .
+message CustomConfig {
+  // (Optional) Collection of tag names and tag expressions to include in the
+  // instance. Conflicts are resolved by the tag name by overriding previously
+  // supplied values.
+  map<string, string> dimensions = 1;
+
+  // (Optional) A list of tags to remove.
+  // Not implemented yet.
+  // $hide_from_docs
+  repeated string tags_to_remove = 2;
+}
+
+// next id: 15
 message PluginConfig {
   // Types of Access logs to export. Does not affect audit logging.
   enum AccessLogging {
@@ -101,4 +115,9 @@ message PluginConfig {
 
   // Optional. Controls what type of logs to export..
   AccessLogging access_logging = 10;
+
+  // (Optional) Collection of tag names and tag expressions to include in the
+  // logs. Conflicts are resolved by the tag name by overriding previously
+  // supplied values. Does not apply to audit logs.
+  CustomConfig custom_log_config = 14;
 }
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -150,6 +150,14 @@ void fillSourceLabels(
   }
 }
 
+void fillExtraLabels(
+    const std::unordered_map<std::string, std::string>& extra_labels,
+    google::protobuf::Map<std::string, std::string>* label_map) {
+  for (const auto& extra_label : extra_labels) {
+    (*label_map)[extra_label.first] = extra_label.second;
+  }
+}
+
 }  // namespace
 
 using google::protobuf::util::TimeUtil;
@@ -167,8 +175,9 @@ constexpr char kClientAuditLogName[] = "client-istio-audit-log";
 void Logger::initializeLogEntryRequest(
     const flatbuffers::Vector<flatbuffers::Offset<Wasm::Common::KeyVal>>*
         platform_metadata,
-    const ::Wasm::Common::FlatNode& local_node_info, bool outbound,
-    bool audit) {
+    const ::Wasm::Common::FlatNode& local_node_info,
+    const std::unordered_map<std::string, std::string>& extra_labels,
+    bool outbound, bool audit) {
   LogEntryType log_entry_type = GetLogEntryType(outbound, audit);
   log_entries_request_map_[log_entry_type]->request =
       std::make_unique<google::logging::v2::WriteLogEntriesRequest>();
@@ -203,10 +212,15 @@ void Logger::initializeLogEntryRequest(
   // Set common labels shared by all client entries or server entries
   outbound ? fillSourceLabels(local_node_info, label_map, audit)
            : fillDestinationLabels(local_node_info, label_map, audit);
+  if (!audit) {
+    fillExtraLabels(extra_labels, label_map);
+  }
 }
 
 Logger::Logger(const ::Wasm::Common::FlatNode& local_node_info,
-               std::unique_ptr<Exporter> exporter, int log_request_size_limit) {
+               std::unique_ptr<Exporter> exporter,
+               const std::unordered_map<std::string, std::string>& extra_labels,
+               int log_request_size_limit) {
   const auto platform_metadata = local_node_info.platform_metadata();
   const auto project_iter =
       platform_metadata ? platform_metadata->LookupByKey(Common::kGCPProjectKey)
@@ -218,28 +232,30 @@ Logger::Logger(const ::Wasm::Common::FlatNode& local_node_info,
   // Initalize the current WriteLogEntriesRequest for client/server
   log_entries_request_map_[LogEntryType::Client] =
       std::make_unique<Logger::WriteLogEntryRequest>();
-  initializeLogEntryRequest(platform_metadata, local_node_info,
+  initializeLogEntryRequest(platform_metadata, local_node_info, extra_labels,
                             true /*outbound */, false /* audit */);
   log_entries_request_map_[Logger::LogEntryType::Server] =
       std::make_unique<Logger::WriteLogEntryRequest>();
-  initializeLogEntryRequest(platform_metadata, local_node_info,
+  initializeLogEntryRequest(platform_metadata, local_node_info, extra_labels,
                             false /* outbound */, false /* audit */);
   log_entries_request_map_[LogEntryType::ClientAudit] =
       std::make_unique<Logger::WriteLogEntryRequest>();
-  initializeLogEntryRequest(platform_metadata, local_node_info,
+  initializeLogEntryRequest(platform_metadata, local_node_info, extra_labels,
                             true /*outbound */, true /* audit */);
   log_entries_request_map_[Logger::LogEntryType::ServerAudit] =
       std::make_unique<Logger::WriteLogEntryRequest>();
-  initializeLogEntryRequest(platform_metadata, local_node_info,
+  initializeLogEntryRequest(platform_metadata, local_node_info, extra_labels,
                             false /* outbound */, true /* audit */);
 
   log_request_size_limit_ = log_request_size_limit;
   exporter_ = std::move(exporter);
 }
 
-void Logger::addTcpLogEntry(const ::Wasm::Common::RequestInfo& request_info,
-                            const ::Wasm::Common::FlatNode& peer_node_info,
-                            long int log_time, bool outbound, bool audit) {
+void Logger::addTcpLogEntry(
+    const ::Wasm::Common::RequestInfo& request_info,
+    const ::Wasm::Common::FlatNode& peer_node_info,
+    const std::unordered_map<std::string, std::string>& extra_labels,
+    long int log_time, bool outbound, bool audit) {
   // create a new log entry
   auto* log_entries = log_entries_request_map_[GetLogEntryType(outbound, audit)]
                           ->request->mutable_entries();
@@ -250,13 +266,15 @@ void Logger::addTcpLogEntry(const ::Wasm::Common::RequestInfo& request_info,
 
   addTCPLabelsToLogEntry(request_info, peer_node_info, new_entry, outbound,
                          audit);
-  fillAndFlushLogEntry(request_info, peer_node_info, new_entry, outbound,
-                       audit);
+  fillAndFlushLogEntry(request_info, peer_node_info, extra_labels, new_entry,
+                       outbound, audit);
 }
 
-void Logger::addLogEntry(const ::Wasm::Common::RequestInfo& request_info,
-                         const ::Wasm::Common::FlatNode& peer_node_info,
-                         bool outbound, bool audit) {
+void Logger::addLogEntry(
+    const ::Wasm::Common::RequestInfo& request_info,
+    const ::Wasm::Common::FlatNode& peer_node_info,
+    const std::unordered_map<std::string, std::string>& extra_labels,
+    bool outbound, bool audit) {
   // create a new log entry
   auto* log_entries = log_entries_request_map_[GetLogEntryType(outbound, audit)]
                           ->request->mutable_entries();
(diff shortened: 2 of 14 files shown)


extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -96,9 +118,9 @@ message PluginConfig {
   // metrics are enabled).
   bool disable_http_size_metrics = 8;
 
-  // Optional. Allows enabling log compression for stackdriver access logs.
-  google.protobuf.BoolValue enable_log_compression = 9;
+  // Optional. Controls whether to export audit log.
```
mandarjog, line 121: This looks like a breaking change with field re assignment.
Was the original config available in 1.7 ?

gargnupur: Yes, the original config is there in 1.7.. Can mark the old one's as deprecated and then remove next cycle?

gargnupur: Actually removed it.. thinking more, we need better organization with Audit logs.. can do in 1.9 rather

extensions/stackdriver/stackdriver.cc
```diff
@@ -578,6 +612,7 @@ inline bool StackdriverRootContext::enableAllAccessLog() {
   return (!config_.disable_server_access_logging() && !isOutbound()) ||
          config_.access_logging() ==
              stackdriver::config::v1alpha1::PluginConfig::FULL;
+  ;
```
kyessenov, line 615: extraneous semicolon

extensions/stackdriver/stackdriver.cc
```diff
@@ -265,6 +268,26 @@ bool StackdriverRootContext::configure(size_t configuration_size) {
   }
 
   if (!logger_ && enableAccessLog()) {
+    std::unordered_map<std::string, std::string> extra_labels;
+    if (config_.has_custom_log_config()) {
+      cleanupExpressions();
+      for (const auto& dimension : config_.custom_log_config().dimensions()) {
+        uint32_t token;
+        if (createExpression(dimension.second, &token) != WasmResult::Ok) {
```
kyessenov, line 276: This is correct: we should create expressions at config time only. Can you add deleteExpression in cleanup to release host resources? Otherwise, we're leaking some data structures.

kyessenov:
Asan test locked up in a go test for 4h. Can you check if the changes to SD test can block forever somewhere?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -27,7 +27,20 @@ package stackdriver.config.v1alpha1;
...
+// Custom instance configuration overrides.
+// Provides a way to customize metrics/logs .
+message CustomConfig {
+  // (Optional) Collection of tag names and tag expressions to include in the
+  // instance. Conflicts are resolved by the tag name by overriding previously
+  // supplied values.
+  map<string, string> dimensions = 1;
+
+  // (Optional) A list of tags to remove.
+  // Not implemeted yet.
```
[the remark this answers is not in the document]
gargnupur: Fixed

extensions/stackdriver/log/logger.cc
```diff
@@ -331,6 +350,11 @@ void Logger::fillAndFlushLogEntry(
     new_entry->set_trace_sampled(request_info.b3_trace_sampled);
   }
 
+  // This is done just before flushing, so that any customized label entry can
+  // override existing one's.
```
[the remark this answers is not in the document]
gargnupur: Fixed

extensions/stackdriver/stackdriver.cc
```diff
@@ -265,6 +268,26 @@ bool StackdriverRootContext::configure(size_t configuration_size) {
...
+                                 dimension.second));
+          continue;
+        }
+        std::string value;
+        if (evaluateExpression(token, &value)) {
+          extra_labels[dimension.first] = value;
+          continue;
+        }
+        // Push back expressions that could not be evaluated at config time, to
+        // be evaluated at request time.
```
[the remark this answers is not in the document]
gargnupur: Some resources can be evaluated using node metadata and hence can be done at config time itself.. EX in test itself "source_workload: "'_' + node.metadata['WORKLOAD_NAME']"
    source_workload_namespace: "'_' + node.metadata['NAMESPACE']"" 
Things like above could be evaluated at config time... wdyt? 
No harm in checking at config time to avoid paying cost in request path?

extensions/stackdriver/stackdriver.cc
gargnupur: Thanks for catching. Fixed

extensions/stackdriver/stackdriver.cc
gargnupur: Fixed

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
bianpengyuan: Are we sure that per request expression cannot be successfully evaluated at config time?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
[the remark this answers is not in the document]
bianpengyuan: Should this be annotated with hide from doc?

extensions/stackdriver/stackdriver.cc
```diff
@@ -645,6 +680,31 @@ void StackdriverRootContext::setConnectionState(
...
+                             expression.expression));
+      continue;
+    }
+    extra_labels[expression.tag] = value;
+  }
+}
+
+void StackdriverRootContext::cleanupExpressions() {
+  expressions_.clear();
+  for (auto const& item : tcp_request_queue_) {
```
bianpengyuan, line 698: I am wondering under what circumstance would clean tcp request expression be useful?

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
gargnupur: Expressions like "request.protocol" can be evaluated at request time only and not at config time...

extensions/stackdriver/stackdriver.cc
gargnupur: This is the question I had in mind.. what happens for a long standing tcp connection when configure is called in between? I thought for those, we need to have a cleanup and evaluate again...

/cc @kyessenov

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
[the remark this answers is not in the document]
gargnupur: Thanks.. done

extensions/stackdriver/stackdriver.cc
bianpengyuan: That will remove and recreate the root context I think.

bianpengyuan: Actually, root context won't be recreated, it is just that onConfigure will be called once again. But looks like here the expr is cleaned but not refill in onConfigure, or did I miss anything? I actually think it is fine to leave existing TCP connection as is to avoid complexity.

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
bianpengyuan: Yeah, I am just worrying that we cannot make sure every request dependent attribute cannot be evaluated on config time. There are quite a lot of attributes, I don't think we can be certain on each of them.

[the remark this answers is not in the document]
gargnupur: so there are scenarios wherein an expression evaluated at config time is different from request time? then I will delete this code of evaluating at config time...

One of the other concerns I had was about source labels being merged at request time but that should be fine too as below:
Things that can be evaluated at config time like source information, in our logger we add it in base WriteLogEntryRequest and then it gets combined into one set of labels on stackdriver side. So if we evaluate custom "source_workload" at request time, it will get added in log entry labels and then on stackdriver side they will keep log entry labels "source_workload" as per `Optional. Default labels that are added to the labels field of all log entries in entries. If a log entry already has a label with the same key as a label in this parameter, then the log entry's label is not changed. See LogEntry.(https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write) `

extensions/stackdriver/stackdriver.cc
gargnupur: expressions will get created again from the proto config in onConfigure..

That's what I thought that, onConfigure will get called again. What about a scenario, wherein we want to debug long standing tcp connection and we add custom access log labels through our Telemetry API to see what is going on with this connection...

Can delete for now to keep TCP labels same and we can add this when we see the need..

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
kyessenov: Can we make this explicit if we want to keep it? The problem is that an error does not indicate that the expression is config vs request time. The expression could have some conditional like `has(request.path) ? request.path : ""` that would make it evaluate at config time, which is not what we want.

extensions/stackdriver/stackdriver.cc
kyessenov: Yeah, it's an issue. Not sure how to handle it except making it fail safe in case some stale reports are referencing deleted expressions.

extensions/stackdriver/stackdriver.cc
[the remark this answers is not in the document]
gargnupur: Changed it to just evaluating at request time..

extensions/stackdriver/stackdriver.cc
gargnupur: Removed as we talked offline..

gargnupur:
> Asan test locked up in a go test for 4h. Can you check if the changes to SD test can block forever somewhere?

It passed now...

gargnupur:
@kyessenov , @bianpengyuan : Can you please take a look at this again?

testdata/filters/stackdriver_outbound.yaml.tmpl
```diff
@@ -18,7 +18,7 @@
           "@type": "type.googleapis.com/google.protobuf.StringValue"
           value: |
             {{- if .Vars.JustSendErrorClientLog }}
```
bianpengyuan, line 20: Can we replace this condition with `if .Vars.StackdriverFilterClientErrorConfig`

testdata/filters/stackdriver_outbound.yaml.tmpl
```diff
@@ -18,7 +18,7 @@
...
-            {"access_logging": "ERRORS_ONLY", "enable_audit_log": true}
+            {{ .Vars.StackdriverFilterClientErrorConfig | fill }}
             {{- else }}
-            {"access_logging": "FULL", "enable_audit_log": true}
-            {{- end }}
+            {{ .Vars.StackdriverFilterClientConfig | fill }}
```
bianpengyuan, line 23: To avoid changing existing code, can we add a default else at the end with `{"access_logging": "FULL", "enable_audit_log": true}` and fill in `StackdriverFilterClientConfig` when it is specified?

test/envoye2e/stackdriver_plugin/stackdriver_test.go
```diff
@@ -766,5 +778,62 @@ func TestStackdriverAttributeGen(t *testing.T) {
...
+			sd,
+			&SecureTokenService{Port: stsPort},
+			&driver.Update{Node: "client", Version: "0", Listeners: []string{driver.LoadTestData("testdata/listener/client.yaml.tmpl")}},
+			&driver.Update{Node: "server", Version: "0", Listeners: []string{driver.LoadTestData("testdata/listener/server.yaml.tmpl")}},
+			&driver.Envoy{Bootstrap: params.LoadTestData("testdata/bootstrap/server.yaml.tmpl")},
+			&driver.Envoy{Bootstrap: params.LoadTestData("testdata/bootstrap/client.yaml.tmpl")},
+			&driver.Sleep{1 * time.Second},
+			&driver.Repeat{N: 10, Step: driver.Get(params.Ports.ClientPort, "hello, world!")},
+			sd.Check(params,
+				[]string{"testdata/stackdriver/client_request_count.yaml.tmpl", "testdata/stackdriver/server_request_count.yaml.tmpl"},
```
bianpengyuan, line 817: Do we need to verify stats?

bianpengyuan (commented):
lgtm, only neats. thanks!

extensions/stackdriver/stackdriver.cc
```diff
@@ -265,6 +268,19 @@ bool StackdriverRootContext::configure(size_t configuration_size) {
   }
 
   if (!logger_ && enableAccessLog()) {
+    std::unordered_map<std::string, std::string> extra_labels;
+    if (config_.has_custom_log_config()) {
+      cleanupExpressions();
```
kyessenov, line 273: Shouldn't we clean up expressions if access log disabled? What happens if the config is updated to remove access logging? onConfigure would be called twice.

kyessenov: Also, it seems that logger_ would be set. Is it possible to re-configure SD?

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -27,7 +27,21 @@ package stackdriver.config.v1alpha1;
...
+// Provides a way to customize metrics/logs .
+message CustomConfig {
+  // (Optional) Collection of tag names and tag expressions to include in the
+  // instance. Conflicts are resolved by the tag name by overriding previously
+  // supplied values.
+  map<string, string> dimensions = 1;
+
+  // (Optional) A list of tags to remove.
+  // Not implemented yet.
+  // $hide_from_docs
```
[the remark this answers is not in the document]
gargnupur: I think we probably need to run `make extension-docs` in some pipeline?

extensions/stackdriver/stackdriver.cc
gargnupur: Thanks for catching..

I think reconfiguring logger would mean loosing the log entries that were in queue.. so either flush those entries and then reconfigure. There is a need now to do this as batch size can be changed in onConfigure...

test/envoye2e/stackdriver_plugin/stackdriver_test.go
gargnupur: Not really.. removed

testdata/filters/stackdriver_outbound.yaml.tmpl
gargnupur: Fixed

testdata/filters/stackdriver_outbound.yaml.tmpl
gargnupur: Good idea.. fixed

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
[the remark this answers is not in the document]
gargnupur: We should build this in a container too.. without that it's complaining about so many things not on machine :(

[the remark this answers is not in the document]
gargnupur: Updated doc.. this is not showing

extensions/stackdriver/stackdriver.cc
kyessenov: This issue doesn't seem to be specific to SD. Let's address this here and in stats separately.

kyessenov (approved):
LGTM. Putting block if others have comments.

bianpengyuan (approved):

gargnupur:
@bianpengyuan , @kyessenov : Thanks for the review!

PR #3048, opened 2020-10-12
grpc: add streaming metrics

kyessenov:

- Adds metrics for request and response message counts.
  These are enabled by default, but I can add a guard.

- Do not report 0 counter values. This is useful for gRPC counts since the metric applies to all HTTP traffic. Peers may interchange HTTP and gRPC traffic on the same connection, so we cannot predict whether gRPC metric is not needed.

- Interval reporting will be in a follow-up.


testdata/metric/client_request_messages.yaml.tmpl
```diff
@@ -0,0 +1,52 @@
...
+    value: ratings
+  - name: destination_canonical_revision
+    value: version-1
+  - name: destination_service_name
+    value: server
+  - name: destination_service_namespace
+    value: default
+  - name: request_protocol
+    value: grpc
+  - name: response_code
```
douglas-reid, line 45: do we need response_code on these metrics?  it is always 200, correct?

testdata/metric/client_request_messages.yaml.tmpl
```diff
@@ -0,0 +1,52 @@
...
+    value: "200"
+  - name: grpc_response_status
+    value: "{{ .Vars.GrpcResponseStatus }}"
+  - name: response_flags
+    value: DC
```
douglas-reid, line 50: do we expect `DC` in the base case for client metrics?  does `response_flags` add value here?

testdata/metric/client_request_messages.yaml.tmpl
```diff
@@ -0,0 +1,52 @@
...
+    value: server.default.svc.cluster.local
+  - name: destination_canonical_service
+    value: ratings
+  - name: destination_canonical_revision
+    value: version-1
+  - name: destination_service_name
+    value: server
+  - name: destination_service_namespace
+    value: default
+  - name: request_protocol
```
douglas-reid, line 43: i wonder, is this useful?  will these metrics already imply grpc?

extensions/stats/plugin.cc
```diff
@@ -206,15 +205,26 @@ const std::vector<MetricFactory>& PluginRootContext::defaultMetrics() {
...
                         -> uint64_t { return request_info.request_size; },
                     false},
       MetricFactory{"response_bytes", MetricType::Histogram,
-
                     [](const ::Wasm::Common::RequestInfo& request_info)
                         -> uint64_t { return request_info.response_size; },
                     false},
+      // GRPC metrics.
+      MetricFactory{
+          "request_messages", MetricType::Counter,
```
douglas-reid, line 217: should we consider making these `request_messages_total` and `response_messages_total` to align with prom best practices: https://prometheus.io/docs/practices/naming/

testdata/metric/client_request_messages.yaml.tmpl
kyessenov: Yeah, it is redundant. It's an artifact of the implementation, that should be fixed I think.

testdata/metric/client_request_messages.yaml.tmpl
kyessenov: DC is downstream closure (client closed connection before server responded, intentionally). I think it's valuable here.

testdata/metric/client_request_messages.yaml.tmpl
kyessenov: Yes. The same problem is with TCP (it's also tcp there). It's not useful.

extensions/stats/plugin.cc
kyessenov: SG.

testdata/metric/client_request_messages.yaml.tmpl
kyessenov: Removed.

testdata/metric/client_request_messages.yaml.tmpl
kyessenov: Removed.

testdata/metric/client_request_messages.yaml.tmpl
kyessenov: Removed.

kyessenov:
Removed a bunch of labels and can remove more. They can be back-filled with metric customization.

extensions/common/context.cc
```diff
@@ -391,6 +395,19 @@ void populateTCPRequestInfo(bool outbound, RequestInfo* request_info,
   request_info->request_protocol = kProtocolTCP;
 }
 
+bool populateGRPCInfo(RequestInfo* request_info) {
```
mandarjog, line 398: 1. This should also be called on a ticker.
2. Is there a way to get the message type? If the underlying protocol does not support it yet, we should add it there too.

kyessenov: 1. Will be in a followup per PR description. 
2. That's not possible in gRPC. We'd need the descriptor.

mandarjog: I think if we can efficiently do proto when descriptor is known, it would be a big value add.

kyessenov: With a descriptor, the message type is already known, so there is nothing extra being added. I think operation name would be useful, but it's not in the standard metrics.

kyessenov:
@douglas-reid @mandarjog is this good to go for non-interval reporting? I'm afraid interval reporting is going to run into some Wasm lifecycle issue, so want to separate it out for now.

extensions/common/context.cc
douglas-reid: operation name for gRPC would likely be ok. it isn't in standard metrics because of unbounded cardinality, but within gRPC, that shouldn't be an issue.

kyessenov: It should be easy to add with customization `operation_name: request.path` for gRPC metrics.

bianpengyuan, line 398: No need to return boolean here? does not seem to be used and it always returns false.

extensions/common/context.cc
```diff
@@ -391,6 +395,19 @@ void populateTCPRequestInfo(bool outbound, RequestInfo* request_info,
...
+  std::string value;
+  if (!getValue({"filter_state", GrpcStatsName}, &value)) {
+    return false;
+  }
+  std::vector<std::string_view> parts = absl::StrSplit(value, ',');
+  if (parts.size() == 2) {
```
bianpengyuan, line 404: Would be great to add a comment here about what are the two parts instead of understanding it from the code. Also is this documented somewhere?

extensions/common/context.cc
kyessenov: It returns true if it parses correctly (SimpleAtoi return bool). It's not used yet, but seems reasonable to have a status return.

extensions/common/context.h
```diff
@@ -231,6 +235,10 @@ void populateExtendedRequestInfo(RequestInfo* request_info);
 void populateTCPRequestInfo(bool outbound, RequestInfo* request_info,
                             const std::string& destination_namespace);
 
+// populateGRPCInfo fills gRPC-related information, such as message counts.
+// Returns true if all informatio is filled.
```
gargnupur, line 239: nit:information

extensions/stats/plugin.cc
```diff
@@ -196,46 +196,56 @@ const std::vector<MetricFactory>& PluginRootContext::defaultMetrics() {
...
                         -> uint64_t { return request_info.response_size; },
-                    false},
+                    false, count_standard_labels},
+      // GRPC metrics.
+      MetricFactory{
+          "request_messages_count", MetricType::Counter,
+          [](const ::Wasm::Common::RequestInfo& request_info) -> uint64_t {
+            return request_info.request_message_count;
+          },
+          false, count_peer_labels},
```
bianpengyuan, line 221: Can you add some comments here about what labels are ignored for this metric?

bianpengyuan: Also would connection_security_policy worth to be added?

bianpengyuan: Or we are considering that it is already tracked by request total, thus it is a duplicated here and this should just focus on the message count.

test/envoye2e/driver/grpc.go
```diff
@@ -74,3 +74,48 @@ func (g *GrpcCall) Run(p *Params) error {
...
+	stream, err := client.EchoStream(context.Background())
+	if err != nil {
+		return err
+	}
+
+	var count uint32
+	i := 0
+
+	for {
+		if i >= len(g.Counts) {
```
bianpengyuan, line 102: Why not just do this and i++ in the top for loop?

test/envoye2e/stats_plugin/stats_test.go
```diff
@@ -309,6 +312,54 @@ func TestStatsGrpc(t *testing.T) {
...
+func TestStatsGrpcStream(t *testing.T) {
+	params := driver.NewTestParams(t, map[string]string{
+		"RequestMessages":            "3",  // 3 from stream
+		"ResponseMessages":           "13", // 13 from stream
+		"MetadataExchangeFilterCode": "inline_string: \"envoy.wasm.metadata_exchange\"",
+		"StatsFilterCode":            "inline_string: \"envoy.wasm.stats\"",
+		"WasmRuntime":                "envoy.wasm.runtime.null",
+		"DisableDirectResponse":      "true",
+		"UsingGrpcBackend":           "true",
+		"GrpcResponseStatus":         "2",
```
bianpengyuan, line 324: This is not used?

extensions/common/context.cc
kyessenov: It's implemented by grpc_stats and should probably be documented there. I'll do that separately.

extensions/common/context.h
kyessenov: Thanks, fixed.

bianpengyuan (commented):
lg, only minor comments.

extensions/stats/plugin.cc
kyessenov: Yeah, I only put peer labels, since the other metrics apply to the outer request rather than individual messages within them. I'm not sure if we want more, so I am starting from the minimal set.

test/envoye2e/driver/grpc.go
kyessenov: Simplified.

test/envoye2e/stats_plugin/stats_test.go
kyessenov: Removed.

bianpengyuan (approved):

extensions/stats/plugin.cc
douglas-reid: looks like we went with `_count` instead of `_total` ?

extensions/stats/plugin.cc
```diff
@@ -192,46 +192,58 @@ const std::vector<MetricFactory>& PluginRootContext::defaultMetrics() {
...
+      MetricFactory{
+          "response_messages_count", MetricType::Counter,
+          [](const ::Wasm::Common::RequestInfo& request_info) -> uint64_t {
+            return request_info.response_message_count;
```
gargnupur, line 223: how about metric for avg size of these messages?

extensions/stats/plugin.cc
kyessenov: Oops, renamed to total.

extensions/stats/plugin.cc
kyessenov: That would be difficult I think. The messages are framed by another filter, so we don't have a way to accumulate their sizes here. Open to any suggestions how to make that possible.

mandarjog: I think it *would* be useful.
@kyessenov is it possible to add it upstream? We need a distribution of message sizes.
We need not  hold this PR for it though.

mandarjog (approved):
lgtm

extensions/stats/plugin.cc
kyessenov: We can report size distribution to native envoy stats, but I don't know how to get it out back here.

gargnupur: not sure either... but can we use grpc stats filter only to get message size and add it in the above stats?

gargnupur (approved):

extensions/stats/plugin.cc
kyessenov: There might be many messages, so we'd be passing a large list of values for sizes. We would need histogram summaries I think, to make it work.

PR #3049, opened 2020-10-13
cleanup: fix TODOs waiting for c++17

kyessenov:

### DIFF
--- a/extensions/stats/plugin.cc
+++ b/extensions/stats/plugin.cc
@@ -623,8 +623,7 @@ bool PluginRootContext::report(::Wasm::Common::RequestInfo& request_info,
   }
 
   incrementMetric(cache_misses_, 1);
-  // TODO: When we have c++17, convert to try_emplace.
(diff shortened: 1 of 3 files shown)


extensions/stats/plugin.cc
```diff
@@ -623,8 +619,7 @@ bool PluginRootContext::report(::Wasm::Common::RequestInfo& request_info,
   }
 
   incrementMetric(cache_misses_, 1);
-  // TODO: When we have c++17, convert to try_emplace.
-  metrics_.emplace(istio_dimensions_, stats);
+  metrics_.try_emplace(istio_dimensions_, stats);
```
gargnupur, line 622: cool.. we should do this for all of our code in a separate PR perhaps...

gargnupur (approved):

extensions/stats/plugin.cc
kyessenov: Agree. Started with Wasm sources since every cycle counts there.

PR #3051, opened 2020-10-14
Parse workload and service from resource labels

bianpengyuan:


extensions/common/context.cc
```diff
@@ -295,13 +269,69 @@ flatbuffers::DetachedBuffer extractLocalNodeFlatBuffer() {
...
 }
 
+PeerNodeInfo::PeerNodeInfo(const std::string_view peer_metadata_id_key,
+                           const std::string_view peer_metadata_key) {
+  found_ = getValue({peer_metadata_id_key}, &peer_id_);
+  if (found_) {
+    getValue({peer_metadata_key}, &peer_node_);
+    return;
+  }
+
```
mandarjog, line 279: If downstreamPeer { return }

No need to check upstream host metadata here.

extensions/stats/plugin.cc
```diff
@@ -537,51 +537,39 @@ void PluginRootContext::onTick() {
...
-      outbound_ ? peer_node
-                : flatbuffers::GetRoot<::Wasm::Common::FlatNode>(
+  const ::Wasm::Common::FlatNode& destination_node_info =
+      outbound_ ? peer_node_info.get()
+                : *flatbuffers::GetRoot<::Wasm::Common::FlatNode>(
                       local_node_info_.data());
   std::string destination_namespace =
-      destination_node_info && destination_node_info->namespace_()
-          ? destination_node_info->namespace_()->str()
+      destination_node_info.namespace_()
```
mandarjog, line 549: I don’t think this is used Or needed anymore. It can be removed if not needed.

extensions/common/context.cc
```diff
@@ -295,13 +269,69 @@ flatbuffers::DetachedBuffer extractLocalNodeFlatBuffer() {
...
+  std::string endpoint_labels;
+  fallback_peer_node_ = extractEmptyNodeFlatBuffer();
+  if (!getValue(
+          {"upstream_host_metadata", "filter_metadata", "istio", "workload"},
+          &endpoint_labels)) {
+    return;
+  }
+  std::vector<std::string_view> parts = absl::StrSplit(endpoint_labels, ';');
+  // workload label should semicolon separated four parts string:
+  // workload_name|namespace|canonical_service|canonical_revision.
```
mandarjog, line 291: Clearer to use workload_name;namespace; ... instead of |

mandarjog: Perhaps create  ‘ flatbuffers::DetachedBuffer parseNodeFromString(std::string)’ ?

extensions/common/context.cc
```diff
@@ -123,28 +88,37 @@ void getDestinationService(const std::string& dest_namespace,
     return;
   }
 
-  std::vector<std::string_view> parts = absl::StrSplit(cluster_name, '|');
-  if (parts.size() == 4) {
-    *dest_svc_host = std::string(parts[3].data(), parts[3].size());
+  // Get destination service name and host from cluster labels.
+  if (getValue({"cluster_metadata", "filter_metadata", "istio", "services", "0",
```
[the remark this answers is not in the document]
bianpengyuan: Done.

extensions/common/context.cc
bianpengyuan: Thanks for catching this! I think it is fine to keep it as is and do not extract it out, since the logic is not shared anywhere.

extensions/stats/plugin.cc
bianpengyuan: Thanks, done.

extensions/common/context.cc
bianpengyuan: Done.

extensions/common/context.cc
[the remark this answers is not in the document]
bianpengyuan: Done.

extensions/common/context.cc
```diff
@@ -123,28 +88,37 @@ void getDestinationService(const std::string& dest_namespace,
...
+                "name"},
+               dest_svc_name)) {
+    getValue({"cluster_metadata", "filter_metadata", "istio", "services", "0",
+              "host"},
+             dest_svc_host);
+  } else if (use_host_header) {
+    // if cluster metadata cannot be found, fallback to host header.
```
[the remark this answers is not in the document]
bianpengyuan: I updated the comment here as well as the comment of this function. I think now this will only happen when no cluster is matched for the request. wildcard cluster is handled above.

extensions/common/context.cc
```diff
@@ -295,13 +277,75 @@ flatbuffers::DetachedBuffer extractLocalNodeFlatBuffer() {
...
+  node.add_workload_name(workload_name);
+  node.add_namespace_(namespace_);
+  node.add_labels(labels_offset);
+  auto data = node.Finish();
+  fbb.Finish(data);
+  fallback_peer_node_ = fbb.Release();
+}
+
+const ::Wasm::Common::FlatNode& PeerNodeInfo::get() const {
+  if (!peer_node_.empty()) {
```
kyessenov, line 336: `if (found_)` may be better here

extensions/common/context.cc
```diff
@@ -295,13 +277,75 @@ flatbuffers::DetachedBuffer extractLocalNodeFlatBuffer() {
   return fbb.Release();
 }
 
+PeerNodeInfo::PeerNodeInfo(const std::string_view peer_metadata_id_key,
+                           const std::string_view peer_metadata_key) {
+  found_ = getValue({peer_metadata_id_key}, &peer_id_);
+  if (found_) {
+    getValue({peer_metadata_key}, &peer_node_);
```
kyessenov, line 284: I think we may want to send just the ID and not node once Metadata Discovery is there. It would be used as a lookup key.

testdata/bootstrap/client.yaml.tmpl
```diff
@@ -40,6 +40,13 @@ static_resources:
     connect_timeout: 1s
     type: STATIC
     http2_protocol_options: {}
+    metadata:
```
kyessenov, line 43: Can you guard this for now with `{{ if }}`, so we make sure that MX is actually exercised? It's hard to tell if it always falls back.

testdata/metric/client_request_total_endpoint_labels.yaml.tmpl
```diff
@@ -0,0 +1,64 @@
...
+    value: grpc
+    {{- else }}
+    value: http
+    {{- end }}
+  - name: response_code
+    value: "503"
+  - name: grpc_response_status
+    value: "{{ .Vars.GrpcResponseStatus }}"
+  - name: response_flags
+    value: "UF"
```
kyessenov, line 54: Why is it failing upstream?

bianpengyuan: I deliberately disabled server setup to not enable metadata exchange. But I guess it is confusing. I will change it to just disable metadata exchange.

extensions/common/context.cc
```diff
@@ -295,13 +277,75 @@ flatbuffers::DetachedBuffer extractLocalNodeFlatBuffer() {
...
+  }
+  std::vector<std::string_view> parts = absl::StrSplit(endpoint_labels, ';');
+  // workload label should semicolon separated four parts string:
+  // workload_name;namespace;canonical_service;canonical_revision.
+  if (parts.size() < 4) {
+    return;
+  }
+
+  flatbuffers::FlatBufferBuilder fbb;
+  flatbuffers::Offset<flatbuffers::String> workload_name, namespace_;
```
gargnupur, line 310: nit: no need of underscore in namespace?

extensions/common/context.cc
```diff
@@ -295,13 +277,75 @@ flatbuffers::DetachedBuffer extractLocalNodeFlatBuffer() {
...
+  }
+  auto labels_offset = fbb.CreateVectorOfSortedTables(&labels);
+
+  FlatNodeBuilder node(fbb);
+  node.add_workload_name(workload_name);
+  node.add_namespace_(namespace_);
+  node.add_labels(labels_offset);
+  auto data = node.Finish();
+  fbb.Finish(data);
+  fallback_peer_node_ = fbb.Release();
```
gargnupur, line 332: should found_ be updated here? we don't want to wait for metadata exchange in this case?

extensions/common/context.cc
bianpengyuan: Done.

testdata/bootstrap/client.yaml.tmpl
bianpengyuan: I think you mean the endpoint labels? I added a wrapper to the endpoint.

testdata/metric/client_request_total_endpoint_labels.yaml.tmpl
bianpengyuan: Fixed.

extensions/common/context.cc
bianpengyuan: namespace is reserved key word, so the generated code suffixed it to work around.

extensions/common/context.cc
bianpengyuan: found_ here means finding it from filter metadata (written by metadata exchange filter). I think we'd still want to wait for metadata exchange to happen, as the fallback peer node only has limited information (workload name and namespace, canonical labels) due to performance concern.

extensions/common/context.cc
bianpengyuan: Yeah, that sounds right to me.

bianpengyuan: Although that relies on metadata service would have all peer information, I am not sure that would be the case, for example multi cluster with multiple control planes.

extensions/common/context.cc
gargnupur: sounds good

testdata/bootstrap/client.yaml.tmpl
kyessenov: Can you change the name then of the cluster? I think you imply that the cluster metadata is mandatory, which is fine, but need to be clear.

kyessenov (commented):
LG from my side. Let's make sure we don't lose coverage and avoid fallback in other cases.

testdata/bootstrap/client.yaml.tmpl
bianpengyuan: Done.

mandarjog (approved):
lgtm

bianpengyuan:
@kyessenov Can I get a approval stamp on this? Thanks!

kyessenov (approved):

PR #3056, opened 2020-10-19
fix: implement metadata fallback on the special value

kyessenov:

- Fix handling of TCP metadata exchange in fallback extraction. Fallback if the metadata ID is the special "not found" value. Fixes a crash with a memory access into an empty string flatbuffer.
- Minor simplifications in stats plugin.

### DIFF
--- a/extensions/common/context.h
+++ b/extensions/common/context.h
@@ -217,7 +217,7 @@ flatbuffers::DetachedBuffer extractEmptyNodeFlatBuffer();
 flatbuffers::DetachedBuffer extractLocalNodeFlatBuffer();
 
 // Extract upstream peer metadata from upstream host metadata.
-// Returns true if the metadat is found in the upstream host metadata.
+// Returns true if the metadata is found in the upstream host metadata.
 bool extractPeerMetadataFromUpstreamHostMetadata(
     flatbuffers::FlatBufferBuilder& fbb);
 
--- a/extensions/common/context.h
+++ b/extensions/common/context.h
@@ -32,6 +32,9 @@ constexpr std::string_view kUpstreamMetadataKey = "upstream_peer";
 constexpr std::string_view kDownstreamMetadataIdKey = "downstream_peer_id";
 constexpr std::string_view kDownstreamMetadataKey = "downstream_peer";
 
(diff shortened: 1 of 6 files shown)


extensions/common/context.h
```diff
@@ -213,19 +216,33 @@ flatbuffers::DetachedBuffer extractEmptyNodeFlatBuffer();
...
   const ::Wasm::Common::FlatNode& get() const;
-  const std::string& getId() const { return peer_id_; }
+  const std::string& id() const { return peer_id_; }
+
+  // Found indicates whether both ID and metadata is available.
   bool found() const { return found_; }
 
+  // Maybe waiting indicates that the metadata is not found but may arrive
+  // later.
+  bool maybeWaiting() const {
```
douglas-reid, line 240: super nit: would this be clearer as something like: `notReceived` or `Unknown`, to eliminate potential confusion around passive vs active waiting?

bianpengyuan: notReceived or unknown seems to be also confusing. Not received implies `peer_id_ == kMetadataNotFoundValue` instead of `!=`. `unknown` makes more sense to me.

kyessenov: It's closer to the usage in plugins, which check whether it's too early to report.

bianpengyuan (approved):
lgtm

extensions/common/context.h
gargnupur: Having `wait` there seems good to me, because we don't report and wait to receive for metadata.. something like `waitToReceive`?

extensions/stats/config.proto
```diff
@@ -68,6 +68,7 @@ message PluginConfig {
   // next id: 7
   // The following settings should be rarely used.
   // Enable debug for this filter.
+  // DEPRECATED.
```
gargnupur, line 71: hide from docs?

kyessenov:
Did this merge too fast?

extensions/stats/config.proto
kyessenov: I think we should keep it to reduce confusion for now. Deprecated fields must stay until we can actually remove usages.

PiotrSikora:
> Did this merge too fast?

No, it was approved and all tests passed.

PR #3058, opened 2020-10-20
generalize interval reporting to HTTP

kyessenov:

Apologizing beforehand for the size of this PR.
This change enables recurrent reporting for all metrics, not just TCP. It is used specifically by gRPC streaming metrics.

There is also a bunch of other blocking issues and cleanups combined:
- define protocol enum to compress storage and enable union/intersect operations
- define recurrent property in metric families, that enables reporting mid-stream
- pre-populate request info with protocol and host before detaching it to root context
- metric extractor function mutates individual info fields instead of resetting in bulk
- await for response headers (not just request headers) before placing on the queue to make sure HTTP metadata is exchanged; upstream host should be selected at this time for the fallback
- cleanup: remove shared pointer for request_info, stream context should always call `onLog` ( if it doesn't, that's a critical bug)
- cleanup: remove special handling of labels in TCP by changing the end index in the metric
- cleanup: delete empty stubs

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -59,31 +59,28 @@ namespace {
 // * Otherwise, try fetching cluster metadata for destination service name and
 //   host. If cluster metadata is not available, set destination service name
 //   the same as destination service host.
-void getDestinationService(bool use_host_header, std::string* dest_svc_host,
-                           std::string* dest_svc_name,
-                           const std::string& cluster_name,
-                           const std::string& route_name) {
-  *dest_svc_host = use_host_header
-                       ? getHeaderMapValue(WasmHeaderMapType::RequestHeaders,
-                                           kAuthorityHeaderKey)
-                             ->toString()
-                       : "unknown";
+void populateDestinationService(bool use_host_header,
+                                RequestInfo* request_info) {
+  request_info->destination_service_host =
+      use_host_header ? request_info->url_host : "unknown";
 
   // override the cluster name if this is being sent to the
   // blackhole or passthrough cluster
+  const std::string& route_name = request_info->route_name;
   if (route_name == kBlackHoleRouteName) {
-    *dest_svc_name = kBlackHoleCluster;
+    request_info->destination_service_name = kBlackHoleCluster;
     return;
   } else if (route_name == kPassThroughRouteName) {
-    *dest_svc_name = kPassThroughCluster;
+    request_info->destination_service_name = kPassThroughCluster;
     return;
   }
 
+  const std::string& cluster_name = request_info->upstream_cluster;
   if (cluster_name == kBlackHoleCluster ||
       cluster_name == kPassThroughCluster ||
       cluster_name == kInboundPassthroughClusterIpv4 ||
       cluster_name == kInboundPassthroughClusterIpv6) {
-    *dest_svc_name = cluster_name;
+    request_info->destination_service_name = cluster_name;
     return;
   }
 
@@ -105,63 +102,57 @@ void getDestinationService(bool use_host_header, std::string* dest_svc_host,
   // identify the intended host.
   if (getValue({"cluster_metadata", "filter_metadata", "istio", "services", "0",
                 "name"},
-               dest_svc_name)) {
+               &request_info->destination_service_name)) {
     getValue({"cluster_metadata", "filter_metadata", "istio", "services", "0",
               "host"},
-             dest_svc_host);
+             &request_info->destination_service_host);
   } else {
     // if cluster metadata cannot be found, fallback to destination service
     // host. If host header fallback is enabled, this will be host header. If
     // host header fallback is disabled, this will be unknown. This could happen
     // if a request does not route to any cluster.
-    *dest_svc_name = *dest_svc_host;
+    request_info->destination_service_name =
+        request_info->destination_service_host;
   }
-}
-
-void populateRequestInfo(bool outbound, bool use_host_header_fallback,
-                         RequestInfo* request_info) {
-  request_info->is_populated = true;
-  getValue({"cluster_name"}, &request_info->upstream_cluster);
-  getValue({"route_name"}, &request_info->route_name);
-  // Fill in request info.
-  // Get destination service name and host based on cluster name and host
-  // header.
-  getDestinationService(
-      use_host_header_fallback, &request_info->destination_service_host,
-      &request_info->destination_service_name, request_info->upstream_cluster,
-      request_info->route_name);
-
-  getValue({"request", "url_path"}, &request_info->request_url_path);
-
-  uint64_t destination_port = 0;
-  if (outbound) {
-    getValue({"upstream", "port"}, &destination_port);
-    getValue({"upstream", "uri_san_peer_certificate"},
-             &request_info->destination_principal);
-    getValue({"upstream", "uri_san_local_certificate"},
-             &request_info->source_principal);
-  } else {
-    getValue({"destination", "port"}, &destination_port);
-
-    bool mtls = false;
-    if (getValue({"connection", "mtls"}, &mtls)) {
-      request_info->service_auth_policy =
-          mtls ? ::Wasm::Common::ServiceAuthenticationPolicy::MutualTLS
-               : ::Wasm::Common::ServiceAuthenticationPolicy::None;
-    }
-    getValue({"connection", "uri_san_local_certificate"},
-             &request_info->destination_principal);
-    getValue({"connection", "uri_san_peer_certificate"},
-             &request_info->source_principal);
-  }
-  request_info->destination_port = destination_port;
-
-  uint64_t response_flags = 0;
-  getValue({"response", "flags"}, &response_flags);
-  request_info->response_flag = parseResponseFlag(response_flags);
 }
 
 }  // namespace
+
+void populateRequestInfo(bool outbound, bool use_host_header_fallback,
+                         RequestInfo* request_info) {
+  if (!request_info->is_populated) {
+    request_info->is_populated = true;
+
+    getValue({"cluster_name"}, &request_info->upstream_cluster);
+    getValue({"route_name"}, &request_info->route_name);
+    // Fill in request info.
+    // Get destination service name and host based on cluster name and host
+    // header.
+    populateDestinationService(use_host_header_fallback, request_info);
+    uint64_t destination_port = 0;
+    if (outbound) {
+      getValue({"upstream", "port"}, &destination_port);
+      getValue({"upstream", "uri_san_peer_certificate"},
+               &request_info->destination_principal);
+      getValue({"upstream", "uri_san_local_certificate"},
+               &request_info->source_principal);
+    } else {
+      getValue({"destination", "port"}, &destination_port);
+
+      bool mtls = false;
+      if (getValue({"connection", "mtls"}, &mtls)) {
+        request_info->service_auth_policy =
+            mtls ? ::Wasm::Common::ServiceAuthenticationPolicy::MutualTLS
+                 : ::Wasm::Common::ServiceAuthenticationPolicy::None;
+      }
+      getValue({"connection", "uri_san_local_certificate"},
+               &request_info->destination_principal);
+      getValue({"connection", "uri_san_peer_certificate"},
+               &request_info->source_principal);
+    }
+    request_info->destination_port = destination_port;
+  }
+}
 
 std::string_view AuthenticationPolicyString(
     ServiceAuthenticationPolicy policy) {
@@ -184,6 +175,20 @@ std::string_view TCPConnectionStateString(TCPConnectionState state) {
       return kConnected;
     case TCPConnectionState::Close:
       return kClose;
+    default:
+      break;
+  }
+  return {};
+}
+
+std::string_view ProtocolString(Protocol protocol) {
+  switch (protocol) {
+    case Protocol::TCP:
+      return kProtocolTCP;
+    case Protocol::HTTP:
+      return kProtocolHTTP;
+    case Protocol::GRPC:
+      return kProtocolGRPC;
     default:
       break;
   }
@@ -362,6 +367,8 @@ const ::Wasm::Common::FlatNode& PeerNodeInfo::get() const {
 // Normally it is ok to use host header within the mesh, but not at ingress.
 void populateHTTPRequestInfo(bool outbound, bool use_host_header_fallback,
                              RequestInfo* request_info) {
+  populateRequestProtocol(request_info);
+  getValue({"request", "url_path"}, &request_info->request_url_path);
   populateRequestInfo(outbound, use_host_header_fallback, request_info);
 
   int64_t response_code = 0;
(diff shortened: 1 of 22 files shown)


extensions/common/context.cc
```diff
@@ -105,64 +102,58 @@ void getDestinationService(bool use_host_header, std::string* dest_svc_host,
...
-             &request_info->source_principal);
-  } else {
-    getValue({"destination", "port"}, &destination_port);
-
-    bool mtls = false;
-    if (getValue({"connection", "mtls"}, &mtls)) {
-      request_info->service_auth_policy =
-          mtls ? ::Wasm::Common::ServiceAuthenticationPolicy::MutualTLS
-               : ::Wasm::Common::ServiceAuthenticationPolicy::None;
+  if (!request_info->is_populated) {
```
gargnupur, line 123: return if it's populated?

extensions/common/context.h
```diff
@@ -62,19 +62,26 @@ constexpr std::string_view kLatest = "latest";
 const std::set<std::string> kGrpcContentTypes{
     "application/grpc", "application/grpc+proto", "application/grpc+json"};
 
-enum class ServiceAuthenticationPolicy : int64_t {
+enum class ServiceAuthenticationPolicy : int8_t {
```
gargnupur, line 65: change to uint?

extensions/common/context.h
```diff
@@ -174,12 +182,12 @@ struct RequestInfo {
   std::string url_scheme;
 
   // TCP variables.
-  int64_t tcp_connections_opened = 0;
-  int64_t tcp_connections_closed = 0;
-  int64_t tcp_sent_bytes = 0;
-  int64_t tcp_received_bytes = 0;
-  int64_t tcp_total_sent_bytes = 0;
-  int64_t tcp_total_received_bytes = 0;
+  uint8_t tcp_connections_opened = 0;
```
gargnupur, line 185: in case of a leak, this can go beyond 255.. -> https://github.com/istio/istio/issues/24720#issuecomment-674147639

extensions/common/context.cc
```diff
@@ -105,64 +102,58 @@ void getDestinationService(bool use_host_header, std::string* dest_svc_host,
...
     }
-    getValue({"connection", "uri_san_local_certificate"},
-             &request_info->destination_principal);
-    getValue({"connection", "uri_san_peer_certificate"},
-             &request_info->source_principal);
+    request_info->destination_port = destination_port;
   }
-  request_info->destination_port = destination_port;
-
-  uint64_t response_flags = 0;
```
gargnupur, line 159: why remove from here?

extensions/stats/plugin.cc
```diff
@@ -190,60 +186,89 @@ const std::vector<MetricTag>& PluginRootContext::defaultTags() {
...
+                    true},
+
       // TCP metrics.
       MetricFactory{"tcp_sent_bytes_total", MetricType::Counter,
-                    [](const ::Wasm::Common::RequestInfo& request_info)
-                        -> uint64_t { return request_info.tcp_sent_bytes; },
-                    true, count_standard_labels},
+                    [](::Wasm::Common::RequestInfo& request_info) -> uint64_t {
+                      uint64_t out = 0;
+                      std::swap(out, request_info.tcp_sent_bytes);
```
gargnupur, line 245: This is cool!

extensions/stats/plugin.cc
```diff
@@ -561,17 +578,32 @@ bool PluginRootContext::report(::Wasm::Common::RequestInfo& request_info,
...
+          outbound_, useHostHeaderFallback(), &request_info);
+    } else {
+      // Await for upstream host selection to resolve destination service.
+      int64_t port;
+      if (!getValue({"upstream", "port"}, &port)) {
+        return;
+      }
+      ::Wasm::Common::populateRequestInfo(outbound_, useHostHeaderFallback(),
+                                          &request_info);
+      if (request_info.request_protocol == Protocol::GRPC) {
```
gargnupur, line 599: curious, can report be called for HTTP onTick?

douglas-reid:
@kyessenov haven't looked through the PR yet, but had a question based on the description: does this potentially unblock the potential to report in-flight request metrics for HTTP (which I believe has been requested elsewhere)?

extensions/common/context.h
douglas-reid, line 185: curious: why drop size?

extensions/stackdriver/edges/edge_reporter.cc
```diff
@@ -155,11 +155,10 @@ void EdgeReporter::addEdge(const ::Wasm::Common::RequestInfo& request_info,
   edge->mutable_destination()->CopyFrom(node_instance_);
 
   auto protocol = request_info.request_protocol;
-  if (protocol == "http" || protocol == "HTTP") {
+  // TODO: add support for HTTPS
```
douglas-reid, line 158: what is the plan here?

extensions/stackdriver/log/logger_test.cc
```diff
@@ -161,7 +161,7 @@ std::string write_audit_request_json = R"({
            "referer":"www.google.com",
            "serverIp":"2.2.2.2",
            "latency":"10s",
-           "protocol":"HTTP"
+           "protocol":"http"
```
douglas-reid, line 164: are we concerned here about the possibility of breaking any filters, etc. ?

extensions/stats/plugin.cc
```diff
@@ -190,60 +186,89 @@ const std::vector<MetricTag>& PluginRootContext::defaultTags() {
...
-          false, count_standard_labels},
+      MetricFactory{"requests_total", MetricType::Counter,
+                    [](::Wasm::Common::RequestInfo&) -> uint64_t { return 1; },
+                    static_cast<uint32_t>(Protocol::HTTP) |
+                        static_cast<uint32_t>(Protocol::GRPC),
+                    count_standard_labels, false},
+      MetricFactory{"request_duration_milliseconds", MetricType::Histogram,
+                    [](::Wasm::Common::RequestInfo& request_info) -> uint64_t {
+                      return request_info.duration /* in nanoseconds */ /
+                             1000000;
```
douglas-reid, line 197: i wonder if we should set this as a const somewhere and name it something like nanos_per_millisecond or something.

extensions/stats/plugin.h
```diff
@@ -75,10 +74,10 @@ const std::string default_stat_prefix = "istio";
   FIELD_FUNC(destination_canonical_service)  \
   FIELD_FUNC(destination_canonical_revision) \
   FIELD_FUNC(request_protocol)               \
-  FIELD_FUNC(response_code)                  \
-  FIELD_FUNC(grpc_response_status)           \
   FIELD_FUNC(response_flags)                 \
-  FIELD_FUNC(connection_security_policy)
+  FIELD_FUNC(connection_security_policy)     \
```
douglas-reid, line 78: since field ordering here is important, can we add some comments / warning here in case we decide to remove fields or add new ones, etc ?

extensions/stats/plugin.cc
```diff
@@ -190,60 +186,89 @@ const std::vector<MetricTag>& PluginRootContext::defaultTags() {
...
-          "request_duration_milliseconds", MetricType::Histogram,
-          [](const ::Wasm::Common::RequestInfo& request_info) -> uint64_t {
-            return request_info.duration /* in nanoseconds */ / 1000000;
-          },
-          false, count_standard_labels},
+      MetricFactory{"requests_total", MetricType::Counter,
+                    [](::Wasm::Common::RequestInfo&) -> uint64_t { return 1; },
+                    static_cast<uint32_t>(Protocol::HTTP) |
+                        static_cast<uint32_t>(Protocol::GRPC),
+                    count_standard_labels, false},
```
douglas-reid, line 193: nit: worth adding `/* not periodic */` or whatever for the `false` and similar for the `true` values in these definitions?

extensions/stats/plugin.cc
```diff
@@ -190,60 +186,89 @@ const std::vector<MetricTag>& PluginRootContext::defaultTags() {
...
+                    static_cast<uint32_t>(Protocol::TCP), count_tcp_labels,
+                    true},
+      MetricFactory{"tcp_connections_opened_total", MetricType::Counter,
+                    [](::Wasm::Common::RequestInfo& request_info) -> uint64_t {
+                      uint8_t out = 0;
+                      std::swap(out, request_info.tcp_connections_opened);
+                      return out;
+                    },
+                    static_cast<uint32_t>(Protocol::TCP), count_tcp_labels,
+                    true},
```
douglas-reid, line 265: curious: why is `connections_opened_total` defined as `recurrent==true` and `connections_closed_total` defined as `recurrent==false` ?

extensions/common/context.cc
kyessenov: Done

extensions/common/context.h
kyessenov: Done

extensions/common/context.cc
kyessenov: This is called repeatedly in report(). The first time it's called, response flag is going to be 0, and be set to 0 till the end.

extensions/stats/plugin.cc
kyessenov: Thanks.

extensions/stats/plugin.cc
kyessenov: Yes, see the test. It's called for gRPC streams only right now.

extensions/common/context.h
kyessenov: This is really a boolean from what I can tell.

extensions/stackdriver/edges/edge_reporter.cc
kyessenov: I haven't seen it being set to HTTPS or https anywhere, so it was dead code. I placed TODO to actually implement it.

extensions/stackdriver/log/logger_test.cc
kyessenov: I think our actual extensions already send lowercase strings based on integration tests. I just made the unit test match it.

extensions/stats/plugin.cc
kyessenov: Yeah, I can do that.

extensions/stats/plugin.h
kyessenov: Done.

extensions/stats/plugin.cc
kyessenov: Connection opened is reported on timer, but connection closed is only set at the end in onLog.

kyessenov:
@douglas-reid Yes, we can report intermediate metrics for long running HTTP calls. We don't have a good metric yet (sizes are histograms so require final count).

extensions/common/context.cc
kyessenov: BTW I moved it both TCP and HTTP populate calls instead.

extensions/stats/plugin.cc
```diff
@@ -561,17 +579,27 @@ bool PluginRootContext::report(::Wasm::Common::RequestInfo& request_info,
     uint64_t response_flags = 0;
     getValue({"response", "flags"}, &response_flags);
```
kyessenov, line 580: @gargnupur I wonder if there is any difference between this check and just looking at `end_stream`. We just want to stop waiting to report until the end, right?

gargnupur: When response flags is set.. `end_stream` would be set right? In that case, it should be fine.
We just wanted to stop waiting in error but stopping when `end_stream` is reached should also work.

kyessenov: Ok, simplified to stop waiting on the end stream.

kyessenov:
This is ready for review. PTAL.

extensions/stats/plugin.h
```diff
@@ -307,45 +325,51 @@ class PluginRootContextInbound : public PluginRootContext {
...
 
-  FilterStatus onNewConnection() override {
-    if (!rootContext()->initialized()) {
-      return FilterStatus::Continue;
+  // HTTP streams start with headers.
+  FilterHeadersStatus onRequestHeaders(uint32_t, bool) override {
+    ::Wasm::Common::populateRequestProtocol(&request_info_);
+    // Save host value for recurrent reporting
+    if (rootContext()->useHostHeaderFallback()) {
+      getValue({"request", "host"}, &request_info_.url_host);
```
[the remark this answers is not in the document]
kyessenov: Yeah, it's the same thing.

extensions/common/context.h
```diff
@@ -62,19 +62,26 @@ constexpr std::string_view kLatest = "latest";
...
 
-enum class ServiceAuthenticationPolicy : int64_t {
+enum class ServiceAuthenticationPolicy : uint8_t {
   Unspecified = 0,
   None = 1,
   MutualTLS = 2,
 };
 
-enum class TCPConnectionState : int64_t {
+enum class TCPConnectionState : int8_t {
```
gargnupur, line 71: nit: uint?

extensions/common/context.h
```diff
@@ -248,6 +258,12 @@ class PeerNodeInfo {
...
 
+// Populate shared information between all protocols.
+// Requires that the connections are established both downstrean and upstream.
+// Caches computation using is_populated field.
+void populateRequestInfo(bool outbound, bool use_host_header_fallback,
+                         RequestInfo* request_info);
+
 // populateHTTPRequestInfo populates the RequestInfo struct. It needs access to
 // the request context.
 void populateHTTPRequestInfo(bool outbound, bool use_host_header,
```
gargnupur, line 269: nit: should this really be populateHTTPAndGrpcRequestInfo?

extensions/common/context.cc
```diff
@@ -449,7 +454,25 @@ void populateTCPRequestInfo(bool outbound, RequestInfo* request_info) {
   // host_header_fallback is for HTTP/gRPC only.
   populateRequestInfo(outbound, false, request_info);
 
-  request_info->request_protocol = kProtocolTCP;
+  uint64_t response_flags = 0;
+  if (getValue({"response", "flags"}, &response_flags)) {
+    request_info->response_flag = parseResponseFlag(response_flags);
+  }
+
+  request_info->request_protocol = Protocol::TCP;
```
gargnupur, line 462: Don't need this? as we set it in onNewConnection and same for HTTP and Grpc.. they are getting set in onRequestHeaders?

gargnupur:
Looks good.. just some minor nits...

extensions/common/context.h
kyessenov: Done.

extensions/common/context.h
kyessenov: We treat gRPC as a subset of HTTP throughout. I think it's OK for now, since it's guarded by protocol field.

extensions/common/context.cc
kyessenov: It's not set in stackdriver. I suppose we could move this line into stackdriver plugin itself.

extensions/stats/plugin.h
```diff
@@ -307,45 +325,51 @@ class PluginRootContextInbound : public PluginRootContext {
...
   };
 
-  FilterStatus onNewConnection() override {
-    if (!rootContext()->initialized()) {
-      return FilterStatus::Continue;
+  // HTTP streams start with headers.
+  FilterHeadersStatus onRequestHeaders(uint32_t, bool) override {
+    ::Wasm::Common::populateRequestProtocol(&request_info_);
+    // Save host value for recurrent reporting
+    if (rootContext()->useHostHeaderFallback()) {
```
bianpengyuan, line 343: It is not clear to me why we fetch request host here? Also since we are fill in url host header here, we should remove it from `populateExtendedHTTPRequestInfo`.

kyessenov: It's filled here because onTick will happen outside onRequestHeaders, and url_host is only available at either onRequestHeader or onLog. URL host is needed for metadata fallback.

I didn't remove it from extended info because StackDriver does not set it separately (same cause as the one @gargnupur pointed out for TCP).

bianpengyuan: Ah I see, I think `url_host is only available at either onRequestHeader or onLog` worths to mentioning in the comment.

kyessenov: Done. TBH there is no reason why it shouldn't be available. Just ABI issue.

bianpengyuan (approved):

gargnupur (approved):

Kyslik:
Is this going to land in Istio 1.6 / Istio 1.7?

kyessenov:
@Kyslik We generally don't backport new features. Only bug fixes for security vulnerabilities.

Kyslik:
> @Kyslik We generally don't backport new features. Only bug fixes for security vulnerabilities.

Isn't this fixing a memory leak reported in https://github.com/istio/istio/issues/24720?

kyessenov:
No, I think you're thinking of the last commit in https://github.com/istio/envoy/pull/284/commits

Kyslik:
On the https://github.com/istio/istio/issues/24720 there is link to this PR 🤷, no other PR was merged so I did not connect the dots correctly. Thank you for pointing me to correct PR.

PR #3076, opened 2020-10-29
Fix useragent in stackdriver access log

gargnupur:

Fix useragent in stackdriver access log -> was using wrong CEL name..

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -414,7 +414,7 @@ void populateExtendedHTTPRequestInfo(RequestInfo* request_info) {
   populateExtendedRequestInfo(request_info);
 
   getValue({"request", "referer"}, &request_info->referer);
-  getValue({"request", "user_agent"}, &request_info->user_agent);
+  getValue({"request", "useragent"}, &request_info->user_agent);
   getValue({"request", "id"}, &request_info->request_id);
   std::string trace_sampled;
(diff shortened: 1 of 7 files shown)


kyessenov (approved):

bianpengyuan (approved):

bianpengyuan:
Can we cherry-pick this to older releases?

gargnupur:
> Can we cherry-pick this to older releases?

Added for 1.7 and 1.6..

extensions/common/context.cc
```diff
@@ -414,7 +414,7 @@ void populateExtendedHTTPRequestInfo(RequestInfo* request_info) {
   populateExtendedRequestInfo(request_info);
 
   getValue({"request", "referer"}, &request_info->referer);
-  getValue({"request", "user_agent"}, &request_info->user_agent);
+  getValue({"request", "useragent"}, &request_info->user_agent);
```
douglas-reid, line 417: @gargnupur @kyessenov is there some sort of linter we could write that would catch these issues at review time?

gargnupur: Yeah something that will sync with source/extensions/filters/common/expr/context.h, to make sure we are using the right fields...

kyessenov: Yeah, although here specifically it's using ABI, not CEL. ABI usage might be hard to statically verify.

PR #3092, opened 2020-11-06
Fix export cache

howardjohn:
Currently it copys a non-writeable file, so next time we export it
fails. Also do not fail if wasm is not there, as its optional.

s**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #


lambdai (approved):
LGTM

PR #3100, opened 2020-11-13
Add Rbac Access Denied Policy Info in Access logs

gargnupur:
Add Rbac Access Denied Policy Info in Access logs

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -428,6 +428,7 @@ void populateExtendedHTTPRequestInfo(RequestInfo* request_info) {
   getValue({"request", "url_path"}, &request_info->url_path);
   getValue({"request", "host"}, &request_info->url_host);
   getValue({"request", "scheme"}, &request_info->url_scheme);
+  getValue({"response", "code_details"}, &request_info->response_code_details);
 }
 
 void populateExtendedRequestInfo(RequestInfo* request_info) {
@@ -448,6 +449,8 @@ void populateExtendedRequestInfo(RequestInfo* request_info) {
       envoy_original_dst_host ? envoy_original_dst_host->toString() : "";
   getValue({"upstream", "transport_failure_reason"},
            &request_info->upstream_transport_failure_reason);
+  getValue({"connection", "termination_details"},
+           &request_info->connection_termination_details);
 }
 
 void populateTCPRequestInfo(bool outbound, RequestInfo* request_info) {
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -428,7 +428,11 @@ void populateExtendedHTTPRequestInfo(RequestInfo* request_info) {
   getValue({"request", "url_path"}, &request_info->url_path);
   getValue({"request", "host"}, &request_info->url_host);
   getValue({"request", "scheme"}, &request_info->url_scheme);
-  getValue({"response", "code_details"}, &request_info->response_code_details);
+  std::string response_details;
+  getValue({"response", "code_details"}, &response_details);
+  if (!response_details.empty()) {
+    request_info->response_details = response_details;
+  }
 }
 
 void populateExtendedRequestInfo(RequestInfo* request_info) {
@@ -449,8 +453,11 @@ void populateExtendedRequestInfo(RequestInfo* request_info) {
       envoy_original_dst_host ? envoy_original_dst_host->toString() : "";
   getValue({"upstream", "transport_failure_reason"},
            &request_info->upstream_transport_failure_reason);
-  getValue({"connection", "termination_details"},
-           &request_info->connection_termination_details);
+  std::string response_details;
+  getValue({"connection", "termination_details"}, &response_details);
+  if (!response_details.empty()) {
+    request_info->response_details = response_details;
+  }
 }
 
 void populateTCPRequestInfo(bool outbound, RequestInfo* request_info) {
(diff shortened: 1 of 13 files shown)


extensions/stackdriver/log/logger.cc
```diff
@@ -31,6 +32,13 @@ namespace Extensions {
 namespace Stackdriver {
 namespace Log {
 namespace {
+// Matches Rbac Access denied string.
+// It is of the format:
+// "rbac_access_denied_matched_policy[ns[NAMESPACE]-policy[POLICY]-rule[POLICY_INDEX]]"
+const RE2 rbac_denied_match(
+    "rbac_access_denied_matched_policy\\[ns\\[(.*)\\]-policy\\[(.*)\\]-rule\\[("
+    ".*)\\]\\]");
+constexpr char kRbacAccessDenied[] = "RbacAccessDenied";
```
yangminzhu, line 41: will this be exposed directly to user on UI? If so we should call it something like "AuthzDenied", otherwise it's fine either way if it is used internally only.

yangminzhu (approved):
Looks great! just a small comment, thanks.

extensions/stackdriver/log/logger.cc
```diff
@@ -341,6 +364,17 @@ void Logger::fillAndFlushLogEntry(
       (*label_map)["upstream_transport_failure_reason"] =
           request_info.upstream_transport_failure_reason;
     }
+    if (!request_info.response_code_details.empty() ||
+        !request_info.connection_termination_details.empty()) {
+      const std::string response_details =
+          !request_info.response_code_details.empty()
+              ? request_info.response_code_details
```
kyessenov, line 371: I think you could re-use the same field for both.

testdata/filters/rbac.yaml.tmpl
```diff
@@ -6,7 +6,7 @@
       rules:
         action: DENY
         policies:
-          "test":
+          ns[foo]-policy[httpbin-deny]-rule[0]:
```
kyessenov, line 9: Do we need an example for network RBAC?

extensions/stackdriver/log/logger.cc
```diff
@@ -31,6 +32,13 @@ namespace Extensions {
 namespace Stackdriver {
 namespace Log {
 namespace {
+// Matches Rbac Access denied string.
+// It is of the format:
+// "rbac_access_denied_matched_policy[ns[NAMESPACE]-policy[POLICY]-rule[POLICY_INDEX]]"
+const RE2 rbac_denied_match(
+    "rbac_access_denied_matched_policy\\[ns\\[(.*)\\]-policy\\[(.*)\\]-rule\\[("
```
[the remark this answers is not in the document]
gargnupur: /cc @yangminzhu 

We wanted to extract policy name and rule, hence the suffix...
what is the exact scenario you thinking Kuat?

extensions/stackdriver/log/logger.cc
gargnupur: Yes, fixed

testdata/filters/rbac.yaml.tmpl
gargnupur: Should be similar to this one with response to extracting stuff? I can add if needed..

extensions/stackdriver/log/logger.cc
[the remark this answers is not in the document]
kyessenov: Custom RBAC rules with EnvoyFilter. I guess we could have a fallback in case regex does not match, just log the whole thing.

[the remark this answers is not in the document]
yangminzhu: I'm fine either way, I think we normally don't do anything special to handle customized config from EnvoyFilter but I think it also doesn't harm anything with a fallback here.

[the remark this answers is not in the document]
gargnupur: yeah we are actually logging the whole thing if regex doesn't match.

extensions/stackdriver/log/logger.cc
gargnupur: Thanks.. Fixed

extensions/common/context.cc
```diff
@@ -428,6 +428,11 @@ void populateExtendedHTTPRequestInfo(RequestInfo* request_info) {
   getValue({"request", "url_path"}, &request_info->url_path);
   getValue({"request", "host"}, &request_info->url_host);
   getValue({"request", "scheme"}, &request_info->url_scheme);
+  std::string response_details;
+  getValue({"response", "code_details"}, &response_details);
```
gargnupur, line 432: @kyessenov : It's not possible for both `connection_termination_details` and `response_code_details` to have values right as one is filled for TCP and other for HTTP

kyessenov (approved):

testdata/filters/rbac.yaml.tmpl
kyessenov: Could be done in a follow-up. Just want to make sure we have coverage for network filter as well.

extensions/common/context.cc
kyessenov: nit: no need to stack allocate a string I think.

testdata/filters/rbac.yaml.tmpl
gargnupur: Sure, will add one

extensions/common/context.cc
gargnupur: will fix in the follow-up pr

PR #3104, opened 2020-11-19
[release-1.6] Update common files

jacob-delgado:
Update commonfiles sha


ericvn:
Wow. This repo never got moved up to the latest common-files. I suspect we want to try changing
```
# Version of image used within build container
IMAGE_VERSION ?= master-2020-03-05T18-27-04
```
in `Makefile` to point to the new version.

I'm not sure how much go code is compiled in this repo, so maybe trying to make sure that we have the fixed version of Golang isn't needed.

brian-avery (approved):

PR #3119, opened 2020-12-04
Move testdata gen to env package.

bianpengyuan:
Testdata folder is not crawled by godoc. This is to make the testdata available at godoc site.


bianpengyuan:
Actually does not seem to be necessary to expose the generated testdata asset on godoc,  since the real template data are hidden as private variable and it might be easier to just point user to the code.

PR #3123, opened 2020-12-09
Add a default rbac catch all regex

gargnupur:
Add a default rbac catch all regex. This is to catch errors like when none of the policies match.
In this scenario, response_code_details will contain `rbac_access_denied_matched_policy[none]`.

### DIFF
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -177,7 +177,7 @@ bool fillAuthInfo(const std::string& response_details,
         absl::StrCat(policy_namespace, ".", policy_name);
     (*label_map)["policy_rule"] = policy_rule_index;
     return true;
-  } else if (RE2::PartialMatch(response_details,rbac_denied_match_default)){
+  } else if (RE2::PartialMatch(response_details, rbac_denied_match_default)) {
     (*label_map)["response_details"] = kRbacAccessDenied;
   }
 
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -15,6 +15,7 @@
 
 #include "extensions/stackdriver/log/logger.h"
 
+#include "absl/strings/match.h"
 #include "extensions/stackdriver/common/constants.h"
 #include "google/logging/v2/log_entry.pb.h"
 #include "google/protobuf/util/time_util.h"
--- a/extensions/stackdriver/log/logger.cc
+++ b/extensions/stackdriver/log/logger.cc
@@ -38,7 +38,7 @@ namespace {
 const RE2 rbac_denied_match(
     "rbac_access_denied_matched_policy\\[ns\\[(.*)\\]-policy\\[(.*)\\]-rule\\[("
     ".*)\\]\\]");
-const RE2 rbac_denied_match_default("rbac_access_denied_matched_policy");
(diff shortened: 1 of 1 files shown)


extensions/stackdriver/log/logger.cc
```diff
@@ -176,6 +177,8 @@ bool fillAuthInfo(const std::string& response_details,
         absl::StrCat(policy_namespace, ".", policy_name);
     (*label_map)["policy_rule"] = policy_rule_index;
     return true;
+  } else if (RE2::PartialMatch(response_details,rbac_denied_match_default)){
```
bianpengyuan, line 180: Should return true inside?

bianpengyuan: Also no need to `else if`? Just `if (...)`.

gargnupur: Thanks..Fixed

bianpengyuan (approved):

extensions/stackdriver/log/logger.cc
```diff
@@ -177,6 +178,10 @@ bool fillAuthInfo(const std::string& response_details,
     (*label_map)["policy_rule"] = policy_rule_index;
     return true;
   }
+  if (RE2::PartialMatch(response_details, rbac_denied_match_default)) {
```
yangminzhu, line 181: Can we refactor the code a little bit:

1. Check if it has the prefix "rbac_denied_match_default"
2. If it has the prefix, set response_details to "AuthzDenied", otherwise just skip
3. Parse the remaining part for the ns, name and rule. Skip if the parse failed

So that you only need 1 regex match, set the response_details once and only need 1 definition of "rbac_denied_match_default".

bianpengyuan:
Added do not merge label in case @gargnupur want to address Yangmin's comment.

extensions/stackdriver/log/logger.cc
gargnupur: @yangminzhu : what is the scenario, we will always have? If it's going to be rbac denied match with full policy name and rule, then we should keep it as is, as it gets away with just 1 regex match. But if the second scenario of none policy match is more common, then we should do what you suggested...

yangminzhu: The second scenario (none policy matched) will be more common because ALLOW action is used much more widely.

gargnupur: Cool.. fixed

extensions/stackdriver/log/logger.cc
```diff
@@ -169,15 +170,17 @@ void fillExtraLabels(
 bool fillAuthInfo(const std::string& response_details,
                   google::protobuf::Map<std::string, std::string>* label_map) {
   std::string policy_name, policy_namespace, policy_rule_index;
-  if (RE2::PartialMatch(response_details, rbac_denied_match, &policy_namespace,
-                        &policy_name, &policy_rule_index)) {
+  if (response_details.find(rbac_denied_match_prefix)) {
```
bianpengyuan, line 173: nit: use absl::StartsWith? https://github.com/abseil/abseil-cpp/blob/fbdff6f3ae0ba977a69f172e85ecaede535e70f6/absl/strings/match.h#L58

gargnupur: Thanks..Fixed

yangminzhu (approved):
thanks for the fix!

PR #3124, opened 2020-12-10
extensions: add a log if done before logged

kyessenov:

Fixes https://github.com/istio/istio/issues/29498

I audited the code, and it seems it should work correctly. Added a log if it goes wrong.

### DIFF
--- a/extensions/stats/plugin.cc
+++ b/extensions/stats/plugin.cc
@@ -542,6 +542,9 @@ std::optional<uint32_t> PluginRootContext::addIntExpression(
 
 bool PluginRootContext::onDone() {
   cleanupExpressions();
+  if (!request_queue_.empty()) {
+    LOG_CRITICAL(absl::StrCat("Request queue is not empty, dropping requests: ", request_queue_.size()));
+  }
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


extensions/stats/plugin.cc
```diff
@@ -542,6 +542,10 @@ std::optional<uint32_t> PluginRootContext::addIntExpression(
 
 bool PluginRootContext::onDone() {
   cleanupExpressions();
+  if (!request_queue_.empty()) {
+    LOG_CRITICAL(absl::StrCat("Request queue is not empty, dropping requests: ",
```
bianpengyuan, line 546: Will this crash envoy? I don't think we want to crash envoy because of stats drop. Should be at warning level I suppose?

Also do we need to do anything for SD plugin?

kyessenov: It shouldn't crash, it's just a log level per https://github.com/proxy-wasm/proxy-wasm-cpp-sdk/blob/7afb39d868a973caa6216a535c24e37fb666b6f3/proxy_wasm_api.h#L88. 

Critical because it should never happen, so it's basically an assert.

I looked over SD, and it seems to be exporting onDone. Let me know if you can think of a situation when it doesn't export.

bianpengyuan (approved):

PR #3126, opened 2020-12-11
Update fake stackdriver to response audit log entry separately.

bianpengyuan:
cc @xulingqing

### DIFF
--- a/test/envoye2e/stackdriver_plugin/fake_stackdriver.go
+++ b/test/envoye2e/stackdriver_plugin/fake_stackdriver.go
@@ -53,11 +53,10 @@ type MetricServer struct {
 
 // LoggingServer is a fake stackdriver server which implements all of logging v2 service method.
 type LoggingServer struct {
-	delay                        time.Duration
-	listServerAccessLogEntryResp logging.ListLogEntriesResponse
-	listServerAuditLogEntryResp  logging.ListLogEntriesResponse
-	RcvLoggingReq                chan *logging.WriteLogEntriesRequest
-	mux                          sync.Mutex
+	delay            time.Duration
+	listLogEntryResp logging.ListLogEntriesResponse
+	RcvLoggingReq    chan *logging.WriteLogEntriesRequest
+	mux              sync.Mutex
 }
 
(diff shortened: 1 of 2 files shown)


test/envoye2e/stackdriver_plugin/fake_stackdriver.go
```diff
@@ -484,6 +502,7 @@ func RunFakeStackdriver(port uint16) error {
 	}
 	http.HandleFunc("/timeseries", fsdms.GetTimeSeries)
 	http.HandleFunc("/logentries", fsdls.GetLogEntries)
+	http.HandleFunc("/auditlogentries", fsdls.GetAuditLogEntries)
```
gargnupur, line 505: why not add the log name in log entry? These are just server entries.. I think we would want to add client side logs too in future and then we would have to add more rest url's for that?

bianpengyuan: Done.

test/envoye2e/stackdriver_plugin/fake_stackdriver.go
```diff
@@ -143,11 +143,12 @@ func (s *LoggingServer) WriteLogEntries(ctx context.Context, req *logging.WriteL
 	s.mux.Lock()
 	defer s.mux.Unlock()
 	for _, entry := range req.Entries {
-		// Add the general labels to every log entry in list logentries response.
+		// Add the general labels and log name to every log entry in list logentries response.
 		tmpEntry := proto.Clone(entry).(*logging.LogEntry)
 		for k, v := range req.Labels {
 			tmpEntry.Labels[k] = v
 		}
+		tmpEntry.Labels["log_name"] = req.LogName
```
gargnupur, line 151: We could also use `LogName` field in LogEntry(https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) ?

bianpengyuan: Thanks done!

gargnupur (approved):

PR #3132, opened 2020-12-18
Use default channel credential when sts is not enabled.

bianpengyuan:
This should be equivalent to google cloud engine when running on gce, and will look up GOOGLE_APPLICATION_CREDENTIALS when running on other platforms.

### DIFF
--- a/extensions/stackdriver/common/utils.cc
+++ b/extensions/stackdriver/common/utils.cc
@@ -66,6 +66,8 @@ void buildEnvoyGrpcService(const StackdriverStubOption &stub_option,
                                   ? kSTSSubjectTokenPath
                                   : stub_option.test_token_path);
   auto initial_metadata = grpc_service->add_initial_metadata();
+  // When using p4sa/sts, google backend needs `x-goog-user-project` in initial
+  // metadata to differentiate which project the call should be accounted for.
   initial_metadata->set_key("x-goog-user-project");
   initial_metadata->set_value(stub_option.project_id);
 
--- a/extensions/stackdriver/common/utils.cc
+++ b/extensions/stackdriver/common/utils.cc
@@ -45,36 +45,37 @@ void buildEnvoyGrpcService(const StackdriverStubOption &stub_option,
     // for testing.
     grpc_service->mutable_google_grpc()->set_target_uri(
         stub_option.insecure_endpoint);
-  } else {
-    grpc_service->mutable_google_grpc()->set_target_uri(
-        stub_option.secure_endpoint.empty() ? stub_option.default_endpoint
-                                            : stub_option.secure_endpoint);
-    if (stub_option.sts_port.empty()) {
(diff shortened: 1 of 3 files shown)


extensions/stackdriver/common/utils.cc
```diff
@@ -45,36 +45,37 @@ void buildEnvoyGrpcService(const StackdriverStubOption &stub_option,
...
+  if (stub_option.sts_port.empty()) {
+    // Security token exchange is not enabled. Use default Google credential.
     grpc_service->mutable_google_grpc()
         ->mutable_channel_credentials()
-        ->mutable_ssl_credentials()
-        ->mutable_root_certs()
-        ->set_filename(stub_option.test_root_pem_path.empty()
-                           ? kDefaultRootCertFile
-                           : stub_option.test_root_pem_path);
+        ->mutable_google_default();
```
mandarjog, line 57: So looks like previously we over specified so GAC did not work?

bianpengyuan: Yeah, logging does not work because it uses this code path. Metric works because OC uses default credential. I verified it fixed the issue.

bianpengyuan:
@mandarjog @kyessenov ptal thanks!

extensions/stackdriver/common/utils.cc
```diff
@@ -45,36 +45,37 @@ void buildEnvoyGrpcService(const StackdriverStubOption &stub_option,
...
   }
+
+  setSTSCallCredentialOptions(grpc_service->mutable_google_grpc()
+                                  ->add_call_credentials()
+                                  ->mutable_sts_service(),
+                              stub_option.sts_port,
+                              stub_option.test_token_path.empty()
+                                  ? kSTSSubjectTokenPath
+                                  : stub_option.test_token_path);
+  auto initial_metadata = grpc_service->add_initial_metadata();
```
mandarjog, line 68: If this is not a cherry pick, please add a comment re ‘x-goog-user-project’

mandarjog (commented):
Is it possible to add a test here?

extensions/stackdriver/common/utils.cc
bianpengyuan: Done.

bianpengyuan:
@mandarjog yeah, I added unit test coverage. Integration test would be hard.

bianpengyuan:
please take a look, thanks!

mandarjog:
@kyessenov PTAL.
@bianpengyuan unit tests look ok, how can we be sure that this actually works as intended talking to stackdriver?

We will have an integration test in istio/istio or ASM, but can use fakeSD to test it?

bianpengyuan:
@mandarjog I am not sure, and the main reason for that is our test is running on google infra, I am not sure if the real google credential will be used or not. If really google credential is used, there is no way for fake backend to test that.

kyessenov (approved):

PR #3134, opened 2020-12-18
Add configuration for stackdriver metric expiry.

bianpengyuan:
Forgot to actually set the expiry while waiting for opencensus change to arrive istio/proxy..


mandarjog (commented):
Is there a way to turn this off in config?
I don’t think it can go in like this.

bianpengyuan:
@mandarjog added configurability. ptal thanks!

mandarjog (commented):
1. It is fine to have the proxy default as never expires and istio config default to a more sane value. 10 mins.
2. Need an integration to ensure that everything is plumbed thru. What is the minimum valid duration ?

bianpengyuan:
@mandarjog thanks! done.

extensions/stackdriver/config/v1alpha1/stackdriver_plugin_config.proto
```diff
@@ -120,4 +120,11 @@ message PluginConfig {
   // logs. Conflicts are resolved by the tag name by overriding previously
   // supplied values. Does not apply to audit logs.
   CustomConfig custom_log_config = 14;
+
+  // Optional. Controls the metric expiry duration. If a metric time series is
+  // not updated for the given duration, it will be purged from time series
+  // cache as well as metric reporting. If this is not set or set to 0, time
+  // series will never be expired. This option is useful to avoid unbounded
```
mandarjog, line 127: Add comment?: a high value here means higher memory consumption, but potentially also a higher tail latency ( if it is true), when purging metrics, does a big purge cost more thAn small purge for cpu? If we are iterating over the whole set, it may not matter.

mandarjog: Especially true for cloud run use case.

bianpengyuan: Opencensus uses its own thread pool, so I think it won't affect tail latency?

bianpengyuan:
please take a look, thanks!

test/envoye2e/stackdriver_plugin/stackdriver_test.go
```diff
@@ -895,3 +895,65 @@ func TestStackdriverRbacAccessDenied(t *testing.T) {
...
+					Port: params.Ports.ClientPort,
+					Body: "hello, world!",
+				},
+			},
+			sd.Check(params,
+				[]string{"testdata/stackdriver/server_request_count.yaml.tmpl"},
+				[]SDLogEntry{}, nil, true,
+			),
+			sd.Reset(),
+			&driver.Sleep{Duration: 10 * time.Second},
```
mandarjog, line 941: any way to speed this up? 10s seems like a lot.

bianpengyuan: Purge happens lazily on report cycle, and the report cycle time is set to 20s in order to accommodate slow asan/tsan build: https://github.com/istio/proxy/pull/2806/commits/9fda46f0d12df049f6c27bf2d64278dda3d7ba06

kyessenov: Yeah, SD test takes about a minute so it's not making it worse. It can be run in parallel if necessary.

kyessenov (approved):

PR #3166, opened 2021-01-27
Update metric stub option builder to make insecure endpoint take precedence.

bianpengyuan:
insecure endpoint should always take precedence, so that intent of testing could be honored.

cc @richardwxn


richardwxn (approved):

mandarjog (approved):
Lgtm

mandarjog:
There is no test associated with this change, but it is only changing the order or precedence.

PR #3173, opened 2021-01-31
Re-enable RBE cache for CentOS.

PiotrSikora:


prow/proxy-common.inc
```diff
@@ -47,17 +47,14 @@ fi
 BAZEL_OUT="$(bazel info ${BAZEL_BUILD_ARGS} output_path)/k8-fastbuild/bin"
 export BAZEL_OUT
 
-# Disable RBE execution due to failures like https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_proxy/2633/release-test_proxy/211
-export BAZEL_BUILD_RBE_JOBS=0
```
PiotrSikora, line 51: We need a few larger machines to build V8 using RBE execution, otherwise it timeouts after 60mins.

@howardjohn do you have access to configure RBE worker pool? If not, do you know who has?

@lizan are there any instructions on how to provision and tag `LARGE_MARCHINE` instances for Envoy? I thought it was documented in Envoy, but I cannot find it now.

PiotrSikora: Actually, Prow has beefy enough machines that we shouldn't need RBE execution.

I'll split this change into separate a PR once we have `LARGE_MACHINE` instances configured, since RBE cache is a big improvement by itself.

PiotrSikora:
Tests should automatically use RBE cache once https://github.com/istio/test-infra/pull/3167 is merged.

prow/proxy-presubmit-centos-release.sh
```diff
@@ -20,8 +20,9 @@ WD=$(cd "$WD" || exit 1 ; pwd)
 #######################################
 # Presubmit script triggered by Prow. #
 #######################################
-# Do not use RBE for this, RBE will run ubuntu instance
-export BAZEL_BUILD_RBE_INSTANCE=""
+
+# Do not use RBE execution with Ubuntu toolchain, but still use RBE cache.
+export BAZEL_BUILD_RBE_JOBS=0
```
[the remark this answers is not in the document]
lambdai: Awesome!

prow/proxy-common.inc
```diff
@@ -57,7 +57,7 @@ if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
 
   # Use RBE when logged in.
   BAZEL_BUILD_RBE_INSTANCE="${BAZEL_BUILD_RBE_INSTANCE-projects/istio-testing/instances/default_instance}"
-  BAZEL_BUILD_RBE_CACHE="${BAZEL_BUILD_RBE_CACHE:-grpcs://remotebuildexecution.googleapis.com}"
+  BAZEL_BUILD_RBE_CACHE="${BAZEL_BUILD_RBE_CACHE-grpcs://remotebuildexecution.googleapis.com}"
```
lambdai, line 60: Use the default rbe setting when BAZEL_BUILD_RBE_CACHE=="" ? NBD but I am curious why you make this change

lambdai (approved):
Looks great!

lambdai:
@PiotrSikora  How does CentOS build select the cache valid for CentOS and not selecting the caches for ubuntu?

prow/proxy-common.inc
PiotrSikora: This is change to match style used in the line above (see: https://github.com/istio/proxy/pull/2995), so that it doesn't override an empty value if someone wants to disable this feature.

PiotrSikora:
> @PiotrSikora How does CentOS build select the cache valid for CentOS and not selecting the caches for ubuntu?

I'm assuming that it's based on hashes of the compiler, system libraries used for linking, etc.

Postsubmit CentOS job is already using RBE cache, so this change doesn't affect published build artifacts at all.

prow/proxy-common.inc
howardjohn: I do have, if you let me know the commands to run I can do it, if still needed

howardjohn (approved):

lizan (approved):

prow/proxy-common.inc
lambdai: So the current behavior is that disable cache only if `BAZEL_BUILD_RBE_CACHE` is unset. I don't have a preference as long as all existing runs are good

PiotrSikora:
Between this PR and https://github.com/istio/test-infra/pull/3167, the time to run presubmits went down significantly:

- `test_proxy` - from 42 mins to 7 mins,
- `test-asan_proxy` - from 1 hour 12 mins to 12 mins,
- `test-tsan_proxy` - from 42 mins to 7 mins,
- `check-wasm_proxy` - from 53 mins to 12 mins,
- `release-test_proxy` - 24 mins (no change, was already using RBE cache),
- `release-centos-test_proxy` - from 1 hour 52 mins to 12 mins.

From ~2 hours down to 24 mins, which we could further improve bring down to ~12 mins if we want to split release build into separate steps for release and debug binaries. cc @stewartbutler

PR #3179, opened 2021-02-03
stackdriver: fix tcp bytes metrics aggregation

douglas-reid:
The aggregation for bytes received/sent metrics for Stackdriver is incorrect. It should be `Sum` and not `Count`. As `Count`, it will merely record the number of times bytes were counted.


bianpengyuan (approved):
thanks!

mandarjog (approved):
Thanks Doug!

PR #3193, opened 2021-02-15
Remove filename from SHA256 output

ericvn:
Fix for the update_envoy script to remove the filename. 

The current automation ends up with (see #3171):
```
ENVOY_SHA256 = "5d8552be5cd2760e2dbabf9c25b7912a002d77486ae9a22d1ebee85449c183cc  96b216696bf69fa83bc65285a0f816cc41b079e6.tar.gz"
``` 
with the filename added.


bianpengyuan (approved):

ericvn:
I'm going to put a DNM on this for now as I don't think the `master` update is doing what it should be doing. This PR fixes the problem with the extra name, but I think it will down level the WORKSPACE variables to the latest istio/envoy in `master` which doesn't appear to be what is wanted. It should probably be the latest commit in the org/rep specified in the WORKSPACE file.

PiotrSikora (approved):

ericvn:
Superseded by https://github.com/istio/proxy/pull/3194

PR #3197, opened 2021-02-19
update_envoy to get latest commit from WORKSPACE REPO/ORG (#3194)

PiotrSikora:
Backport of #3194, since `release-1.9` branch is broken.

PR #3198, opened 2021-02-19
Update Envoy SHA to latest with Wasm bugfix.

PiotrSikora:
Fixes proxy-wasm/proxy-wasm-cpp-host#125.


PiotrSikora:
Manual update, because automator is broken (see: https://github.com/istio/proxy/pull/3197).

shamsher31 (approved):

PR #3202, opened 2021-02-19
Update Envoy SHA to latest.

PiotrSikora:

### DIFF
--- a/src/envoy/extensions/wasm/wasm.cc
+++ b/src/envoy/extensions/wasm/wasm.cc
@@ -51,11 +51,12 @@ class IstioWasm : public Wasm {
  public:
   IstioWasm(absl::string_view runtime, absl::string_view vm_id,
             absl::string_view vm_configuration, absl::string_view vm_key,
+            proxy_wasm::AllowedCapabilitiesMap allowed_capabilities,
             const Stats::ScopeSharedPtr& scope,
             Upstream::ClusterManager& cluster_manager,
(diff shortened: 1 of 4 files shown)


bianpengyuan (approved):

src/envoy/extensions/wasm/wasm.cc
```diff
@@ -101,15 +102,23 @@ class IstioWasmExtension : public EnvoyWasm {
...
+            const Stats::ScopeSharedPtr& scope,
             Upstream::ClusterManager& cluster_manager,
             Event::Dispatcher& dispatcher,
             Server::ServerLifecycleNotifier& lifecycle_notifier,
             absl::string_view vm_key) -> WasmHandleBaseSharedPtr {
-    auto wasm =
-        std::make_shared<IstioWasm>(vm_config.runtime(), vm_config.vm_id(),
-                                    anyToBytes(vm_config.configuration()),
-                                    vm_key, scope, cluster_manager, dispatcher);
+    // TODO(rapilado): make this transformation in Proxy-Wasm C++ Host.
```
PiotrSikora, line 112: @ryanapilado it looks that this leaks quite a bit and each extension needs to convert `CapabilityRestrictionConfig` to `allowed_capabilities`, which seems a bit error prone. We should do it earlier in the process (either in Proxy-Wasm C++ host or Envoy) to avoid this repetition.

cc @mathetake to catch similiar issues in the future.

PR #2833, opened 2020-04-28
[release-1.6] fix location tag, update proto format (#2831)

richardwxn:
manual cherry-pick to resolve conflict(conflict arise from the comments part only)

PR #3214, opened 2021-02-26
[release-1.6] Update config.pb.html

hzxuzhonghu:
https://github.com/istio/istio.io/pull/9050#issuecomment-785983960

### DIFF
--- a/extensions/metadata_exchange/config.pb.html
+++ b/extensions/metadata_exchange/config.pb.html
@@ -1,7 +1,7 @@
 ---
 title: Metadata Exchange Config
 description: Configuration for Metadata Exchange Filter.
-location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.
+location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.html
 layout: protoc-gen-docs
 generator: protoc-gen-docs
 weight: 20
--- a/extensions/metadata_exchange/config.pb.html
+++ b/extensions/metadata_exchange/config.pb.html
@@ -1,7 +1,7 @@
 ---
 title: Metadata Exchange Config
 description: Configuration for Metadata Exchange Filter.
-location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.html.
+location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.
 layout: protoc-gen-docs
 generator: protoc-gen-docs
 weight: 20


hzxuzhonghu:
@ericvn Is this right

ericvn:
@hzxuzhonghu No. You need to change: 

extensions/metadata_exchange/config.proto in the release-1.6 branch.

But...

It looks like the change was actually made already in https://github.com/istio/proxy/pull/2833.

So maybe we just need to do some doc workgroup PRs in the istio.io repo.

ericvn:
It's still not correct here:

The line you changed should be just have the trailing `.` removed.  Normally, it's updated in the proto (which is missing the period) and the make gen will fix the pb.html to match.  For some reason the pb.html doesn't match here.

Not sure if it's a `make gen` or something else. So back to actually needing a change in the pb.html. Just not sure if we can do that manually or not.

extensions/metadata_exchange/config.pb.html
```diff
@@ -1,7 +1,7 @@
 ---
 title: Metadata Exchange Config
 description: Configuration for Metadata Exchange Filter.
-location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.html.
+location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.
```
ericvn, line 4:
```suggestion
location: https://istio.io/docs/reference/config/proxy_extensions/metadata_exchange.html
```

ericvn (dismissed):
add additional text back.

jacob-delgado:
cc @ericvn @hzxuzhonghu I'm a bit confused about this PR. 1.6 closed last November. I see it's just html, so it has something to do with the old istio archives for 1.6?

hzxuzhonghu:
@ericvn can you take it over, i am not familiar with the docs at all.

@jacob-delgado It is to fix in windows, we can not clone this repo, because the dir name is `metadata_exchange.html`

ericvn:
@jacob-delgado @hzxuzhonghu opened https://github.com/istio/istio.io/pull/9050 for an issue where a directory ended in a `.` prevents a Windows user from cloning the istio.io repo. Discussion there, but summary follows.

The problem is the directory is created when the reference docs are imported into istio.io (it uses the location specified in the doc to determine where to put the file). istio/proxy had a trailing period on the location in release-1.6, but was corrected in 1.7.  To fix the issue correctly, we want to update the doc in istio/proxy release-1.6, then I'll pull a new set of reference docs in the istio.io release-1.6 brach which should change the directory to no longer have the trailing period.  Then when we rebuild the 1.6 archive in istio.io master, the directory should have the period removed as well. Then, Windows users should be able to clone the master branch. Question still outstanding if we want to update the archive in the istio.io 1.9 branch.

ericvn:
As @hzxuzhonghu asked me to take over, I committed my suggestion so that only the trailing period is removed (this is a cherry pick of one line from later releases (see https://github.com/istio/proxy/blob/master/extensions/metadata_exchange/config.pb.html#L4)).

brian-avery (approved):

brian-avery:
Thanks for the explanation on here @ericvn

PR #3222, opened 2021-03-01
Use smaller Wasm heap and stack at compile time

mandarjog:
Continuation of https://github.com/istio/proxy/pull/2965

### DIFF
--- a/extensions/BUILD
+++ b/extensions/BUILD
@@ -27,9 +27,7 @@ load(
 envoy_package()
 
 WASM_LINKOPTS = [
-    "--js-library external/proxy_wasm_cpp_sdk/proxy_wasm_intrinsics.js",
-    "-s TOTAL_MEMORY=8MB",
-    "-s TOTAL_STACK=5MB",
+    "-s INITIAL_MEMORY=8MB",
 ]
 
 envoy_wasm_cc_binary(
--- a/extensions/BUILD
+++ b/extensions/BUILD
@@ -26,6 +26,12 @@ load(
 
 envoy_package()
 
+WASM_LINKOPTS = [
+    "--js-library external/proxy_wasm_cpp_sdk/proxy_wasm_intrinsics.js",
+    "-s TOTAL_MEMORY=8MB",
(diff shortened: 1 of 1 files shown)


mandarjog:
@bianpengyuan we need to do the same in the istio-extensions repo.

extensions/BUILD
```diff
@@ -26,6 +26,12 @@ load(
 
 envoy_package()
 
+WASM_LINKOPTS = [
+    "--js-library external/proxy_wasm_cpp_sdk/proxy_wasm_intrinsics.js",
+    "-s TOTAL_MEMORY=8MB",
+    "-s TOTAL_STACK=5MB",
```
PiotrSikora, line 32: `TOTAL_STACK=5MB` is the default value, so we might skip it.

extensions/BUILD
```diff
@@ -26,6 +26,12 @@ load(
 
 envoy_package()
 
+WASM_LINKOPTS = [
+    "--js-library external/proxy_wasm_cpp_sdk/proxy_wasm_intrinsics.js",
+    "-s TOTAL_MEMORY=8MB",
```
PiotrSikora, line 31: `TOTAL_MEMORY` was renamed to `INITIAL_MEMORY` a year ago.

extensions/BUILD
```diff
@@ -26,6 +26,12 @@ load(
 
 envoy_package()
 
+WASM_LINKOPTS = [
+    "--js-library external/proxy_wasm_cpp_sdk/proxy_wasm_intrinsics.js",
```
PiotrSikora, line 30: Is this needed here?

PiotrSikora (changes requested):

extensions/BUILD
bianpengyuan: This is not needed.

PiotrSikora (approved):
Thanks!

PiotrSikora:
@bianpengyuan PTAL

bianpengyuan (approved):

PR #3225, opened 2021-03-02
[test] Do not override downloaded envoy if already exists.

bianpengyuan:

### DIFF
--- a/test/envoye2e/driver/envoy.go
+++ b/test/envoye2e/driver/envoy.go
@@ -233,7 +233,7 @@ func downloadEnvoy(ver string) (string, error) {
 	}
 	defer os.RemoveAll("usr/")
 
-	cpCmd := exec.Command("cp", src, dst)
+	cpCmd := exec.Command("cp", "-n", src, dst)
 	cpCmd.Stderr = os.Stderr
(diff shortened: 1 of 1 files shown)


test/envoye2e/driver/envoy.go
```diff
@@ -233,7 +233,7 @@ func downloadEnvoy(ver string) (string, error) {
 	}
 	defer os.RemoveAll("usr/")
 
-	cpCmd := exec.Command("cp", src, dst)
+	cpCmd := exec.Command("cp", "-n", src, dst)
```
PiotrSikora, line 236: Shouldn't this check be performed before we download Envoy? It's a bit wasteful to download all those bytes if we're not going to use the binary.

PiotrSikora (changes requested):

test/envoye2e/driver/envoy.go
bianpengyuan: There is already a check at line 209, but duplicate download still happens some time due to test runs parallel.

PiotrSikora: Ah, OK.

PiotrSikora (approved):
.

PR #3229, opened 2021-03-04
Codify cluster id in stats filter.

bianpengyuan:
Since now we unconditionally add cluster labels with EnvoyFilter configuration, it is better to just codify them within the stats filter. Corresponding istio PR: https://github.com/istio/istio/pull/31226

Also make a fix for https://github.com/istio/istio/issues/29373. Corresponding istio PR: https://github.com/istio/istio/pull/31214


extensions/common/context.cc
```diff
@@ -316,11 +317,19 @@ bool extractPeerMetadataFromUpstreamHostMetadata(
...
+    // name if there are the fifth part.
+    cluster_id = fbb.CreateString(parts[4]);
+  }
   auto labels_offset = fbb.CreateVectorOfSortedTables(&labels);
 
   FlatNodeBuilder node(fbb);
   node.add_workload_name(workload_name);
   node.add_namespace_(namespace_);
+  if (!cluster_id.IsNull()) {
+    node.add_cluster_id(cluster_id);
```
douglas-reid, line 331: qq: do we need to default this to something if cluster id is null (like "kubernetes" or "unknown")?

bianpengyuan: Currently, with Envoyfilter config, I think the cluster labels will just be `unknown` when it is unknown, and stats filter will fill in unknown for empty labels by default. For example in one of the test, you can find that source cluster is set to unknown when mx is not happening:
```
  - name: source_cluster
    value: unknown
```

douglas-reid: ok. as long as stats_filter continues to populate `unknown` for missing values, we should be good.

testdata/bootstrap/client.yaml.tmpl
```diff
@@ -60,7 +60,7 @@ static_resources:
           metadata:
             filter_metadata:
               istio:
-                workload: ratings-v1;default;ratings;version-1
+                workload: ratings-v1;default;ratings;version-1;test-cluster
```
kyessenov, line 63: nit: would be good to have a different value for destination cluster than source.

kyessenov:
/lgtm
/approve
Nit to improve the test, DNM if Doug has more concerns.

kyessenov (approved):

douglas-reid (approved):

testdata/bootstrap/client.yaml.tmpl
bianpengyuan: Done.

PR #3239, opened 2021-03-13
Update Envoy SHA to latest.

PiotrSikora:

### DIFF
--- a/testdata/stackdriver/client_access_log_entry.yaml.tmpl
+++ b/testdata/stackdriver/client_access_log_entry.yaml.tmpl
@@ -1,10 +1,6 @@
 http_request:
   request_method: "GET"
-  {{- if eq .Vars.ServiceAuthenticationPolicy "MUTUAL_TLS" }}
-  request_url: "https://127.0.0.1:{{ .Ports.ClientPort }}/{{ .Vars.RequestPath }}"
-  {{- else }}
   request_url: "http://127.0.0.1:{{ .Ports.ClientPort }}/{{ .Vars.RequestPath }}"
-  {{- end }}
   server_ip: "127.0.0.1:{{ .Ports.ClientPort }}"
   protocol: "http"
   status: {{ .Vars.SDLogStatusCode }}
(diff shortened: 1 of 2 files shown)


testdata/stackdriver/client_access_log_entry.yaml.tmpl
```diff
@@ -1,10 +1,6 @@
 http_request:
   request_method: "GET"
-  {{- if eq .Vars.ServiceAuthenticationPolicy "MUTUAL_TLS" }}
-  request_url: "https://127.0.0.1:{{ .Ports.ClientPort }}/{{ .Vars.RequestPath }}"
-  {{- else }}
   request_url: "http://127.0.0.1:{{ .Ports.ClientPort }}/{{ .Vars.RequestPath }}"
-  {{- end }}
```
PiotrSikora, line 7: As a result of changes in envoyproxy/envoy#15321, the `:scheme` is now persistent and does not change depending on the security of the underlying transport socket.

For example, assuming Istio service mesh with mTLS between sidecars:
```
[client] -- cleartext --> [C-sidecar] -- mTLS --> [S-sidecar] -- cleartext --> [server]
```
Previously, when client sent request `http://server`, the `:scheme` would be reported as `http` between `client` and `C-sidecar`, and between `S-sidecar` and `server`, but it would be reported as `https` between sidecars, because the underlying transport socket was considered secure and it "upgraded" the scheme.

However, this was wrong, since the requested URI didn't change, and starting with envoyproxy/envoy#15321, `:scheme` now correctly reported as `http` across all intermediates.

This might have some unintended side-effects if we use it to monitor for mTLS, etc.

cc @mandarjog @bianpengyuan @kyessenov @myidpt

bianpengyuan: Scheme is not used to monitor mTLS, it is only used to assembly a request URL in access log. So I think this should be fine.

bianpengyuan (approved):

PR #3245, opened 2021-03-17
envoy: initial boilerplate/skeleton for metadata upstream

tbarrella:
Initial commit for https://github.com/istio/istio/issues/31444

Based on https://github.com/lambdai/envoy-dai/pull/7

I'd rather keep the commits small to make things easier to review. This is meant to be the smallest testable unit, establishing the code location and some naming

### DIFF
--- a/src/envoy/upstreams/http/metadata/BUILD
+++ b/src/envoy/upstreams/http/metadata/BUILD
@@ -1,0 +1,50 @@
+# Copyright 2021 Istio Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


lambdai (changes requested):
Nice move!

Can you add an integration test to prove that this upstream extension can be injected into the envoy cluster?

PiotrSikora (commented):
Is there any reason why this is added to Istio Proxy and not Envoy itself?

tbarrella:
> Nice move!
> 
> Can you add an integration test to prove that this upstream extension can be injected into the envoy cluster?

That can't be done yet. It will require a lot more code (connection pool, connection pool factory). I think that should be left for future PRs to keep size small

> Is there any reason why this is added to Istio Proxy and not Envoy itself?

I think adding this to Envoy would require aligning on an API that would be extensible enough to be used for needs in addition to just Istio. https://github.com/envoyproxy/envoy/issues/15472 is for discussing this. If that discussion blocked starting this, I don't think there would be time to finish this before Istio 1.10. This code would not include the extensible API, but just the headers that are needed for Istio (BTS, maybe temeletry)

lambdai (approved):
LGTM. Thanks!

PiotrSikora (changes requested):
Consensus from the offline discussion is to work on this upstream.

/hold

PiotrSikora (approved):
Apparently, Envoy doesn't want this.

PR #3250, opened 2021-03-18
Avoid specifying root cert at Stackdriver gRPC client.

bianpengyuan:
I've tested that this works fine with real sd backend.

cc @williamaronli @howardjohn

### DIFF
--- a/extensions/stackdriver/common/constants.h
+++ b/extensions/stackdriver/common/constants.h
@@ -96,7 +96,6 @@ constexpr std::string_view kGCECreatedByKey = "gcp_gce_instance_created_by";
 // Misc
 constexpr char kIstioProxyContainerName[] = "istio-proxy";
 constexpr double kNanosecondsPerMillisecond = 1000000.0;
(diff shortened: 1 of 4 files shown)


mandarjog:
Is this done because it is not necessary or because it does not work sometimes?

bianpengyuan:
It does not work on some VM distro where system root is not at `/etc/ssl/certs/ca-certificates.crt`, such as centos. Google grpc already handled different root ca file path gracefully.

mandarjog (approved):
Lgtm

howardjohn:
Just to verify, this means we will use system certs, right? Not that we
will not verify?

bianpengyuan:
> Just to verify, this means we will use system certs, right? Not that we
will not verify?

Right. To double check, I removed the logic to set test root ca cert, and integration test fails:
```
E0318 15:32:41.284894724   19837 ssl_transport_security.cc:1446] Handshake failed with fatal error SSL_ERROR_SSL: error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED.
```

PR #3256, opened 2021-03-23
authz: add initial authz dry-run implementation

yangminzhu:

Adds the initial support of dry-run authorization policy.

Currently 3 custom labels are added if dry-run is used: `dry_run_result`, `dry_run_policy_name` and `dry_run_policy_rule`, this follows the same name used in the authz deny logging, could be updated later if needed.

### DIFF
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -25,6 +25,7 @@
 #include "extensions/stackdriver/edges/mesh_edges_service_client.h"
 #include "extensions/stackdriver/log/exporter.h"
 #include "extensions/stackdriver/metric/registry.h"
+#include "re2/re2.h"
 
 #ifndef NULL_PLUGIN
 #include "api/wasm/cpp/proxy_wasm_intrinsics.h"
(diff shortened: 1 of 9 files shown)


extensions/stackdriver/stackdriver.cc
```diff
@@ -62,6 +63,20 @@ constexpr int kDefaultTickerMilliseconds = 10000;  // 10s
 
 namespace {
 
+const RE2 authz_name_match("ns\\[(.*)\\]-policy\\[(.*)\\]-rule\\[(.*)\\]");
+constexpr char kRbacAccessAllowed[] = "AuthzAllowed";
+constexpr char kRbacAccessDenied[] = "AuthzDenied";
+constexpr char kRBACHttpFilterName[] = "envoy.filters.http.rbac";
+constexpr char kRBACNetworkFilterName[] = "envoy.filters.network.rbac";
+constexpr char kDryRunDenyShadowEngineResult[] =
```
bianpengyuan, line 71: These keys are defined in control plane?

yangminzhu: the prefix `istio_dry_run_deny_` is defined in istiod and the remaining part `shadow_engine_result` is defied in Envoy RBAC filter.

extensions/stackdriver/stackdriver.cc
```diff
@@ -206,6 +221,78 @@ flatbuffers::DetachedBuffer getLocalNodeMetadata() {
...
+      // If denied by ALLOW policy, the shadow_effective_policy shold be from
+      // the ALLOW policy.
+      shadow_effective_policy = shadow_allow_policy;
+    }
+  }
+
+  extra_labels["dry_run_result"] =
+      shadow_result ? kRbacAccessAllowed : kRbacAccessDenied;
+  std::string policy_name, policy_namespace, policy_rule_index;
+  if (RE2::PartialMatch(shadow_effective_policy, authz_name_match,
```
bianpengyuan, line 288: I am a bit worried about the performance impact of this. Is it possible to do a customized parsing instead of regex?

Also could you add an example policy name in the comment?

extensions/stackdriver/stackdriver.cc
```diff
@@ -534,6 +622,7 @@ bool StackdriverRootContext::recordTCP(uint32_t id) {
       evaluateExpressions(record_info.extra_log_labels);
       record_info.expressions_evaluated = true;
     }
+    fillAuthzDryRunInfo(record_info.extra_log_labels);
```
bianpengyuan, line 625: Can you add a test for TCP as well?

yangminzhu: added TCP test mostly based on the TestStackdriverTCPMetadataExchange.

extensions/stackdriver/stackdriver.cc
yangminzhu: I think it should be fine as the same regex is already used in the logger.cc for authz deny logging parsing. I updated to not to use regex anyway as it's a simple change, I can update the usage in logger.cc in a follow up PR if needed.

bianpengyuan (approved):

PR #3263, opened 2021-03-29
Make fake sd trace service work the same as real sd.

bianpengyuan:
For https://github.com/istio/istio/issues/29284, so that we can consolidate real and fake sd code path.


douglas-reid (approved):

lambdai (approved):
LGTM

PR #3282, opened 2021-04-10
WiP: detect downstream first in tcp stream

lambdai:

### DIFF
--- a/extensions/stackdriver/stackdriver.h
+++ b/extensions/stackdriver/stackdriver.h
@@ -209,6 +209,9 @@ class StackdriverContext : public Context {
 
   // Called on onData call, so counting the data that is received.
   FilterStatus onDownstreamData(size_t size, bool) override {
+    if (!downstream_first_.has_value() && is_tcp_ &&  size > 0) {
+      downstream_first_ = true;
+    }
     if (!is_initialized_) {
       return FilterStatus::Continue;
     }
@@ -219,6 +222,9 @@ class StackdriverContext : public Context {
   }
   // Called on onWrite call, so counting the data that is sent.
   FilterStatus onUpstreamData(size_t size, bool) override {
+    if (!downstream_first_.has_value() && is_tcp_ && size > 0) {
+      downstream_first_ = false;
+    }
     if (!is_initialized_) {
       return FilterStatus::Continue;
     }
@@ -236,6 +242,7 @@ class StackdriverContext : public Context {
   bool is_tcp_;
   uint32_t context_id_;
   const bool is_initialized_;
+  absl::optional<bool> downstream_data_first_{};
 };
 
 class StackdriverOutboundRootContext : public StackdriverRootContext {


extensions/stackdriver/stackdriver.h
```diff
@@ -219,6 +222,9 @@ class StackdriverContext : public Context {
   }
   // Called on onWrite call, so counting the data that is sent.
   FilterStatus onUpstreamData(size_t size, bool) override {
+    if (!downstream_first_.has_value() && is_tcp_ && size > 0) {
+      downstream_first_ = false;
```
mandarjog, line 226: Just so I am clear, this is the server writing the first byte like MySQL.
Also I would like to confirm that we only need to do this for non-tls connections, since tls is a normal client first protocol.

mandarjog (commented):
Thanks this is very simple and clever.
How do we make it available ?

mandarjog:
@bianpengyuan @douglas-reid we want to know how many users use server first protocol. How do we collecting info?
We may not need to in perpetuity.

bianpengyuan:
Through TCP log labels?

extensions/stackdriver/stackdriver.h
lambdai: I am not sure if I understand your concern. 
Assume your imagined scenario is that the downstream connection is *plaintext* protocol, and upstream is *tls*  protocol. Since it's the proxy that promote the tls(through transport socket), the arg of onUpstreamData() is only called if there is data within the tls/plaintext channel. To phrase: Tls hand shake bytes won't trigger onUpstreamData().

This PR will detect MySQL as server-first regardless the downstream/upstream are tls or plaintext

lambdai:
> Through TCP log labels?

@bianpengyuan @douglas-reid Can you take over the rest? It's unclear to me how we'd like to export the metric. 
TBH I don't have the context if the detection code is expected to live in SD wasm filter.

extensions/stackdriver/stackdriver.h
```diff
@@ -236,6 +242,7 @@ class StackdriverContext : public Context {
   bool is_tcp_;
   uint32_t context_id_;
   const bool is_initialized_;
+  absl::optional<bool> downstream_data_first_{};
```
adiprerepa, line 245: new to c++, why do we end the variable name with `_`?

extensions/stackdriver/stackdriver.h
```diff
@@ -209,6 +209,9 @@ class StackdriverContext : public Context {
 
   // Called on onData call, so counting the data that is received.
   FilterStatus onDownstreamData(size_t size, bool) override {
+    if (!downstream_first_.has_value() && is_tcp_ &&  size > 0) {
```
adiprerepa, line 212: extra space in front of size
```suggestion
    if (!downstream_first_.has_value() && is_tcp_ && size > 0) {
```

extensions/stackdriver/stackdriver.h
lambdai: Not a must in C++ but a code style in Envoy. https://github.com/envoyproxy/envoy/blob/main/STYLE.md

mandarjog:
@kyessenov can you look at this, possible take it over?

kyessenov:
We technically already record that (connection open with sent > 0 and received = 0). But it needs to hit the timer at the right time.

zirain:
outdated

PR #3311, opened 2021-04-27
Fix sha verification breakage

howardjohn:
See https://github.com/bazelbuild/rules_docker/issues/1814

tl;dr we cannot build proxy due to a SHA mismatch. This is because some
dependency has some magic file that dynamically sets a value when a
github archive is created.

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -208,7 +208,7 @@ def istioapi_dependencies():
 def docker_dependencies():
     http_archive(
         name = "io_bazel_rules_docker",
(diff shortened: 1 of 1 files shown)


kyessenov (approved):

lambdai (approved):
Thank you for the fix!

We need another one after as per 

https://github.com/bazelbuild/rules_docker/issues/1814#issuecomment-827171708

lambdai:
It's a supply chain error so we need to backport. See https://github.com/istio/proxy/pull/3312

PR #3322, opened 2021-05-03
Makefile: fix shell target to be phony

rmstoi:

This PR makes "shell" target to be phony. This fixes the following issue if there is a file with name shell:

```
$ BUILD_WITH_CONTAINER=1 make shell
make: 'shell' is up to date.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ default:
 shell:
 	@$(RUN) /bin/bash
 
-.PHONY: default
+.PHONY: default shell
 
 else
 


bianpengyuan (changes requested):
This is automatically copied from common-files repo: https://github.com/istio/common-files/blob/master/files/Makefile#L54, you should change it directly here.

Also I don't think `shell` is really a target available in proxy repo.

rmstoi:
> This is automatically copied from common-files repo: https://github.com/istio/common-files/blob/master/files/Makefile#L54, you should change it directly here.
> 
> Also I don't think `shell` is really a target available in proxy repo.

Thanks for the info, I can try to make PR to common-files repo.
How often changes from common-files repo are merged?

If I understand correctly `shell` target is available if [BUILD_WITH_CONTAINER=1](https://github.com/istio/proxy/blob/master/Makefile#L33) is set
I've been using `shell` target of istio proxy to get shell access to the build-tools-proxy docker image. As here:

`BUILD_WITH_CONTAINER=1 make shell`

I agree, probably not a conventional approach, nevertheless, good way to learn how the build system works.

bianpengyuan:
> How often changes from common-files repo are merged?

There is no frequency limit on commit merge to that repo AFAIK..

rmstoi:
Closing PR since moved to [common-files](https://github.com/istio/common-files/pull/411). Thanks @bianpengyuan

PR #3331, opened 2021-05-06
stackdriver: harden against invalid utf8

kyessenov:

Per my experiments, Envoy hardens host and path already, so we sanitize the remaining headers.

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -436,10 +436,11 @@ void populateExtendedHTTPRequestInfo(RequestInfo* request_info) {
   if (getValue({"request", "headers", "x-b3-sampled"}, &trace_sampled) &&
       trace_sampled == "1") {
     if (getValue({"request", "headers", "x-b3-traceid"},
-             &request_info->b3_trace_id)) {
+                 &request_info->b3_trace_id)) {
       sanitizeBytes(&request_info->b3_trace_id);
     }
-    if (getValue({"request", "headers", "x-b3-spanid"}, &request_info->b3_span_id)) {
+    if (getValue({"request", "headers", "x-b3-spanid"},
+                 &request_info->b3_span_id)) {
       sanitizeBytes(&request_info->b3_span_id);
     }
     request_info->b3_trace_sampled = true;
@@ -534,8 +535,8 @@ bool getAuditPolicy() {
 }
 
 bool sanitizeBytes(std::string* buf) {
-  char *start = buf->data();
-  const char *const end = start + buf->length();
+  char* start = buf->data();
+  const char* const end = start + buf->length();
   bool modified = false;
   while (start < end) {
     if (flatbuffers::FromUTF8(const_cast<const char**>(&start)) < 0) {
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -539,10 +539,13 @@ bool sanitizeBytes(std::string* buf) {
   const char* const end = start + buf->length();
   bool modified = false;
   while (start < end) {
-    if (flatbuffers::FromUTF8(const_cast<const char**>(&start)) < 0) {
+    char* s = start;
+    if (flatbuffers::FromUTF8(const_cast<const char**>(&s)) < 0) {
       *start = ' ';
       start += 1;
       modified = true;
+    } else {
+      start = s;
     }
   }
   return modified;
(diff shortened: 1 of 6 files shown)


extensions/common/context.cc
```diff
@@ -520,5 +534,19 @@ bool getAuditPolicy() {
   return shouldAudit;
 }
 
+bool sanitizeBytes(std::string* buf) {
```
lambdai, line 537: std::string&

extensions/common/context.cc
```diff
@@ -520,5 +534,19 @@ bool getAuditPolicy() {
...
+  char* start = buf->data();
+  const char* const end = start + buf->length();
+  bool modified = false;
+  while (start < end) {
+    if (flatbuffers::FromUTF8(const_cast<const char**>(&start)) < 0) {
```
lambdai, line 542: is it possible that `flatbuffers::FromUTF8` access the bytes beyond the last '\0'?

extensions/common/context.cc
```diff
@@ -422,15 +423,26 @@ std::string_view nodeInfoSchema() {
...
+    sanitizeBytes(&request_info->referer);
+  }
+  if (getValue({"request", "useragent"}, &request_info->user_agent)) {
+    sanitizeBytes(&request_info->user_agent);
+  }
+  if (getValue({"request", "id"}, &request_info->request_id)) {
+    sanitizeBytes(&request_info->request_id);
+  }
   std::string trace_sampled;
   if (getValue({"request", "headers", "x-b3-sampled"}, &trace_sampled) &&
```
lambdai, line 436: Should `x-b3-sampled` be sanitized as well?

lambdai (commented):
It's not clear to me what fields need sanitation and what do not.

I am wondering if `getValue` should by default execute sanitization and explicitly do getValueUnsafe.

extensions/common/context.cc
kyessenov: Using references to mutate is not consistent with the rest of this file.

extensions/common/context.cc
kyessenov: No, you can see that flatbuffers has an identical loop https://github.com/google/flatbuffers/blob/16836ff95a7f112f66a8c5994a508e6bfec010c9/src/idl_parser.cpp#L76

extensions/common/context.cc
kyessenov: No, since this just compares against "1".

extensions/common/context.cc
lambdai: I see

kyessenov:
@lambdai There is no requirement that Envoy provides utf-8 for anything. In fact, latin1 encoding is perfectly functional in Envoy. We just don't want to deal with latin1 in stackdriver.

extensions/common/context.cc
```diff
@@ -520,5 +534,19 @@ bool getAuditPolicy() {
...
 }
 
+bool sanitizeBytes(std::string* buf) {
+  char* start = buf->data();
+  const char* const end = start + buf->length();
+  bool modified = false;
+  while (start < end) {
+    if (flatbuffers::FromUTF8(const_cast<const char**>(&start)) < 0) {
+      *start = ' ';
+      start += 1;
```
lambdai, line 544: It seem the start could be advanced if the first byte is legit (e.g. the first byte indicate len = 2  b'11000000').
I *think* your goal is to replace b'1100000' by SPACE and start from the byte after b'11000000'. 

https://github.com/google/flatbuffers/blob/df2df21ec1be2468106fed10c1533eacc1cf0d2e/include/flatbuffers/util.h#L528


my imagined nagative case is a two byte string: b'11000000' b'11000000' 

Can you confirm?

kyessenov: Good point. Missed this with the cast. PTAL again, and updated test case with \C0.

lambdai: LGTM!

lambdai (approved):
Thank you for the update!

bianpengyuan:
@kyessenov  curious, how does envoy deal invalid utf8 at host or url path?

kyessenov:
@bianpengyuan Envoy issues a codec error and returns 400 Bad Request

PR #3332, opened 2021-05-06
envoy: http connection pool factory for metadata upstream extension

tbarrella:
**What this PR does / why we need it**: Next step for istio/istio#31444

**Special notes for your reviewer**: Basic structure/testing before implementing the metadata -> headers mapping. See https://github.com/envoyproxy/envoy/blob/v1.18.0/test/integration/upstreams/per_host_upstream_config.h and https://github.com/envoyproxy/envoy/blob/v1.18.0/test/integration/cluster_upstream_extension_integration_test.cc

### DIFF
--- a/src/envoy/upstreams/http/metadata/config.cc
+++ b/src/envoy/upstreams/http/metadata/config.cc
@@ -36,10 +36,7 @@ MetadataGenericConnPoolFactory::createGenericConnPool(
     const Router::RouteEntry& route_entry,
     absl::optional<Envoy::Http::Protocol> downstream_protocol,
     Upstream::LoadBalancerContext* ctx) const {
-  if (is_connect) {
-    // TODO(tbarrella)
-    return nullptr;
-  }
+  ASSERT(!is_connect);
   auto ret = std::make_unique<MetadataConnPool>(
       thread_local_cluster, is_connect, route_entry, downstream_protocol, ctx);
   return ret->valid() ? std::move(ret) : nullptr;
--- a/src/envoy/upstreams/http/metadata/config.cc
+++ b/src/envoy/upstreams/http/metadata/config.cc
@@ -1,0 +1,53 @@
+/* Copyright 2021 Istio Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


src/envoy/upstreams/http/metadata/config.h
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Config registration for the MetadataConnPool.
+ * @see Router::GenericConnPoolFactory
+ */
+class MetadataGenericConnPoolFactory : public Router::GenericConnPoolFactory {
+ public:
+  std::string name() const override {
+    return "istio.filters.connection_pools.http.metadata";
+  }
+  std::string category() const override { return "istio.upstreams"; }
```
lambdai, line 42: "istio.upstreams" is probably not the best category. 

The MetadataGenericConnPoolFactory is compatible with the Router::GenericConnPoolFactory and TcpProxy::GenericConnPoolFactory so the category should be "envoy.upstreams"

src/envoy/upstreams/http/metadata/config.cc
```diff
@@ -0,0 +1,53 @@
...
+namespace Metadata {
+
+Router::GenericConnPoolPtr
+MetadataGenericConnPoolFactory::createGenericConnPool(
+    Upstream::ThreadLocalCluster& thread_local_cluster, bool is_connect,
+    const Router::RouteEntry& route_entry,
+    absl::optional<Envoy::Http::Protocol> downstream_protocol,
+    Upstream::LoadBalancerContext* ctx) const {
+  if (is_connect) {
+    return nullptr;
```
lambdai, line 40: It's ok to return nullptr at this comment. 

But in the HBONE project this MetadataGenericConnPool has to support user originated CONNECT as well.
We cannot let the envoy default connpool handle it because the user originated CONNECT need HBONE headers. 
This header manipulation must be executed by MetadataGenericConnPool.

Please add comment this is temporary or fix it in this PR.

lambdai (changes requested):
Thank you! Only a few comments

src/envoy/upstreams/http/metadata/config.cc
```diff
@@ -0,0 +1,54 @@
...
+namespace Metadata {
+
+Router::GenericConnPoolPtr
+MetadataGenericConnPoolFactory::createGenericConnPool(
+    Upstream::ThreadLocalCluster& thread_local_cluster, bool is_connect,
+    const Router::RouteEntry& route_entry,
+    absl::optional<Envoy::Http::Protocol> downstream_protocol,
+    Upstream::LoadBalancerContext* ctx) const {
+  if (is_connect) {
+    // TODO(tbarrella)
```
lambdai, line 40: Just ASSERT since it's not supported

lambdai (approved):
LGTM
It would be good if you can add the comment or integration test to declare the extension only handle http upstream

lambdai:
meanwhile cc @lizan

tbarrella:
@lizan do you have a chance to review?

lizan (approved):

PR #3339, opened 2021-05-10
use canonical name as fallback for destination_service

douglas-reid:

Provides a more useful fallback for failed metadata exchange.

https://github.com/istio/istio/issues/32756


extensions/common/context.cc
```diff
@@ -551,5 +550,21 @@ bool sanitizeBytes(std::string* buf) {
   return modified;
 }
 
+// Used for `destination_service` fallback. Unlike elsewhere when that fallback
+// to workload name, this falls back to "unknown" when the canonical name label
+// is not found. This preserves the existing behavior for `destination_service`
+// labeling. Using a workload name as a service name could be potentially
+// problematic.
+std::string getServiceNameFallback() {
```
bianpengyuan, line 558: This will happen on both inbound and outbound right? I think we only want this fallback on inbound.

extensions/common/context.cc
```diff
@@ -551,5 +550,21 @@ bool sanitizeBytes(std::string* buf) {
...
+  const auto local_node_buffer = extractLocalNodeFlatBuffer();
```
mandarjog, line 559: this is not used in the code, is it done for a side-effect?

douglas-reid: it was sloppy typing without a compiler while participating in that meeting. i wanted to get general clearance before i found a way to a machine that could easily build the proxy.

extensions/common/context.cc
kyessenov: +1. It's not correct for outbound (although gateways are outbound).

douglas-reid: good point. handling that will require more grace.

douglas-reid: fixed.

extensions/common/context.cc
douglas-reid: adopted @kyessenov 's suggestion.

nrjpoddar:
@douglas-reid is this possibly a minor semantic breaking change?

douglas-reid:
> @douglas-reid is this possibly a minor semantic breaking change, and if yes, are we going to give users an option to revert to old behavior?

this is a minor semantic breaking change. it pairs with disabling host header fallback by default, per request in the linked issue. re-enabling host header fallback is one way to revert. i don't know if we want another.

extensions/common/context.cc
```diff
@@ -551,5 +551,22 @@ bool sanitizeBytes(std::string* buf) {
   return modified;
 }
 
+// Used for `destination_service` fallback. Unlike elsewhere when that fallback
+// to workload name, this falls back to "unknown" when the canonical name label
+// is not found. This preserves the existing behavior for `destination_service`
+// labeling. Using a workload name as a service name could be potentially
+// problematic.
+std::string_view getServiceNameFallback() {
```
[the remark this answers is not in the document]
douglas-reid: done.

douglas-reid:
@mandarjog thoughts on trying to get this in for 1.10?

mandarjog:
I have a mild preference for getting it in if we are not too late.

. @liminw do you consider a must have?

mandarjog:
@kyessenov @bianpengyuan ptal. We should get this in.

extensions/common/context.cc
```diff
@@ -551,5 +554,21 @@ bool sanitizeBytes(std::string* buf) {
...
+// Used for `destination_service` fallback. Unlike elsewhere when that fallback
+// to workload name, this falls back to "unknown" when the canonical name label
+// is not found. This preserves the existing behavior for `destination_service`
+// labeling. Using a workload name as a service name could be potentially
+// problematic.
+std::string getServiceNameFallback() {
+  auto buf = getProperty({"node", "metadata", "LABELS"});
+  if (buf.has_value()) {
+    for (const auto& [key, val] : buf.value()->pairs())
+      if (key == ::Wasm::Common::kCanonicalServiceLabelName.data()) {
```
howardjohn, line 566: Why is canonical service name the right thing to use? For the majority of workloads there is 1 Service per workload, which we encode in the cluster metadata. Would it be better to use that?

bianpengyuan: This happen on inbound passthrough, destination service cannot be decided.

PR #3352, opened 2021-05-12
Fix stackdriver requestUrl mapping

mandarjog:
[Cloud Loggin.LogEntry#HttpRequest](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest) specifies that `requestUrl` should be `The scheme (http, https), the host name, the path and the query portion of the URL that was requested` 

The current version did not include the query string.

Note: Also removed `request_url_path` which was duplicated in `url_path`


douglas-reid (commented):
this seems reasonable. probably need to update the tests too, though.

kyessenov (approved):

PR #3361, opened 2021-05-20
fix the stackdriver logging for dry-run policy

yangminzhu:
This PR fixes the stackdriver logging when only one of the authorization ALLOW or DENY policy is used.

Added e2e tests to cover the fix:
1) only ALLOW policy;
2) only DENY policy; and
3) both policies;


test/envoye2e/stackdriver_plugin/stackdriver_test.go
```diff
@@ -842,63 +842,88 @@ func TestStackdriverRbacAccessDenied(t *testing.T) {
...
-		"RbacDryRun":                  "true",
-	}, envoye2e.ProxyE2ETests)
+	rbacCases := []struct {
+		name             string
+		rbacDryRunKey    string
+		rbacDryRunFilter string
+	}{
+		{
+			name:             "ActionBoth",
+			rbacDryRunKey:    "RbacDryRunDenied",
```
bianpengyuan, line 852: nit: it is a bit weird to specify the key instead of the value here. Can you instead make the value the variable? Like make the key `bacDryRunResult`.

yangminzhu: sure, updated.

bianpengyuan (approved):

PR #3376, opened 2021-06-15
update envoy

kyessenov:

Upstream PR https://github.com/envoyproxy/envoy/pull/16465 appears to have broken some telemetry. Until we debug it out, this PR allows us to move forward.


bianpengyuan (approved):
Looks like a manual PR in istio/istio will also be needed. We should create an issue at least in istio/istio to track this and mark that as 1.11 release blocker.

PR #3386, opened 2021-06-24
envoy: pick a bind_to_port config fix to unblock istio proxy CI

lambdai:
The fix is on top of envoyproxy main commit 420d2b9d51ab66587647e450f8e958fbb52ac950 on 06/21.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -38,12 +38,12 @@ bind(
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
 # Note: this is needed by release builder to resolve envoy dep sha to tag.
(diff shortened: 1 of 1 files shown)


kyessenov:
Do we have an ETA on upstream fix? I don't sense urgency in carrying a fork right now but I might be wrong.

lambdai:
> Do we have an ETA on upstream fix? I don't sense urgency in carrying a fork right now but I might be wrong.

Good point. There are string_view issues that @bianpengyuan is looking so we cannot catch up with the latest build yet.

lambdai:
Not needed: the source of this PR is landed in envoy upstream

PR #3387, opened 2021-06-25
Update envoy SHA to 6/18

bianpengyuan:


bianpengyuan:
@PiotrSikora @lambdai Any idea about why centos has `range-loop-analysis` check enabled? Should we turn it off? https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_proxy/3387/release-centos-test_proxy/1408575965637906432

lambdai:
Add a "-Wno-range-loop-analysis" around https://github.com/istio/proxy/blob/master/.bazelrc#L51-L54 ?

lambdai:
> Add a "build --cxxopt -Wno-range-loop-analysis" around https://github.com/istio/proxy/blob/master/.bazelrc#L51-L54 ?

I want to make it clear the -Wno-range-loop-analysis is following "--cxxopt" and *NOT* in `action_env`

PiotrSikora:
I wonder if the issue is outdated compiler and a false positive warning? It's not happening on Ubuntu nor upstream.

extensions/common/util.h
```diff
@@ -17,11 +17,28 @@
...
 namespace Wasm {
 namespace Common {
 
+// None response flag.
+const std::string NONE = "-";
+
 // Parses an integer response flag into a readable short string.
 const std::string parseResponseFlag(uint64_t response_flag);
 
+// Used for converting sanctioned uses of std string_view (e.g. extensions) to
```
lambdai, line 31: Could you use source/common/common/stl_helpers.h ?

extensions/common/util.h
```diff
@@ -17,11 +17,28 @@
 
 #include <string>
 
+#include "absl/strings/string_view.h"
+
 namespace Wasm {
 namespace Common {
 
+// None response flag.
+const std::string NONE = "-";
```
lambdai, line 26: global variable std::string is not recommended in prod code. The string can be destroyed at exiting, and access log(or any other unknown components) can refers to the heap allocated "-".

Since it's the tech debt of the existing code, could you fix or add a comment to address it later?

extensions/common/wasm/json_util.h
```diff
@@ -38,6 +39,7 @@ template <typename T>
 std::pair<std::optional<T>, JsonParserResultDetail> JsonValueAs(
     const JsonObject&) {
   static_assert(true, "Unsupported Type");
+  return std::make_pair(std::nullopt, JsonParserResultDetail::UNKNOWN);
```
lambdai, line 42: Where does it come from? 
Does `static_assert(true,...) make sense? It seems it should be `static_assert(`*false* in the past and static_assert should be removed now...

extensions/metadata_exchange/plugin.cc
```diff
@@ -130,7 +132,13 @@ bool PluginRootContext::updatePeer(std::string_view key,
     }
   }
 
-  auto bytes = Base64::decodeWithoutPadding(peer_header);
+#ifndef NULL_PLUGIN
+  auto peer_header_view = peer_header;
+#else
+  auto peer_header_view = Wasm::Common::toAbslStringView(peer_header);
```
lambdai, line 138: I am not fully sure. Should `Wasm::Common::toAbslStringView` wraps NULL_PLUGIN? Otherwise, Wasm::Common::toAbslStringView is duplicating the envoy version

lambdai (commented):
Thanks! A few questions though

extensions/common/util.h
kyessenov: We probably can't because of Wasm compilation method. We're trying to keep code minimal since Wasm is not as fast.

extensions/common/wasm/json_util.h
bianpengyuan: I thought I got a build error because of no return, but looks like that was not the case. Updated.

extensions/common/util.h
bianpengyuan: Done.

extensions/common/util.h
bianpengyuan: Yeah we'd need to copy those two functions for Wasm module build. It is hard to introduce any Envoy source dep.

extensions/metadata_exchange/plugin.cc
bianpengyuan: Not sure how could you wrap this?

bianpengyuan:
> > Add a "build --cxxopt -Wno-range-loop-analysis" around https://github.com/istio/proxy/blob/master/.bazelrc#L51-L54 ?
> 
> I want to make it clear the -Wno-range-loop-analysis is following "--cxxopt" and _NOT_ in `action_env`

@lambdai does not seems to work..

extensions/metadata_exchange/plugin.cc
lambdai: NVM

lambdai:
```
-Wno-range-loop-analysis -D_GLIBCXX_USE_CXX11_ABI=1'...
-Wrange-loop-analysis
```

Looks like conflict args are provided and the order matters.
Need to eliminate the latter one or place our override behind

extensions/stackdriver/common/utils.cc
```diff
@@ -106,9 +107,12 @@ std::string getGCEInstanceUID(const ::Wasm::Common::FlatNode &node) {
   }
 
   if (name.size() > 0 && project && location) {
-    return absl::StrCat("//compute.googleapis.com/projects/",
-                        project->value()->string_view(), "/zones/",
-                        location->value()->string_view(), "/instances/", name);
+    return absl::StrCat(
+        "//compute.googleapis.com/projects/",
+        ::Wasm::Common::toAbslStringView(project->value()->string_view()),
```
kyessenov, line 112: nit: doesn't absl StrCat handle std::string_view natively?

kyessenov (approved):
I think Wasm needs to switch to absl string_view to make this cleaner.

extensions/stackdriver/common/utils.cc
PiotrSikora: It doesn't - it supports `absl::string_view`, which by default is equivalent to `std::string_view` in C++17, but Envoy switched to non-standard null-terminated variants in envoyproxy/envoy#16029, so `std::string_view` is no longer supported.

lambdai: @PiotrSikora Could you clarify wrt null?
is it null-terminated or nullptr constructed?

PiotrSikora: Good catch - `nullptr`-constructed.

PR #3395, opened 2021-07-01
Update Envoy SHA to 07-12.

bianpengyuan:


kyessenov:
I think you can undo new connection pool override in https://github.com/istio/proxy/pull/3376/files

bianpengyuan:
@kyessenov  I am not sure how `TestStackdriverPayloadGateway` ever worked. This is the important part of the failure log: https://gist.github.com/bianpengyuan/f35002b956375179fa22dbee86d6219f. The client log entry misses destination workload metadata.

I look at the test, it simulates a middle proxy with both inbound and outbound listener configured without metadata exchange, but how does server/client access log ever gets peer metadata? Is the dynamic metadata somehow preserved across streams?

kyessenov:
@bianpengyuan I think there is metadata exchange but they are both installed in the same proxy so source and destination are the same? Stream metadata could be preserved, it depends on how we configured it. I think it should be per-request now but it's been awhile I looked into it.

bianpengyuan:
@kyessenov Ah yeah sorry, misread the code. Turns out the MX filter always consider itself outbound regardless of which listener it is on, and it happens after https://github.com/envoyproxy/envoy/commit/5a8f89a90ab7c18f336f7d6b89686ca878fde3c4.

@mathetake Could you please take a look? metadata exchange filter get direction information with property API: https://github.com/istio/proxy/blob/3a9cc090479baacd31f87b62f3313493aa76dce5/extensions/common/context.cc#L208. After that change, if both inbound and outbound listeners are configured, metadata exchange filters will just pick one of them to get listener information.

mathetake:
OK let me take a look in a few days

mathetake:
ok it looks like the test has depended on the bag I fixed in that PR. 

When I looked thorough the log, I see that the exactly the same configuration both for inbound/outbound mx filter:


**outbound**
```
      - name: mx_outbound162
        typed_config:
          "@type": type.googleapis.com/udpa.type.v1.TypedStruct
          type_url: envoy.extensions.filters.http.wasm.v3.Wasm
          value:
            config:
              vm_config:
                runtime: envoy.wasm.runtime.null
                code:
                  local: { inline_string: "envoy.wasm.metadata_exchange" }
                allow_precompiled: true
              configuration:
                "@type": "type.googleapis.com/google.protobuf.StringValue"
                value: |
                  { "max_peer_cache_size": 20 }
```

**inbound**

```
      - name: mx_inbound162
        typed_config:
          "@type": type.googleapis.com/udpa.type.v1.TypedStruct
          type_url: envoy.extensions.filters.http.wasm.v3.Wasm
          value:
            config:
              vm_config:
                runtime: envoy.wasm.runtime.null
                code:
                  local: { inline_string: "envoy.wasm.metadata_exchange" }
                allow_precompiled: true
              configuration:
                "@type": "type.googleapis.com/google.protobuf.StringValue"
                value: |
                  { "max_peer_cache_size": 20 }
```

which results in using the same plugin first created. This is the expected behavior.


I would recommend to set different root_id or vm_id for each of them, and believe that would fix the test.

mathetake:
^ cc @PiotrSikora

mathetake:
maybe we could do better, like, use the streamInfo of where the plugin is used, rather than embedding streamInfo into pluginBase (which is how it behaves now).

bianpengyuan:
> I would recommend to set different root_id or vm_id for each of them, and believe that would fix the test.

@mathetake we originally had separated root_id/vm_id for inbound and outbound MX filters, in order to get the correct listener direction. Current behavior was considered as a "bug" and fixed, so that Wasm filter can get the listener context that it is on. It feels unnecessary to have different root / VM to just differentiate listener direction. There can be other listener properties, and I don't think we want to have different root context for each of them? It might be fine for us, but likely this will be surprised to the Wasm filter developers.

mathetake:
>  It feels unnecessary to have different root / VM to just differentiate listener direction. There can be other listener properties, and I don't think we want to have different root context for each of them? It might be fine for us, but likely this will be surprised to the Wasm filter developers.

yeah actually it is unnecessary if you use `getProperty` on stream context, then this should work correctly without setting vm_id or whatever. But the point is that mx filter plugin uses `getProperty` in the constructor of RootContext, which is only called once the plugin handle is created, and cache it: https://github.com/istio/proxy/blob/e02cbbe2e3f390afd9711c5d5e9f4bb5462059ea/extensions/metadata_exchange/plugin.h#L74
This is the root cause here. (stackdriver also uses `getProperty` in `::configure` so same goes for it.)

  This is by the original design of native Proxy-Wasm filters that a single in-VM plugin can be used at multiple extension points of the host implementation. But yeah we should reconsider this "original" behavior since as you said this is confusing and error-prone. Please refer to @PiotrSikora's comment here https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/175#issuecomment-869987292

mathetake:
ah - I misread the code, `getTrafficDirection` is called on each stream context.. seems weird.

mathetake:
update: successfully reproduced with Go SDK + a small Envoy config - dig into now.

mathetake:
OK I found the root cause which has existed since the beginning and just been caught and surfaced here due to my bug fix in that PR. The problem is much more complex than I expected. anyway thanks for reporting! now working on a fix.

mathetake:
So I raised a PR in Envoy https://github.com/envoyproxy/envoy/pull/17243 . It was much more complex than I thought, and I think it's going to take some time to land. Thanks.

kyessenov:
@mathetake We have several fixes after the Wasm change that need to land. Do you think we need a deprecation flag to guard your change or we can get your fix PR soon?

mathetake:
I think the fix can get merged soon though the stackdriver extension needs to be revised accordingly - I see that it caches the direction in root context: https://github.com/istio/proxy/blob/master/extensions/stackdriver/stackdriver.cc#L379. But it all depends on how @PiotrSikora review the PR.

lizan (approved):

PiotrSikora:
@bianpengyuan could you bump Envoy SHA? The Wasm fix was merged.

kyessenov (approved):

douglas-reid (approved):

PR #3403, opened 2021-07-08
Update the upstream envoy version as well as to fix some breakage.

incfly:
https://github.com/istio/istio/issues/33809

https://github.com/envoyproxy/envoy/pull/16795/files updated the constructor so we need to refactor internal usage of the context constructor accordingly.


bianpengyuan:
An update is ongoing, which is blocked by upstream changes: https://github.com/istio/proxy/pull/3395

incfly:
noticed when clicking other open PRs to understand the test failures..
should take a look of existing dep updaets pr next time.

PR #3404, opened 2021-07-11
wasm: delete unused code.

mathetake:
Wasm is already upstreamed to Envoy and this PR removes the unused codes from here. Notably this removes `IstioContext`, `IstioWasm` and thus the factory overrides over EnvoyWasm class.
Ref: https://github.com/envoyproxy/envoy/pull/17268

Signed-off-by: Takeshi Yoneda <takeshi@tetrate.io>


mathetake:
cc @PiotrSikora

bianpengyuan (approved):

PR #3413, opened 2021-07-15
wasm: remove unused Istio specific Wasm code under src.

mathetake:
Combined with #3404, this PR removes unused Istio specific Wasm extension code under src. Necessary change per https://github.com/envoyproxy/envoy/pull/17268.

Signed-off-by: Takeshi Yoneda <takeshi@tetrate.io>


bianpengyuan (approved):

mathetake:
OK I found that the IstioExtension class calls `registerPrometheusNamespace("istio");` in its constructor so we have to do it somewhere. This is related to https://github.com/envoyproxy/envoy/issues/14920 https://github.com/istio/istio/issues/27635 and maybe we should support that via foregin functions.

mathetake:
so for now, leave the namespace registration as-is, will come back to this after I fixed  https://github.com/envoyproxy/envoy/issues/14920

mathetake:
fyi: created a patch https://github.com/envoyproxy/envoy/pull/17357 to allow plugins to register namespaces

PR #3439, opened 2021-08-01
Add check to prevent proxy response flag out of sync with upstream

ChristophBleyer:
Istio proxy has a response flag parsing logic, which needs to keep in sync with a set of response flags from the upstream.

This PR adds a presubmit check script into istio/proxy, which verifies the LastFlag is the same in istio/proxy and envoyproxy/envoy.

fixes istio/istio#33678.


ChristophBleyer:
@googlebot I signed it!

scripts/verify-last-flag-matches-upstream.sh
```diff
@@ -0,0 +1,74 @@
...
+curl -sSL "https://raw.githubusercontent.com/${ENVOY_ORG}/${ENVOY_REPO}/${ENVOY_LATEST_SHA}/${ENVOY_PATH_LASTFLAG_SPEC}" > "${PATH_LASTFLAG_SPEC_UPSTREAM}"
+
+# Extract the LastFlag specification from both sources and trim all white spaces.
+
+if grep -q "^\s*LastFlag\s*=.*$" "${PATH_LASTFLAG_SPEC_DOWNSTREAM}"
+then
+    DOWNSTREAM_LASTFLAG=$(grep -m1 "^\s*LastFlag\s*=.*$" "${PATH_LASTFLAG_SPEC_DOWNSTREAM}"|tr -d '[:space:]')
+else
+    echo "istio/proxy: LastFlag not specified in ${PATH_LASTFLAG_SPEC_DOWNSTREAM}"
+    rm "${PATH_LASTFLAG_SPEC_UPSTREAM}"
```
bianpengyuan, line 42: nit: you can use trap on exit, to avoid adding rm on every branch: `trap 'rm -rf ${PATH_LASTFLAG_SPEC_UPSTREAM}' EXIT`

bianpengyuan (commented):
Thanks for the change! You can add the check to the lint target for now: https://github.com/istio/proxy/blob/683f2d07e648d55bdddbe5cf125abebc3316248c/Makefile.core.mk#L149

bianpengyuan:
@ChristophBleyer I think you need to `chmod +x` on the script file.

bianpengyuan (approved):

PR #3458, opened 2021-08-16
fix: handle result to unknown when istio_version is missing

Xunzhuo:

fix: handle result to unknown when istio_version is missing

**Special notes for your reviewer**: cc @bianpengyuan

### DIFF
--- a/extensions/stats/plugin.cc
+++ b/extensions/stats/plugin.cc
@@ -450,9 +450,9 @@ bool PluginRootContext::initializeDimensions(const json& j) {
   Metric build(MetricType::Gauge, absl::StrCat(stat_prefix, "build"),
                {MetricTag{"component", MetricTag::TagType::String},
                 MetricTag{"tag", MetricTag::TagType::String}});
+  std::string istio_version = flatbuffers::GetString(local_node.istio_version()) == "" ? absl::StrCat(unknown, ";") : absl::StrCat(istio_version, ";");
   build.record(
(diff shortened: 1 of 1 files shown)


bianpengyuan (commented):
lgtm, although likely there will be lint error for line too long.

extensions/stats/plugin.cc
```diff
@@ -450,9 +450,9 @@ bool PluginRootContext::initializeDimensions(const json& j) {
   Metric build(MetricType::Gauge, absl::StrCat(stat_prefix, "build"),
                {MetricTag{"component", MetricTag::TagType::String},
                 MetricTag{"tag", MetricTag::TagType::String}});
+  std::string istio_version = flatbuffers::GetString(local_node.istio_version()) == "" ? absl::StrCat(unknown, ";") : absl::StrCat(istio_version, ";");
```
bianpengyuan, line 453: This seems wrong.. istio_version is used before it is defined.

bianpengyuan (changes requested):

extensions/stats/plugin.cc
Xunzhuo: @bianpengyuan Thanks Peter~ I think it`s ok now

Xunzhuo:
cc @bianpengyuan  I think it`s ok now, should I rebase these commits?

bianpengyuan (approved):

PR #3477, opened 2021-09-03
clear route cache after JWT validation

yangminzhu:
This is needed for the JWT claim based routing.

The validated JWT claims are stored in the dynamic metadata and later used for routing, we need to clear the route cache after the JWT validation as it may change the route result.

Tested manually and will add full e2e test in Istio repo.

Approved design doc: https://docs.google.com/document/d/1s8Lykf8-KUZTLiG4YMSCHvxBDwZzHlrrHyZ7cJymc_8/edit#

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -104,8 +104,8 @@ cc_library(
 # 1) find the ISTIO_API SHA you want in git
 # 2) wget https://github.com/istio/api/archive/$ISTIO_API_SHA.tar.gz && sha256sum $ISTIO_API_SHA.tar.gz
 #
-ISTIO_API = "31d048906d97fb7f6b1fa8e250d3fa07456c5acc"
-ISTIO_API_SHA256 = "5bf68ef13f4b9e769b7ca0a9ce83d9da5263eed9b1223c4cbb388a6ad5520e01"
+ISTIO_API = "75bb24b620144218d26b92afedbb428e4d84e506"
+ISTIO_API_SHA256 = "c72602c38f7ab10e430618e4ce82fee143f4446d468863ba153ea897bdff2298"
 
 def istioapi_repositories(bind = True):
     BUILD = """
--- a/src/envoy/http/authn/http_filter.cc
+++ b/src/envoy/http/authn/http_filter.cc
@@ -86,6 +86,9 @@ FilterHeadersStatus AuthenticationFilter::decodeHeaders(
     decoder_callbacks_->streamInfo().setDynamicMetadata(
         Utils::IstioFilterName::kAuthentication, data);
     ENVOY_LOG(debug, "Saved Dynamic Metadata:\n{}", data.DebugString());
+    // Clear the route cache after saving the dynamic metadata for routing
+    // based on JWT claims.
+    decoder_callbacks_->clearRouteCache();
   }
   state_ = State::COMPLETE;
   return FilterHeadersStatus::Continue;
--- a/src/envoy/http/authn/http_filter.cc
+++ b/src/envoy/http/authn/http_filter.cc
@@ -86,9 +86,12 @@ FilterHeadersStatus AuthenticationFilter::decodeHeaders(
     decoder_callbacks_->streamInfo().setDynamicMetadata(
         Utils::IstioFilterName::kAuthentication, data);
     ENVOY_LOG(debug, "Saved Dynamic Metadata:\n{}", data.DebugString());
-    // Clear the route cache after saving the dynamic metadata for routing
-    // based on JWT claims.
-    decoder_callbacks_->clearRouteCache();
+    if (!filter_config_.disable_clear_route_cache()) {
+      // Clear the route cache after saving the dynamic metadata for routing
+      // based on JWT claims.
+      decoder_callbacks_->clearRouteCache();
+      ENVOY_LOG(debug, "Istio authn filter cleared route cache.");
+    }
   }
   state_ = State::COMPLETE;
   return FilterHeadersStatus::Continue;


bianpengyuan (commented):
Does this mean for every request, regardless of whether dynamic routing is configured or not, we will calculate rout twice? That seems unnecessary.. Can we make this configurable?

yangminzhu:
> Does this mean for every request, regardless of whether dynamic metadata routing is configured or not, we will calculate rout twice? That seems unnecessary.. Can we make this configurable?

It's only when JWT validation is enabled, and the istio_authn filter will always write to the dynamic metadata. So there is no change when JWT validation is not used.

We could add a flag to make this configurable but I'm afraid the benefit does not worth the efforts:
1. the extra performance overhead is already negligible compared to the JWT token cryptographic validation and the write to the dynamic metadata.  
2. It will really complicates the control plane logic as it will require the security code to have a hard dependency on the routing API if we want to set this optionally.

bianpengyuan:
> > Does this mean for every request, regardless of whether dynamic metadata routing is configured or not, we will calculate rout twice? That seems unnecessary.. Can we make this configurable?
> 
> It's only when JWT validation is enabled, and the istio_authn filter will always write to the dynamic metadata. So there is no change when JWT validation is not used.
> 
> We could add a flag to make this configurable but I'm afraid the benefit does not worth the efforts:
> 
> 1. the extra performance overhead is already negligible compared to the JWT token cryptographic validation and the write to the dynamic metadata.
> 2. It will really complicates the control plane logic as it will require the security code to have a hard dependency on the routing API if we want to set this optionally.

I see. Did not realize that it was added only when Jwt is enabled. My understanding is that overhead from route computation is low, so this looks fine to me, but better let @lambdai double check on this.

kyessenov:
We probably need some way to disable this because users might be already changing routes based on other filters (e.g. Lua), so it would be unexpected if that stops working. Is there any problem with ordering filters? If some action based on route happens before this filter, then they need to be re-evaluated on the new route as in a proper redirect.

lambdai:
IMHo The outbound route is not trivial. But that is unavoidable.

yangminzhu:
> We probably need some way to disable this because users might be already changing routes based on other filters (e.g. Lua), so it would be unexpected if that stops working. 

I don't know how will this affect other filters in the first place, can you clarify a little bit? If the Lua filter just changes the header and then clear the route cache, then this filter should be independent from it. 

> Is there any problem with ordering filters? If some action based on route happens before this filter, then they need to be re-evaluated on the new route as in a proper redirect.

AFAIK, the standard behavior in Envoy is to `If a filter has modified the headers in a way that affects routing, clearRouteCache() must be called to clear the cache.`, the `ext-authz` filter implements this without any ordering problem, so I think it should be the same for istio_authn filter here (though we are doing it for dynamic metadata instead of headers)

kyessenov:
Yeah, I understand that clear route cache works in isolation. But this behavior does not compose well. What if both ext_authz and authn want to change route? Which one wins and why? Basically, we used to have a way to reconcile route changes with mixer, but now it's just whatever order filters are chained, and that's not always right.

lambdai:
> We probably need some way to disable this because users might be already changing routes based on other filters (e.g. Lua), so it would be unexpected if that stops working. Is there any problem with ordering filters? If some action based on route happens before this filter, then they need to be re-evaluated on the new route as in a proper redirect.

This is technically possible with EnvoyFilter, but user can always add their new filter and may shoot on their foot.

We can only coordinate known istio filters

yangminzhu:
> Yeah, I understand that clear route cache works in isolation. But this behavior does not compose well. What if both ext_authz and authn want to change route? Which one wins and why? Basically, we used to have a way to reconcile route changes with mixer, but now it's just whatever order filters are chained, and that's not always right.

I think the right behavior should be to use the route decision from whatever the last filter that made any changes affecting the route decision. This is kind similar to the authz, the authorization policy does not take user's custom filters into consideration when making access control decisions.

Also note, the authn filter modifies the dynamic metadata only which should really not cause any conflict at all because we just added the dynamic metadata based routing a week ago so it should not be used by any existing customers.

And if the user plans to use this in the future, they need to configure the dynamic metadata matcher in the routing config which requires to specify the namespace (filter name) of the dynamic metadata, it's extremely unlikely they would use the same same as the Istio built-in filter (istio_authn) here.

kyessenov:
One dangerous side effect is that the user might not want route reset. E.g. if they have some append/delete header operations prior to authn filter but want to keep the route the same (e.g. they are doing true redirect). It seems like this PR will interfere with that. We need a way to limit the blast radius of this change, since it's difficult to anticipate what is going to break.

yangminzhu:
> One dangerous side effect is that the user might not want route reset. E.g. if they have some append/delete header operations prior to authn filter but want to keep the route the same (e.g. they are doing true redirect). It seems like this PR will interfere with that. We need a way to limit the blast radius of this change, since it's difficult to anticipate what is going to break.

So the user has a filter to change the header that would change the routing result but intentionally want to keep the old route result? 

This seems conflict with the [Envoy comment](https://github.com/envoyproxy/envoy/blob/main/envoy/http/filter.h#L206-L207) that explicitly says `If a filter has modified the headers in a way that affects routing, clearRouteCache() must be called to clear the cache.`. It is also confusing to me, if they do not need the routing, they should not configure it in virtual service in the first place.

And the user should also add their filter after authn filter (they won't be affected if they do so). Otherwise their filter is making decisions on an unauthenticated request that is insecure and should be discouraged.

My concern is that it seems we are adding a flag (or whatever config) for an internal implementation detail to a very few highly customized users (if any). I'm skeptical about these use cases in the first place and I prefer not to do it because it's not needed by majority customers and there is an alternative solution (put their filter after istio authn) that not only solve the problem but also a more safer way in general.

And I agree with @lambdai, we can only coordinate known istio filters.

bianpengyuan:
Making it possible to be turned off always sounds safer to me.. As a alternative can we add the turn off noob to the filter config, but do not implement setting it first class? User can use EnvoyFilter to overwrite it if they hit the edge cases.

mandarjog:
I think this is the wrong place to clear cache.
Lua filters clears cache when it changes headers and so does the wasm sandbox.
That means it is the responsibility of those filters  to clear cache or sometimes make it configurable.

kyessenov:
> > One dangerous side effect is that the user might not want route reset. E.g. if they have some append/delete header operations prior to authn filter but want to keep the route the same (e.g. they are doing true redirect). It seems like this PR will interfere with that. We need a way to limit the blast radius of this change, since it's difficult to anticipate what is going to break.
> 
> So the user has a filter to change the header that would change the routing result but intentionally want to keep the old route result?

Yes, this is a change in behavior since none of Istio filters clear route cache at the moment. It's possible to write a Wasm filter that appends a header that affects routing without clearing the cache, and then insert it prior to authn with Wasm API. Remember, users don't have to use Istio authn/authz, they can write their own with Wasm or Lua.  

> 
> This seems conflict with the [Envoy comment](https://github.com/envoyproxy/envoy/blob/main/envoy/http/filter.h#L206-L207) that explicitly says `If a filter has modified the headers in a way that affects routing, clearRouteCache() must be called to clear the cache.`. It is also confusing to me, if they do not need the routing, they should not configure it in virtual service in the first place.

It's not always intentional. The config is complex, and we should keep its behavior intact.

> And the user should also add their filter after authn filter (they won't be affected if they do so). Otherwise their filter is making decisions on an unauthenticated request that is insecure and should be discouraged.

This is totally fine, we don't require istio authn for everything.

yangminzhu:
> I think this is the wrong place to clear cache.
> Lua filters clears cache when it changes headers and so does the wasm sandbox.
> That means it is the responsibility of those filters to clear cache or sometimes make it configurable.

@mandarjog  the JWT claim dynamic metadata does not exist until authn filter writes to it, and authn filter will always write to it after successful JWT validation. In other words, the authn filter will always change the dynamic metadata affecting the routing decision. Also if we do not clear cache here, where should we do it?

We can add a flag to make this configurable so that users can opt-out (do not clear route cache) for the corner cases mentioned by @kyessenov. But I want to make it clear we are not going to expose the flag in any first-class Istio API or MeshConfig, the user will likely only use the EnvoyFilter to set it.

mandarjog:
On Fri, Sep 3, 2021 at 10:16 AM Yangmin Zhu ***@***.***>
wrote:

> I think this is the wrong place to clear cache.
> Lua filters clears cache when it changes headers and so does the wasm
> sandbox.
> That means it is the responsibility of those filters to clear cache or
> sometimes make it configurable.
>
> @mandarjog <https://github.com/mandarjog> the JWT claim dynamic metadata
> does not exist until authn filter writes to it, and authn filter will
> always write to it after successful JWT validation. In other words, the
> authn filter will always change the dynamic metadata affecting the routing
> decision. Also if we do not clear cache here, where should we do it?
>
ok ack. Since we generating the pseudo header inside the jwt filter, I will
need to clear cache.

In the future. I think a clearer way though is for envoy to be able to just
use claims directly from the header to decide route. Jwt filter will come
in and can reject the request if validation fails.
In absence of this support, we need the flush.

We can add a flag to make this configurable so that users can opt-out (do

lambdai:
@kyessenov @mandarjog We and user must understand the filters are cooperative mode. The StreamInfo::route() is volatile. That says,
the beginning filters sees the route X by StreamInfo::route() , made the decision, then the latter filters updated to route Y. 

Every http filter must accept it. Istio should not promise more than it.

On the other hands, we might want to keep compatibility with old istio behavior so adding a flag for users who wants the old istio version if JWT based routing is not adopted.

bianpengyuan (approved):

PR #3493, opened 2021-09-14
Update Envoy SHA to 09-13.

mathetake:
Thanks to https://github.com/envoyproxy/envoy/pull/17357, registerPrometheusNamespace has been deleted and we no longer need to statically register prometheus stat namespaces. Not only that, this also resolves https://github.com/istio/istio/issues/27635 since the change in Envoy is handling any stat namespaces transparently to users and therefore users are free to determine the namespace they want to use in their own metrics exposed by Wasm plugins.

This completes the refactoring for removing all the Istio specific Wasm code base in Envoy (i.e. C++ code under src/**) combined with https://github.com/istio/proxy/pull/3413 https://github.com/istio/proxy/pull/3404 https://github.com/envoyproxy/envoy/pull/17268


mathetake:
Looks like stackdriver is implicitly relying on the past Envoy's behavior of prefixing all the metrics with "envoy_" .. need to make change to the plugin so that is explicitly prefix it.

extensions/stackdriver/common/metrics.cc
```diff
@@ -34,7 +34,7 @@ uint32_t newExportCallMetric(const std::string& type, bool success) {
   // base VM and thread local VM would cause host side thread local VM root
   // context missing metric definition. This is not going to be a problem with
   // real Wasm VM due to memory isolation.
-  Metric export_call(MetricType::Counter, "export_call",
+  Metric export_call(MetricType::Counter, "envoy_export_call",
                      {MetricTag{"wasm_filter", MetricTag::TagType::String},
                       MetricTag{"type", MetricTag::TagType::String},
                       MetricTag{"success", MetricTag::TagType::Bool}});
```
mathetake, line 40: @bianpengyuan is this actually used?

bianpengyuan: Yeah we occasionally use this to debug that stackdriver export call is made or not.

bianpengyuan (approved):

bianpengyuan:
@mathetake actually, curious how do we generate istio stats with `istio_` prefix now? I'd expect that we will have some change in the stats filter but looks like there is not.

mathetake:
The "istio" prefix has already been added at extensions side:
https://github.com/istio/proxy/blob/e02cbbe2e3f390afd9711c5d5e9f4bb5462059ea/extensions/stats/plugin.h#L55
so we don't need to modify that. Previously what we have done so far is that using `registerPrometheusNamespace` in the Envoy codebase to _register_ "istio" __statically_, and make Envoy opt-out "istio_**" metrics in the prometheus endpoint, meaning Envoy doesn't append "envoy_" prefix to these "istio" prefixed metrics when they are exposed in /stats/prometheus.

The change in https://github.com/envoyproxy/envoy/pull/17357 removed the need for registerPrometheusNamespace and made the prometheus stat endpoint (/stats/prometheus) **expose all the Wasm-program-defined metrics as-is**. That means, not only "istio_**" metrics but also any other user-defined custom metrics are exposed in the /stats/prometheus without being modified (which is great for all the Wasm extension users!).

bianpengyuan:
I see, thanks for detailed explanation! @mathetake

PR #3501, opened 2021-09-17
Allow skipping reporting of individual prometheus metrics

douglas-reid:
Adds the ability to remove reporting of an individual stat in the `stats` filter configuration.

This PR is meant to allow `stats` filter configuration to match the full capabilities exposed in the experimental Telemetry API. In particular, it allows matching of `Telemetry.metrics.overrides.disabled`.

It assumes the `MetricsConfig` will include a `name` field, as disabling all metrics can be achieved in the control plane by simply not sending any filter config.

### DIFF
--- a/extensions/stats/plugin.cc
+++ b/extensions/stats/plugin.cc
@@ -356,7 +356,6 @@ bool PluginRootContext::initializeDimensions(const json& j) {
         std::sort(tags.begin(), tags.end());
 
         auto name = JsonGetField<std::string>(metric, "name").value_or("");
-        // for (const auto& factory_it : factories) {
         for (auto factory_it = factories.begin();
              factory_it != factories.end();) { /*do not advance iterator here*/
           if (!name.empty() && name != factory_it->first) {
--- a/extensions/stats/plugin.cc
+++ b/extensions/stats/plugin.cc
@@ -15,6 +15,8 @@
 
 #include "extensions/stats/plugin.h"
 
+#include <iterator>
+
 #include "absl/strings/ascii.h"
 #include "absl/time/time.h"
 #include "extensions/common/util.h"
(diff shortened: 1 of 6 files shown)


testdata/stats/client_config_customized.yaml.tmpl
```diff
@@ -3,10 +3,15 @@ definitions:
...
   value: "10"
   type: COUNTER
+- name: request_duration_milliseconds
+  value: "100"
+  type: COUNTER
 - name: custom
   value: "1"
   type: COUNTER
 metrics:
+  - name: request_duration_milliseconds
```
howardjohn, line 13: I probably just don't understand the test but why does this show up on L6? It won't show up in /stats/promtheus I assume?

douglas-reid: good catch: leftover from early testing iterations. removed.

extensions/stats/plugin.cc
```diff
@@ -358,6 +358,14 @@ bool PluginRootContext::initializeDimensions(const json& j) {
           if (!name.empty() && name != factory_it.first) {
             continue;
           }
+
+          bool skip_reporting =
+              JsonGetField<bool>(metric, "skip_reporting").value_or(false);
+          if (skip_reporting) {
+            factories.erase(factory_it.first);
```
mandarjog, line 365: isn't it better to do this in line  293 instead? That is when the factories are created from `defaultMetrics()`

douglas-reid: That would require moving the parsing of the `metrics` config before L293. But I'm open to better ways (including moving the config around to different spots).

bianpengyuan: Probably not, the current logic also accomodates the case which defines a new metric at higher level (e.g. mesh wide) and skip reporting that at a lower level (e.g. per workload). Also keeping the logic for per metric configuration processing at one place seems cleaner to me.

bianpengyuan, line 365: This is likely problematic, Wasm test failed because of invalid memory access. Better record the factory to be removed and remove it after the iteration.

mandarjog: Actually modifying inside iteration is what I was looking at when I thought we could do it at factory construction time. 
Are you saying that it is not possible because overrides will get lost?
I don’t see things added into factories anywhere else.

bianpengyuan: New factory might be added at line 315.

lambdai:
```
        for (auto factory_it =factories.begin(); factory_it != factories.end(); /*do not advance iterator here */) {

          if (skip_reporting) {
            // erase returns the increased iterator
            factory_it = factories.erase(factory_it /* erase iterator to avoid another lookup */);
            continue;
          }
       }
```
Most importantly, we need to advance iterator at other continue or end of the loop.

douglas-reid: ACK. I was planning on updating once i'd run the full test suite. I've adopted @lambdai 's suggested changes here.

douglas-reid:
Please take another look. I've updated `skip_reporting` to `drop`, as I think that's clearer and more accurate. And I've adopted the suggested loop changes.

extensions/stats/plugin.cc
```diff
@@ -354,11 +356,21 @@ bool PluginRootContext::initializeDimensions(const json& j) {
         std::sort(tags.begin(), tags.end());
 
         auto name = JsonGetField<std::string>(metric, "name").value_or("");
-        for (const auto& factory_it : factories) {
-          if (!name.empty() && name != factory_it.first) {
+        // for (const auto& factory_it : factories) {
```
bianpengyuan, line 359: nit: clean this up.

PR #3505, opened 2021-09-22
Add support for configurability of stackdriver metrics

douglas-reid:
This PR is an attempt to bring some configurability to the existing Stackdriver extension for monitoring. It is intended to bring Stackdriver to _near_ (ex: Stackdriver does not support adding new tags) feature-parity with the Telemetry API.

This adds two basic capabilities:
1. Dropping metrics
2. Overriding tag expressions

---
Work still left:

- [ ] Validate tags exist and are well-known before handling
- [ ] Clean-up comments, etc.

### DIFF
--- a/extensions/stackdriver/metric/record.cc
+++ b/extensions/stackdriver/metric/record.cc
@@ -244,29 +244,70 @@ void addHttpSpecificTags(const ::Wasm::Common::RequestInfo& request_info,
                        std::to_string(response_code));
 }
 
+TagKeyValueList getMetricTagMap(
+    TagKeyValueList input_map,
+    std::vector<std::pair<std::string, std::string>> overrides) {
+  std::cout << "overrides: " << overrides.size() << "\n";
+
+  if (overrides.size() == 0) {
+    return input_map;
+  }
+
+  TagKeyValueList out;
+  for (auto in : input_map) {
+    auto it = std::find_if(overrides.begin(), overrides.end(),
+                           [&in](const auto& override) {
+                             return override.first == in.first.name();
+                           });
+    if (it != overrides.end()) {
+      auto tag_key = opencensus::tags::TagKey::Register(it->first);
+      out.emplace_back(tag_key, it->second);
+    } else {
+      out.emplace_back(in.first, in.second);
+    }
+  }
+
+  return out;
+}
+
 }  // namespace
 
 void record(bool is_outbound, const ::Wasm::Common::FlatNode& local_node_info,
             const ::Wasm::Common::FlatNode& peer_node_info,
             const ::Wasm::Common::RequestInfo& request_info,
-            bool record_http_size_metrics) {
+            bool record_http_size_metrics,
+            std::unordered_map<std::string,
+                               std::vector<std::pair<std::string, std::string>>>
+                overrides) {
   double latency_ms = request_info.duration /* in nanoseconds */ / 1000000.0;
   if (is_outbound) {
     TagKeyValueList tagMap =
         getOutboundTagMap(local_node_info, peer_node_info, request_info);
     addHttpSpecificTags(request_info, tagMap);
+
     if (record_http_size_metrics) {
       opencensus::stats::Record(
-          {{clientRequestCountMeasure(), 1},
-           {clientRoundtripLatenciesMeasure(), latency_ms},
-           {clientRequestBytesMeasure(), request_info.request_size},
-           {clientResponseBytesMeasure(), request_info.response_size}},
-          tagMap);
+          {{clientRequestCountMeasure(), 1}},
+          getMetricTagMap(tagMap, overrides[Common::kClientRequestCountView]));
+      opencensus::stats::Record(
+          {{clientRoundtripLatenciesMeasure(), latency_ms}},
+          getMetricTagMap(tagMap,
+                          overrides[Common::kClientRoundtripLatenciesView]));
+      opencensus::stats::Record(
+          {{clientRequestBytesMeasure(), request_info.request_size}},
+          getMetricTagMap(tagMap, overrides[Common::kClientRequestBytesView]));
+      opencensus::stats::Record(
+          {{clientResponseBytesMeasure(), request_info.response_size}},
+          getMetricTagMap(tagMap, overrides[Common::kClientResponseBytesView]));
+
     } else {
       opencensus::stats::Record(
-          {{clientRequestCountMeasure(), 1},
-           {clientRoundtripLatenciesMeasure(), latency_ms}},
-          tagMap);
+          {{clientRequestCountMeasure(), 1}},
+          getMetricTagMap(tagMap, overrides[Common::kClientRequestCountView]));
+      opencensus::stats::Record(
+          {{clientRoundtripLatenciesMeasure(), latency_ms}},
+          getMetricTagMap(tagMap,
+                          overrides[Common::kClientRoundtripLatenciesView]));
     }
 
     return;
@@ -277,33 +318,56 @@ void record(bool is_outbound, const ::Wasm::Common::FlatNode& local_node_info,
   addHttpSpecificTags(request_info, tagMap);
   if (record_http_size_metrics) {
     opencensus::stats::Record(
-        {{serverRequestCountMeasure(), 1},
-         {serverResponseLatenciesMeasure(), latency_ms},
-         {serverRequestBytesMeasure(), request_info.request_size},
-         {serverResponseBytesMeasure(), request_info.response_size}},
-        tagMap);
+        {{serverRequestCountMeasure(), 1}},
+        getMetricTagMap(tagMap, overrides[Common::kServerRequestCountView]));
+    opencensus::stats::Record(
+        {{serverResponseLatenciesMeasure(), latency_ms}},
+        getMetricTagMap(tagMap,
+                        overrides[Common::kServerResponseLatenciesView]));
+    opencensus::stats::Record(
+        {{serverRequestBytesMeasure(), request_info.request_size}},
+        getMetricTagMap(tagMap, overrides[Common::kServerRequestBytesView]));
+    opencensus::stats::Record(
+        {{serverResponseBytesMeasure(), request_info.response_size}},
+        getMetricTagMap(tagMap, overrides[Common::kServerResponseBytesView]));
   } else {
-    opencensus::stats::Record({{serverRequestCountMeasure(), 1},
-                               {serverResponseLatenciesMeasure(), latency_ms}},
-                              tagMap);
+    opencensus::stats::Record(
+        {{serverRequestCountMeasure(), 1}},
+        getMetricTagMap(tagMap, overrides[Common::kServerRequestCountView]));
+    opencensus::stats::Record(
+        {{serverResponseLatenciesMeasure(), latency_ms}},
+        getMetricTagMap(tagMap,
+                        overrides[Common::kServerResponseLatenciesView]));
   }
 }
 
-void recordTCP(bool is_outbound,
-               const ::Wasm::Common::FlatNode& local_node_info,
-               const ::Wasm::Common::FlatNode& peer_node_info,
-               const ::Wasm::Common::RequestInfo& request_info) {
+void recordTCP(
+    bool is_outbound, const ::Wasm::Common::FlatNode& local_node_info,
+    const ::Wasm::Common::FlatNode& peer_node_info,
+    const ::Wasm::Common::RequestInfo& request_info,
+    std::unordered_map<std::string,
+                       std::vector<std::pair<std::string, std::string>>>
+        overrides) {
   if (is_outbound) {
     TagKeyValueList tagMap =
         getOutboundTagMap(local_node_info, peer_node_info, request_info);
     opencensus::stats::Record(
         {{clientConnectionsOpenCountMeasure(),
-          request_info.tcp_connections_opened},
-         {clientConnectionsCloseCountMeasure(),
-          request_info.tcp_connections_closed},
-         {clientReceivedBytesCountMeasure(), request_info.tcp_received_bytes},
-         {clientSentBytesCountMeasure(), request_info.tcp_sent_bytes}},
-        tagMap);
+          request_info.tcp_connections_opened}},
+        getMetricTagMap(tagMap,
+                        overrides[Common::kClientConnectionsOpenCountView]));
+    opencensus::stats::Record(
+        {{clientConnectionsCloseCountMeasure(),
+          request_info.tcp_connections_closed}},
+        getMetricTagMap(tagMap,
+                        overrides[Common::kClientConnectionsCloseCountView]));
+    opencensus::stats::Record(
+        {{clientReceivedBytesCountMeasure(), request_info.tcp_received_bytes}},
+        getMetricTagMap(tagMap,
+                        overrides[Common::kClientReceivedBytesCountView]));
+    opencensus::stats::Record(
+        {{clientSentBytesCountMeasure(), request_info.tcp_sent_bytes}},
+        getMetricTagMap(tagMap, overrides[Common::kClientSentBytesCountView]));
 
     return;
   }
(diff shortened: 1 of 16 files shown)


extensions/stackdriver/stackdriver.cc
```diff
@@ -432,6 +456,20 @@ bool StackdriverRootContext::configure(size_t configuration_size) {
     tcp_log_entry_timeout_ = getTcpLogEntryTimeoutNanoseconds();
   }
 
+  // Extract metric tags expressions
+  cleanupMetricsExpressions();
+  for (const auto& override : getMetricsTagOverrides(config_)) {
+    uint32_t token;
+    auto metric = std::get<0>(override);
```
bianpengyuan, line 463: Why do we have this intermediate tuple vector data structure, instead of directly constructing the expression and push it to `metrics_expressions_`? We can make `getMetricsTagOverrides()` take `metrics_expressions_` as input and build everything there.

extensions/stackdriver/stackdriver.cc
```diff
@@ -186,6 +186,30 @@ absl::Duration getMetricExpiryDuration(
          absl::Nanoseconds(duration.nanos());
 }
 
+std::vector<std::string_view> getDroppedMetrics(
+    const stackdriver::config::v1alpha1::PluginConfig& config) {
+  std::vector<std::string_view> dropped_metrics;
```
bianpengyuan, line 191: Should we make this a map?

extensions/stackdriver/metric/registry.cc
```diff
@@ -201,66 +201,86 @@ StackdriverOptions getStackdriverOptions(
...
-            .set_expiry_duration(expiry_duration)            \
-            .set_aggregation(Aggregation::Count()) ADD_TAGS; \
-    View view(view_descriptor);                              \
-    view_descriptor.RegisterForExport();                     \
+#define REGISTER_COUNT_VIEW(_v)                                            \
+  void register##_v##View(absl::Duration expiry_duration,                  \
+                          std::vector<std::string_view> dropped_metrics) { \
+    auto iter = std::find(dropped_metrics.begin(), dropped_metrics.end(),  \
+                          k##_v##View);                                    \
+    if (iter != dropped_metrics.end()) return;                             \
```
bianpengyuan, line 209: nit: wrap return with `{}`

extensions/stackdriver/metric/registry.cc
```diff
@@ -201,66 +201,86 @@ StackdriverOptions getStackdriverOptions(
...
+    const ViewDescriptor view_descriptor =                                 \
+        ViewDescriptor()                                                   \
+            .set_name(k##_v##View)                                         \
+            .set_measure(k##_v##Measure)                                   \
+            .set_expiry_duration(expiry_duration)                          \
+            .set_aggregation(Aggregation::Count()) ADD_TAGS;               \
+    View view(view_descriptor);                                            \
+    view_descriptor.RegisterForExport();                                   \
```
bianpengyuan, line 217: I don't think this will work if user configure dropping metric in the middle of proxy lifetime. Opencensus maintains everything (measurement, views) with a singleton, skip registering won't revert previous registered view. User will need to restart proxy to make the config effective. Not sure if Opencensus provides a way to deregister or not (I don't think so..), so probably need to document this clearly.

extensions/stackdriver/stackdriver.cc
```diff
@@ -515,9 +553,15 @@ void StackdriverRootContext::record() {
   ::Wasm::Common::RequestInfo request_info;
   ::Wasm::Common::populateHTTPRequestInfo(outbound, useHostHeaderFallback(),
                                           &request_info);
+  std::unordered_map<std::string,
```
bianpengyuan, line 556: Can we add a typedef for this data structure?

extensions/stackdriver/stackdriver.cc
```diff
@@ -732,13 +780,38 @@ void StackdriverRootContext::evaluateExpressions(
...
+    std::cout << "adding expression: " << expression.tag
+              << ", for metric: " << expression.metric << "\n";
+    std::string value;
+    if (!evaluateExpression(expression.token, &value)) {
+      LOG_WARN(absl::StrCat("Could not evaluate expression: ",
+                            expression.expression));
+      continue;
+    }
+    overrides[expression.metric].emplace_back(
+        std::make_pair(expression.tag, value));
```
bianpengyuan, line 797: I don't think you need to make_pair? Should work if you directly `emplace_back(expression.tag, value)`, so that we can avoid a copy.

bianpengyuan (commented):
looks good!

extensions/stackdriver/stackdriver.cc
douglas-reid: yeah, some of this is an artifact of just trying to piece together a set of working code. there is a definite need for cleanup and streamlining, but I wanted to make sure the broad strokes looked ok first.

extensions/stackdriver/metric/registry.cc
douglas-reid: This is a good point. I'll look into deregistration.

extensions/stackdriver/stackdriver.cc
douglas-reid: i don't think it needs to be a map. there will never be many in the list in any case.

extensions/stackdriver/metric/registry.cc
douglas-reid: done.

extensions/stackdriver/metric/registry.cc
douglas-reid: added deregistration support.

extensions/stackdriver/stackdriver.cc
douglas-reid: done.

extensions/stackdriver/stackdriver.cc
douglas-reid: This should be in fewer places now. Please let me know if you think we should still typedef it.

extensions/stackdriver/metric/registry.cc
```diff
@@ -356,22 +390,28 @@ void registerViews(absl::Duration expiry_duration) {
...
+  registerServerConnectionsCloseCountView(expiry_duration, dropped_metrics);
+  registerServerReceivedBytesCountView(expiry_duration, dropped_metrics);
+  registerServerSentBytesCountView(expiry_duration, dropped_metrics);
+  registerClientConnectionsOpenCountView(expiry_duration, dropped_metrics);
+  registerClientConnectionsCloseCountView(expiry_duration, dropped_metrics);
+  registerClientReceivedBytesCountView(expiry_duration, dropped_metrics);
+  registerClientSentBytesCountView(expiry_duration, dropped_metrics);
+}
+
+void dropViews(std::vector<std::string> dropped_metrics) {
```
bianpengyuan, line 411: nit: `const std::vector<std::string>&`

extensions/stackdriver/metric/registry.cc
```diff
@@ -356,22 +390,28 @@ void registerViews(absl::Duration expiry_duration) {
...
+  for (auto metric : dropped_metrics) {
```
bianpengyuan, line 412: nit: `const auto&`

extensions/stackdriver/metric/registry.h
```diff
@@ -39,7 +39,10 @@ opencensus::exporters::stats::StackdriverOptions getStackdriverOptions(
         stub_option);
 
 // registers Opencensus views
-void registerViews(absl::Duration);
+void registerViews(absl::Duration, std::vector<std::string>);
```
bianpengyuan, line 42: nit: `const std::vector<std::string>`

extensions/stackdriver/metric/record.h
```diff
@@ -27,14 +27,18 @@ namespace Metric {
 void record(bool is_outbound, const ::Wasm::Common::FlatNode& local_node_info,
             const ::Wasm::Common::FlatNode& peer_node_info,
             const ::Wasm::Common::RequestInfo& request_info,
-            bool record_http_size_metrics);
+            bool record_http_size_metrics,
+            std::unordered_map<
```
bianpengyuan, line 31: nit: make this const reference, and same to the function below.

extensions/stackdriver/metric/record.cc
```diff
@@ -244,29 +244,82 @@ void addHttpSpecificTags(const ::Wasm::Common::RequestInfo& request_info,
...
     addHttpSpecificTags(request_info, tagMap);
+
     if (record_http_size_metrics) {
       opencensus::stats::Record(
-          {{clientRequestCountMeasure(), 1},
-           {clientRoundtripLatenciesMeasure(), latency_ms},
-           {clientRequestBytesMeasure(), request_info.request_size},
-           {clientResponseBytesMeasure(), request_info.response_size}},
-          tagMap);
+          {{clientRequestCountMeasure(), 1}},
```
bianpengyuan, line 302: Record is expensive. It will make a copy of the tagmap: https://github.com/census-instrumentation/opencensus-cpp/blob/afe0460f92fb78e6d6cf8c8a30ced9bc5e2e57d3/opencensus/stats/recording.h#L44.

We should only call with separated record function if the metric has customization.

bianpengyuan (commented):
lgtm

extensions/stackdriver/metric/record.cc
douglas-reid: That's gonna be tough, given the structure of the code. I'll see what can be done here.

douglas-reid: I'm at a loss, given that `Record()` only takes initializer lists.  At best, if there are no customizations, we could use the old code path, but pay the penalty if there are _any_ customizations. Thoughts?

extensions/stackdriver/common/constants.h
```diff
@@ -69,6 +69,36 @@ constexpr char kClientConnectionsCloseCountView[] =
 constexpr char kClientReceivedBytesCountView[] = "client/received_bytes_count";
 constexpr char kClientSentBytesCountView[] = "client/sent_bytes_count";
 
+const std::vector<std::string_view> kDefinedLabels = {
```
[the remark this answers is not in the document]
lambdai: constexpr array can be used here
Ref: https://github.com/envoyproxy/envoy/blob/main/source/common/stream_info/utility.h#L49

extensions/stackdriver/metric/record.cc
bianpengyuan: Yeah that sgtm.

extensions/stackdriver/metric/record.cc
```diff
@@ -244,29 +244,82 @@ void addHttpSpecificTags(const ::Wasm::Common::RequestInfo& request_info,
                        std::to_string(response_code));
 }
 
+TagKeyValueList getMetricTagMap(
+    TagKeyValueList input_map,
+    std::vector<std::pair<std::string, std::string>> overrides) {
```
lambdai, line 249: I feel `const&` is good for the two args input_map and overrides

douglas-reid: done.

extensions/stackdriver/metric/record.cc
```diff
@@ -244,29 +244,82 @@ void addHttpSpecificTags(const ::Wasm::Common::RequestInfo& request_info,
...
+  if (overrides.size() == 0) {
+    return input_map;
+  }
+
+  TagKeyValueList out;
+  for (auto in : input_map) {
```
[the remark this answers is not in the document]
douglas-reid: done.

extensions/stackdriver/metric/record.cc
```diff
@@ -244,29 +244,82 @@ void addHttpSpecificTags(const ::Wasm::Common::RequestInfo& request_info,
                        std::to_string(response_code));
 }
 
+TagKeyValueList getMetricTagMap(
+    TagKeyValueList input_map,
+    std::vector<std::pair<std::string, std::string>> overrides) {
+  if (overrides.size() == 0) {
```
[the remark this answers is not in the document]
douglas-reid: done.

extensions/stackdriver/common/constants.h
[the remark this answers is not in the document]
douglas-reid: done.

extensions/stackdriver/metric/record.cc
douglas-reid: updated. please review to make sure i haven't borked anything in the rewrite.

extensions/stackdriver/metric/record.h
douglas-reid: done.

extensions/stackdriver/metric/registry.cc
douglas-reid: done.

extensions/stackdriver/metric/registry.h
douglas-reid: done.

extensions/stackdriver/metric/registry.cc
douglas-reid: done.

douglas-reid:
Thanks for the good reviews. I think this is getting close now. PTAL.

extensions/stackdriver/metric/record.cc
```diff
@@ -245,166 +245,309 @@ void addHttpSpecificTags(const ::Wasm::Common::RequestInfo& request_info,
...
       [](const auto& override) { return override.first == "api_name"; });
-  if (it != overrides.end()) {
+  if (it != tag_overrides.end()) {
     out.emplace_back(apiNameKey(), it->second);
   }
 
   return out;
 }
 
+bool hasOverridesMatching(const override_map& overrides, std::string metric) {
```
bianpengyuan, line 285: `const std::string&`

extensions/stackdriver/metric/record.cc
```diff
@@ -245,166 +245,309 @@ void addHttpSpecificTags(const ::Wasm::Common::RequestInfo& request_info,
...
-  for (auto in : input_map) {
-    auto it = std::find_if(overrides.begin(), overrides.end(),
-                           [&in](const auto& override) {
-                             return override.first == in.first.name();
-                           });
-    if (it != overrides.end()) {
-      auto tag_key = opencensus::tags::TagKey::Register(it->first);
-      out.emplace_back(tag_key, it->second);
+  for (const auto& [tag_key, value_list] : input_map) {
+    auto name = tag_key.name();
```
bianpengyuan, line 256: const auto&

extensions/stackdriver/metric/record.cc
```diff
@@ -245,166 +245,309 @@ void addHttpSpecificTags(const ::Wasm::Common::RequestInfo& request_info,
...
+    auto it = std::find_if(
+        tag_overrides.begin(), tag_overrides.end(),
+        [&name](const auto& override) { return override.first == name; });
+    if (it != tag_overrides.end()) {
+      auto key = opencensus::tags::TagKey::Register(it->first);
```
bianpengyuan, line 261: This feels awkward.. Can we build a search map for tagKey at registration time?

douglas-reid: opencensus already does that for us: https://github.com/census-instrumentation/opencensus-cpp/blob/master/opencensus/tags/internal/tag_key.cc#L58-L69

extensions/stackdriver/metric/record.cc
douglas-reid: done.

extensions/stackdriver/metric/record.cc
douglas-reid: done.

extensions/stackdriver/metric/record.cc
bianpengyuan: It looks very heavy. It has a lock acquisition, and this happen on every call. We should really avoid that.

douglas-reid: Ok. moved the call to configure().

douglas-reid:
PTAL.

bianpengyuan (approved):

PR #3514, opened 2021-09-30
Update Envoy SHA 0930.

bianpengyuan:


kyessenov (approved):
Did we update clang to 12 (https://github.com/envoyproxy/envoy/pull/18220/files)?

bianpengyuan:
No I don't think so. Thanks for reminder, will update that.

PR #3525, opened 2021-10-07
wasm: unambigious GetFromStringView()

lambdai:
Fix the sync failure https://github.com/istio/proxy/pull/3524

It seems flatbuffer lib introduce another GetStringView.
That one should not be used. 
I renamed ours because flatbuffers one is introduced by Argument-dependent lookup

### DIFF
--- a/extensions/common/context.h
+++ b/extensions/common/context.h
@@ -290,7 +290,8 @@ bool populateGRPCInfo(RequestInfo* request_info);
 bool getAuditPolicy();
 
 // Returns a string view stored in a flatbuffers string.
-static inline std::string_view GetStringView(const flatbuffers::String* str) {
+static inline std::string_view GetFromFbStringView(
+    const flatbuffers::String* str) {
   return str ? std::string_view(str->c_str(), str->size()) : std::string_view();
 }
 
--- a/extensions/stats/plugin.cc
+++ b/extensions/stats/plugin.cc
@@ -44,7 +44,7 @@ namespace Stats {
 const uint32_t kDefaultTCPReportDurationMilliseconds = 15000;  // 15s
 
 using ::nlohmann::json;
-using ::Wasm::Common::GetStringView;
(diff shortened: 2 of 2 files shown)


lambdai:
For reference: https://github.com/google/flatbuffers/blob/master/include/flatbuffers/flatbuffers.h#L732-L738

```

extensions/stats/plugin.cc:95:48: error: call to 'GetStringView' is ambiguous
    instance[destination_workload_namespace] = GetStringView(node.namespace_());
                                               ^~~~~~~~~~~~~
bazel-out/k8-fastbuild/bin/external/com_github_google_flatbuffers/_virtual_includes/runtime_cc/flatbuffers/flatbuffers.h:660:40: note: candidate function
static inline flatbuffers::string_view GetStringView(const String *str) {
                                       ^
./extensions/common/context.h:293:32: note: candidate function
static inline std::string_view GetStringView(const flatbuffers::String* str) {
```

lambdai:
> This looks like C++20 was enabled? I think flatbuffers guards its string_view aliasing to std with some macro.

I don't see C++20 related PR in the past 2 days. There is only a FlatBuffer version bump to 2.0.0. The issue you proposed enabling partial c++ 20 features should not break the build :)

kyessenov:
Ah, it looks like flatbuffers just defined this helper. Why not delete the one here to avoid double definition?

lambdai:
> Ah, it looks like flatbuffers just defined this helper. Why not delete the one here to avoid double definition?

Because the flatbuffers helper is returning flatbuffers::string_view and it's guarded by macro. IMHO the flatbuffers is not robust enough. 

I am happy to change in a follow up version.

kyessenov:
Sure, let's merge this.

lambdai:
Thanks.

Do you know how to interrupt the ongoing test-tsan_proxy? I remember `/test` doesn't work

lambdai:
rerun all test because the stuck one need another 45 minutes to timeout

PR #3548, opened 2021-10-26
bazel: add extensions config

lambdai:
Disable tcp_stats and add mysql_proxy 

### DIFF
--- a/bazel/extension_config/extensions_build_config.bzl
+++ b/bazel/extension_config/extensions_build_config.bzl
@@ -371,6 +371,6 @@ EXTENSIONS = dict([(k,v) for k,v in ENVOY_EXTENSIONS.items() if not k in ISTIO_D
 
 # These can be changed to ["//visibility:public"], for  downstream builds which
 # need to directly reference Envoy extensions.
-EXTENSION_CONFIG_VISIBILITY = ["//:extension_config"]
-EXTENSION_PACKAGE_VISIBILITY = ["//:extension_library"]
-CONTRIB_EXTENSION_PACKAGE_VISIBILITY = ["//:contrib_library"]
+EXTENSION_CONFIG_VISIBILITY = ["//visibility:public"]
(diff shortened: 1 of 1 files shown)


anuragagarwal561994:
`envoy.transport_sockets.tcp_stats` how do we enable this if one wants to?

zirain:
it's enabled.
https://github.com/istio/proxy/blob/ff99990ec25610e697462d78c741e624ca2640dc/bazel/extension_config/extensions_build_config.bzl#L256

anuragagarwal561994:
@zirain it is disabled https://github.com/istio/proxy/blob/ff99990ec25610e697462d78c741e624ca2640dc/bazel/extension_config/extensions_build_config.bzl#L485

zirain:
so, you cannot use it without a custom proxy build.

anuragagarwal561994:
As per the disucssion I followed, it was done because some of the kernel versions hindered with this filter and as per the above conversation a different binary was may be expected to be released where these would be applicable.

Are the above conditions still valid based on the latest linux versions and the base image being used and is there any plan to support a custom build with this extension enabled?

zirain:
AFAIK, there isn't a plan about this.
need context about why should istio do that? is it a common use case?

anuragagarwal561994:
@zirain so the idea is we wanted to track metrics for how much rtt time is spent between envoy to upstream and envoy to downstream.

To do so we wanted to expose these metrics, it is available in the envoyproxy but then because of the custom build of enovy in istio this is currently not possible.

I tried using envoyproxy without the custom version but then istio specific customisations like `metadata-exchange` extension is missing.

I tried building do a custom build of envoy, but now to build it I am using

```
docker run -it --rm -w /work -v $PWD:/work gcr.io/istio-testing/build-tools-proxy:master-latest bash
> make build_envoy
```

But the base image is using ubuntu bionic which doesn't support the required kernel for correct tcp.h and hence the build is failing. I think upgrading the base image should be able to solve the issue, so trying to now figure this out. I wanted to know if there is an easier way.

PR #3557, opened 2021-10-29
Update to envoyproxy v1.20 and update extensions config

ericvn:


ericvn:
Per Slack, I no longer revert the extensions config, but updated the file:
```
I think the correct thing is to update the proxy's extension_build_config.bzl following the above file in envoy 1.20
```

bazel/extension_config/extensions_build_config.bzl
```diff
@@ -356,17 +345,17 @@ ENVOY_CONTRIB_EXTENSIONS = {
 }
 
 
-ISTIO_DISABLED_EXTENSIONS = [ 
+ISTIO_DISABLED_EXTENSIONS = [
     # ISTIO disable tcp_stats by default because this plugin must be built and running on kernel >= 4.6
     "envoy.transport_sockets.tcp_stats",
```
ericvn, line 350: Maybe this can be removed since it was removed above. Not sure if leaving it here will cause a build issue.

lambdai: It's ok to keep or delete.

lambdai (approved):
LGTM. Thanks!

dgn (approved):

PR #3560, opened 2021-11-01
contrib: enable all contrib extensions

lambdai:
Starting from envoy 1.20 the contrib extensions are built via allow list.

In istio 1.12 we build all known contrib extensions as was in istio 1.11.


Kmoneal (approved):
LGTM

lambdai:
Thank you @Kmoneal

PR #3591, opened 2021-11-18
Add gen compilation database tool like envoy

kebe7jun:
This tool was added to make it easier for us to generate a compilation database for the project and to improve the efficiency of reading the code, this tool comes from envoy: https://github.com/envoyproxy/envoy/blob/main/tools/gen_compilation_database.py

Run `tools/gen_compilation_database.py --vscode` in vscode can generate compilation database.

### DIFF
--- a/tools/gen_compilation_database.py
+++ b/tools/gen_compilation_database.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import os
+import shlex
+import subprocess
+from pathlib import Path
+
+
+# This method is equivalent to https://github.com/grailbio/bazel-compilation-database/blob/master/generate.sh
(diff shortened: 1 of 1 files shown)


bianpengyuan (commented):
Does not seem like it has any dep on this repo so no need to duplicate here? Feel better to just point to the Envoy code base, download and run that.

kebe7jun:
> Does not seem like it has any dep on this repo so no need to duplicate here? Feel better to just point to the Envoy code base, download and run that.

The 'bazel_targets' is not different with envoy, but we can use arguments override it.

bianpengyuan:
I see. Yeah override bazel_targets sound better. We can document something at top level readme, instead of duplicating a script here, which is unlikely going to be well maintained.

PR #3614, opened 2021-12-02
fix string includes for header

tbarrella:
**What this PR does / why we need it**: This was causing issues for me when including this header

### DIFF
--- a/extensions/stackdriver/common/constants.h
+++ b/extensions/stackdriver/common/constants.h
@@ -13,6 +13,10 @@
  * limitations under the License.
  */
 
+#include <string>
+
+#include "absl/strings/string_view.h"
+
 namespace Extensions {
 namespace Stackdriver {
 namespace Common {
@@ -69,7 +73,7 @@ constexpr char kClientConnectionsCloseCountView[] =
 constexpr char kClientReceivedBytesCountView[] = "client/received_bytes_count";
 constexpr char kClientSentBytesCountView[] = "client/sent_bytes_count";
 
-constexpr std::string_view kDefinedLabels[] = {
+constexpr absl::string_view kDefinedLabels[] = {
     "request_protocol",
     "service_authentication_policy",
     "mesh_uid",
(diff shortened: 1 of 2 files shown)


extensions/stackdriver/common/constants.h
```diff
@@ -13,6 +13,9 @@
  * limitations under the License.
  */
 
+#include <string>
+#include <string_view>
```
kyessenov, line 17: FYI Envoy standardized on absl::string_view which is not aliased to std::string_view but easily convertible.

tbarrella: I wouldn't mind replacing the `std::string_view` in this file

mandarjog:
Curious, what does this pr do?

tbarrella:
These includes should be in this header because it uses `std::string` and `std::string_view`. Without `<string_view>` I was getting an error from including this header in a new build target.

kyessenov (approved):
Can you confirm that the size of output Wasm files did not change? I think it's fine if you just touch stackdriver, but the common files will affect stats.wasm.

tbarrella:
Sorry, how do I check that? Theoretically, I don't think this could impact compilation of any currently existing dependency, because all targets that depend on this must already have these includes somewhere else in the build tree, and they shouldn't be duplicated

kyessenov:
The build script publishes the wasm files somewhere. Agree this is not necessary in this PR but if you touch extensions/common, we'd need to do that.

tbarrella:
Ok, I won't touch that for now; I just did here because I was working in `stackdriver/`

tbarrella:
```
extensions/stackdriver/stackdriver.cc:172:8: error: no matching function for call to 'getValue'
  if (!getValue({"node", "metadata", "PLATFORM_METADATA", kGCPProjectNumberKey},
       ^~~~~~~~
external/proxy_wasm_cpp_sdk/proxy_wasm_api.h:596:13: note: candidate function template not viable: no known conversion from 'const absl::string_view' to 'std::string_view' for 1st argument
inline bool getValue(const std::initializer_list<std::string_view> &parts, T *out) {
```

interesting... seems I should keep these as `std::string_view`

PR #3630, opened 2021-12-14
stackdriver(logging): add support for cel logging filters

douglas-reid:
This PR adds support to the Stackdriver extension for CEL filtering in order to support Telemetry API additions. In https://github.com/istio/api/pull/2110, the ability to filter access logs via a CEL expression was added to the Telemetry API.

### DIFF
--- a/extensions/stackdriver/stackdriver.cc
+++ b/extensions/stackdriver/stackdriver.cc
@@ -368,6 +368,18 @@ bool StackdriverRootContext::onConfigure(size_t size) {
   return true;
 }
 
+bool StackdriverRootContext::initializeLogFilter() {
+  uint32_t token = 0;
+  if (createExpression(config_.access_logging_filter_expression(), &token) !=
+      WasmResult::Ok) {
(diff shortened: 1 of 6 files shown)


extensions/stackdriver/stackdriver.cc
```diff
@@ -368,6 +368,18 @@ bool StackdriverRootContext::onConfigure(size_t size) {
   return true;
 }
 
+bool StackdriverRootContext::initializeLogFilter() {
+  uint32_t token = 0;
+  if (createExpression(config_.access_logging_filter_expression(), &token) !=
```
bianpengyuan, line 373: Do we need to check if config_ has filter expression first?

bianpengyuan (commented):
Looks good to me!

extensions/stackdriver/stackdriver.cc
douglas-reid: added check.

bianpengyuan (approved):

PR #3640, opened 2021-12-25
add envoy.filters.thrift.header_to_metadata

zirain:

sync extensions_build_config.bzl
prevent issues like istio/istio#36620

### DIFF
--- a/bazel/extension_config/extensions_build_config.bzl
+++ b/bazel/extension_config/extensions_build_config.bzl
@@ -175,6 +175,7 @@ ENVOY_EXTENSIONS = {
     #
 
     "envoy.filters.thrift.router":                      "//source/extensions/filters/network/thrift_proxy/router:config",
+    "envoy.filters.thrift.header_to_metadata":          "//source/extensions/filters/network/thrift_proxy/filters/header_to_metadata:config",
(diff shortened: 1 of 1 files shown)


bianpengyuan (commented):
I don't think we necessarily want to keep in sync with upstream filters. More filters means more possibility of vulnerabilities. Since thrift is not supported as first class protocol at Istio, I think it is fine to exclude this new filter. cc @lambdai @kyessenov

zirain:
> I don't think we necessarily want to keep in sync with upstream filters. More filters means more possibility of vulnerabilities. Since thrift is not supported as first class protocol at Istio, I think it is fine to exclude this new filter. cc @lambdai @kyessenov

Is there any rules about what kind of filters should be added or not ?

kyessenov:
@zirain The rule is that we have it must fulfill a supported feature in Istio. Using EnvoyFilters to invoke some upstream filter from Envot is not a supported feature. E.g. if you want first-class Thrift support, then it must be done as a first-class Pilot feature.

zirain:
> @zirain The rule is that it must fulfill a supported feature in Istio. Using EnvoyFilters to invoke some upstream filter from Envot is not a supported feature. E.g. if you want first-class Thrift support, then it must be done as a first-class Pilot feature.

Got it，thx！

PR #3666, opened 2022-01-21
extensions: sync full envoy extension list with envoy release 1.21

lambdai:
Will build below in proxy binary 
envoy.filters.http.stateful_session
envoy.filters.thrift.header_to_metadata
envoy.http.stateful_session.cookie

### DIFF
--- a/bazel/extension_config/extensions_build_config.bzl
+++ b/bazel/extension_config/extensions_build_config.bzl
@@ -107,6 +107,7 @@ ENVOY_EXTENSIONS = {
     "envoy.filters.http.set_metadata":                  "//source/extensions/filters/http/set_metadata:config",
     "envoy.filters.http.tap":                           "//source/extensions/filters/http/tap:config",
     "envoy.filters.http.wasm":                          "//source/extensions/filters/http/wasm:config",
+    "envoy.filters.http.stateful_session":              "//source/extensions/filters/http/stateful_session:config",
 
     #
     # Listener filters
@@ -175,6 +176,7 @@ ENVOY_EXTENSIONS = {
     #
 
     "envoy.filters.thrift.router":                      "//source/extensions/filters/network/thrift_proxy/router:config",
+    "envoy.filters.thrift.header_to_metadata":          "//source/extensions/filters/network/thrift_proxy/filters/header_to_metadata:config",
     "envoy.filters.thrift.ratelimit":                   "//source/extensions/filters/network/thrift_proxy/filters/ratelimit:config",
 
     #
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

bianpengyuan (commented):
Is the deal to always sync up with Envoy extension list? FWIW we turned down a sync PR for thrift.header_to_metadata at master https://github.com/istio/proxy/pull/3640 :). cc @kyessenov

howardjohn:
> Is the deal to always sync up with Envoy extension list? FWIW we turned down a sync PR for thrift.header_to_metadata at master #3640 :). cc @kyessenov

IMO the status quo is we have all upstream extensions. Unless there is a design doc suggesting otherwise we should maintain that

>Using EnvoyFilters to invoke some upstream filter from Envot is not a supported feature.

I don't fully agree with this. There are absolutely benefits to a minimal Envoy binary, but EnvoyFilter is a supported feature and its intended purpose _is_ to invoke Envoy Filters (hence the name). We should explore other options like contrib image or upstream Envoy, but I don't think we should freeze all extensions without broader agreement.

kyessenov:
The problem is that puts expectations for the support of unstable features in envoy on us. We are not willing to commit to that so as long as we need understanding from users that issues with `EnvoyFilter` (including CVEs) will not be addressed.

kyessenov:
@howardjohn I don't think we ever said EnvoyFilter is supported. It's a loop hole API that is unstable by design. If someone triggers CVEs with EnvoyFilters, that's is not our responsibility.

kyessenov:
Lastly, I want to iterate that _supported_ really means users do not have to worry about operations and upgradability of a feature. It does not mean there is some code that lets you do arbitrary things. Given our push to promote alpha to stable or get it out, I don't really see a future for EnvoyFilter. We can never say EnvoyFilter is safe to use and will not cause random outage, security holes, or difficult to diagnose stability problems.

howardjohn:
Envoy Filter has pretty defined guidelines on the risks in https://istio.io/latest/docs/reference/config/networking/envoy-filter/ (and if its not clear/scary enough we can make it scarier). That doesn't mean we should artificially hinder it. Things don't need to be safe to be accessible. For example, even rust has `unsafe` - there is a huge list of precautions to using it, but its an escape hatch if you need it, just like EnvoyFilter.

kyessenov:
@howardjohn Again, I have no problem enabling the kitchen sink in the OSS project. That's the difference between public projects and vendor products. Users who are concerned about the supply chain risks should not place too much trust in the OSS distribution.

howardjohn:
After some discussion with Kuat we came to the conclusion we should continue to do this (enabling extensions in upstream) but encourage pushing custom extensions to first class APIs

lambdai:
The future noise includes more report if the extra extensions are vulnerable. We probably won't upgrade them as eagerly as must-have extension.

If we concede that the extra extension are safe if the EnvoyFilter doesn't enable them, the risk is controlled by the EnvoyFilter user.

Plus the extension list can be redefined by individual proxy providers

stevenctl (approved):

bianpengyuan (approved):

PR #3720, opened 2022-02-19
Push Wasm extensions as OCI image at postsubmit.

bianpengyuan:
For wasm image pull integration test.


scripts/release-binary.sh
```diff
@@ -211,6 +211,14 @@ trap 'rm -rf ${TMP_WASM}' EXIT
 make build_wasm
 if [ -n "${DST}" ]; then
   for extension in "${extensions[@]}"; do
+    if [ "${PUSH_DOCKER_IMAGE}" -eq 1 ]; then
```
zirain, line 214: can we use this image when `telemetry.metadataExchange.wasmEnabled: true`

zirain: BTW, do we have any data about the performance between `wasm metadataExchange` and `null metadataExchange`?

bianpengyuan: We used to measure it a long time ago. It is roughly 10% ~ 20% more expensive both cpu and latency. Although I heard it has been improved, so a new measurement probably would be needed.

zirain: once https://perf.dashboard.istio.io/ is fixed, we can setup a measurement daily.

kyessenov (approved):

otterley:
@bianpengyuan Unfortunately this broke the build on the arm64 processor architecture; see https://github.com/istio/istio/issues/37913

PR #3721, opened 2022-02-19
Use RBE when prow runs with workload identity.

bianpengyuan:

### DIFF
--- a/prow/proxy-common.inc
+++ b/prow/proxy-common.inc
@@ -54,8 +54,7 @@ BAZEL_BUILD_RBE_INSTANCE="${BAZEL_BUILD_RBE_INSTANCE-projects/istio-testing/inst
 BAZEL_BUILD_RBE_CACHE="${BAZEL_BUILD_RBE_CACHE-grpcs://remotebuildexecution.googleapis.com}"
 
 METADATA_SERVER_URL="http://169.254.169.254/computeMetadata/v1/instance/service-accounts/"
-METADATA_QUERY_HEADER="Metadata-Flavor: Google"
-WORKLOAD_IDENTITY="istio-prow-build.svc.id.goog"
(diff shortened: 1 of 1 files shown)


prow/proxy-common.inc
```diff
@@ -68,4 +72,17 @@ if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
       export BAZEL_BUILD_ARGS="${BAZEL_BUILD_ARGS} --google_credentials=${GOOGLE_APPLICATION_CREDENTIALS} --remote_instance_name=${BAZEL_BUILD_RBE_INSTANCE} --remote_cache=${BAZEL_BUILD_RBE_CACHE}"
     fi
   fi
+elif curl -H "${METADATA_QUERY_HEADER}" "${METADATA_SERVER_URL}" | grep "${WORKLOAD_IDENTITY_SUFFIX}"; then
```
howardjohn, line 75: i think it's bound to a GSA so may need to change the check to account for that? `gcloud auth list --filter=status:ACTIVE --format="value(account)"` may be simpler as well

prow/proxy-common.inc
```diff
@@ -50,14 +50,18 @@ export BAZEL_OUT
 # Disable RBE execution until we have LARGE_MACHINE instances provisioned for building V8.
 export BAZEL_BUILD_RBE_JOBS="${BAZEL_BUILD_RBE_JOBS:-0}"
```
howardjohn, line 51: nit- the comment here is outdated, the new issue is the tool chain used in RBE is too new for us. We will need to see if we can switch it over to our image

bianpengyuan:
Let me dig a bit into the failure..

howardjohn (approved):

kyessenov (approved):

PR #3752, opened 2022-03-04
Enable CryptoMB PrivateKeyProvider extension

rveerama1:

Envoy cryptomb contrib extension is an extension which brings tls private key providers features to Istio.

Currently, it can be activated via ProxyConfig in Istio service-mesh environment, the config looks like below:
To set the mesh wide defaults, configure the `defaultConfig` section of `meshConfig`. For example:
```
    meshConfig:
      defaultConfig:
        privateKeyProvider:
          name: "cryptomb"
          config:
            "@type": type.googleapis.com/envoy.extensions.private_key_providers.cryptomb.v3alpha.CryptoMbPrivateKeyMethodConfig
            pollDelay: 0.01s
```
This can also be configured on a per-workload basis by configuring the `proxy.istio.io/config` annotation on the pod.
For example:
```
    annotations:
      proxy.istio.io/config: |
        privateKeyProvider:
          name: "cryptomb"
          config:
            "@type": type.googleapis.com/envoy.extensions.private_key_providers.cryptomb.v3alpha.CryptoMbPrivateKeyMethodConfig
            pollDelay: 0.01s
```

istio-api addition PR: https://github.com/istio/api/pull/2261
istio related changes PR: https://github.com/istio/istio/pull/37681

### DIFF
--- a/bazel/extension_config/extensions_build_config.bzl
+++ b/bazel/extension_config/extensions_build_config.bzl
@@ -374,6 +374,7 @@ ISTIO_ENABLED_CONTRIB_EXTENSIONS = [
     "envoy.filters.network.mysql_proxy",
     "envoy.filters.network.sip_proxy",
     "envoy.filters.sip.router",
+    "envoy.tls.key_providers.cryptomb",
 ]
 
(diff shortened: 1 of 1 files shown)


lambdai (approved):
It's ok to me that the proxy build with this contrib extension. 

I have concern whether to introduce istio api on top of contrib feature, but it is discussed in istio/api issue

PR #3771, opened 2022-03-15
stackdriver: add support for generic_node MR

douglas-reid:
This PR adds support for detection of non-Kubernetes, non-GCE workloads to the Stackdriver extension. This will enable reporting of telemetry against non-GCP resources when the extension is used in non-GCP environments.

It uses the following logic for the [`generic_node`](https://cloud.google.com/monitoring/api/resources#tag_generic_node) MR:

- `project` <--  `platform_metadata.gcp_project`
- `location` <-- `platform_metadata.gcp_location`
- `namespace` <-- `namespace`
- `node_id` <-- `instance_ips`

/cc @liminw


douglas-reid:
test_proxy failure looks unrelated to this change.

lambdai (commented):
LGTM. I will leave the rest to @kyessenov

extensions/common/proto_util.cc
```diff
@@ -105,6 +118,7 @@ flatbuffers::DetachedBuffer extractNodeFlatBufferFromStruct(
   node.add_labels(labels_offset);
   node.add_platform_metadata(platform_metadata_offset);
   node.add_app_containers(app_containers_offset);
+  node.add_instance_ips(ip_addrs_offset);
```
kyessenov, line 121: ip_addrs_offset is not initialized when size() == 0.

kyessenov: This needs addition to unit tests I think.

douglas-reid: Are the same changes needed for `app_containers_offset` ?

kyessenov: Not sure. Library might already be handling uninitialized offsets. Adding a test would make it clear.

extensions/stackdriver/common/utils.cc
```diff
@@ -35,6 +36,23 @@ const std::string getContainerName(
...
+  if (!ip_addrs || ip_addrs->size() == 0) {
+    return "istio-proxy";
+  }
+
+  std::string node_id;
+  auto it = ip_addrs->begin();
+  node_id = flatbuffers::GetString(*it);
+  ++it;
+  for (; it != ip_addrs->end(); ++it) {
+    node_id = absl::StrCat(node_id, ",", flatbuffers::GetString(*it));
```
kyessenov, line 51: nit: StrJoin is probably faster since you are allocating a string each time here.

extensions/stackdriver/common/utils.cc
```diff
@@ -35,6 +36,23 @@ const std::string getContainerName(
   return kIstioProxyContainerName;
 }
 
+const std::string getNodeID(
```
kyessenov, line 39: const std::string is a tautology. just a string is fine.

kyessenov (approved):
/hold
Just minor nits.

extensions/stackdriver/common/utils.cc
douglas-reid: do mean instead of the loop?  or instead of `StrCat` inside of the loop?  `StrJoin` had issues taking ip_addrs directly, iirc.

extensions/stackdriver/common/utils.cc
douglas-reid: done.

extensions/stackdriver/common/utils.cc
douglas-reid: done. ptal.

PR #3935, opened 2022-07-13
Do not hardcode arch to `k8`

howardjohn:


scripts/release-binary.sh
```diff
@@ -115,6 +115,11 @@ if [ -n "${DST}" ]; then
     || echo 'Building a new binary.'
 fi
 
+ARCH_NAME="k8"
+case "$(uname -m)" in
+  aarch64) ARCH_NAME"aarch64";;
```
douglas-reid, line 120: my shell scripting knowledge is as close to zero as it can reasonably be, but should this include an `=` ?

howardjohn: I hope so... or my knowledge is a lot less than I thought :-) thanks

PR #3937, opened 2022-07-14
release: fixed missed location for `${ARCH_SUFFIX}`

howardjohn:
Note: currently testing this internally for now but this will be part of standard CI, I hope, soon in OSS

### DIFF
--- a/scripts/release-binary.sh
+++ b/scripts/release-binary.sh
@@ -162,8 +162,8 @@ do
   export BUILD_CONFIG=${config}
 
   echo "Building ${config} proxy"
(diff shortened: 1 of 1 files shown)


kyessenov (commented):
Do you plan to add pre-submit for ARM as well? Envoy has a lot of arch-dependent code so it's not safe to release something we don't test in pre-submit.

howardjohn:
Yep. We don't yet have an ARM cluster (should be in ~1 week), but once we have that we can

kyessenov:
We'll need ARM tests here as well.

howardjohn:
Yes that is what I mean. We will have istio/proxy ARM tests.

kyessenov (approved):

PR #3977, opened 2022-07-29
Minor fix for arm64 build

howardjohn:
Our arm images do not run Ubuntu 16 so we disable the check

### DIFF
--- a/scripts/push-debian.sh
+++ b/scripts/push-debian.sh
@@ -34,6 +34,11 @@ if [[ "$(uname)" != "Darwin" && "${BAZEL_BUILD_ARGS}" != *"--config=libc++"* ]];
   BAZEL_BUILD_ARGS="${BAZEL_BUILD_ARGS} --config=libc++"
 fi
 
+# ARCH_SUFFIX allows optionally appending a -{ARCH} suffix to published binaries.
+# For backwards compatibility, Istio skips this for amd64.
+# Note: user provides "arm64"; we expand to "-arm64" for simple usage in script.
+export ARCH_SUFFIX="${ARCH_SUFFIX+-${ARCH_SUFFIX}}"
+
 set -o errexit
 set -o nounset
 set -o pipefail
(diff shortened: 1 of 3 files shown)


kyessenov:
I'm worried that downstream repos will get broken (etc. https://github.com/istio-ecosystem/wasm-extensions/blob/master/go.mod). Is there any value at all to bump go version requirements?

howardjohn:
@kyessenov it doesn't require Go 1.18 to build `go.mod` files with `go 1.18` -- it will just warm if there are build errors and you are on an older version. It also optimizes the modules downloaded (some improvements in recent go versions).

I don't actually care that much though, I just wanted to make _some_ change to trigger CI. I am fine without changing this if you prefer

kyessenov:
If I remember, we broke gRPC by bumping go-control-plane, because they have to support old Go versions. Wouldn't it complain if you try to build with go1.15 when go.mod wants go1.18?

howardjohn:
Only if you actually depend on go 1.18 features afaik. Like I said, I don't
mind dropping this though, Ill just change some comment somewhere instead

howardjohn:
@kyessenov I kept the same PR but entirely changed the contents btw

scripts/push-debian.sh
```diff
@@ -34,6 +34,11 @@ if [[ "$(uname)" != "Darwin" && "${BAZEL_BUILD_ARGS}" != *"--config=libc++"* ]];
   BAZEL_BUILD_ARGS="${BAZEL_BUILD_ARGS} --config=libc++"
 fi
 
+# ARCH_SUFFIX allows optionally appending a -{ARCH} suffix to published binaries.
+# For backwards compatibility, Istio skips this for amd64.
+# Note: user provides "arm64"; we expand to "-arm64" for simple usage in script.
+export ARCH_SUFFIX="${ARCH_SUFFIX+-${ARCH_SUFFIX}}"
```
kyessenov, line 40: Never seen this syntax. What does this do?

scripts/push-debian.sh
```diff
@@ -63,21 +68,27 @@ fi
...
-BAZEL_OUT="$(bazel info ${BAZEL_BUILD_ARGS} output_path)/k8-opt/bin"
+BAZEL_OUT="$(bazel info ${BAZEL_BUILD_ARGS} output_path)/${ARCH_NAME}-opt/bin"
 BAZEL_BINARY="${BAZEL_OUT}/tools/deb/istio-proxy"
 
 # shellcheck disable=SC2086
 bazel build ${BAZEL_BUILD_ARGS} --config=release ${BAZEL_TARGET}
 
 if [[ -n "${GCS_PATH}" ]]; then
-  gsutil -m cp -r "${BAZEL_BINARY}.deb" "${GCS_PATH}"/
+  gsutil -m cp -r "${BAZEL_BINARY}.deb" "${GCS_PATH}/istio-proxy${ARCH_SUFFIX}.deb"
```
kyessenov, line 88: This changes name for x86 artifacts?

scripts/push-debian.sh
howardjohn: X="${Y+blah}" means

```
if y is set {
 x=blah
} else
 x=""
}
```

So this just adds a `-` in front of ARCH_SUFFIX unless its empty. To achieve `user provides "arm64"; we expand to "-arm64" for simple usage in script.`

scripts/push-debian.sh
howardjohn: No, since ARCH_SUFFIX will not be set for the x86 stuff for backwards compat

kyessenov (approved):

PR #4003, opened 2022-08-10
telemetry: fix handling of off-GCP generic node cases

douglas-reid:
In certain off-GCP testing scenarios, the Stackdriver extension can be used with no platform metadata (no `GCP_METADATA`). This can lead to proxy crashes when trying to lookup platform metadata when building a Monitored Resource for reporting.

This PR adds a check to ensure platform metadata exists when building `generic_node` MRs.  A test is added to ensure no crashing.

Fixes istio/istio#40201


kyessenov (approved):
Thanks.

PR #4047, opened 2022-09-06
deflake: combine split metric point values during comparison

kyessenov:

Sum up metrics that got split into separate points for whatever reason.

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -21,7 +21,7 @@ BAZEL_TARGETS ?= //...
 # Don't build Debian packages and Docker images in tests.
 BAZEL_TEST_TARGETS ?= ${BAZEL_TARGETS}
 E2E_TEST_TARGETS ?= $$(go list ./...)
-E2E_TEST_FLAGS ?=
+E2E_TEST_FLAGS ?= -p=1 -parallel=1
 HUB ?=
 TAG ?=
 repo_dir := .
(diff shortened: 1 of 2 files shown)


test/envoye2e/stackdriver_plugin/stackdriver.go
```diff
@@ -120,11 +128,16 @@ func (sd *Stackdriver) Cleanup() {
...
 func (sd *Stackdriver) Check(p *driver.Params, tsFiles []string, lsFiles []SDLogEntry, verifyLatency bool) driver.Step {
 	// check as sets of strings by marshaling to proto
-	twant := make(map[string]struct{})
+	twant := make(map[string]int64)
 	for _, t := range tsFiles {
 		pb := &monitoring.TimeSeries{}
 		p.LoadTestProto(t, pb)
-		twant[prototext.Format(pb)] = struct{}{}
+		if len(pb.Points) != 1 || pb.Points[0].Value.GetInt64Value() == 0 {
+			panic("malformed metric golden")
```
douglas-reid, line 136: should we have a `ValidateGolden` pre-step that can fail (t.Fatalf) tests without panics as a better option here?

kyessenov: Seems like too much work for just one extension. I think panics will be caught early anyways.

douglas-reid: could we `log.Fatalf` here instead?  i know that doesn't make a ton of difference in the end, but I have a strong aversion to seeing `panic` in code (for better or worse).

kyessenov: Done.

Makefile.core.mk
```diff
@@ -21,7 +21,7 @@ BAZEL_TARGETS ?= //...
 # Don't build Debian packages and Docker images in tests.
 BAZEL_TEST_TARGETS ?= ${BAZEL_TARGETS}
 E2E_TEST_TARGETS ?= $$(go list ./...)
-E2E_TEST_FLAGS ?=
+E2E_TEST_FLAGS ?= -p=1 -parallel=1
```
kyessenov, line 24: Golang log muxing is absolutely atrocious. Making it single threaded so we can actually parse the logs.

douglas-reid (approved):

PR #4058, opened 2022-09-13
Ambient branch merge

kyessenov:
Adds extensions from the ambient branch. They don't matter on the data plane unless used by the control plane.


douglas-reid:
Should we add a new file somewhere with a TODO list of things that need to be addressed for ambient extensions (and filter state objects)? I feel like we would ultimately want some coherent scheme for namespacing, etc., as part of this, but I'm not sure that list of things lives anywhere outside our heads.

kyessenov:
@douglas-reid we should have a follow up that puts them into proper hierarchy `source/extensions/...`. Otherwise, this is just plain merge as-is.

douglas-reid:
As long as that work is being tracked somewhere, I'm fine with it.  There are changes in this PR to the extensions used in _Istio Classic_ :tm:. They seem harmless, but I just want to note there is the potential for some impact.

douglas-reid (approved):

PR #4077, opened 2022-09-22
merge metadata object library from ambient

kyessenov:

Remove ssl from metadata object: TLS forwarding is handled by tls_passthrough filter combo.
Merge metadata object into main.


kyessenov:
/assign @stevenctl 
for confirmation about control plane usage
/assign @douglas-reid 
as code owner

kyessenov:
/assign @briansonnenberg
for C++ readability

src/envoy/common/metadata_object_test.cc
```diff
@@ -98,7 +92,6 @@ TEST_F(WorkloadMetadataObjectTest, FromBaggage) {
   EXPECT_EQ(gotDeploy->workloadName(), "foo");
```
kyessenov, line 92: @douglas-reid this breaks hashing, doesn't it? there's is no instance for deployment.

stevenctl:
Seems fine from CP standpoint.

src/envoy/common/metadata_object_test.cc
kyessenov: @howardjohn why isn't instance set for deployments?

howardjohn: Where is it trying to read from? WorkloadMetadataResource over ECDS?

If so, as far as I can tell we should always have the instance name?

if its from baggage, then I don't think that is included as part of baggage, the format we have is just `k8s.deployment.name=<deployment_name>`

kyessenov:
@zirain Want to take a look?

We're slowly merging the new baggage encoding for peer metadata into main code base.

kyessenov:
@douglas-reid can you approve if you think this is improving your code? There's no impact on istio existing filters.

zirain:
> @zirain Want to take a look?
> 
> We're slowly merging the new baggage encoding for peer metadata into main code base.

Is there any design doc for this?

kyessenov:
@zirain I checked but couldn't find any in the community groups. The gist is very simple, for HTTP CONNECT, we are using a different HTTP header to pass the peer metadata.

**Before:**: Header `x-envoy-peer-metadata`, which is a base64 encoded protobuf struct, closely aligned to schema https://github.com/istio/proxy/blob/master/extensions/common/node_info.fbs

**After**: Header `baggage`, which is a standard HTTP encoding `key1=value1,key2=value2`. The schema is defined in `WorkloadMetadataObject` in this PR.

The primary advantage is standardization on the RFC https://www.w3.org/TR/baggage/.

zirain:
glad to see removal of flatbuff, which's hard to use in my personal experience
I used to think istio use flatbuff because of performace, correct me if I'm wrong
so, after moving to `baggage`, will performace become better or worse?

kyessenov:
Micro-benchmark result (rough, do not assume precise numbers):
*Before*: header length 356, 4456ns
*After*: header length 163, 572ns

So the new encoding is 2x more compact and 10x faster to decode.

extensions/common/metadata_object.cc
```diff
@@ -0,0 +1,227 @@
...
+    } else if (parts.first == "k8s.cronjob.name") {
+      workload_type = WorkloadType::KUBERNETES_CRONJOB;
+      workload = parts.second;
+    } else if (parts.first == "app.name") {
+      app_name = parts.second;
+    } else if (parts.first == "app.version") {
+      app_version = parts.second;
+    }
+  }
+  return WorkloadMetadataObject(instance, cluster, namespace_name, workload,
```
zirain, line 71: add comments for last two `empty`?

extensions/common/metadata_object.cc
```diff
@@ -0,0 +1,227 @@
...
+  absl::string_view app_version;
+  WorkloadType workload_type = WorkloadType::KUBERNETES_POD;
+  const std::vector<std::string> empty;
+
+  std::vector<absl::string_view> properties =
+      absl::StrSplit(baggage_header_value, ',');
+  for (absl::string_view property : properties) {
+    std::pair<absl::string_view, absl::string_view> parts =
+        absl::StrSplit(property, "=");
+    if (parts.first == "k8s.namespace.name") {
```
zirain, line 43: change these `if-else` to `switch-case`, WDYT?

kyessenov: You can't switch on strings :) But yes, we can tokenize here with enums, will do that.

extensions/common/metadata_object.cc
```diff
@@ -0,0 +1,227 @@
...
+  const auto* version_iter = labels->LookupByKey("version");
+  const auto* version = version_iter ? version_iter->value() : nullptr;
+  const absl::string_view app_version = toAbslStringView(version);
+
+  Istio::Common::WorkloadType workload_type =
+      Istio::Common::WorkloadType::KUBERNETES_POD;
+  // Strip "/workload_name" and check for workload type.
+  absl::string_view owner = toAbslStringView(node.owner());
+  owner.remove_suffix(workload.size() + 1);
+  if (absl::EndsWith(owner, "/deployments")) {
```
zirain, line 210: auto owners = absl.StrSplit(owner, "/")];
switch (owners[len-1]){
case 'deployments':
case 'cornjobs':
}

kyessenov: This would be doing more work since it's a long string. Probably easier to find last index and just tokenize/switch on the last part.

zirain: avoid multi `absl:EndsWith`

kyessenov: Yeah, will fix this with switch. Thanks.

extensions/common/metadata_object.cc
kyessenov: Removed those two fields since they are unused.

extensions/common/metadata_object.cc
kyessenov: Done.

kyessenov:
@zirain I'll need your approval from P&T group since there is no one else around :)

zirain:
sorry, I'm OOO for national holidays. will take a look when BTO.

zirain (approved):
LGTM, thanks

PR #4079, opened 2022-09-23
stats: rewrite as native extension

kyessenov:

This is a quick prototype of Istio stats extension written as a native extension. The advantages are numerous:

* Unblocks ability to implement metric expiry, memory accounting for multi tenancy, and opens paths for Envoy improvements in that area;
* Removes requirement to rely on regex parsing to strip tags. Tags are produced directly.
* Removes dependency on Wasm. 
* Removes dependency on flatbuffers.
* CPU improvements. It is well known that Istio proxy is not frugal with CPU. This PR removes a lot of abstraction overhead (via Wasm, flatbuffers, regexes), and can be made extremely efficient with regards to memory / CPU cycles (Envoy uses highly efficient packed strings which are sub-pointer size).

*Update*:
This extension conforms to the original extensions for the common case. The main gaps are:
* support for configurable dimensions and metrics via CEL;
* support for gRPC streaming metrics;
* some flatbuffers dependencies are still present to allow seamless transition.

Neither of those are difficult in principle to add, but this PR is sufficient for the purpose of rebasing for Ambient.

*Update 2*:
Added limited CEL support and removal/modification of metrics. I am not sure how far we really want to support that in the end, but removal/little modifications are fairly safe.


zirain:
@kyessenov do you plan to land this in 1.16?
another question is can user use this in previous version(1.15 etc)?

kyessenov:
I might be able to get this in 1.16 as an option. No, I'm not planning to backport to previous versions. I think this is mostly need for Ambient, since Wasm is hard to make multi-tenant and shared proxies are, well, shared.

douglas-reid:
Of all of the benefits:

> Removes requirement to rely on regex parsing to strip tags. Tags are produced directly.

would probably be the biggest UX improvement for Istio customers. Dynamic ability to add tags would reduce so much pain (and reduce the learning curve).

zirain:
> I might be able to get this in 1.16 as an option. No, I'm not planning to backport to previous versions. I think this is mostly need for Ambient, since Wasm is hard to make multi-tenant and shared proxies are, well, shared.

the new filter can run in previous envoy?

kyessenov:
> the new filter can run in previous envoy?

I think the internal stats APIs have not changed much for awhile, so yes, you could backport it to older Envoy releases.

kyessenov:
Benchmark runs:
- [stats-old.pdf](https://github.com/istio/proxy/files/9737341/stats-old.pdf)
- [stats-new.pdf](https://github.com/istio/proxy/files/9737340/stats-new.pdf)
(load test at 3kqps for 30s for single threaded envoy pinned to a core, no peer info)

Stats overhead:
- before: PluginRootContext::report 0.03s (0.49%) of 1.17s (19.02%)
- after: HttpFilters::IstioStats 0.06s (1.37%) of 0.55s (12.53%)

So we save about 40% CPU, but could do better if we reduce per-request overhead further.

zirain (commented):
small nit, otherwise, lgtm

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -0,0 +1,669 @@
...
+      tags_.push_back({context_.request_protocol_, context_.grpc_});
+    } else {
+      tags_.push_back({context_.request_protocol_, context_.http_});
+    }
+
+    // TODO: copy Http::CodeStatsImpl version for status codes and flags.
+    tags_.push_back(
+        {context_.response_code_,
+         config_->resolve(absl::StrCat(info.responseCode().value_or(0)))});
+    if (is_grpc) {
```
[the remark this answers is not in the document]
kyessenov: Thanks for taking a look. The order of tags is preserved so I can't move the lines together without changing how vector is appended.

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -0,0 +1,669 @@
...
+      const auto* headers = info.getRequestHeaders();
+      if (headers && headers->Host()) {
+        service_host = headers->Host()->value().getStringView();
+        service_host_name = service_host;
+      }
+    }
+    switch (config_->reporter()) {
+      case Reporter::ServerSidecar: {
+        tags_.push_back({context_.source_workload_,
+                         peer ? config_->resolve(peer->workload_name_)
```
sschepens, line 503: could there maybe be some opportunity optimizing this not to use a map lookup?

for example, can't we initialize these into variables when the connection is first established? so then we can directly access the variable and not do map operations

kyessenov: Yes, I think that's the only way to further optimize this, but it requires some architectural changes to make lifecycle of metadata more accurate. There are two options:
* require TCP metadata for HTTP traffic, and then move the peer population to the filter constructor;
* require CONNECT tunneling (from ambient), which is equivalent to the first with a different protocol.

Without knowing a-priori that connections cannot be shared between peers, we can't do much but expect every request to have different per-request data.

kyessenov: https://github.com/istio/istio/issues/36420

I think the performance is going to be better with native rewrite.

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -0,0 +1,669 @@
...
+        break;
+      }
+      default:
+        break;
+    }
+  }
+
+  ConfigSharedPtr config_;
+  Context& context_;
+  Stats::StatNameTagVector tags_;
```
sschepens, line 624: I don't know what the default vector capacity is, maybe should initialize this vector with the worst-case scenario? to avoid needing to grow the vector in most cases

source/extensions/filters/http/istio_stats/istio_stats.cc
sschepens: Hmm yes, it would be great to rewrite TCP MX in native and use that instead of HTTP where possible, i'm sure there would be CPU usage benefits as well as latency improvements.

source/extensions/filters/http/istio_stats/istio_stats.cc
kyessenov: Good idea, reserved it for maximum 25 (number of HTTP dimensions).

source/extensions/filters/http/istio_stats/istio_stats.cc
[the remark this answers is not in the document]
zirain: > The order of tags is preserved so I can't move the lines together without changing how vector is appended.

I cannot 100% understand this, can share more details?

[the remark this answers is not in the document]
kyessenov: The order of prometheus tags between the old and new implementations matches. I don't think it's important, but it's easier not to change the test framework to ignore the order.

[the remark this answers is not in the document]
zirain: got it, thanks

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -0,0 +1,670 @@
...
+      tags_.push_back({context_.grpc_response_status_,
+                       optional_status ? config_->resolve(absl::StrCat(
+                                             optional_status.value()))
+                                       : context_.empty_});
+    } else {
+      tags_.push_back({context_.grpc_response_status_, context_.empty_});
+    }
+    populateFlagsAndConnectionSecurity(info);
+
+    Stats::Utility::counterFromElements(
```
ymesika, line 305: Why don't you use `counterFromStatNames` if all elements are of type _StatName_? It's mentioned to be faster.

ymesika:
@kyessenov I'm trying to use this with Ambient's ztunnel and added it both as http and filter listener of the inbound.
While I see the `counterFromElements` being called with elements and tags I don't see anything when getting the metrics from the `:15020/stats`.
Is there some extra configuration that needs to be set? Not providing any configuration to the filters atm.

ymesika:
> @ymesika Maybe you need to call /stats/prometheus to see the tagged metrics?

Sorry, that's what I called.. Copy-paste got truncated

source/extensions/filters/http/istio_stats/istio_stats.cc
kyessenov: Thanks, yes I should, changed.

kyessenov:
@ymesika Not sure, maybe something to do with stats_config allow-list in the bootstrap? Try to clear it and see if it shows up.

zirain:
@ymesika @sschepens do your guys have any further options, I'd like merged this first.

ymesika:
> @ymesika @sschepens do your guys have any further options, I'd like merged this first.

I don't have more comments. Thanks.

kyessenov:
Thanks, want to finish the expressions part so we can substitute in general.

PR #4082, opened 2022-09-26
Update go.mod go directive

ericvn:

Update the go.mod go directive to match that in istio/istio.

Note that there really isn't a need to update this directive. We haven't added anything the needs Go 1.16+ functionality (and not even sure if we need 1.15). 

We do pick up lint errors when specifying a later Go version as there are library deprecations. Note that the changes to fix this lint would require a go 1.16 as the library functions were added in 1.16 (I suspect that leaving the go directive as 1.15 will cause compilation errors trying to access the 1.16 functions).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,11 @@
 module istio.io/proxy
 
-go 1.15
+go 1.19
 
 require (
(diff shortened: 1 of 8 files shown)


zirain (approved):
thanks!

PR #4086, opened 2022-09-29
Changes to filters to support ztunnel tcp metrics

ymesika:

The PR has changes to some stats relates filters to support adding them to ztunnel inbound listener to enable TCP metrics.
The labels of the source workload are being set based on the baggage header received in the tunnel connection request.
The labels of the destination workload (local) are being taken from the workloads metadata mapping retrieved by the workload metadata filter through ECDS.

The changes are using the existing mechanisms.

NA

Please notice that I've kept the cluster/host metadata modes in the stats configuration because refactoring the names (to ztunnel/waypoint) will require a sync with a PR on the Istio repo as well.
Changes in this PR shouldn't break the waypoint L7 metrics support or current ambient branch on the Istio repo.


ymesika:
Opened https://github.com/istio/proxy/pull/4090 agains master.
@kyessenov The changes in PR are all in if clauses and shouldn't be executed if not running in a ztunnel mode. So current ambient/istio shouldn't be affected by these changes.
However, I will run ambient/istio integration tests with this proxy to make sure.

PR #4117, opened 2022-10-17
stats(native): use SAN namespace by default

kyessenov:

Make workload metadata namespace optional by parsing SAN first.


source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -761,6 +762,34 @@ class IstioStatsFilter : public Http::PassThroughFilter,
         service_host_name = service_host;
       }
     }
+
+    // Implements fallback from using the namespace from SAN if available to
```
lei-tang, line 766: In the existing implementation without this PR, where does SAN come from (e.g., from peer metadata or from peer certificate)? Is peer metadata from metadata exchange or from the metadata in endpoint of an EDS response?

kyessenov: SAN always comes from a certificate. Peer metadata is from metadata exchange. EDS response only contains destination service I think.

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -761,6 +762,34 @@ class IstioStatsFilter : public Http::PassThroughFilter,
...
+      peer_san = ssl_info->uriSanPeerCertificate()[0];
+    }
+    absl::string_view peer_namespace;
+    if (!peer_san.empty()) {
+      const auto san_namespace = Utils::GetNamespace(peer_san);
+      if (san_namespace) {
+        peer_namespace = san_namespace.value();
+      }
+    }
+    if (peer_namespace.empty() && peer) {
```
lei-tang, line 785: In which scenarios, will the variable peer be nil? 
If peer_namespace=="foo" and peer == nil, peer_namespace will be unset. 
Maybe checking peer != nil in a separate checking statement that is immediately after peer is set?

kyessenov: Peer is nil when there is no metadata exchange (either header, or TCP prefix), and it's set on L732 and used extensively below. It's OK to have peer_namespace empty if there is no MX and no SAN namespace, it's handled with `!peer_namespace.empty()`.

lei-tang (approved):

PR #4125, opened 2022-10-19
stats(native): fixes to align with wasm stats

kyessenov:

Fix an error in parsing in Wasm stats for endpoint encoding (parts == 4 leads to crash).
Adds fallbacks to native stats for endpoint encoding, blackhole/passthrough.
Handle empty values properly (should be unknown).
Finish CEL expression support.


testdata/stats/client_config_customized.yaml.tmpl
```diff
@@ -1,8 +1,5 @@
 field_separator: ";.;"
 definitions:
-- name: requests_total
```
lei-tang, line 3: Why is requests_total deleted?

kyessenov: This test case was changing the value for a standard metric. I don't think we want to support this feature.

lei-tang (approved):

PR #4164, opened 2022-11-04
authn: network peer identity

kyessenov:

Define a new filter state to carry the peer identity (SPIFFE URI SAN).

This will deprecate the "TLS capture" hack used in ambient, and generalize it for https://github.com/istio/istio/issues/41585.

Future PRs will transition points-of-use to this filter state:
* RBAC will refer to the filter state instead of direct "authenticated principal" from TLS. Ambient uses plaintext when applying RBAC policies, so this is needed. Ref: https://github.com/envoyproxy/envoy/pull/23828.
* Stats will refer to the filter state instead of re-computing the principal. This is needed to make sure policy&telemetry apply to the same principal.


lei-tang (approved):
LGTM. Consider adding tests when the feature is complete.

PR #4175, opened 2022-11-07
stats: add server gateway reporter for shared proxies

kyessenov:

Ambient changes:
* remove experimental changes to the stable wasm stats filter.
* re-introduce them to the new stats filter, xDS must set `reporter: SERVER_GATEWAY` to read the endpoint metadata for server reporting.
* `CLUSTER_ID` and `NAMESPACE` are derived from the bootstrap for efficiency, must be set correctly on the gateway proxies.

xDS change:
* switch to native stats for Waypoints;
* set `reporter: SERVER_GATEWAY` as config.


source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -58,17 +58,37 @@ absl::string_view extractMapString(const ProtobufWkt::Struct& metadata,
...
+      }
+    }
+  }
+  return {};
+}
+
 enum class Reporter {
   ClientSidecar,
   ServerSidecar,
+  ServerGateway,
```
lei-tang, line 82: What is a ServerGateway? Is ServerGateway for ingress/egress gateway?

kyessenov: It's "waypoint" proxy in Ambient mode, which is a server proxy shared between backends.
Basically, it's a sidecar that is shared between deployment pods.

kyessenov: This is usually called internal gateway, but "server" sort of implies internal since it's not cluster gateway.

kyessenov:
@stevenctl FYI, this will need control plane changes.

source/extensions/filters/http/istio_stats/istio_stats.cc
zirain: can you add some comment for this?

lei-tang (approved):

PR #4194, opened 2022-11-11
authn: use filter state for principal on the server-side

kyessenov:

* Use network `istio_authn` filter state in the stats filter. This is needed for https://github.com/istio/istio/issues/41585, e.g. if the principal is obtained from the tunnel rather than direct TLS observation. Stats filter will fall back to the original behavior in order to maintain compatibility.

* Add unit and integration test for `istio_authn`.  RBAC integration depends on https://github.com/envoyproxy/envoy/pull/23828.

_Ambient breaking change:_
As a cleanup, we can now remove the experimental filters for capturing the TLS context. xDS should now install `istio_authn` filter on the tunnel.


kyessenov:
/retest
The really annoying flake in stackdriver.

source/extensions/filters/network/istio_authn/config.cc
```diff
@@ -28,39 +26,63 @@ namespace Extensions {
 namespace NetworkFilters {
 namespace IstioAuthn {
 
-absl::optional<uint64_t> PeerPrincipal::hash() const {
+absl::optional<uint64_t> Principal::hash() const {
```
lei-tang, line 29: There are a few terms in this PR related to principal, e.g., Principal, PeerPrincipal (e.g., PeerPrincipalKey), LocalPrincipal (e.g., LocalPrincipalKey).

- Why change PeerPrincipal to Principal? 
- Are you generalizing PeerPrincipal into Principal? If so, will PeerPrincipalKey and LocalPrincipalKey be factored too? 
- Are there any comments in the code describing the differences between Principal, PeerPrincipal, LocalPrincipal?

kyessenov: Principal is a shared representation for peer and local principal. The keys refers to peer vs local, but both values are Principal structs. The reason why local is important is multi-tenancy (which we don't have enabled yet), if the proxy can present multiple certificates, we need to detect and propagates the local principal as well.

The refactor was needed because I didn't have local one before, and then realized it's important and needed.

lei-tang (approved):

PR #4211, opened 2022-11-16
stats: support block_all and allow_any routes

kyessenov:

Last missing fallback for "block_all" route names.
Should make https://github.com/istio/istio/pull/41991 pass.


kyessenov:
All the fallbacks are poorly documented and temporary unfortunately. We probably need a better mechanism with metadata discovery down the line.

lei-tang (approved):

PR #4214, opened 2022-11-16
remove ambient extensions

kyessenov:

Since these are not tested and don't receive work to promote further, delete them to prevent tech debt.
Can be recovered from git if needed.

/assign @stevenctl


stevenctl (commented):
Do we have them on some other branch or are we relying on git history?

For the ambient experimental branch we need to temporarily pin to a version of envoy that has these.

kyessenov:
Well, the author is gone so git history is the best we can do. Ambient branch is already pinned, we already replaced a couple of extensions.

zirain (approved):
@stevenctl feel free to remove `DNM`

stevenctl (approved):

PR #4191, opened 2022-11-10
fix static cluster filter

zirain:
fix: https://github.com/istio/proxy/pull/4190

PR #4234, opened 2022-11-21
auto update extensions_build_config.bzl

zirain:
avoid issue like https://github.com/istio/proxy/pull/4231

cc @kyessenov @lei-tang


zirain:
the original thought is preventing manual sync this configuration with upstream, like https://github.com/istio/istio/issues/42080 
https://github.com/istio/proxy/pull/4216 https://github.com/istio/proxy/pull/4191 https://github.com/istio/proxy/pull/4200

kyessenov:
@zirain That doesn't happen that often. The cluster work was to reduce the binary size for envoy mobile, which is a one-time thing. I think that's a good question for TOC, whether we auto-import new extensions in Envoy into Istio. I'm mostly concerned about the support, some extensions are experimental and not very stable and may later be removed from Envoy.

zirain:
> @zirain That doesn't happen that often. The cluster work was to reduce the binary size for envoy mobile, which is a one-time thing. I think that's a good question for TOC, whether we auto-import new extensions in Envoy into Istio. I'm mostly concerned about the support, some extensions are experimental and not very stable and may later be removed from Envoy.

another question is why some extensions have a different name with upstream? 

if we do not want to support some extensions, we should add them to `ISTIO_DISABLED_EXTENSIONS `?

PR #4255, opened 2022-11-30
waypoint: metadata discovery service client

kyessenov:

This PR adds a workload discovery client support to Envoy, which allows Envoy to retrieve additional metadata by an IP address using delta ADS .

Sample config:
```yaml
dynamic_resources:
  ads_config:
    api_type: DELTA_GRPC
    transport_api_version: V3
    grpc_services:
      envoy_grpc:
        cluster_name: xds_cluster
bootstrap_extensions:
- name: metadata_discovery
  typed_config:
    "@type": type.googleapis.com/udpa.type.v1.TypedStruct
    type_url: type.googleapis.com/istio.workload.BootstrapExtension
    value:
      config_source:
        ads: {}
```

`AGGREGATED_DELTA_GRPC` needs to be completed in Envoy to support both delta and sotw ADS simulatenously.

Context: https://github.com/istio/istio/issues/42070


zirain:
can your share doc/PR about `AGGREGATED_DELTA_GRPC `?

kyessenov:
@zirain It's a hidden option in Envoy that's not fully implemented for muxing delta and sotw over ADS. I think it's probably easier to separate the delta stream from ADS initially, but need a PR in envoy to allow new delta services on top of builtin xDS.

kyessenov:
CC @costinm @howardjohn 

This implements a waypoint telemetry discovery service using the shared WDS from istiod. This is strictly guarded to:
* waypoint, backend IPs only;
* optional, enabled via bootstrap extension;
* confirmed to work with delta ADS, does not work per service delta xDS (upstream envoy issue);

With this change, we have uniform handling of "destination" metadata in the waypoints, that no longer depends on EDS metadata or some other ECDS tables. 

Some future work may include:
* add on-demand support;
* change workload key away from IP into something more trustworthy;
* support for source telemetry to eliminate "baggage" request;
* inclusion into sidecar xDS;

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -975,7 +1001,10 @@ class IstioStatsFilter : public Http::PassThroughFilter,
                                                      : context_.unknown_});
       switch (config_->reporter()) {
       case Reporter::ServerGateway: {
```
kyessenov, line 1003: Main guard: only Waypoints set ServerGateway.

lei-tang (approved):
LGTM. The PR is guarded for Ambient waypoints.

PR #4202, opened 2022-11-15
Enable QAT PrivateKeyProvider extension

rveerama1:
**What this PR does / why we need it**:
An addition to CryptoMB private key provider extension https://github.com/istio/proxy/pull/3752. 

QAT support in Envoy is merged and available from https://github.com/envoyproxy/envoy/pull/21984. 

Next generation Intel® QAT support with Intel® Xeon® Scalable processors will feature an Intel® QAT cryptography and compression acceleration engine.

QAT private key provider extension will use qatlib library (https://github.com/intel/qatlib) to accelerate RSA operations in handshakes. The extension will look a bit like the existing cryptomb private key provider. The use case is to move the expensive cryptographic operations away from the CPU to the accelerator device, leaving CPU cycles for other use.

Additional Description:
Support for Intel® QAT is already present in the mainline Linux kernel and in Kubernetes device plugins (to expose the device files to containers). There are previous generations of Intel QAT® hardware devices, but they are not supported by this extension.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

PR #4320, opened 2023-01-06
enable DLB connection balance extension

daixiang0:
Signed-off-by: Loong <loong.dai@intel.com>

It is to enable DLB connection balance extension.

DLB support in Envoy is merged and available from https://github.com/envoyproxy/envoy/pull/20268

Intel® Dynamic Load Balancer (Intel® DLB) is a hardware managed system of queues and arbiters connecting producers and consumers. It is a PCI device envisaged to live in the server CPU uncore and can interact with software running on cores, and potentially with other devices. More details please refer to https://www.intel.com/content/www/us/en/download/686372/intel-dynamic-load-balancer.html.

With EnvoyFilter, this feature can work, no more extra integration work in Istio side.

### DIFF
--- a/bazel/extension_config/extensions_build_config.bzl
+++ b/bazel/extension_config/extensions_build_config.bzl
@@ -395,6 +395,12 @@ ENVOY_CONTRIB_EXTENSIONS = {
     #
 
     "envoy.bootstrap.vcl":                                      "//contrib/vcl/source:config",
+
+    #
+    # Connection Balance extensions
+    #
+
+    "envoy.network.connection_balance.dlb":                     "//contrib/network/connection_balance/dlb/source:connection_balancer",
 }
 
 
@@ -409,6 +415,7 @@ ISTIO_ENABLED_CONTRIB_EXTENSIONS = [
     "envoy.filters.sip.router",
     "envoy.tls.key_providers.cryptomb",
     "envoy.tls.key_providers.qat",
+    "envoy.network.connection_balance.dlb",
 ]
 
 EXTENSIONS = dict([(k,v) for k,v in ENVOY_EXTENSIONS.items() if not k in ISTIO_DISABLED_EXTENSIONS] +


daixiang0:
wasm test failed

/retest

daixiang0:
@istio/wg-policies-and-telemetry-maintainers please review.

zirain:
@kfaseela have same request in https://github.com/istio/istio/issues/42380.

I think istio need a clear rule about contrib filter import policy.

cc @howardjohn @lei-tang @kyessenov

daixiang0:
> @kfaseela have same request in [istio/istio#42380](https://github.com/istio/istio/issues/42380).
> 
> I think istio need a clear rule about contrib filter import policy.
> 
> cc @howardjohn @lei-tang @kyessenov

They are different cases: Postgres has been moved from core to contrib, but no move history for DLB.

Also, **it is not a filter**, actually, it accelerates connection balance inside Envoy, which means all key applications like ingress gateway can get benefit from it.

It's more likes https://github.com/istio/proxy/pull/4202.

zirain (commented):
seems reasonable, left to @kyessenov for another look.

kyessenov:
I don't think we can say "no" after accepting cryptomb and other contrib accelerators.

kyessenov (approved):

kyessenov:
Per-usual: please work with Istio community to define a first-class API (not EnvoyFilter). Without that, users cannot assume this extension works for managed versions of Istio.

daixiang0:
@kyessenov Thanks, I will do it:)

ramaraochavali:
IMO, crytoMB is different. We have an [API](https://github.com/istio/api/blob/master/mesh/v1alpha1/proxy.proto#L301) for it already. I certainly think we need a process around bringing  Envoy contrib extensions to Istio Proxy. Otherwise it has potential to destabilize Istio proxy.
cc: @howardjohn @hzxuzhonghu

kyessenov:
@ramaraochavali That's quite surprising to have a de-facto stable API (Mesh) to depend on a contrib extension in Envoy (which is technically "experimental"). What matters is that they should be aligned: an experimental Istio API can depend on contrib Envoy extension, but a stable Istio API should not depend on contrib. 

@ramaraochavali Does Istio have a way to define experimental APIs? I think almost all versions have been pretty much stuck under v1alpha1.

ramaraochavali:
> That's quite surprising to have a de-facto stable API (Mesh) to depend on a contrib extension in Envoy (which is technically "experimental"). What matters is that they should be aligned: an experimental Istio API can depend on contrib Envoy extension, but a stable Istio API should not depend on contrib.

I agree.

daixiang0:
@kyessenov do you know which release would include this feature?

PR #4328, opened 2023-01-10
enable postgres filter extension

kfaseela:
Signed-off-by: Faseela K <faseela.k@est.tech>

Fixes #42380

### DIFF
--- a/bazel/extension_config/extensions_build_config.bzl
+++ b/bazel/extension_config/extensions_build_config.bzl
@@ -411,6 +411,7 @@ ISTIO_DISABLED_EXTENSIONS = [
 
 ISTIO_ENABLED_CONTRIB_EXTENSIONS = [
     "envoy.filters.network.mysql_proxy",
+    "envoy.filters.network.postgres_proxy",
     "envoy.filters.network.sip_proxy",
     "envoy.filters.sip.router",
     "envoy.tls.key_providers.cryptomb",


zirain (approved):
approved after accepting DLB and other contrib accelerators.

ramaraochavali:
I thought we agreed to deploy this in prod using contrib build https://github.com/istio/istio/issues/42380#issuecomment-1353097872 and then based on that we want to move forward. Has that changed? @howardjohn @hzxuzhonghu WDYT?

kfaseela:
> I thought we agreed to deploy this in prod using contrib build [istio/istio#42380 (comment)](https://github.com/istio/istio/issues/42380#issuecomment-1353097872) and then based on that we want to move forward. Has that changed? @howardjohn @hzxuzhonghu WDYT?

@ramaraochavali This was tested by adding in ISTIO_ENABLED_CONTRIB_EXTENSIONS as in this PR, and is working fine, I will add the details on how the testing was done to the original issue. Once this is done, probably we have to working on Istio standard APIs to use postgres, but that can be a different PR and a bigger activity? Until then we can have in CONTRIB_EXTENSIONS which we already have for many other extensions?

ramaraochavali:
> Until then we can have in CONTRIB_EXTENSIONS which we already have for many other extensions? If I have to add atleast a task in istio documentation showing how this works, I would need this PR.

It does not really matter whether it is in contrib extensions or not - it will be part of main Istio proxy build. All I am saying is we need a process around brining Envoy contrib extensions to Istio (they are there in contrib in Envoy for a reason). Some of the extensions already have an API (for example cryptoMB) and IMO whatever contrib extensions we bring here, we should have a plan to work with Envoy community to move it main so that they have proper test coverage etc.etc.

PR #4406, opened 2023-02-06
istio_authn: share multiple times

howardjohn:
This is needed to sync ambient branch with master. While there is another PR doing this its also changing a bunch of other things, so to just have a minimal "sync to master" this is needed because we need to pass through multiple listener hops.

### DIFF
--- a/source/extensions/filters/network/istio_authn/config.cc
+++ b/source/extensions/filters/network/istio_authn/config.cc
@@ -63,7 +63,7 @@ void IstioAuthnFilter::populate() const {
             PeerPrincipalKey, std::make_shared<Principal>(san),
             StreamInfo::FilterState::StateType::ReadOnly,
             StreamInfo::FilterState::LifeSpan::Connection,
-            StreamInfo::FilterState::StreamSharing::SharedWithUpstreamConnectionOnce);
+            StreamInfo::FilterState::StreamSharing::SharedWithUpstreamConnection);
         break;
       }
     }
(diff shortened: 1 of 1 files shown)


kyessenov:
This would break my PR https://github.com/istio/istio/pull/42961?

howardjohn:
Does it break it? I thought it just made it more liberal (can share across 1 hop or many)?

I don't 100% understand the field, so maybe my understanding is wrong though

kyessenov:
Could you just point Envoy sha back to prior PR https://github.com/istio/proxy/pull/4396? We certainly don't want to share more than once, but I don't see why we need the tip to pass the build.

kyessenov:
Yes, I understand, but instead of reverting recent master changes, I think we need to point experimental-ambient to an old master commit, instead of the tip. We're making breaking changes on istio/proxy for the ambient, and right now, the tips of proxy and experimental-ambient don't work together without https://github.com/istio/istio/pull/42961

howardjohn:
I am not sure how to reconcile your suggestion to point to an old master with the goals I stated, which seem to require we could to `master`. I have a PR on ambient that merges `master` and just needs this patch to pass all tests: https://github.com/istio/istio/pull/43168

kyessenov:
@howardjohn I don't understand the point of making this change so that we revert immediately after. My suggestion is to change istio.deps in your PR to point to an older commit of istio/proxy. Is that not clear?

howardjohn:
Which commit? You pointed to https://github.com/istio/proxy/pull/4396 but that is a completely different filter?

If we can use an old proxy against istio/istio master that SGTM I just don't know which commit?

kyessenov:
I think https://github.com/istio/proxy/pull/4386 was the commit, or anything prior to that from this year is probably fine.

PR #4495, opened 2023-03-07
Enable Hyperscan regex engine extension

zhxie:

Enable Hyperscan regex engine extension.

Hyperscan support in Envoy is merged and available from envoyproxy/envoy#21973

[Hyperscan](https://github.com/intel/hyperscan) is a high-performance multiple regex matching library, using hybrid automata techniques to allow simultaneous matching of large numbers of regular expressions across streams of data, which can be used in situations with large quantities of regex including router and WAF.

Hyperscan regex engine can be enabled with custom bootstrap config.


zhxie:
Looks like the CI uses older-version Clang, which is different from Envoy 🧐

zhxie:
Will close first and expect upstream (Envoy build or Istio CI) change.

kyessenov:
@ericvn Is anybody tracking update of clang to 14 or something more recent? Are we stuck with old clang because of centos?

PR #4519, opened 2023-03-20
algin tcp_reporting_duration default value

zirain:

### DIFF
--- a/source/extensions/filters/http/istio_stats/config.proto
+++ b/source/extensions/filters/http/istio_stats/config.proto
@@ -100,7 +100,7 @@ message PluginConfig {
   bool disable_host_header_fallback = 6;
 
   // Optional. Allows configuration of the time between calls out to for TCP
-  // metrics reporting. The default duration is `15s`.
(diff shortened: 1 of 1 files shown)


source/extensions/filters/http/istio_stats/config.proto
```diff
@@ -100,7 +100,7 @@ message PluginConfig {
   bool disable_host_header_fallback = 6;
 
   // Optional. Allows configuration of the time between calls out to for TCP
-  // metrics reporting. The default duration is `15s`.
+  // metrics reporting. The default duration is `5s`.
```
lei-tang, line 103: Where does the 5s value come from? Why does it need to be changed from 15s to 5s? Any performance impact for changing from 15s to 5s?

kyessenov (approved):

source/extensions/filters/http/istio_stats/config.proto
lei-tang: Thanks!

lei-tang (approved):

PR #4544, opened 2023-03-28
Try and fix erorrs with updated Envoy

ericvn:

Attempt to fix the failures in the pipeline

### DIFF
--- a/test/integration/exchanged_token_integration_test.cc
+++ b/test/integration/exchanged_token_integration_test.cc
@@ -220,10 +220,9 @@ public:
   }
 };
 
-INSTANTIATE_TEST_SUITE_P(
-    Protocols, ExchangedTokenIntegrationTest,
-    testing::ValuesIn(HttpProtocolIntegrationTest::getProtocolTestParams()),
-    HttpProtocolIntegrationTest::protocolTestParamsToString);
+INSTANTIATE_TEST_SUITE_P(Protocols, ExchangedTokenIntegrationTest,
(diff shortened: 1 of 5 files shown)


kyessenov (approved):

kyessenov:
Alright, let me take a look.

test/integration/exchanged_token_integration_test.cc
```diff
@@ -220,9 +220,10 @@ class ExchangedTokenIntegrationTest : public HttpProtocolIntegrationTest {
   }
 };
 
-INSTANTIATE_TEST_SUITE_P(Protocols, ExchangedTokenIntegrationTest,
-                         testing::ValuesIn(HttpProtocolIntegrationTest::getProtocolTestParams()),
-                         HttpProtocolIntegrationTest::protocolTestParamsToString);
+INSTANTIATE_TEST_SUITE_P(
+    Protocols, ExchangedTokenIntegrationTest,
+    testing::ValuesIn(HttpProtocolIntegrationTest::getProtocolTestParamsWithoutHTTP3()),
```
kyessenov, line 225: I don't think we should exclude h3?

kyessenov: The issue is that requests have ":authority" set to "host", so if you change them there, it'll pass with h3.

ericvn: Looks like the tests were passing (maybe an asan flake). Will try reverting the withoutH3 to see how that does,

PR #4545, opened 2023-03-29
Enable Hyperscan input matcher and regex engine extension

zhxie:

This is a follow up of #4495. Current version of Hyperscan is compatible to both Clang 14 (Envoy CI) and Clang 12 (Istio CI).

Enable Hyperscan input matcher and regex engine extension.

Hyperscan support in Envoy is merged and available from envoyproxy/envoy#21973

[Hyperscan](https://github.com/intel/hyperscan) is a high-performance multiple regex matching library, using hybrid automata techniques to allow simultaneous matching of large numbers of regular expressions across streams of data, which can be used in situations with large quantities of regex including router and WAF.

Hyperscan regex engine can be enabled with custom bootstrap config.


zhxie:
May need to check again on Hyperscan's dependency. Close.

PR #4563, opened 2023-04-05
Update Envoy and add changes that came from the private build

ericvn:

See if some changes are needed due to some upstream changes.


ericvn:
This would supersede https://github.com/istio/proxy/pull/4562

kyessenov (approved):

PR #4427, opened 2023-02-10
rotate the scopes to support metrics expiry

ZhengzheYang:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

PR #4597, opened 2023-04-19
stats: implement metric expiry

kyessenov:
Implements a rotation of metric scopes for long-living listeners. Disabled by default.

A graceful drain keeps both active and previous scopes for the additional 5m to accommodate asynchronous metric pull sinks. This work can be used for Otel metric sink.

Subsumes https://github.com/istio/proxy/pull/4427.
/assign @zirain 
/assign @lei-tang

### DIFF
--- a/source/extensions/filters/http/istio_stats/config.proto
+++ b/source/extensions/filters/http/istio_stats/config.proto
@@ -112,9 +112,11 @@ message PluginConfig {
   // Proxy deployment type.
   Reporter reporter = 10;
 
-  // Metric expiry duration. Defaults to 12h. Set to 0 to disable the metric expiry.
-  google.protobuf.Duration expiry_duration = 11;
+  // Metric scope rotation interval. Set to 0 to disable the metric scope rotation.
+  // Defaults to 0.
+  google.protobuf.Duration rotation_interval = 11;
 
-  // Metric expiry graceful drain duration. Defaults to 5m. No-op if the metric expiry is disabled.
-  google.protobuf.Duration graceful_drain_duration = 12;
+  // Metric expiry graceful deletion interval. No-op if the metric rotation is disabled.
+  // Defaults to 5m. Must be >=1s.
+  google.protobuf.Duration graceful_deletion_interval = 12;
 }
--- a/source/extensions/filters/http/istio_stats/config.proto
+++ b/source/extensions/filters/http/istio_stats/config.proto
@@ -111,4 +111,10 @@ message PluginConfig {
 
   // Proxy deployment type.
   Reporter reporter = 10;
+
+  // Metric expiry duration. Defaults to 12h. Set to 0 to disable the metric expiry.
+  google.protobuf.Duration expiry_duration = 11;
+
+  // Metric expiry graceful drain duration. Defaults to 5m. No-op if the metric expiry is disabled.
+  google.protobuf.Duration graceful_drain_duration = 12;
 }
--- a/source/extensions/filters/http/istio_stats/istio_stats.cc
+++ b/source/extensions/filters/http/istio_stats/istio_stats.cc
@@ -103,8 +103,7 @@ enum class Reporter {
   ClientSidecar,
   // Regular inbound listener on a sidecar.
   ServerSidecar,
-  // Inbound listener on a shared proxy. The destination properties are derived
-  // from the endpoint metadata instead of the proxy bootstrap.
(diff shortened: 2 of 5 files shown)


source/extensions/filters/http/istio_stats/config.proto
```diff
@@ -111,4 +111,10 @@ message PluginConfig {
 
   // Proxy deployment type.
   Reporter reporter = 10;
+
+  // Metric expiry duration. Defaults to 12h. Set to 0 to disable the metric expiry.
+  google.protobuf.Duration expiry_duration = 11;
+
+  // Metric expiry graceful drain duration. Defaults to 5m. No-op if the metric expiry is disabled.
+  google.protobuf.Duration graceful_drain_duration = 12;
```
lei-tang, line 119: - What does draining do?
- As graceful_drain_duration is used as drain_interval_ms,  should graceful_drain_duration be called graceful_drain_interval?

source/extensions/filters/http/istio_stats/config.proto
```diff
@@ -111,4 +111,10 @@ message PluginConfig {
 
   // Proxy deployment type.
   Reporter reporter = 10;
+
+  // Metric expiry duration. Defaults to 12h. Set to 0 to disable the metric expiry.
+  google.protobuf.Duration expiry_duration = 11;
```
lei-tang, line 116: - Upon expiration, what happens for a metric?
- If expiry_duration is used to assign the value of rotate_interval_ms, should expiry_duration be called rotate_interval?

zirain: I'm thinking make this default to `0` in the first release, so that the behavior is same as prev  release.
We can change the default value later, WDYT?

zirain:
what about short-lived metrics?

source/extensions/filters/http/istio_stats/config.proto
kyessenov: @lei-tang I'll add more code comments to explain how this works.
@zirain That would be fine.

source/extensions/filters/http/istio_stats/config.proto
kyessenov: This is not HTTP drain, this is metrics drain. See the latest code comments.

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -457,6 +458,73 @@ struct MetricOverrides : public Logger::Loggable<Logger::Id::filter>,
   absl::flat_hash_map<std::string, uint32_t> expression_ids_;
 };
 
+// Self-managed scope with active rotation. Envoy stats scope controls the
+// lifetime of the individual metrics. Because the scope is attached to xDS
+// resources, metrics with data derived from the requests can accumulate and
+// grow indefinitely for long-living xDS resources. To limit this growth,
+// this class implements a rotation mechanism, whereas a new scope is created
+// periodically to replace the current scope.
```
lei-tang, line 466: - What is the reason behind "a new scope is created periodically to replace the current scope"? As scope controls the lifetime of a metric, it seems an original scope (not a newly created scope) should be associated with a metric.
- If "a new scope is created periodically to replace the current scope", when will a metric expire and be cleared to save memory?

kyessenov: A metric (a set of symbols for the name and the tags) is cleared when its parent reference count reaches 0.
A new scope is needed to emit new metrics - this is happening in active listener handling requests.
An old scope is kept for a short window (both old and new are present then), as explained in the comment, before deleting it and all its metrics (that are not used elsewhere).

zirain:
> @zirain Short-lived listeners already rotate metrics. Listeners (or any xDS resource) that live <5s lose their stats altogether, which is a known problem with envoy.

can you link upstream issue if there's one.

zirain (commented):
Generally lgtm, it's disabled default, the risk is very low.

leave to @lei-tang for second review.

lei-tang (approved):

kyessenov:
> > @zirain Short-lived listeners already rotate metrics. Listeners (or any xDS resource) that live <5s lose their stats altogether, which is a known problem with envoy.
> 
> can you link upstream issue if there's one.

There's none since it almost never happens that xDS resource (cluster?) is added and removed within 5s.

zirain:
@kyessenov we need API to enable this for use who's using Telemetry API.

PR #4623, opened 2023-04-28
stats: fix destination service name for service entries

kyessenov:
Fixes https://github.com/istio/istio/issues/44602


lei-tang (approved):

zirain:
need a test for this

PR #4674, opened 2023-05-22
updates to workload API per design doc

GregHanson:

related to https://github.com/istio/istio/pull/44199 and https://github.com/istio/ztunnel/pull/486

contains implementations of https://docs.google.com/document/d/1V5wkeBHbLSLMzAMbwFlFZNHdZPyUEspG4lHbnB0UaCg/edit#heading=h.qpcs4a7wyicb


GregHanson:
placing hold for testing

source/extensions/common/workload_discovery/discovery.proto
```diff
@@ -65,10 +70,10 @@ message Workload {
   reserved "network";
   reserved 4;
 
-  reserved "protocol";
+  reserved "tunnel_protocol";
   reserved 5;
 
-  reserved "waypoint_addresses";
+  reserved "waypoint";
```
lei-tang, line 76: "protocol" and "waypoint_addresses" are existing fields. Will renaming them impact existing users for these fields? 
An alternative is to add "tunnel_protocol" and "waypoint" as new fields.

GregHanson: yes, this is a breaking change.  upgrade is not supported for ztunnel yet

kyessenov: No, you cannot do that since we already shipped this code. Never change field numbers or names, even in the experimental protos.

bleggett: Changing names is fine, I thought? Just indexes are the issue, from a proto perspective?

howardjohn: We have made it very clear in the development that the API is not stable and has no guarantees - protobuf or otherwise. I don't think there is a need to permanently complicate our APIs because of experimentation we did in the pre-alpha timeframe.

kyessenov: This is a different level of compatibility. Data wire compatibility is extremely hard to debug - and if it ever happens that a version of proxy mixes fields/numbers with another istiod, it's going to be impossible to debug. Unfortunately, Istio chose to rely on data-wire compatibility for CRDs which means that the breaking changes in CRDs are breaking on the wire.

@bleggett Yeah, field names are not important, but Envoy will emit a warning when it sees a new name, which bothers some.

kyessenov (approved):
The diff is small enough not to cause issues - but I don't think it's OK to re-number fields ever. Just use new numbers.

GregHanson:
@kyessenov this is the more up-to-date proto based on the design doc

source/extensions/common/workload_discovery/discovery.proto
```diff
@@ -27,64 +27,152 @@ option go_package = "test/envoye2e/workloadapi";
...
+    Workload workload = 1;
+    // Service represents a service - a group of workloads that can be accessed together.
+    Service service = 2;
+  }
+}
+
+// Service represents a service - a group of workloads that can be accessed together.
+// The primary key is "namespace/hostname".
+// Secondary (alias) keys are the unique `network/IP` pairs that the service can be reached at.
+message Service {
```
kyessenov, line 64: We don't need this here, delete this proto and make field 2 "service" reserved.

source/extensions/common/workload_discovery/discovery.proto
```diff
@@ -27,64 +27,152 @@ option go_package = "test/envoye2e/workloadapi";
...
+  // TODO: when this is repeated, update xds primary key from network/IP to network/UID
   bytes address = 3;
 
+  // The hostname for the workload to be resolved by the ztunnel.
+  // DNS queries are sent on-demand by default.
+  // If the resolved DNS query has several endpoints, the request will be forwarded
+  // to the first response.
+  // Each workload must have at least either an address or hostname; not both.
+  // TODO: support this field
+  string hostname = 21;
```
kyessenov, line 114: All these fields are not needed and just cause overhead. Can you just reserve their fields to avoid warnings and not define them?

source/extensions/common/workload_discovery/discovery.proto
GregHanson: 👍 I wasn't sure that notation would work in a `oneOf` block

source/extensions/common/workload_discovery/api.cc
```diff
@@ -145,9 +157,15 @@ class WorkloadMetadataProviderImpl : public WorkloadMetadataProvider, public Sin
                         const std::string&) override {
       AddressIndexSharedPtr added = std::make_shared<AddressIndex>();
       for (const auto& resource : added_resources) {
-        const auto& workload =
-            dynamic_cast<const istio::workload::Workload&>(resource.get().resource());
-        added->emplace(workload.address(), convert(workload));
+        const auto& address =
+            dynamic_cast<const istio::workload::Address&>(resource.get().resource());
+        switch (address.type()) {
+        case istio::workload::Service:
```
kyessenov, line 163: add default clause instead.

source/extensions/common/workload_discovery/api.cc
```diff
@@ -31,9 +31,15 @@
 namespace Envoy::Extensions::Common::WorkloadDiscovery {
 
 namespace {
-Istio::Common::WorkloadMetadataObject convert(const istio::workload::Workload& workload) {
+Istio::Common::WorkloadMetadataObject convert(const istio::workload::Address& address) {
+  switch (address.type()) {
+    case istio::workload::Service:
+      return;
```
kyessenov, line 37: This is incorrect, it still returns something. Maybe put ASSERT to ensure to ensure the type is right.

source/extensions/common/workload_discovery/discovery.proto
kyessenov: Experiments can never break production. It's ok to make changes but breaking data plane compatibility (running old Envoy with new istiod) should not be allowed.

source/extensions/common/workload_discovery/discovery.proto
kyessenov: I think you can remove oneof safely as long as the server doesn't set them both (which it shouldn't).

kdorosh (approved):

PR #4691, opened 2023-05-30
update workload api to use repeated addresses

GregHanson:

updates to use primary key as network/uid since addresses are repeated

related ztunnel PR: https://github.com/istio/ztunnel/pull/529
related control plane PR: https://github.com/istio/istio/pull/45170


source/extensions/common/workload_discovery/api.cc
```diff
@@ -138,7 +138,9 @@ class WorkloadMetadataProviderImpl : public WorkloadMetadataProvider, public Sin
             dynamic_cast<const istio::workload::Address&>(resource.get().resource());
         switch (address.type_case()) {
         case istio::workload::Address::kWorkload:
-          index->emplace(address.workload().address(), convert(address.workload()));
+          for (const auto& addr : address.workload().addresses()) {
+            index->emplace(addr, convert(address.workload()));
```
kyessenov, line 142: Doesn't this prevent lookup-by-address for the on-demand mode?
The xDS resource key must be the lookup key.

source/extensions/common/workload_discovery/api.cc
```diff
@@ -156,7 +158,9 @@ class WorkloadMetadataProviderImpl : public WorkloadMetadataProvider, public Sin
             dynamic_cast<const istio::workload::Address&>(resource.get().resource());
         switch (address.type_case()) {
         case istio::workload::Address::kWorkload:
-          added->emplace(address.workload().address(), convert(address.workload()));
+          for (const auto& addr : address.workload().addresses()) {
+            added->emplace(addr, convert(address.workload()));
```
kyessenov, line 162: You should save convert() into temporary instead of converting in a loop.

source/extensions/common/workload_discovery/api.cc
kdorosh: xDS supports primary and alias keys: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#resource-names

we can still lookup on demand by any of the alias keys (the IP addresses) but the primary key is a single primary name we can depend on (since which IP address to use in e.g. dual stack is confusing)

kdorosh (approved):
lgtm

source/extensions/common/workload_discovery/api.cc
kyessenov: Are you sure about that? Last time I looked at the code, if a client requests resource X it has to receive resource X with possible alias Y. If a client request Y, then it gets Y with an alias X. There's a lot of diffing logic that may not handle a different key from what is requested.

source/extensions/common/workload_discovery/api.cc
```diff
@@ -137,9 +137,14 @@ class WorkloadMetadataProviderImpl : public WorkloadMetadataProvider, public Sin
         const auto& address =
             dynamic_cast<const istio::workload::Address&>(resource.get().resource());
         switch (address.type_case()) {
-        case istio::workload::Address::kWorkload:
-          index->emplace(address.workload().address(), convert(address.workload()));
+        case istio::workload::Address::kWorkload: {
+          const auto& metadata = convert(address.workload());
+          index->emplace(address.workload().uid(), metadata);
```
kyessenov, line 142: If UID is the resource key, then you need to update line 121 to refer to "uid" field instead of "address".

kyessenov (changes requested):

source/extensions/common/workload_discovery/api.cc
kdorosh: it's my understanding of @howardjohn's proposal here https://github.com/istio/istio/pull/45170#discussion_r1207338408

we still need to minorly update ztunnel / istio to handle this, but this is the plan

kyessenov: It could be a limitation of the implementation right now. In any case, we're not doing on-demand right now so this can be deferred.

source/extensions/common/workload_discovery/api.cc
kyessenov: This might be a problem with the structure of the proto. The key must be a top-level field I think.

GregHanson: how is this working at all right now after https://github.com/istio/proxy/pull/4674, there is no longer any field named `address` in Address, Service or Workload

kyessenov: It's using some reflection I think so it evaluates to nothing. That means the entire collection is collapsed to one resource.

kdorosh: > how is this working at all right now after https://github.com/istio/proxy/pull/4674, there is no longer any field named address in Address, Service or Workload

technically the `address` field still exists on the `Workload`, and in this PR we change `address` to `addresses` on the `Workload`. that said, the subscription was updated in https://github.com/istio/proxy/pull/4674 to watch `Address`es instead of `Workload`s, which don't have an `address` field. so the last PR should have broken the subscription..

---

agree that we should update this to a field in the top-level proto.. it gets used when we try to get resource name. see https://github.com/envoyproxy/envoy/blob/08b7a8922702b49c0f2191ee58a48a4850206a35/source/common/config/opaque_resource_decoder_impl.h#L27-L29 and https://github.com/GregHanson/proxy/blob/b0b9e0436a4be19dad0304a703a9fbd48c17b120/source/extensions/common/workload_discovery/api.cc#L126

@howardjohn any appetite for moving UID up a level and perhaps even renaming to "name" (which is what envoy has been using, and also needs to be globally unique per envoy precedent)

howardjohn: I may be misunderstanding the code here but can we just override `getResourceName` to not use the default `Envoy::Config::getResourceName` which calls `MessageUtil::getStringField` and instead use our own implementation?

It is not ideal to make our ~permanent API worse to appease to envoy limitations if we can avoid it.

Alternatively, why do we need to subscribe to Address in envoy? Don't we only need Workload and not `Workload|Service`?

Istiod can serve both

kyessenov: The problem is that xDS requires resources to be "named", so it's not an issue with Envoy but the proto itself.
Besides, it's much easier to change WDS than Envoy so it makes sense to follow conventions if there is no good reason not to. I'd try to avoid creating two separate xDSes as well, that's doubling the cost of serving when you have more than one xDS server (proxies, federation).

GregHanson: > Alternatively, why do we need to subscribe to Address in envoy? Don't we only need Workload and not Workload|Service? Istiod can serve both

I thought I addressed something similar here: https://github.com/istio/ztunnel/pull/537#issue-1736861482
control plane only serves Workload|Service wrapped in Address, never the Workload|Service directly

howardjohn: It is named, its just named in a top level field.

I am not sure what you mean about 2 separate XDS's. A client can subscribe to what they want, they would not subscribe to both

kyessenov: What I mean is that if service is set in oneof, there's no way to identify the resource name. That's not compatible with xDS. Why are two very different protos under one umbrella anyways? If we have a problem with Envoy, the same will likely happen with other clients (gRPC?) so it might be better to split them into xDSes everywhere.

howardjohn: > What I mean is that if service is set in oneof, there's no way to identify the resource name. 

Service also has a name

>  Why are two very different protos under one umbrella anyways?

Its strictly to allow a somewhat niche case of needing to on-demand lookup an IP to find info about it, where the IP could be a service or workload. 

> so it might be better to split them into xDSes everywhere.

There should be 3 - merged, workload, service. Client can subscribe to what they want. In envoy we should subscribe to workloads only

kyessenov: oneof is not exhaustive in protobuf, it's legal to have none set. I think this is easily solved by adding name to a top-level proto.

howardjohn: `string` can be unset to, not sure why that is an issue. I do not get why we want to duplicate a field for the 3rd time (its already in the type and in `Resource` wrapper) when we don't need to. Envoy shouldn't even subscribe to Address

GregHanson: @kyessenov @howardjohn modified proxy to only request workload types and updated my control plane PR to handle it, PTAL

source/extensions/common/workload_discovery/discovery.proto
```diff
@@ -58,8 +58,7 @@ message Service {
 }
 
 message Workload {
-  reserved "uid";
-  reserved 20;
+  string uid = 20;
```
kyessenov, line 61: Could you clean up before submitting by removing unused protos (service and address)

kyessenov (approved):

kyessenov (approved):

source/extensions/common/workload_discovery/discovery.proto
GregHanson: done

kyessenov (approved):

PR #4728, opened 2023-06-16
ci: fix glibc error

zirain:
this's a temporary solution to fix the build breakage


ericvn:
Hopefully fixed when https://github.com/istio/tools/pull/2514 and https://github.com/istio/tools/pull/2515 get into the build-tools-proxy image.

PR #4757, opened 2023-06-29
infra: allow better proxy override

howardjohn:
See comments in code for rationale

### DIFF
--- a/prow/proxy-common.inc
+++ b/prow/proxy-common.inc
@@ -35,11 +35,22 @@ export BAZEL_BUILD_ARGS="--local_ram_resources=12288 --local_cpu_resources=8 --v
 
 # Override envoy.
 if [[ "${ENVOY_REPOSITORY:-}" && "${ENVOY_PREFIX:-}" ]]; then
+  # Legacy path, keep around for a few releases to allow folks to migrate
+  # The other path is preferred as it uses an API intended for programmatic access and allows fine-grained access token usage.
   TMP_DIR=$(mktemp -d -t envoy-XXXXXXXXXX)
   trap 'rm -rf ${TMP_DIR:?}' EXIT
   ENVOY_SHA="${ENVOY_SHA:-$(grep -Pom1 "^ENVOY_SHA = \"\K[a-zA-Z0-9]{40}" "$WORKSPACE")}"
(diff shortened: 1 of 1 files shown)


prow/proxy-common.inc
```diff
@@ -35,11 +35,22 @@ export BAZEL_BUILD_ARGS="--local_ram_resources=12288 --local_cpu_resources=8 --v
...
   trap 'rm -rf ${TMP_DIR:?}' EXIT
   ENVOY_SHA="${ENVOY_SHA:-$(grep -Pom1 "^ENVOY_SHA = \"\K[a-zA-Z0-9]{40}" "$WORKSPACE")}"
   BAZEL_BUILD_ARGS="${BAZEL_BUILD_ARGS} --override_repository=envoy=${TMP_DIR}/${ENVOY_PREFIX}-${ENVOY_SHA}"
   curl -nsSfL "${ENVOY_REPOSITORY}/archive/${ENVOY_SHA}.tar.gz" | tar -C "${TMP_DIR}" -xz
+elif [[ "${ENVOY_REPOSITORY:-}" ]]; then
+  TMP_DIR=$(mktemp -d -t envoy-XXXXXXXXXX)
+  trap 'rm -rf ${TMP_DIR:?}' EXIT
+  ENVOY_SHA="${ENVOY_SHA:-$(grep -Pom1 "^ENVOY_SHA = \"\K[a-zA-Z0-9]{40}" "$WORKSPACE")}"
+  BAZEL_BUILD_ARGS="${BAZEL_BUILD_ARGS} --override_repository=envoy=${TMP_DIR}"
+  # User passes a URl like https://github.com/foo/bar, we translate it to https://api.github.com/foo/bar/repos.
```
kyessenov, line 50: bazel fetches a ton of stuff using github URLs. does this change make a difference with RBE?

howardjohn: When this is used we are only ever overriding Envoy, and we override it to a local path. I don't know how RBE handles local path overrides... but this is only used in private builds where we disable RBE as well.

The other URLs we don't care about since they are all public. In theory we could have a CVE where we need to pull in something other than envoy that is private, but I doubt it will happen in practice - we only get early info from Envoy typically. 

So I agree this is not super robust but I think its also good enough for what we use it for?

kyessenov: Ok, I have no problem with it. Might be better to use git clone --depth 1 for private repo clone.

kyessenov (approved):

PR #4768, opened 2023-06-30
Disable CPU limit

howardjohn:
Currently we give each job limit=64 CPUs but then tell Bazel to limit to 8 cores. I think it will be much faster if we remove the limit..?


kyessenov:
Yeah, not sure why it's doing limiting twice.

kyessenov (approved):

PR #4789, opened 2023-07-07
metadata_exchange: combine into native implementation

kyessenov:
Implements HTTP metadata exchange as a native filter on top of the existing CONNECT baggage filter.

The main benefit is a smooth migration from Istio-specific headers via fallback mechanisms.

The recommended setting for 1.19 sidecars would be:
- for discovery: use WDS first, HTTP MX headers second. 
- for propagation: use HTTP MX headers.
- allow turning-off WDS in case of instability (as it's coming from Ambient alpha code)

The recommended setting for 1.20 sidecars would be:
- for discovery: use WDS.
- for propagation: none.

The recommended setting for 1.19+ waypoints would be:
- for discovery: use WDS, HTTP CONNECT second;
- for propagation: none
- turn off response CONNECT baggage since it is likely not used and is a waste


zirain:
so this means there's connect filter will replace metadata_exchange in the future?

kyessenov:
> so this means there's connect filter will replace metadata_exchange in the future?

Yeah, seems like it's best to combine all the reasonable ways into one filter (HTTP MX, baggage, and WDS). It's impossible to extend Wasm version to support WDS.

zirain:
+1 for using a new native filter replace the wasm one.

will you make this landing in 1.19, it will cut branch at end of July, see https://github.com/istio/istio/wiki/Istio-Release-1.19

zirain:
another thought: can we add an option to propagate metadata into filterstate, with that it will be easy to get metadata in accesslog

@kyessenov WDYT?

kyessenov:
@zirain metadata is already stored in the filter state in many cases, but it's a bit messy:
- istio_stats can call WDS directly if it doesn't find it
- ambient doesn't have upstream
- TCP/HTTP MX store them in different filter states (upstream vs downstream).

Ideally, I'd like standard Otel logging work for ambient, hence we need a dedicated metadata extension that's not coupled with stats/stackdriver.

zirain:
> @zirain metadata is already stored in the filter state in many cases, but it's a bit messy:
> 
> * istio_stats can call WDS directly if it doesn't find it
> * ambient doesn't have upstream
> * TCP/HTTP MX store them in different filter states (upstream vs downstream).
> 
> Ideally, I'd like standard Otel logging work for ambient, hence we need a dedicated metadata extension that's not coupled with stats/stackdriver.

still use x-envoy-peer-metadata encoding by flatbuff?

kyessenov:
> still use x-envoy-peer-metadata encoding by flatbuff?

Yes, but it doesn't matter since native Envoy filters can reflect on it with CEL.

kyessenov:
FYI @costinm . This should be ready for review.

zirain:
> > still use x-envoy-peer-metadata encoding by flatbuff?
> 
> Yes, but it doesn't matter since native Envoy filters can reflect on it with CEL.

can you add an example about how to use?

kyessenov:
@zirain 

I think `%CEL(filter_state["wasm.downstream_peer"].namespace)% should work.

source/extensions/filters/http/peer_metadata/config.proto
```diff
@@ -0,0 +1,72 @@
...
+  // Use Istio HTTP metadata exchange headers. Removes these headers if found.
+  message IstioHeaders {
+  }
+
+  // An exhaustive list of the derivation methods.
+  message DiscoveryMethod {
+    oneof method_specifier {
+      Baggage baggage = 1;
+      WorkloadDiscovery workload_discovery = 2;
+      IstioHeaders istio_headers = 3;
```
hzxuzhonghu, line 47: nit: from the pr desc, WorkloadDiscovery will be the choice in the future, why donot we need to set discovery method. I mean it can do this by default

kyessenov: WorkloadDiscovery is still alpha since it's a part of ambient so we need a way to disable it if it goes wrong.

kyessenov:
Ref: https://github.com/istio/istio/issues/43937

source/extensions/filters/http/peer_metadata/config.proto
```diff
@@ -0,0 +1,72 @@
...
+
+  // An exhaustive list of the metadata propagation methods.
+  message PropagationMethod {
+    oneof method_specifier {
+      IstioHeaders istio_headers = 1;
+    }
+  }
+
+  // Downstream injection of the metadata via a response header.
+  repeated PropagationMethod downstream_propagation = 3;
```
zirain, line 65: why this repeated?

source/extensions/filters/http/peer_metadata/config.proto
hzxuzhonghu: IC, thanks

source/extensions/filters/http/peer_metadata/config.proto
kyessenov: There's a discussion to use JWT tokens for header-based propagation, so we need another option.
We need more than one for migration, since we may not know which versions is supported by the peer.

kyessenov:
Ping for review @zirain.

zirain (approved):
lgtm, add DNM for someone want to take a review.

also, we should add more test in istio repo to make sure everything works fine.

extensions/common/metadata_object.cc
```diff
@@ -231,7 +231,8 @@ flatbuffers::DetachedBuffer convertWorkloadMetadataToFlatNode(const WorkloadMeta
   node.add_labels(labels_offset);
   auto data = node.Finish();
   fbb.Finish(data);
-  return fbb.Release();
+  auto fb = fbb.Release();
+  return std::string(reinterpret_cast<const char*>(fb.data()), fb.size());
```
edtanous, line 235: nit, can you use bit_cast here instead?

source/extensions/filters/http/peer_metadata/filter.cc
```diff
@@ -0,0 +1,344 @@
...
+absl::optional<PeerInfo> MXMethod::lookup(absl::string_view id, absl::string_view value) const {
+  // This code is copied from:
+  // https://github.com/istio/proxy/blob/release-1.18/extensions/metadata_exchange/plugin.cc#L116
+  if (max_peer_cache_size_ > 0 && !id.empty()) {
+    auto it = cache_.find(id);
+    if (it != cache_.end()) {
+      return it->second;
+    }
+  }
+  const auto bytes = Base64::decodeWithoutPadding(value);
```
edtanous, line 171: Why "WithoutPadding" here?

source/extensions/filters/http/peer_metadata/filter.cc
```diff
@@ -0,0 +1,344 @@
...
+    return {};
+  }
+  const auto fb = ::Wasm::Common::extractNodeFlatBufferFromStruct(metadata);
+  std::string out(reinterpret_cast<const char*>(fb.data()), fb.size());
+  if (max_peer_cache_size_ > 0 && !id.empty()) {
+    // do not let the cache grow beyond max cache size.
+    if (static_cast<uint32_t>(cache_.size()) > max_peer_cache_size_) {
+      cache_.erase(cache_.begin(), std::next(cache_.begin(), max_peer_cache_size_ / 4));
+    }
+    cache_.emplace(id, out);
```
edtanous, line 183: Not sure if this matters or not, but if id is already in the cache, this might unnecessarily cause the cache to clear if id already exists in the cache.

Maybe:
```cpp
if(!id.empty()){
  auto id_it = cache_.find(id);
  if (id_it != cache.end()){
      id_it->second = out;
      return;
  }
  if (max_peer_cache_size_ > 0 && !id.empty()) {
    // do not let the cache grow beyond max cache size.
    if (static_cast<uint32_t>(cache_.size()) > max_peer_cache_size_) {
      cache_.erase(cache_.begin(), std::next(cache_.begin(), max_peer_cache_size_ / 4));
    }
    cache_.emplace(id, out);
}
return out;
```

Might be bikeshedding.  Up to you if you want to change it.

source/extensions/filters/http/peer_metadata/filter.cc
kyessenov: For backwards compatibility with "Wasm" implementation of the same logic [here](https://github.com/istio/proxy/blob/master/extensions/metadata_exchange/plugin.cc#L133). Without padding is more forgiving so we have to keep it lax like that.

source/extensions/filters/http/peer_metadata/filter.cc
kyessenov: Same reason as above: backwards compatibility https://github.com/istio/proxy/blob/master/extensions/metadata_exchange/plugin.cc#L133. There's no proper automated test for this caching logic so I am hesitant to touch it in any way.

extensions/common/metadata_object.cc
kyessenov: We're still in C++17 here, but we can switch to C++20 shortly.

kyessenov:
@briansonnenberg @edtanous or anyone else let me know if you have further comments, I will create a follow-up to respond. Need to have a test in another repo, so letting it merge here.

kyessenov:
/retest
CustomAccessLog appears to be another flake

source/extensions/filters/http/peer_metadata/filter.cc
kyessenov: The problem is thread management, will send a follow-up.

PR #4835, opened 2023-07-25
metadata_exchange: stop waiting for data when upstream closes

blancobrenden:
This fixes a bug found in a test environment when the upstream TCP server is starting up and the server sidecar receives a connection too early (possible if there is no k8s startupProbe defined, for instance). In such a case, the server sidecar should reset the TLS connection (2) once the SNI is parsed. This should normally cause the client connection to close as well, but in some cases the connection (1) would hang instead.
client -(1)-> sidecar -(2)-> sidecar -> -(3)-> server

This was observed only with TLS1.3 connections. With TLS1.2, the upstream reset would abort filter processing during the handshake, whereas with TLS1.3 the metadata_exchange filter was allowed to send data into the upstream buffer and move its state machine forward.

The reproducer included in this PR uses a slightly different mechanism to reliably trigger the issue, since the real-world reproduction is timing/latency dependent.

### DIFF
--- a/source/extensions/filters/network/metadata_exchange/metadata_exchange.cc
+++ b/source/extensions/filters/network/metadata_exchange/metadata_exchange.cc
@@ -107,7 +107,7 @@ Network::FilterStatus MetadataExchangeFilter::onData(Buffer::Instance& data, boo
         // Upstream has entered a half-closed state, and will be sending no more data.
         // Since this plugin would expect additional headers, but none is forthcoming,
         // do not block the tcp_proxy downstream of us from draining the buffer.
-        ENVOY_LOG(warn, "Upstream closed early, aborting istio-peer-exchange");
+        ENVOY_LOG(debug, "Upstream closed early, aborting istio-peer-exchange");
         conn_state_ = Invalid;
         return Network::FilterStatus::Continue;
       }
--- a/source/extensions/filters/network/metadata_exchange/metadata_exchange.cc
+++ b/source/extensions/filters/network/metadata_exchange/metadata_exchange.cc
@@ -69,7 +69,7 @@ MetadataExchangeConfig::MetadataExchangeConfig(const std::string& stat_prefix,
     : scope_(scope), stat_prefix_(stat_prefix), protocol_(protocol),
       filter_direction_(filter_direction), stats_(generateStats(stat_prefix, scope)) {}
 
-Network::FilterStatus MetadataExchangeFilter::onData(Buffer::Instance& data, bool) {
+Network::FilterStatus MetadataExchangeFilter::onData(Buffer::Instance& data, bool end_stream) {
   switch (conn_state_) {
   case Invalid:
     FALLTHRU;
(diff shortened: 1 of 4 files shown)


zirain (commented):
lgtm, defer to @kyessenov for second review.

source/extensions/filters/network/metadata_exchange/metadata_exchange.cc
```diff
@@ -103,6 +103,14 @@ Network::FilterStatus MetadataExchangeFilter::onData(Buffer::Instance& data, boo
   case NeedMoreDataInitialHeader: {
     tryReadInitialProxyHeader(data);
     if (conn_state_ == NeedMoreDataInitialHeader) {
+      if (end_stream) {
+        // Upstream has entered a half-closed state, and will be sending no more data.
+        // Since this plugin would expect additional headers, but none is forthcoming,
+        // do not block the tcp_proxy downstream of us from draining the buffer.
+        ENVOY_LOG(warn, "Upstream closed early, aborting istio-peer-exchange");
```
kyessenov, line 110: This should be debug or trace level.

kyessenov (approved):
Great find! The change looks good to me. Does the difference with TLS 1.3 have to do with [0-RTT](https://datatracker.ietf.org/doc/html/rfc8446#section-2.3)?

source/extensions/filters/network/metadata_exchange/metadata_exchange.cc
kyessenov: Please fix log level to be less noisy - will remove the label then.

blancobrenden: Yes, I believe the difference is due to the 0-RTT behavior of TLS 1.3 changing the timing of the this extension relative to the handshake. Specifically, with 1.2, mx is never invoked because the upstream always closes early with a handshake failure.
Regarding the log level, I can certainly change that and will do so. My rationale had been that this failure is rare and would happen only when one does not define something like a startupProbe in the pod spec, and so users that debugged this may have been more likely to see the warning were it to happen, as it did for us.

kyessenov: The log level is driven by Envoy conventions, which usually limits per-connection and per-request lines to be <= debug, to avoid overwhelming the log service.

If I understand correctly, this issue can happen with any mis-configured mesh using TLS 1.3. Any client sidecar connection that is accepted and then half-closed would cause a connection leak from the app to the client sidecar. This may happen when trying to use mTLS to external servers (out-of-mesh). If it's truly 0-RTT timing then the server doesn't even have to be TLS to cause an issue.

blancobrenden: Misconfigured external servers may trigger this issue, although I believe we only get to the problematic point in the state machine if the server advertises the correct magic string in the ALPN sent back in the first packet.
Thank you for the review!

PR #4840, opened 2023-07-26
bump envoy sha256 to be consistent

edieYoung:
the envoy sha256 doesn't match, bump it to be the correct one:
```
 wget https://github.com/envoyproxy/envoy/archive/da77eab7d2f758ac4380467775ec29e47ff28205.tar.gz && sha256sum da77eab7d2f758ac4380467775ec29e47ff28205.tar.gz

593055446f358fe3af389d8fef191b9e011b9ef05689437745e210c1600fe897
```
https://github.com/istio/proxy/commit/5382f1d9645b73f7a8aff075a966200a6b9f7169#r122780476
@howardjohn


howardjohn (commented):
Hmm. We can do this but not sure the end result. The point of 1.17.5 patch branch is that it is building 1.17.5 specifically. If we merge changes to it, we end up with " not really 1.17.5 but kind of close"

Obviously we messed this up and shouldn't do that again (will update the private CVE process), but not sure the best path forward.

Do you expect to update istio/istio as well?

jacob-delgado:
We should leave this alone and change the ENVOY_ORG and ENVOY_REPO to istio/envoy instead if we followed the process. We should not change the sha.

edieYoung:
I don't know what's the best process neither, but this inconsistency might break our Istio upgrade to 1.17.5(we have custom envoy build based on Istio) :(

edieYoung:
> Do you expect to update istio/istio as well?

Yes, I think so, you mean here right? https://github.com/istio/istio/blame/1.17.5/istio.deps#L7

edieYoung:
> Hmm. We can do this but not sure the end result. The point of 1.17.5 patch branch is that it is building 1.17.5 specifically. If we merge changes to it, we end up with " not really 1.17.5 but kind of close"

I guess we probably will end up a new patch release.

PR #4865, opened 2023-08-04
Fix skipping ALPN overwriting

jewertow:

I noticed in logs `override with 3 ALPNs`, while cluster had set `istio.alpn_override: false`, so I think that the previous implementation was not correct.

**Which issue this PR fixes**
fixes https://github.com/istio/istio/issues/40680


jewertow:
`test_proxy` check is failing with the following error message:
```
Checking file format ...
Source file(s) not formatted:
source/extensions/filters/http/alpn/alpn_test.cc
Makefile.core.mk:133: recipe for target 'lint' failed
make: *** [lint] Error 1
```
I cannot fix it, because when I run `clang-format -style=Google -i source/extenstions/filters/http/alpn/alpn_filter.cc`, it changes the entire file. Does anyone know how can I get to know what's exactly wrong with the current code format?

howardjohn (approved):

zirain (approved):

PR #4876, opened 2023-08-07
metadata exchange: fix cache data race

kyessenov:
Follow up to https://github.com/istio/proxy/pull/4789 to address cache concurrency issue. The cache should be thread local. Adds an integration test to exercise concurrent access.


kyessenov:
CC @edtanous

lei-tang (approved):

PR #4885, opened 2023-08-09
metadata_exchange: fix layered TCP/HTTP write conflict

kyessenov:
It appears Istio enables both HTTP and TCP MX filters at the same time in some cases (e.g. for dry run RBAC Stackdriver test). When that happens, there's a data write conflict:
- TCP MX writes "peer not found" at the connection life span, because ALPN is not set;
- HTTP MX writes actual peer ID at the request life span for the same key.

Because they use the same key, the second write fails in the native implementation of HTTP MX. On the other hand, Wasm HTTP MX uses a controversial trick where it overwrites the TCP MX value with HTTP MX value that makes this problem hidden. That trick causes different issues because now requests are accidentally sharing their filter state (it's per connection, not per request anymore).

The fix is to prevent sharing in this case by 1) making TCP MX write a different "peer not found" key, 2) gracefully accepting TCP MX over HTTP MX when they are truly duplicate (it's not clear if that ever happens), since transport level metadata is likely more trusted.

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -347,7 +347,7 @@ PeerNodeInfo::PeerNodeInfo(const std::string_view peer_metadata_id_key,
                            const std::string_view peer_metadata_key) {
   // Attempt to read from filter_state first.
   found_ = getValue({peer_metadata_id_key}, &peer_id_);
-  if (found_ && peer_id_ != kMetadataNotFoundValue) {
+  if (found_) {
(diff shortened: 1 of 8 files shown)


zirain:
can you add a test for this?

kyessenov:
@zirain Added a test that makes sure TCP MX is no-op with HTTP MX when there's no mTLS+ALPN.

zirain (approved):

PR #4891, opened 2023-08-11
build: remove wasm metadata exchange

kyessenov:
Code will move to istio-ecosystem/wasm-extensions for archiving.


zirain (approved):

hzxuzhonghu:
Will this proy be landed on 1.20, then we will already repalce it with native filter

zirain:
> Will this proy be landed on 1.20, then we will already repalce it with native filter

see https://github.com/istio/istio/pull/46498

ericvn:
See https://github.com/istio/istio/pull/46517 for a change since this is no longer being built.

PR #4895, opened 2023-08-16
do not send mx headers for services out of mesh

GregHanson:

add logic in proxy to strip `x-envoy-peer-metadata` and `x-envoy-peer-metadata-id` headers for request to services outside of the mesh.  These headers contain values such as IP's and istio version which block some users from adopting istio due to security audits

 - outside of the mesh is detected on best effort basis
 - behavior is opt-in to prevent any breaking changes
 - affects HTTP traffic + traffic directed to PassthroughCluster or non-eds clusters

detection logic is limited to settings exposed in Envoy XDS cluster API

### DIFF
--- a/source/extensions/filters/http/peer_metadata/config.proto
+++ b/source/extensions/filters/http/peer_metadata/config.proto
@@ -71,4 +71,7 @@ message Config {
 
   // True to enable sharing with the upstream.
   bool shared_with_upstream = 5;
+
+  // Strip x-envoy-peer-metadata and x-envoy-peer-metadata-id headers on HTTP requests to services outside the mesh
+  bool strip_headers = 6;
 }
--- a/source/extensions/filters/http/peer_metadata/config.proto
+++ b/source/extensions/filters/http/peer_metadata/config.proto
@@ -37,6 +37,9 @@ message Config {
 
   // This method uses Istio HTTP metadata exchange headers, e.g. `x-envoy-peer-metadata`. Removes these headers if found.
   message IstioHeaders {
+    // Strip x-envoy-peer-metadata and x-envoy-peer-metadata-id headers on HTTP requests to services outside the mesh.
+    // Detects upstream clusters with `istio` and `external` filter metadata fields
+    bool skip_external_clusters = 1;
   }
 
   // An exhaustive list of the derivation methods.
@@ -71,7 +74,4 @@ message Config {
 
   // True to enable sharing with the upstream.
   bool shared_with_upstream = 5;
-
-  // Strip x-envoy-peer-metadata and x-envoy-peer-metadata-id headers on HTTP requests to services outside the mesh
-  bool strip_headers = 6;
 }
--- a/source/extensions/filters/http/peer_metadata/filter.cc
+++ b/source/extensions/filters/http/peer_metadata/filter.cc
@@ -349,13 +349,8 @@ bool MXPropagationMethod::skipMXHeaders(const StreamInfo::StreamInfo& info) cons
     if (it != filter_metadata.end()) {
       const auto& skip_mx = it->second.fields().find("external");
       if (skip_mx != it->second.fields().end()) {
-        const auto skip_mx_value = skip_mx->second.bool_value();
-        if (skip_mx_value) {
-          return true;
-        }
+        return skip_mx->second.bool_value();
       }
-    } else {
-      return true;
     }
   }
   return false;
--- a/source/extensions/filters/http/peer_metadata/filter.cc
+++ b/source/extensions/filters/http/peer_metadata/filter.cc
@@ -354,6 +354,8 @@ bool MXPropagationMethod::skipMXHeaders(const StreamInfo::StreamInfo& info) cons
           return true;
         }
       }
+    } else {
+      return true;
     }
   }
   return false;
--- a/source/extensions/filters/http/peer_metadata/filter.cc
+++ b/source/extensions/filters/http/peer_metadata/filter.cc
@@ -346,8 +346,8 @@ bool MXPropagationMethod::skipMXHeaders(StreamInfo::StreamInfo& info) const {
     if (it != filter_metadata.end()) {
       const auto& skip_mx = it->second.fields().find("external");
       if (skip_mx != it->second.fields().end()) {
-        const auto skip_mx_value = skip_mx->second.string_value();
-        if (skip_mx_value == "true") {
+        const auto skip_mx_value = skip_mx->second.bool_value();
+        if (skip_mx_value) {
           return true;
         }
       }
--- a/source/extensions/filters/http/peer_metadata/filter.cc
+++ b/source/extensions/filters/http/peer_metadata/filter.cc
@@ -180,8 +180,11 @@ absl::optional<PeerInfo> MXMethod::lookup(absl::string_view id, absl::string_vie
 }
 
 MXPropagationMethod::MXPropagationMethod(
-    Server::Configuration::ServerFactoryContext& factory_context)
-    : id_(factory_context.localInfo().node().id()), value_(computeValue(factory_context)) {}
+    Server::Configuration::ServerFactoryContext& factory_context,
+    io::istio::http::peer_metadata::Config_IstioHeaders istioHeaders)
+    : id_(factory_context.localInfo().node().id()), value_(computeValue(factory_context)) {
+  skip_external_clusters_ = istioHeaders.skip_external_clusters();
+}
 
 std::string MXPropagationMethod::computeValue(
     Server::Configuration::ServerFactoryContext& factory_context) const {
@@ -195,14 +198,16 @@ std::string MXPropagationMethod::computeValue(
   return Base64::encode(metadata_bytes.data(), metadata_bytes.size());
 }
 
-void MXPropagationMethod::inject(Http::HeaderMap& headers) const {
-  headers.setReference(Headers::get().ExchangeMetadataHeaderId, id_);
-  headers.setReference(Headers::get().ExchangeMetadataHeader, value_);
+void MXPropagationMethod::inject(StreamInfo::StreamInfo& info, Http::HeaderMap& headers) const {
+  if (!skip_external_clusters_ || !(skip_external_clusters_ && skipMXHeaders(info))) {
+    headers.setReference(Headers::get().ExchangeMetadataHeaderId, id_);
+    headers.setReference(Headers::get().ExchangeMetadataHeader, value_);
+  }
 }
 
 FilterConfig::FilterConfig(const io::istio::http::peer_metadata::Config& config,
                            Server::Configuration::FactoryContext& factory_context)
-    : strip_headers_(config.strip_headers()), shared_with_upstream_(config.shared_with_upstream()),
+    : shared_with_upstream_(config.shared_with_upstream()),
       downstream_discovery_(
           buildDiscoveryMethods(config.downstream_discovery(), true, factory_context)),
       upstream_discovery_(
@@ -249,8 +254,8 @@ std::vector<PropagationMethodPtr> FilterConfig::buildPropagationMethods(
     switch (method.method_specifier_case()) {
     case io::istio::http::peer_metadata::Config::PropagationMethod::MethodSpecifierCase::
         kIstioHeaders:
-      methods.push_back(
-          std::make_unique<MXPropagationMethod>(factory_context.getServerFactoryContext()));
+      methods.push_back(std::make_unique<MXPropagationMethod>(
+          factory_context.getServerFactoryContext(), method.istio_headers()));
       break;
     default:
       break;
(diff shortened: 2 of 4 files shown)


GregHanson:
Current logic is toggled via env var on sidecar container. I think there is a valid use case for controlling this behavior on a per proxy basis to limit any potential breaking changes.

Alternative would be to add a new value to the config.proto which could be set via istiod. I'm not sure on the longevity of this fox, since switching to ambient will resolve any issues in the long term - so I'm not sure we want to add a field for this to the proto

source/extensions/filters/http/peer_metadata/filter.cc
```diff
@@ -210,7 +210,14 @@ FilterConfig::FilterConfig(const io::istio::http::peer_metadata::Config& config,
       downstream_propagation_(
           buildPropagationMethods(config.downstream_propagation(), factory_context)),
       upstream_propagation_(
-          buildPropagationMethods(config.upstream_propagation(), factory_context)) {}
+          buildPropagationMethods(config.upstream_propagation(), factory_context)) {
+  strip_headers_ = false;
+  if (const char* env_p = std::getenv("ISTIO_META_STRIP_HEADERS")) {
```
kyessenov, line 215: This should be a protobuf field IMHO for compliance with the rest of xDS. Envoy doesn't use environment variables for configuration.

GregHanson:
> can we do this in control plane like best effort infer svc mtls mode

@hzxuzhonghu maybe some of the detection logic, but not all?  Maybe a metadata field added to the cluster's filter_metadata that the proxy would check for?  The headers will still have to be stripped inside the proxy - since that is where they are added.

source/extensions/filters/http/peer_metadata/filter.cc
linsun: I am not sure about making this an official API, the consensus from upstream was this is a bug and we don't need to declare it in an API thus do it as an env var. cc a few others who may want to chime in.

@keithmattix @howardjohn @costinm

linsun:
> > can we do this in control plane like best effort infer svc mtls mode
> 
> @hzxuzhonghu maybe some of the detection logic, but not all? Maybe a metadata field added to the cluster's filter_metadata that the proxy would check for? The headers will still have to be stripped inside the proxy - since that is where they are added.

Given this is only a sidecar problems and only opt in, I am not convinced we want to change control plane xDS for this as that would have bigger impact.

source/extensions/filters/http/peer_metadata/filter.cc
kyessenov: Lin, this is data plane APIs, not user-facing APIs. xDS does not use environment variables normally, only in rare cases (backwards compatibility). This should be a field in xDS that's controlled by env variable in istiod.

GregHanson: @kyessenov @hzxuzhonghu a POC for istiod to append a new filter_metadata field for services marked as external, and sending a new field in the XDS config.proto. PTAL, draft PR here: https://github.com/istio/istio/pull/46579

linsun: @kyessenov @GregHanson thanks for the clarification! I had thought this is an env var user configures. Sounds like it is not!

linsun: Chatting with Greg a bit, I don't think pod label makes sense as first pass, our users want to do this globally.

source/extensions/filters/http/peer_metadata/config.proto
```diff
@@ -71,4 +71,7 @@ message Config {
 
   // True to enable sharing with the upstream.
   bool shared_with_upstream = 5;
+
+  // Strip x-envoy-peer-metadata and x-envoy-peer-metadata-id headers on HTTP requests to services outside the mesh
+  bool strip_headers = 6;
```
kyessenov, line 76: Can we move it to IstioHeaders message and call it "skip_external_clusters"? The comment should say it's detecting upstream clusters with "istio" "external" field.

kyessenov (changes requested):
Suggest moving the field into MX headers, since this doesn't apply to baggage or what not. Also, we need a unit test.

source/extensions/filters/http/peer_metadata/filter.cc
```diff
@@ -180,8 +180,11 @@ absl::optional<PeerInfo> MXMethod::lookup(absl::string_view id, absl::string_vie
 }
 
 MXPropagationMethod::MXPropagationMethod(
-    Server::Configuration::ServerFactoryContext& factory_context)
-    : id_(factory_context.localInfo().node().id()), value_(computeValue(factory_context)) {}
+    Server::Configuration::ServerFactoryContext& factory_context,
+    io::istio::http::peer_metadata::Config_IstioHeaders istioHeaders)
```
kyessenov, line 184: istio_headers per style

source/extensions/filters/http/peer_metadata/filter.cc
```diff
@@ -195,9 +198,11 @@ std::string MXPropagationMethod::computeValue(
   return Base64::encode(metadata_bytes.data(), metadata_bytes.size());
 }
 
-void MXPropagationMethod::inject(Http::HeaderMap& headers) const {
-  headers.setReference(Headers::get().ExchangeMetadataHeaderId, id_);
-  headers.setReference(Headers::get().ExchangeMetadataHeader, value_);
+void MXPropagationMethod::inject(StreamInfo::StreamInfo& info, Http::HeaderMap& headers) const {
+  if (!skip_external_clusters_ || !(skip_external_clusters_ && skipMXHeaders(info))) {
```
kyessenov, line 202: This is hard to read, suggest:

```
if (skip_external_clusters_) {
  if (skipMXHeaders(info)) {
    return;
  }
}
```

source/extensions/filters/http/peer_metadata/config.proto
```diff
@@ -37,6 +37,9 @@ message Config {
 
   // This method uses Istio HTTP metadata exchange headers, e.g. `x-envoy-peer-metadata`. Removes these headers if found.
   message IstioHeaders {
+    // Strip x-envoy-peer-metadata and x-envoy-peer-metadata-id headers on HTTP requests to services outside the mesh.
+    // Detects upstream clusters with `istio` and `external` filter metadata fields
```
kyessenov, line 41: And based on PASSTHROUGH name it appears?

source/extensions/filters/http/peer_metadata/filter.cc
```diff
@@ -323,13 +330,37 @@ void FilterConfig::setFilterState(StreamInfo::StreamInfo& info, bool downstream,
 
 Http::FilterHeadersStatus Filter::decodeHeaders(Http::RequestHeaderMap& headers, bool) {
   config_->discoverDownstream(decoder_callbacks_->streamInfo(), headers);
-  config_->injectUpstream(headers);
+  config_->injectUpstream(decoder_callbacks_->streamInfo(), headers);
   return Http::FilterHeadersStatus::Continue;
 }
 
+bool MXPropagationMethod::skipMXHeaders(StreamInfo::StreamInfo& info) const {
```
kyessenov, line 337: this can be `const StreamInfo&` I think

source/extensions/filters/http/peer_metadata/filter.h
```diff
@@ -74,20 +74,23 @@ class MXMethod : public DiscoveryMethod {
...
 };
 
 using PropagationMethodPtr = std::unique_ptr<PropagationMethod>;
 
 class MXPropagationMethod : public PropagationMethod {
 public:
-  MXPropagationMethod(Server::Configuration::ServerFactoryContext& factory_context);
-  void inject(Http::HeaderMap&) const override;
+  MXPropagationMethod(Server::Configuration::ServerFactoryContext& factory_context,
+                      io::istio::http::peer_metadata::Config_IstioHeaders);
```
kyessenov, line 85: `const IstioHeaders&` please

source/extensions/filters/http/peer_metadata/filter.cc
```diff
@@ -180,8 +180,11 @@ absl::optional<PeerInfo> MXMethod::lookup(absl::string_view id, absl::string_vie
 }
 
 MXPropagationMethod::MXPropagationMethod(
-    Server::Configuration::ServerFactoryContext& factory_context)
-    : id_(factory_context.localInfo().node().id()), value_(computeValue(factory_context)) {}
+    Server::Configuration::ServerFactoryContext& factory_context,
+    io::istio::http::peer_metadata::Config_IstioHeaders istioHeaders)
+    : id_(factory_context.localInfo().node().id()), value_(computeValue(factory_context)) {
+  skip_external_clusters_ = istioHeaders.skip_external_clusters();
```
kyessenov, line 186: initialize in the constructor like `id_`, then you can make it `const`

source/extensions/filters/http/peer_metadata/filter.h
```diff
@@ -74,20 +74,23 @@ class MXMethod : public DiscoveryMethod {
 class PropagationMethod {
 public:
   virtual ~PropagationMethod() = default;
-  virtual void inject(Http::HeaderMap&) const PURE;
+  virtual void inject(StreamInfo::StreamInfo&, Http::HeaderMap&) const PURE;
```
kyessenov, line 77: Suggest making this `const StreamInfo&` here and in `injectUpstream/Downstream`.

kyessenov (changes requested):
Logic is fine, please address c++ comments and add tests.

source/extensions/filters/http/peer_metadata/filter.cc
GregHanson: done

source/extensions/filters/http/peer_metadata/filter.cc
GregHanson: done

source/extensions/filters/http/peer_metadata/filter.cc
GregHanson: done

source/extensions/filters/http/peer_metadata/filter.h
GregHanson: done

source/extensions/filters/http/peer_metadata/filter.cc
GregHanson: done

source/extensions/filters/http/peer_metadata/filter.h
GregHanson: done

source/extensions/filters/http/peer_metadata/config.proto
GregHanson: I can either explicitly check for it here, or I can add the 
```
filter_metadata: 
  istio:
    external: true
```
to the passthrough cluster in the controlplane

source/extensions/filters/http/peer_metadata/filter.cc
```diff
@@ -323,13 +333,37 @@ void FilterConfig::setFilterState(StreamInfo::StreamInfo& info, bool downstream,
...
+    if (it != filter_metadata.end()) {
+      const auto& skip_mx = it->second.fields().find("external");
+      if (skip_mx != it->second.fields().end()) {
+        const auto skip_mx_value = skip_mx->second.bool_value();
+        if (skip_mx_value) {
+          return true;
+        }
+      }
+    } else {
+      return true;
```
kyessenov, line 358: This should be `false`? It doesn't have istio metadata.

source/extensions/filters/http/peer_metadata/filter.cc
```diff
@@ -323,13 +333,37 @@ void FilterConfig::setFilterState(StreamInfo::StreamInfo& info, bool downstream,
...
+    const auto& cluster_name = cluster_info.value()->name();
+    if (cluster_name == "PassthroughCluster") {
+      return true;
+    }
+    const auto& filter_metadata = cluster_info.value()->metadata().filter_metadata();
+    const auto& it = filter_metadata.find("istio");
+    if (it != filter_metadata.end()) {
+      const auto& skip_mx = it->second.fields().find("external");
+      if (skip_mx != it->second.fields().end()) {
+        const auto skip_mx_value = skip_mx->second.bool_value();
```
kyessenov, line 352: `return skip_mx->second.bool_value()`.

kyessenov (changes requested):
I think there's one wrong conditional, please fix, and then LG.

source/extensions/filters/http/peer_metadata/config.proto
GregHanson: leaving logic in proxy for now

source/extensions/filters/http/peer_metadata/filter.cc
GregHanson: done

source/extensions/filters/http/peer_metadata/filter.cc
GregHanson: done

kyessenov (approved):

PR #4958, opened 2023-09-18
delete forward sni, cluster rewrite, and sni verifier filters

kyessenov:
No longer used in Istio, and should be rewritten as Wasm if still needed.

Deletes some unused bazel dependencies.

Issue: https://github.com/envoyproxy/envoy/issues/29681


lei-tang (commented):
LGTM as Istio no longer uses these three filters. Beside Istio, are there other projects/products that use Istio proxy and these filters? Consider providing an advance notice (e.g., in Istio release notes, or on istio.io) about removing these filters to avoid breaking their users.

kyessenov:
The only way to use those is via EnvoyFilter, and there's no integration test, no doc, and no reference in the istio code base to them. We'll post a release note, but I think it's OK to try to remove them for now, and if there's a strong complaint we can get them back.

lei-tang (approved):

frankbu:
@kyessenov Removing forward_downstream_sni will break IBM (and presumably other) production clusters that are still using the [old egress wildcard pattern](https://istio.io/v1.13/docs/tasks/traffic-management/egress/wildcard-egress-hosts/#setup-egress-gateway-with-sni-proxy). We are going to switch to an alternative approach, but I think advance notice of this change is needed, before deleting it, because it was previously documented.

kyessenov:
@frankbu Ack, I checked with @howardjohn first, but didn't see `EnvoyFilter` in the docs. I think we can get https://github.com/envoyproxy/envoy/pull/29844 from upstream merged by the release of 1.20, and it's going to be functionally equivalent. If not, we'll bring it back. I hope there's at least some testing somewhere in istio.

PR #4994, opened 2023-10-05
tcp mx: implement WDS fallback

kyessenov:
Adds fallback to WDS in TCP MX filter. Gated by config.

This can be useful for a variety of uses:
* deprecate custom ALPN for MX
* TCP MX for plaintext
* removes the need for HTTP MX in majority of cases


kyessenov:
Issue https://github.com/istio/istio/issues/43937

lei-tang (approved):

PR #5032, opened 2023-10-12
replace connect authority with set_filter_state

kyessenov:
Replacing an experimental ambient filter with an upstream extension.


kyessenov:
@ericvn This also fixed the sync job. Any idea why EasyCLA is stuck?

lei-tang (approved):

PR #5085, opened 2023-10-24
add BAZEL_BIN_PATH back to fix builds

sspaeth-r7:
This PR adds back an important line that was removed between the 1.18 and 1.19 releases: https://github.com/istio/proxy/compare/release-1.18..release-1.19#diff-662262a00ad44baa70fb14405b688c43143c3b4635ff8766030dd34b328f8306L67

Without this line, Istio fails to build locally. The `exportcache` target fails with a `/envoy: no such file or directory` error.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: I can't see issue list so not sure if there is one for this already.

I notice this line was removed in the following commit, but I'm not an expert on Bazel nor the build process so I'm sure I'm not understanding everything here: https://github.com/istio/proxy/commit/b6a4cab8f74cf38a77b9d02626e326c2cee2524a

I'm not sure if the `TEST_ENVOY_PATH` was necessary too. It seems like the `exportcache` target is working again after only adding that one line back, but not sure why any of those lines were removed in the first place, so maybe someone could help explain to me that and which lines need to be there and which are ok to leave out.

Credit to @jeesmon for finding this to be the issue. More info in Istio slack thread in #test-and-release: https://istio.slack.com/archives/C6FCV6WN4/p1696463622534729

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -64,6 +64,7 @@ BAZEL_CONFIG_TSAN = # no working config
 endif
 BAZEL_CONFIG_CURRENT ?= $(BAZEL_CONFIG_DEV)
 
+BAZEL_BIN_PATH ?= $(shell bazel info $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_CURRENT) bazel-bin)
 TEST_ENVOY_TARGET ?= //:envoy
 TEST_ENVOY_DEBUG ?= trace
 


ericvn:
Normally, one would create the PR against the master branch first and then we would cherry-pick the change back as needed.

The original line was removed in https://github.com/istio/proxy/pull/4763.

sspaeth-r7:
Thank you @ericvn. I made a new PR based on master into master: https://github.com/istio/proxy/pull/5087

PR #4469, opened 2023-02-25
make: use bazel info bin path for exportcache

unicell:
Avoid making assumptions about bazel-bin path and honor the source of truth from bazel info commaand.

**What this PR does / why we need it**:

`/work/bazel-bin` doesn't always exist inside the container (e.g. when $XDG_CACHE_HOME is set to leverage external cache)

```
BUILD_WITH_CONTAINER=1 DOCKER_RUN_OPTIONS="-v ${XDG_CACHE_HOME}:${XDG_CACHE_HOME}" make shell
```

This small change tries to avoid making assumptions and takess bazel info result (that already exists in Makefile) as source of truth.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

N/A

**Special notes for your reviewer**:

N/A

PR #5087, opened 2023-10-24
turn previously removed BAZEL_BIN_PATH in to target-specific variable to fix exportcache builds

sspaeth-r7:
This PR adds back an important definition that was removed between the 1.18 and 1.19 releases: https://github.com/istio/proxy/compare/release-1.18..release-1.19#diff-662262a00ad44baa70fb14405b688c43143c3b4635ff8766030dd34b328f8306L67

Without this definition, Istio fails to build locally. The `exportcache` target fails with a `/envoy: no such file or directory` error since `BAZEL_BIN_PATH` is still referenced in a couple places in that target.

Instead of defining the variable globally, this PR defines it as a target-specific one so that Bazel is only started when a target requiring it is run. This is to retain the optimizations in speed added by @howardjohn in PR https://github.com/istio/proxy/pull/4763.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: I can't see issue list so not sure if there is one for this already.

I notice this line was removed in the following commit, but I'm not an expert on Bazel nor the build process so I'm sure I'm not understanding everything here: https://github.com/istio/proxy/commit/b6a4cab8f74cf38a77b9d02626e326c2cee2524a

I'm not sure if the `TEST_ENVOY_PATH` was necessary too. It seems like the `exportcache` target is working again after only adding that one line back, but not sure why any of those lines were removed in the first place, so maybe someone could help explain to me that and which lines need to be there and which are ok to leave out.

Credit to @jeesmon for finding this to be the issue. More info in Istio slack thread in #test-and-release: https://istio.slack.com/archives/C6FCV6WN4/p1696463622534729

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -49,6 +49,7 @@ BAZEL_CONFIG_TSAN = # no working config
 endif
 BAZEL_CONFIG_CURRENT ?= $(BAZEL_CONFIG_DEV)
 
+BAZEL_BIN_PATH ?= $(shell bazel info $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_CURRENT) bazel-bin)
 TEST_ENVOY_TARGET ?= //:envoy
 TEST_ENVOY_DEBUG ?= trace
 
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -49,7 +49,6 @@ BAZEL_CONFIG_TSAN = # no working config
 endif
 BAZEL_CONFIG_CURRENT ?= $(BAZEL_CONFIG_DEV)
 
-BAZEL_BIN_PATH ?= $(shell bazel info $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_CURRENT) bazel-bin)
 TEST_ENVOY_TARGET ?= //:envoy
 TEST_ENVOY_DEBUG ?= trace
 
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -151,6 +151,7 @@ else
 endif
 
 # Used by build container to export the build output from the docker volume cache
+exportcache: BAZEL_BIN_PATH ?= $(shell bazel info $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_CURRENT) bazel-bin)
 exportcache:
 	@mkdir -p /work/out/$(TARGET_OS)_$(TARGET_ARCH)
 	@cp -a $(BAZEL_BIN_PATH)/envoy /work/out/$(TARGET_OS)_$(TARGET_ARCH)


lei-tang (commented):
LGTM. As John made the original PR, will let John to take a look before approving.

sspaeth-r7:
Thanks @lei-tang. @howardjohn if you could please advise on whether `TEST_ENVOY_PATH` is needed too since I don't see it referenced in the makefile anymore (unlike `BAZEL_BIN_PATH` which like @jeesmon said is still referenced in the `exportcache` target), but not sure if it's used elsewhere. A quick search throughout the repo didn't come back with anything but I don't know what all is happening inside the make process so just want to be 100% sure.

sspaeth-r7:
Also, does anyone think there would be value in adding a test that makes sure `exportcache` stage works as expected? I bet if there was one it could've caught this before us.

ericvn:
Per the prior PR, the`BAZEL_BIN_PATH` variable was removed so that the bazel command is not run every time the make is parsed (made lazy). Simply adding this line back in will revert that, which isn't wanted. I think what you want to do is set it up so `BAZEL_BIN_PATH` is evaluated inside the `exportcache` target

ericvn (changes requested):
Requesting changes since simply reverts the prior PR. I think we just want to add a line within the target so the `bezel` command isn't run for all targets (which the original PR tried to do (just didn't fix one target)).

ericvn:
Also for past history: https://github.com/istio/proxy/pull/4469

howardjohn:
Its removed to avoid making a `shell bazel` call for 100% of makefile targets. This means you cannot even do trivial operations with a super slow bazel call (and requiring bazel).

If we add it back (which seems good), we should add it so it only runs for `exportcache` where it is needed

ericvn:
Some trial work:
If I run `BUILD_WITH_CONTAINER=1 make shell` and within the shell with no changes run `make check` I quickly get an message back:
```
build-tools:/work$ make check
make[1]: Entering directory '/work'
Please use "make lint" instead.
make[1]: *** [Makefile.core.mk:106: check] Error 1
make[1]: Leaving directory '/work'
```

If I add the line of code specified here, I end up trying to call `bazel`:
```
build-tools:/work$ make check
make[1]: Entering directory '/work'
2023/10/24 20:46:52 Downloading https://releases.bazel.build/6.3.2/release/bazel-6.3.2-linux-arm64...
^C
```

If I instead move the line of added code so it's like:
```
# Used by build container to export the build output from the docker volume cache
exportcache:
	@BAZEL_BIN_PATH ?= $(shell bazel info $(BAZEL_BUILD_ARGS) $(BAZEL_CONFIG_CURRENT) bazel-bin)
```
The `make check` reverts to the same quick call, and `make exportcache` tries to call `bazel`

I haven't checked it this new placement works with the build.

sspaeth-r7:
Understood. I have next to no experience writing makefiles, so I'm not familiar with how to do stuff like this, but I did a little bit of experimentation on my end and found the following:

Moving the definition just into the target block won't work because it expects all these commands to be shell commands, and `BAZEL_BIN_PATH` is obviously not one of them. Is anyone familiar with how to create a variable whose contents is only evaluated when the target is called?

So far the only workaround that I found which works is using a shell variable as opposed to a Makefile one, which requires some escaping but seems to only sleep when I call `mytarget` instead of `defaulttarget`. When I call `defaulttarget`, it immediately builds and shows the expected output. Guessing this isn't an acceptable solution but I'm not sure how to work around the evaluation time problems since I have so little make experience. I did a little bit of research but couldn't find much off the bat.
```make
defaulttarget:
    echo hi default

mytarget:
  @MYVAR=$$(sleep 10; echo this is just here to act like bazel running) ; echo hi $$MYVAR
```
I will keep doing research on how to solve this correctly in the mean time.

ericvn:
Maybe:
```
	$(eval BAZEL_BIN_PATH ?= ${shell bazel info ${BAZEL_BUILD_ARGS} ${BAZEL_CONFIG_CURRENT} bazel-bin})
```
If I tried:
```
# Used by build container to export the build output from the docker volume cache
exportcache:
	$(eval BAZEL_BIN_PATH ?= ${shell bazel info ${BAZEL_BUILD_ARGS} ${BAZEL_CONFIG_CURRENT} bazel-bin})
	echo BAZEL_BIN_PATH=$$BAZEL_BIN_PATH
	@mkdir -p /work/out/$(TARGET_OS)_$(TARGET_ARCH)
	@cp -a $(BAZEL_BIN_PATH)/envoy /work/out/$(TARGET_OS)_$(TARGET_ARCH)
```
and then I got things like:
```
build-tools:/work$ make exportcache
make[1]: Entering directory '/work'
echo BAZEL_BIN_PATH=$BAZEL_BIN_PATH
BAZEL_BIN_PATH=/home/.cache/bazel/_bazel_user/1e0bb3bee2d09d2e4ad3523530d3b40c/execroot/io_istio_proxy/bazel-out/aarch64-fastbuild/bin
cp: cannot stat '/home/.cache/bazel/_bazel_user/1e0bb3bee2d09d2e4ad3523530d3b40c/execroot/io_istio_proxy/bazel-out/aarch64-fastbuild/bin/envoy': No such
```
and
```
AZEL_BIN_PATH=ttt make exportcache
make[1]: Entering directory '/work'
echo BAZEL_BIN_PATH=$BAZEL_BIN_PATH
BAZEL_BIN_PATH=ttt
cp: cannot stat 'ttt/envoy': No such file or directory
```

sspaeth-r7:
Added a commit using target-specific variables to define it in the same context it was being done in prior to 1.19. Apparently with target-specific variables, you must define variables in the target first, then in a separate block, start the target (with the same name) again, and put the shell code you want to run there. It will automatically detect which is make and which is shell code based off the order they came in (IMU). From the docs I linked above:
>There is one more special feature of target-specific variables: when you define a target-specific variable that variable value is also in effect for all prerequisites of this target, and all their prerequisites, etc. (unless those prerequisites override that variable with their own target-specific variable value). So, for example, a statement like this:
> ```make
> prog : CFLAGS = -g
> prog : prog.o foo.o bar.o
> ```
> will set CFLAGS to ‘-g’ in the recipe for prog, but it will also set CFLAGS to ‘-g’ in the recipes that create prog.o, foo.o, and bar.o, and any recipes which create their prerequisites. 

This way it's eval-less which from the SO posts I was reading is probably safer for several reasons.

sspaeth-r7:
About this excert from previous comment, I'm not sure how make infers which recipes create the ones used in our target's recipe so going to test to make sure it's not running the Bazel command when we run `build_envoy` or `build`.
> will set CFLAGS to ‘-g’ in the recipe for prog, but it will also set CFLAGS to ‘-g’ in the recipes that create prog.o, foo.o, and bar.o, and any recipes which create their prerequisites.

sspaeth-r7:
Looks like everything works as expected. I put some code I knew would fail inside the target-specific variable definition and
```bash
(flags...) make build_wasm build build_envoy
```
didn't get hung up on it but
```bash
(flags...) make build_wasm build build_envoy exportcache
```
did. So I think this is the best best of both worlds. It won't start Bazel at all now unless the target that requires it is called, but we can still use that target so we don't get failing builds 🙂

howardjohn (approved):

zirain (approved):

howardjohn (approved):

zirain:
@ericvn can you review again?

ericvn (approved):
Verified as well.

PR #5117, opened 2023-11-03
move extension check to postsubmit

zirain:
cc @kyessenov is this what we talked about via slack?


kyessenov (approved):
Yeah, let's not block the sync again.

PR #5134, opened 2023-11-08
Update golang.org/x/net dependency

tedjpoole:
CVE update backport


ksubrmnn:
Hi, @tedjpoole! Thanks for your contribution! Looks like this update was already handled in https://github.com/istio/tools/pull/2729/files so I will close this PR.

Thanks again!

PR #5150, opened 2023-11-16
use absl::StatusOr

zirain:
fixes: https://github.com/istio/proxy/pull/5149

xref: https://github.com/envoyproxy/envoy/pull/30765

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -33,10 +33,10 @@ bind(
 # 1. Determine SHA256 `wget https://github.com/envoyproxy/envoy/archive/$COMMIT.tar.gz && sha256sum $COMMIT.tar.gz`
 # 2. Update .bazelversion, envoy.bazelrc and .bazelrc if needed.
 #
-# Commit date: 2023-11-14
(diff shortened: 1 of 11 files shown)


kyessenov:
Can you make sure there are no exceptions as well? StatusOr is meant to replace exceptions.

zirain:
> Can you make sure there are no exceptions as well? StatusOr is meant to replace exceptions.

I change this because of https://github.com/envoyproxy/envoy/pull/30765, I don't think need any logic change.

kyessenov (approved):

kyessenov:
/retest

I think it's starting to flake locally too, so maybe we should finally root cause it.

zirain:
can we make these optional before figure out what happen?

kyessenov:
Yeah, put SkipTsanAsan on the test TestStackdriverRbacTCPDryRun, it'll let this merge. The test fails on non-Asan so it's just a borked test probably.

zirain:
open https://github.com/istio/test-infra/pull/5093

PR #5154, opened 2023-11-17
osx support

zirain:
this PR aim to make it possible to run `make build` on m1 mac pro

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -13,6 +13,9 @@ build:remote --remote_timeout=7200
 # Istio specific Bazel build/test options.
 # ========================================
 
+# Enable libc++ and C++20 by default.
+build:linux --config=libc++20
+
 # Need for CI image to pickup docker-credential-gcloud, PATH is fixed in rbe-toolchain-* configs.
 build:remote-ci --action_env=PATH=/usr/local/google-cloud-sdk/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/llvm/bin
 
--- a/.bazelrc
+++ b/.bazelrc
@@ -13,9 +13,6 @@ build:remote --remote_timeout=7200
 # Istio specific Bazel build/test options.
 # ========================================
 
-# Enable libc++ and C++20 by default.
-build --config=libc++20
-
 # Need for CI image to pickup docker-credential-gcloud, PATH is fixed in rbe-toolchain-* configs.
 build:remote-ci --action_env=PATH=/usr/local/google-cloud-sdk/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/llvm/bin
 
@@ -25,7 +22,8 @@ build --define path_normalization_by_default=true
 
 # Heap profiler is supported only with gperf tcmalloc, not google tcmalloc.
 # See: https://github.com/istio/istio/issues/28233
-build --define tcmalloc=gperftools
+build:linux --define tcmalloc=gperftools
+build:macos --define tcmalloc=disabled
 
 # Build with embedded V8-based WebAssembly runtime.
 build --define wasm=v8
@@ -67,3 +65,9 @@ build:clang-asan-ci --linkopt=-l:libclang_rt.ubsan_standalone_cxx.a
 build:clang-tsan-ci --config=clang-tsan
 build:clang-tsan-ci --linkopt=-L/opt/libcxx_tsan/lib
 build:clang-tsan-ci --linkopt=-Wl,-rpath,/opt/libcxx_tsan/lib
+
+# get from https://github.com/Homebrew/homebrew-core/blob/master/Formula/e/envoy.rb
+build:macos --cxxopt=-Wno-range-loop-analysis
+build:macos --host_cxxopt=-Wno-range-loop-analysis
+build:macos --cxxopt=-Wno-deprecated-declarations
+build:macos --host_cxxopt=-Wno-deprecated-declarations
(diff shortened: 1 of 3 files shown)


bazel/extension_config/extensions_build_config.bzl
```diff
@@ -486,7 +486,7 @@ ISTIO_ENABLED_CONTRIB_EXTENSIONS = [
     "envoy.filters.sip.router",
     "envoy.tls.key_providers.cryptomb",
     "envoy.tls.key_providers.qat",
-    "envoy.network.connection_balance.dlb",
+#    "envoy.network.connection_balance.dlb",
```
zirain, line 489: `select` cannot work here, @kyessenov any suggestion to make this filter optional on macos?

zirain: https://github.com/envoyproxy/envoy/pull/30951 may give some help.

.bazelrc
```diff
@@ -14,7 +14,7 @@ build:remote --remote_timeout=7200
 # ========================================
 
 # Enable libc++ and C++20 by default.
-build --config=libc++20
+build:linux --config=libc++20
```
kyessenov, line 17: why disable on macos? we have c++20 code already I think.

.bazelrc
```diff
@@ -25,7 +25,8 @@ build --define path_normalization_by_default=true
 
 # Heap profiler is supported only with gperf tcmalloc, not google tcmalloc.
 # See: https://github.com/istio/istio/issues/28233
-build --define tcmalloc=gperftools
+build:linux --define tcmalloc=gperftools
+build:macos --define tcmalloc=disabled
```
kyessenov, line 29: Same - why disable gperftools?

bazel/extension_config/extensions_build_config.bzl
kyessenov: Yeah, you have to fix it upstream to guard the code on macos.

.bazelrc
zirain: always fails with following error:
```
configure: error: in `/private/var/tmp/_bazel_zirain/351f889f26744009d08525dc5ecb1daa/sandbox/darwin-sandbox/371/execroot/io_istio_proxy/bazel-out/darwin_arm64-fastbuild/bin/external/envoy/bazel/foreign_cc/gperftools_build.build_tmpdir':
configure: error: C compiler cannot create executables
```

.bazelrc
zirain:
```
clang: error: invalid linker name in argument '-fuse-ld=lld'
```

bazel/extension_config/extensions_build_config.bzl
zirain: https://github.com/envoyproxy/envoy/pull/30951 merged

.bazelrc
kyessenov: you should use clang's lld on macos? what linker are you using?

.bazelrc
kyessenov: ok, it's fine but you won't get heap profile and some tools. what does tcmalloc disabled mean? what malloc is it using?

zirain: envoy has same line, this's a known limitation. https://github.com/envoyproxy/envoy/blob/05457bb047b87f4cc1a25635400ddcaa7fccc923/.bazelrc#L138

I won't encourage anyone use proxy on macos for production, all this line make that easier to develop on macos.

.bazelrc
zirain: I think the default one setting xcode.

> I won't encourage anyone use proxy on macos for production, all these lines make that easier to develop on macos.

.bazelrc
kyessenov: I don't recall, does `linux` configuration automatically kick-in?

.bazelrc
kyessenov: I have no problems with allowing macos development. Just want to make sure we're not breaking release on linux.

kyessenov (approved):

.bazelrc
zirain: if not, there must be huge problem.

.bazelrc
zirain: I do my best to make it won't happen, and there's a long time before next release, so I think it's pretty safe.

jcetkov:
@zirain the mac build is failing for me with
```
./lib/findprog-in.c:137:25: error: call to undeclared function 'eaccess'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
                    if (eaccess (progpathname, X_OK) == 0)
                        ^
./lib/findprog-in.c:137:25: note: did you mean 'access'?
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/unistd.h:431:6: note: 'access' declared here
int      access(const char *, int);
         ^
./lib/findprog-in.c:211:21: error: call to undeclared function 'eaccess'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
                if (eaccess (progpathname, X_OK) == 0)
```

seems to be related to https://github.com/bazelbuild/rules_foreign_cc/issues/859, however I'm not very proficient neither with bazel nor with c, would you have any tips how to work around this?

Seems to be some fairly recent change (bazel version change?) as I was able to build the proxy @ 1.17.8, but it's failing at HEAD as well as @1.19.4 tag.

Thank you!

zirain:
I test these only for master on m1 mac pro, I will give a try when I have bandwidth.

jcetkov:
thanks! this is also m1 mac pro (ventura 13.6)

jcetkov:
@zirain I don't think this is the root of my problem, rather an update of the `rules_foreign_cc` in upstream envoy.
[This](https://github.com/envoyproxy/envoy/commit/f76cc4da2816759ed141ba71db8090b6aff52239#diff-a2ca502474bb8ec5f092176f7ee6d2f11fc395ebb00e1717d6fbad9b16ee8fd8R873) change which then makes me run into [this](https://github.com/bazelbuild/rules_foreign_cc/issues/859)

PR #5169, opened 2023-11-25
update envoy

zirain:
fixes: https://github.com/istio/proxy/pull/5168


joaopgrassi (approved):
I worked on that extension in Envoy, so will give my meaningless approval here :D

lei-tang (approved):

PR #5174, opened 2023-11-29
Enable zstd

zirain:
zstd is a core feature, should enable it


odubajDT:
Thank you for your effort!

lei-tang (approved):

PR #5219, opened 2023-12-19
stats: fix memory leak in periodic reporting

kyessenov:
Change-Id: Ibd473bb14c51786c34f2d459ce5e1eea675e3d13

Fixes: https://github.com/istio/istio/issues/48023

For periodically reported metrics (TCP bytes and gRPC), the expressions were evaluated repeatedly every 5s. That requires allocating the memory for storing the expression result even when it does not change. The fix is to evaluate it at most twice (at the start and the end of a stream). This happens when some expression overrides are applied on TCP or gRPC metrics.

Tested locally. Without the change I see an increase from 22MB to 50MB in allocated bytes with 500 open connections over a few hours. With the change, the memory remains stable. The setup is 3 envoys, client sidecar, envoy sidecar and envoy echo backend. The test is:
```
for f in `seq 1 500`; do echo test | nc localhost 8081 & done
```

### DIFF
--- a/source/extensions/filters/http/istio_stats/istio_stats.cc
+++ b/source/extensions/filters/http/istio_stats/istio_stats.cc
@@ -619,18 +619,21 @@ struct Config : public Logger::Loggable<Logger::Id::filter> {
 
   // RAII for stream context propagation.
   struct StreamOverrides : public Filters::Common::Expr::StreamActivation {
-    StreamOverrides(Config& parent, Stats::StatNameDynamicPool& pool,
-                    const StreamInfo::StreamInfo& info,
-                    const Http::RequestHeaderMap* request_headers = nullptr,
-                    const Http::ResponseHeaderMap* response_headers = nullptr,
-                    const Http::ResponseTrailerMap* response_trailers = nullptr)
-        : parent_(parent) {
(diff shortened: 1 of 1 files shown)


lei-tang (approved):

zirain:
@kyessenov possible to add a test for this?

ldemailly:
Thanks Kuat! you mention in the description that instead of every 5s it's done once at start and end, but... that wouldn't quite be the definition of a leak fix, it'd leak 'less'?  While unnecessary (it's great to make it do less unnecessary job) can you summarize what makes it leak vs not leak (is that the _pool / stream_ thing?) - a leak sort of implies adding new keys constantly to some map or similar ?

kyessenov:
@ldemailly It was doing it more than twice before this fix. Every 5s, it'd re-evaluate and store symbols in a per-request pool, which is what makes it a leak. With the fix, the overhead of stats is bounded per-request.

hanxiaop:
This works great for me, thanks!

jezhang2014:
whether istio 1.8 and 1.13 might face the similar problem?

jezhang2014:
In addition, if possible, please help to explain in detail how to reproduce the memory leak problem before the fix on 1.18. Thanks!

PR #5238, opened 2024-01-03
wasm: align on xds attributes

kyessenov:
Change-Id: Iaeb55c0dd5cb5801af28209bae2f669f42bc554d

Follow-up to https://github.com/envoyproxy/envoy/pull/31365 since we want to deprecate wasm-specific attributes.

### DIFF
--- a/extensions/common/context.cc
+++ b/extensions/common/context.cc
@@ -124,8 +124,8 @@ void populateRequestInfo(bool outbound, bool use_host_header_fallback, RequestIn
 
   request_info->is_populated = true;
 
-  getValue({"cluster_name"}, &request_info->upstream_cluster);
-  getValue({"route_name"}, &request_info->route_name);
+  getValue({"xds", "cluster_name"}, &request_info->upstream_cluster);
+  getValue({"xds", "route_name"}, &request_info->route_name);
   // Fill in request info.
   // Get destination service name and host based on cluster name and host
   // header.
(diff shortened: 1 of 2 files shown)


zirain (approved):

kyessenov:
Needs https://github.com/envoyproxy/envoy/pull/31620

zirain:
seems still not working @kyessenov

kyessenov:
Root caused it to https://github.com/envoyproxy/envoy/pull/32077

zirain:
one more try

/retest

zirain:
no client accesslog found

extensions/common/context.cc
```diff
@@ -193,7 +193,7 @@ std::string_view ProtocolString(Protocol protocol) {
 // Retrieves the traffic direction from the configuration context.
 TrafficDirection getTrafficDirection() {
   int64_t direction;
-  if (getValue({"listener_direction"}, &direction)) {
+  if (getValue({"xds", "listener_direction"}, &direction)) {
```
zirain, line 196: looks like there's a bug, `xds.listener_direction` didn't return the right result, always return `TrafficDirection`.

zirain: see https://github.com/istio/proxy/pull/5300

kyessenov: Yeap, I know about this case. It doesn't actually make sense to read listener direction during config time with ECDS since the same filter instance can be used for both directions.

PR #5284, opened 2024-01-24
network filter: set identiy from PROXY TLV header

stevenctl:

This is the most basic filter for propagating identity from the sandwich 

// TODO link the zTunnel PR here that sets the TLV header

**Which issue this PR fixes** 

First step towards https://github.com/istio/istio/issues/48362

This uses 0xD0 as the key for the spiffe string. We could support multiple formats here, e.g.

0xD0 -> SPIFFE string directly into peer_principal
0XD1 -> A JWT that also sets some metadata

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -39,7 +39,7 @@ envoy_cc_binary(
         "//source/extensions/filters/http/alpn:config_lib",
         "//source/extensions/filters/http/istio_stats",
         "//source/extensions/filters/http/peer_metadata:filter_lib",
-        "//source/extensions/filters/network/istio_tlv_authn:config_lib", # Very experimental: ambient/sandwich
+        "//source/extensions/filters/network/istio_tlv_authn:config_lib",  # Very experimental: ambient/sandwich
         "//source/extensions/filters/network/metadata_exchange:config_lib",
         "@envoy//source/exe:envoy_main_entry_lib",
     ],
--- a/BUILD
+++ b/BUILD
@@ -39,6 +39,7 @@ envoy_cc_binary(
         "//source/extensions/filters/http/alpn:config_lib",
         "//source/extensions/filters/http/istio_stats",
         "//source/extensions/filters/http/peer_metadata:filter_lib",
+        "//source/extensions/filters/network/istio_tlv_authn:config_lib", # Very experimental: ambient/sandwich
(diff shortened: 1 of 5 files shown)


kyessenov (changes requested):
Do we really need a new filter? Can we use set_filter_state filter to copy TLV record from ProxyProto filter state into the principal? I thought someone has already accomplished this upstream.

stevenctl:
How can we grab one element of [tlv_vector](https://github.com/envoyproxy/envoy/blob/4f0965e9c61f8daa3519dc0298d4833cf8730307/envoy/network/proxy_protocol.h#L22) matching the "type" byte using that API?

stevenctl:
I could not get the `on_tlv_present -> some other key -> set_filter_state(%FILTER_STATE(thatotherkey)%)` working. Will provide yaml to reproduce this.

stevenctl:
Not a standalone yaml, but grabbed some from configdumps:

On the listener it's this and http inspector:

```
            listener_filters:
              - name: envoy.filters.listener.proxy_protocol
                typed_config:
                  '@type': type.googleapis.com/envoy.extensions.filters.listener.proxy_protocol.v3.ProxyProtocol
                  rules:
                    - tlv_type: 208
                      on_tlv_present:
                        key: peer_principal
                  pass_through_tlvs:
                    tlv_type:
                      - 208
```

And then before HCM I have

```
                  - name: proxy_protocol_authority
                    typed_config:
                      '@type': type.googleapis.com/envoy.extensions.filters.network.set_filter_state.v3.Config
                      on_new_connection:
                        - object_key: io.istio.peer_principal
                          format_string:
                            text_format_source:
                              inline_string: '%DYNAMIC_METADATA(envoy.filters.listener.proxy_protocol:peer_principal)%'
                          shared_with_upstream: ONCE
```

kyessenov:
Is it because `on_new_connection` is called too early? Can you check from trace logs? If so, we might need another trigger for `onEvent`.

kyessenov:
Also, `set_filter_state` needs to be on the same listener as proxy proto handler.

stevenctl:
Set_filter_state is working now. I'm not sure what I was doing wrong originally.. possibly unrelated to set filter state or maybe I had one of the connect_authority ones in there before?

PR #5339, opened 2024-02-15
fips: force stackdriver to use TLSv1.2

kyessenov:
Change-Id: Icd1cd577c039512bb90234642719a8b5d3523567

Unconditional enforcement of https://github.com/istio/istio/issues/49081 for GCP Stackdriver, since TLSv1.2 is sufficient.

### DIFF
--- a/extensions/stackdriver/metric/registry.cc
+++ b/extensions/stackdriver/metric/registry.cc
@@ -21,6 +21,7 @@
 #include "extensions/stackdriver/common/constants.h"
 #include "google/api/monitored_resource.pb.h"
 #include "grpcpp/grpcpp.h"
+#include "grpcpp/security/tls_certificate_provider.h"
 
 namespace Extensions {
 namespace Stackdriver {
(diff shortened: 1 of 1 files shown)


extensions/stackdriver/metric/registry.cc
```diff
@@ -80,16 +81,19 @@ getStackdriverOptions(const Wasm::Common::FlatNode& local_node_info,
     }
   }
 
-  auto ssl_creds_options = grpc::SslCredentialsOptions();
+  grpc::experimental::TlsChannelCredentialsOptions tls_options;
+  tls_options.set_max_tls_version(grpc_tls_version::TLS1_2);
```
lei-tang, line 85: Is the intension to 1) only use TLS 1.2 or 2) avoid using TLS 1.3?
From the following code in the CL, the intension seems to be 2) but the title of the CL seems to indicate 1).

 tls_options.set_max_tls_version(grpc_tls_version::TLS1_2);

kyessenov: It's the same. gRPC sets TLS 1.2 as a minimum version.

extensions/stackdriver/metric/registry.cc
```diff
@@ -80,16 +81,19 @@ getStackdriverOptions(const Wasm::Common::FlatNode& local_node_info,
     }
   }
 
-  auto ssl_creds_options = grpc::SslCredentialsOptions();
+  grpc::experimental::TlsChannelCredentialsOptions tls_options;
```
lei-tang, line 84: Based on its name, "grpc::experimental::TlsChannelCredentialsOptions" seems to be an experimental feature. Do we have a more stable API to set the max TLS version?  If this experimental feature becomes deprecated, we need to find a replacement for it.

kyessenov: gRPC team specifically backported this experimental feature for us. A similar change in Envoy was approved https://github.com/envoyproxy/envoy/pull/32315.

lei-tang (approved):

PR #5399, opened 2024-03-11
Build envoy with debug

nirvanagit:


envoy.bazelrc
```diff
@@ -64,6 +64,7 @@ common --experimental_allow_tags_propagation
 build:linux --copt=-fdebug-types-section
 build:linux --copt=-fPIC
 build:linux --copt=-Wno-deprecated-declarations
+build:linux --copt=-fno-limit-debug-info
```
lei-tang, line 67: The PR description is currently empty. Can you add to the PR description of 1) why this option needs to be added; 2) pros and cons of adding this option?

kyessenov:
Also, this will be overwritten by Envoy sync anyways. We have debug builds of Envoy but they are O(Gb) sized. Why do you want this on all builds?

zirain:
feel free to reopen.

PR #5425, opened 2024-03-27
Enable QAT key provider

soulxu:

The qat key provider was disabled due to CI breaking https://github.com/istio/proxy/pull/5411#issuecomment-2017030825, now the build issue was fixed by https://github.com/envoyproxy/envoy/pull/33143, so enable it again.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #5411

### DIFF
--- a/bazel/extension_config/extensions_build_config.bzl
+++ b/bazel/extension_config/extensions_build_config.bzl
@@ -507,7 +507,7 @@ ISTIO_ENABLED_CONTRIB_EXTENSIONS = [
     "envoy.filters.network.sip_proxy",
     "envoy.filters.sip.router",
     "envoy.tls.key_providers.cryptomb",
-    # "envoy.tls.key_providers.qat", # TODO enable this later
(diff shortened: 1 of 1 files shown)


zirain:
wait for https://github.com/istio/proxy/pull/5426

zirain (approved):

zirain:
@soulxu can you share the root cause? is there anything we can do at envoy repo to prevent this happen again?

soulxu:
> @soulxu can you share the root cause? is there anything we can do at envoy repo to prevent this happen again?

@zirain, Ifind some useless building parameters were added by [PR](https://github.com/envoyproxy/envoy/pull/32166). And those parameters lead to we build some shared library which qat provider key provider doesn't need actually. But I can't figure out why those build error only happened on the istio ci, since I also can't reproduce this compile error in my local.

PR #5466, opened 2024-04-13
sync with upstream

zirain:
fix: #5462

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -53,6 +53,9 @@ build --cxxopt -Wformat-security
 # Link pthread for flatbuffers
 build --host_linkopt=-pthread
 
+build:clang --host_action_env=CC=
+build:clang --host_action_env=CXX=
+
 # CI sanitizer configuration
 #
 build:clang-asan-ci --config=clang-asan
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -55,7 +55,6 @@ build --host_linkopt=-pthread
 
 # CI sanitizer configuration
 #
-build:clang-asan-ci --config=clang-asan
```
kyessenov, line 58: That simply disables asan, which is not the right fix.

envoy.bazelrc
```diff
@@ -103,7 +105,6 @@ build:clang-tidy --output_groups=report
 build:clang-tidy --build_tag_filters=-notidy
 
 # Basic ASAN/UBSAN that works for gcc
-build:asan --action_env=ENVOY_ASAN=1
```
zirain, line 106: seems cause by this line, let me give a try.

.bazelrc
```diff
@@ -53,17 +53,15 @@ build --cxxopt -Wformat-security
 # Link pthread for flatbuffers
 build --host_linkopt=-pthread
 
+build:clang --host_action_env=CC=
+build:clang --host_action_env=CXX=
```
zirain, line 57: not sure why need this, cc @kyessenov

kyessenov: I see, so this is an environment pollution I think from Istio CI.

kyessenov (approved):

PR #5473, opened 2024-04-16
Add option for local caching

keithmattix:
If the build container ever gets stopped/updated, the local cache is lost. This PR adds a flag to persist the local cache to the user's .cache directory to save bazel artifacts across builds


common/scripts/run.sh
lei-tang, line 1: Please follow the instruction from line 3 to 8 of this file.

keithmattix: Ack, didn't realize this was a copy

PR #5514, opened 2024-05-01
ambient: Fix destination principal with WDS

keithmattix:
Grab the service account and trust domain from WDS and use that to create the destination principal

### DIFF
--- a/source/extensions/common/workload_discovery/api.cc
+++ b/source/extensions/common/workload_discovery/api.cc
@@ -29,8 +29,9 @@
 #include "source/extensions/common/workload_discovery/extension.pb.validate.h"
 
 namespace Envoy::Extensions::Common::WorkloadDiscovery {
-
 namespace {
+constexpr absl::string_view DefaultNamespace = "default";
+constexpr absl::string_view DefaultTrustDomain = "cluster.local";
 Istio::Common::WorkloadMetadataObject convert(const istio::workload::Workload& workload) {
   auto workload_type = Istio::Common::WorkloadType::Deployment;
   switch (workload.workload_type()) {
@@ -46,10 +47,23 @@ Istio::Common::WorkloadMetadataObject convert(const istio::workload::Workload& w
   default:
     break;
   }
+
+  absl::string_view ns = workload.namespace_();
+  absl::string_view trust_domain = workload.trust_domain();
+  // Trust domain may be elided if it's equal to "cluster.local"
+  if (trust_domain.empty()) {
+    trust_domain = DefaultTrustDomain;
+  }
+  // The namespace may be elided if it's equal to "default"
+  if (ns.empty()) {
+    ns = DefaultNamespace;
+  }
+  const auto identity = absl::StrCat("spiffe://", trust_domain, "/ns/", workload.namespace_(),
+                                     "/sa/", workload.service_account());
   return Istio::Common::WorkloadMetadataObject(
       workload.name(), workload.cluster_id(), workload.namespace_(), workload.workload_name(),
       workload.canonical_name(), workload.canonical_revision(), workload.canonical_name(),
-      workload.canonical_revision(), workload_type);
+      workload.canonical_revision(), workload_type, identity);
 }
 } // namespace
 
--- a/source/extensions/filters/http/istio_stats/istio_stats.cc
+++ b/source/extensions/filters/http/istio_stats/istio_stats.cc
@@ -1147,7 +1147,7 @@ private:
                              ? pool_.add(endpoint_peer->namespace_name_)
                              : context_.unknown_});
         tags_.push_back({context_.destination_principal_,
-                         !local_san.empty() ? pool_.add(local_san) : context_.unknown_});
+                         endpoint_peer ? pool_.add(endpoint_peer->identity_) : context_.unknown_});
         // Endpoint encoding does not have app and version.
         tags_.push_back(
             {context_.destination_app_, endpoint_peer && !endpoint_peer->app_name_.empty()
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,3 +7,16 @@ to find out how you can help.
 ## Prerequisites
 
 To make sure you're ready to build Envoy, clone and follow the upstream Envoy [build instructions](https://github.com/envoyproxy/envoy/blob/main/bazel/README.md). Be sure to copy clang.bazelrc into this directory after running the setup_clang script. Confirm that your environment is ready to go by running `bazel build --config=clang --define=wasm=disabled :envoy` (in this directory).
+
+## How to use a devcontainer
+
+1. Change the image in .devcontainer.json to `build-tools-proxy` instead of `build-tools`
+2. Open the directory in a container with the Remote - Containers extension
+3. Install the following extensions:
+    - [clangd](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd)
+    - [bazel-stack-vscode-cc](https://marketplace.visualstudio.com/items?itemName=StackBuild.bazel-stack-vscode-cc)
+4. Update clangd and reload the container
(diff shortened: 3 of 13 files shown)


source/extensions/common/workload_discovery/api.cc
```diff
@@ -46,10 +46,22 @@ Istio::Common::WorkloadMetadataObject convert(const istio::workload::Workload& w
...
   }
+  std::string ns = workload.namespace_();
+  std::string trust_domain = workload.trust_domain();
+  // Trust domain may be elided if it's equal to "cluster.local"
+  if (trust_domain.empty()) {
+    trust_domain = "cluster.local";
+  }
+  // The namespace may be elided if it's equal to "default"
+  if (ns.empty()) {
+    ns = "default";
```
kyessenov, line 57: This sounds dangerous to imply default.

source/extensions/common/workload_discovery/api.cc
```diff
@@ -46,10 +46,22 @@ Istio::Common::WorkloadMetadataObject convert(const istio::workload::Workload& w
   default:
     break;
   }
+  std::string ns = workload.namespace_();
+  std::string trust_domain = workload.trust_domain();
+  // Trust domain may be elided if it's equal to "cluster.local"
+  if (trust_domain.empty()) {
+    trust_domain = "cluster.local";
```
kyessenov, line 53: cluster.local is strongly discouraged - I think it should be explicit.

extensions/common/context.cc
```diff
@@ -293,9 +293,13 @@ namespace {
 bool extractPeerMetadataFromUpstreamMetadata(const std::string& metadata_type,
                                              flatbuffers::FlatBufferBuilder& fbb) {
   std::string endpoint_labels;
+  std::string identity;
```
kyessenov, line 296: Undo these changes?

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -95,6 +95,11 @@ extractEndpointMetadata(const StreamInfo::StreamInfo& info) {
     const auto& filter_metadata = upstream_host->metadata()->filter_metadata();
     const auto& it = filter_metadata.find("istio");
     if (it != filter_metadata.end()) {
+      std::string identity = "";
```
kyessenov, line 98: I'm confused - this is using EDS, not WDS?

source/extensions/common/workload_discovery/api.cc
```diff
@@ -46,10 +46,22 @@ Istio::Common::WorkloadMetadataObject convert(const istio::workload::Workload& w
   default:
     break;
   }
+  std::string ns = workload.namespace_();
```
kyessenov, line 49: `const std::string&` to avoid a copy.

source/extensions/common/workload_discovery/api.cc
keithmattix: The comments are straight from the WDS proto

source/extensions/common/workload_discovery/api.cc
keithmattix: This is also from the proto. In my tests the trust domain was elided as well

extensions/common/context.cc
keithmattix: Good catch; this was from an older implementation

source/extensions/filters/http/istio_stats/istio_stats.cc
keithmattix: Also an older implementation I forgot to scrub

source/extensions/common/workload_discovery/api.cc
keithmattix: I think this is necessary if we do need the defaulting logic?

keithmattix:
Fixes istio/istio##50225

keithmattix:
@kyessenov please see my responses above; if we need to change WDS across waypoint, ztunnel, and istiod, we can probably tackle that later, but I think the PR as-is represents the current state of the API

source/extensions/common/workload_discovery/api.cc
kyessenov: I think you can use `string_view` here, and pull out a constant into `constexpr`?

source/extensions/common/workload_discovery/api.cc
```diff
@@ -46,10 +46,22 @@ Istio::Common::WorkloadMetadataObject convert(const istio::workload::Workload& w
...
+  std::string trust_domain = workload.trust_domain();
+  // Trust domain may be elided if it's equal to "cluster.local"
+  if (trust_domain.empty()) {
+    trust_domain = "cluster.local";
+  }
+  // The namespace may be elided if it's equal to "default"
+  if (ns.empty()) {
+    ns = "default";
+  }
+  const auto identity = "spiffe://" + trust_domain + "/ns/" + workload.namespace_() + "/sa/" +
```
kyessenov, line 59: Use `absl::StrCat`.

kyessenov (changes requested):
Minor code comments.

source/extensions/common/workload_discovery/api.cc
keithmattix: I think that worked thanks!

source/extensions/common/workload_discovery/api.cc
```diff
@@ -46,10 +46,25 @@ Istio::Common::WorkloadMetadataObject convert(const istio::workload::Workload& w
   default:
     break;
   }
+
+  constexpr absl::string_view defaultNs = "default";
```
kyessenov, line 50: can you pull them out of and use correct style, e.g. `constexpr absl::string_view DefaultNamespace = ...`

source/extensions/common/workload_discovery/api.cc
```diff
@@ -46,10 +46,25 @@ Istio::Common::WorkloadMetadataObject convert(const istio::workload::Workload& w
...
+  absl::string_view trust_domain = workload.trust_domain();
+  // Trust domain may be elided if it's equal to "cluster.local"
+  if (trust_domain.empty()) {
+    trust_domain = defaultTrustDomain;
+  }
+  // The namespace may be elided if it's equal to "default"
+  if (ns.empty()) {
+    ns = defaultNs;
+  }
+  const auto identity = absl::StrCat("spiffe://", trust_domain, "/ns/" + workload.namespace_(),
```
kyessenov, line 62: "/ns/" + <--- please fix

test/envoye2e/driver/otel.go
```diff
@@ -114,6 +114,7 @@ func (v *verify) Run(p *Params) error {
 		return fmt.Errorf("timed out waiting for all metrics to match")
 	}
 }
+
```
kyessenov, line 117: revert?

test/envoye2e/tcp_metadata_exchange/tcp_metadata_exchange_test.go
```diff
@@ -66,27 +66,30 @@ func TestTCPMetadataExchange(t *testing.T) {
...
 workload_name: productpage-v1
 workload_type: DEPLOYMENT
 canonical_name: productpage-v1
 canonical_revision: version-1
 cluster_id: client-cluster
 uid: //v1/pod/default/productpage
-`}, {
-						Address: "127.0.0.2",
-						Metadata: `
+`,
```
kyessenov, line 80: revert superfluous changes please

test/envoye2e/driver/otel.go
keithmattix: They were added by make lint; I'll try and revert

source/extensions/common/workload_discovery/api.cc
keithmattix: Pull them out to where? I skimmed through the Envoy style guide and didn't see what the proper scoping was. I'll add them to the namespace, but let me know if they need to be elsewhere

kyessenov (approved):

PR #5546, opened 2024-05-14
separate Stackdriver code

kyessenov:
Issue: https://github.com/istio/istio/issues/50808
Minor refactoring to enable removal of Stackdriver + Opencensus code from OSS build.

The plan is to follow-up with removal of the following directories:
- extensions/access_log_policy
- extensions/stackdriver
- test/envoye2e/stackdriver
- testdata/stackdriver

The notable change is to move "peer info" associated with GCP to a private implementation. That means things like "mesh_id" are no longer populated for peer metadata. They were never used in GCP or OSS code anyways, so this helps with removing the overhead.

It is still necessary to use flatbuffers for "peer info" because Wasm cannot easily lookup native structs and that requires a major refactoring to address.


extensions/stackdriver/stackdriver.cc
```diff
@@ -217,25 +217,16 @@ void clearTcpMetrics(::Wasm::Common::RequestInfo& request_info) {
...
+    if (absl::StartsWith(mesh_id, "proj-") || absl::StartsWith(mesh_id, "projects/")) {
+      // do nothing
+      return mesh_id;
+    }
     // Insert or update mesh id to default format as it is missing, empty, or
     // not properly set.
     auto project_number = getProjectNumber();
-    auto* mesh_id_field = (*node.mutable_fields())["MESH_ID"].mutable_string_value();
-    if (!project_number.empty()) {
-      *mesh_id_field = absl::StrCat("proj-", project_number);
```
lei-tang, line 235: In extensions/stackdriver/metric/record.cc, local_node_info.mesh_id() is used in the "mesh_uid" metric label: 

  TagKeyValueList outboundMap = {
      {meshUIDKey(), unknownIfEmpty(flatbuffers::GetString(local_node_info.mesh_id()))},

Will the code update in this PR change the behavior of the "mesh_uid" metric label?

howardjohn (approved):

extensions/stackdriver/stackdriver.cc
kyessenov: No, `local_node_info` is derived differently than peer node info and is fully populated.
We don't report the peer mesh ID so we don't have to send it to peers.

zirain (commented):
LGTM, defer to @lei-tang .

lei-tang (approved):

PR #5547, opened 2024-05-14
stats: turn down v8 testing

kyessenov:
Change-Id: I1da35dd245bbb1ebe7578f9077efad9de20b2724

Running stats.wasm is no longer recommended and not supported by Istio.


keithmattix:
If stats.wasm == https://istio.io/latest/docs/reference/config/proxy_extensions/wasm_telemetry/ then yeah I'm 100% on board with removing it. We should probably change the warnings/explicitly not recommend it on this page anymore as well

lei-tang (approved):

PR #5573, opened 2024-05-23
Re-implement RotatingScope with thread local storage

wangjian-pg:

The current implementation of metric rotation relies on timing to ensure that the scope pointer used by the worker thread is valid, which is probably not a good practice or design as far as I know.

After digging into the envoy's implementation of `ThreadLocalStoreImpl`(which faces similar issues of how to propagate scoped metrics to worker threads and expire them when the scope is released), I found that storing the active scope reference in thread local storage and replacing it with the new one when rotation occurs might be a better design. With this approach, we eliminate the use of raw pointers, making rotation behavior more predictable and still lock-free.

I re-implemented the rotation as described and tested the code in my environment,  it works as expected. Please take a look. Looking forward to any feedback.

### DIFF
--- a/source/extensions/filters/http/istio_stats/istio_stats.cc
+++ b/source/extensions/filters/http/istio_stats/istio_stats.cc
@@ -427,11 +427,10 @@ public:
                 uint64_t delete_interval_ms)
       : parent_scope_(factory_context.scope()), active_scope_(parent_scope_.createScope("")),
         tls_scope_(factory_context.serverFactoryContext().threadLocal()),
-        rotate_interval_ms_(rotate_interval_ms),
-        delete_interval_ms_(delete_interval_ms) {
+        rotate_interval_ms_(rotate_interval_ms), delete_interval_ms_(delete_interval_ms) {
 
-    tls_scope_.set([&scope = *active_scope_](Event::Dispatcher&){
-        return std::make_shared<TlsCachedScope>(scope);
+    tls_scope_.set([&scope = *active_scope_](Event::Dispatcher&) {
+      return std::make_shared<TlsCachedScope>(scope);
     });
 
     if (rotate_interval_ms_ > 0) {
@@ -456,8 +455,8 @@ public:
   Stats::Scope& scope() { return tls_scope_->_scope; }
 
 private:
-  struct TlsCachedScope:ThreadLocal::ThreadLocalObject{
-    TlsCachedScope(Stats::Scope& scope):_scope(scope) {};
+  struct TlsCachedScope : ThreadLocal::ThreadLocalObject {
+    TlsCachedScope(Stats::Scope& scope) : _scope(scope){};
     std::reference_wrapper<Stats::Scope> _scope;
   };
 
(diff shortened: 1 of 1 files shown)


wangjian-pg:
@lei-tang I have formatted the changes with clang-lint, and all tests have passed. Could you please review the code?

lei-tang (commented):
@wangjian-pg Thank you for your PR!

Can you provide a concrete example to describe the problems of current implementation?

wangjian-pg:
@lei-tang The current implementation generally works well for most regular cases and under normal circumstances. 

However, consider a scenario where thousands of processes are running on a host with limited resources, such as a system with only two CPU cores. In this case,  the operating system may schedule the worker thread off the CPU core when it attempts to allocate a new gauge metric with the currently active scope pointer through the function call `Stats::Utility::counterFromStatName`. Meanwhile, the main thread continues to run on another CPU core. In a heavily loaded system, there is a possibility that the worker thread may not get any CPU time for a prolonged period (more than 1 second). Subsequently, when the worker thread is scheduled back onto a CPU core and continues the logic processing of the function `Stats::Utility::counterFromStatName`, the scope pointer has already expired and been released by the main thread. This leads to a problem of working with a dangling pointer.

The real problem is that it relies on timing to ensure that the scope pointer used by the worker threads is valid and lacks any other synchronization mechanism. This behavior is non-deterministic in a time-sharing operating system.

By the way, as far as I know, using std::atomic may introduce a memory barrier and could potentially lead to performance degradation. However, I haven't yet benchmarked this and it shouldn't be a big deal.

jezhang2014:
If the http.stats plugin is used as the statistics plugin, when the onDelete callback is triggered,  Is it possible that all the historical statistical data will be cleared? if so, this could lead to a sudden change in the statistical data of the monitoring system, potentially triggering alerts.

To avoid this issue, it would be better to only clear the historical statistical data for targets (e.g., Pods) that have not received requests for a long time, while leaving the statistical data for targets that are still actively receiving requests unaffected. This would help maintain the continuity of the statistical data and prevent unnecessary alerts.

wangjian-pg:
@jezhang2014
I believe it's not a problem of dropping historical stats data. Whether the envoy proxy is deployed as a sidecar or a gateway, it can be terminated at any time for reasons like upgrades or scale-down. This is common in dynamic cloud environments, such as with k8s HPA. The in-memory stats data would be lost whenever the envoy proxy is terminated or restarted. This shouldn't be a problem, and it's the responsibility of the monitoring system to handle this situation. 

AFAIK, the widely adopted monitoring system `Prometheus` has a built-in function to detect stats data reset. You may find more information here: https://prometheus.io/docs/prometheus/latest/querying/functions/#resets

wangjian-pg:
@kyessenov Cloud you please take a look at this PR?

wangjian-pg:
@kyessenov Could you please take a look at this PR?

zirain:
not stale

kyessenov:
Expiration has been added to the core Envoy stats, so this PR is no longer needed.

PR #5604, opened 2024-06-11
Allow versionless cluster name match in stats

howardjohn:
For https://github.com/istio/istio/pull/51503, but also required even if
we don't change; for inbound HBONE we need to add initial support for
passthrough (https://github.com/istio/istio/issues/51336). I don't want
to start with legacy hacks on new code, so that will for sure use InboundPassthroughCluster


zirain (approved):

howardjohn:
/cherrypick release-1.22

I think would be nice since its extremely low risk, but will ensure proper stats handling when users upgrade the control plane first.

zirain:
> /cherrypick release-1.22
> 
> I think would be nice since its extremely low risk, but will ensure proper stats handling when users upgrade the control plane first.

this's a special case for backporting new feature.

howardjohn:
not really part of a feature but of a bug fix for HBONE (https://github.com/istio/istio/issues/51336), but its ok if we have broken telemetry for this state I suppose.

PR #5617, opened 2024-06-17
Use cluster metadata for destination_service_name

keithmattix:
**What this PR does / why we need it**: Fixes istio/istio#49842

~I'm setting the field to unknown right now just to confirm that we're retrieving the namespace from metadata correctly. Before we merge, it should probably fall back to peer metadata~ The only two tests that were failing were for clusters with no metadata; we're good to go here.

### DIFF
--- a/source/extensions/filters/http/istio_stats/istio_stats.cc
+++ b/source/extensions/filters/http/istio_stats/istio_stats.cc
@@ -1234,8 +1234,11 @@ private:
       tags_.push_back({context_.destination_service_name_, service_host_name.empty()
                                                                ? context_.unknown_
                                                                : pool_.add(service_host_name)});
-      tags_.push_back({context_.destination_service_namespace_,
-                       !service_namespace.empty() ? pool_.add(service_namespace) : context_.unknown_});
+      tags_.push_back({context_.destination_service_namespace_, !service_namespace.empty()
+                                                                    ? pool_.add(service_namespace)
+                                                                    : (peer ?
+                                                                        pool_.add(peer->namespace_name_)
+                                                                        : context_.unknown_)});
       tags_.push_back({context_.destination_cluster_, peer && !peer->cluster_name_.empty()
                                                           ? pool_.add(peer->cluster_name_)
                                                           : context_.unknown_});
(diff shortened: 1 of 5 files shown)


howardjohn (commented):
LGTM with fallback to peer_namespace

zirain:
LGTM, can you add a test for it?

keithmattix:
The test is already kind of there; do you just want to see a diff between endpoint metadata and cluster metadata to prove the latter takes priority?

zirain:
> The test is already kind of there; do you just want to see a diff between endpoint metadata and cluster metadata to prove the latter takes priority?

yeah, that's what I mean.

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -1230,7 +1235,9 @@ class IstioStatsFilter : public Http::PassThroughFilter,
                                                                ? context_.unknown_
                                                                : pool_.add(service_host_name)});
       tags_.push_back({context_.destination_service_namespace_,
-                       !peer_namespace.empty() ? pool_.add(peer_namespace) : context_.unknown_});
+                       !service_namespace.empty()
+                           ? pool_.add(service_namespace)
+                           : (peer ? pool_.add(peer->namespace_name_) : context_.unknown_)});
```
zirain, line 1240: why change this from `!peer_namespace.empty() ? pool_.add(peer_namespace) : context_.unknown_`?

keithmattix: We were relying exclusively on endpoint metadata for destination_service_namespace. This makes sense for workload based metrics, but for service metrics, it has the unfortunate side effect of losing information when there are no healthy backends. So instead, we petard with using cluster metadata if it exists, falling back to endpoint metadata after that

zirain: just L1240, why not use prev expression?

keithmattix: Ah I see what you mean; I'll switch it around so it's consistent

keithmattix:
> > The test is already kind of there; do you just want to see a diff between endpoint metadata and cluster metadata to prove the latter takes priority?
> 
> yeah, that's what I mean.

Done!

zirain (approved):

PR #5712, opened 2024-08-02
CI: Add bazel extra args variable to prow scripts

kb000:
**What this PR does / why we need it**:  Add BAZEL_BUILD_EXTRA_ARGS as a prow build customization point. In at least one use case, this is used to configure a bazel remote cache:

Per https://github.com/envoyproxy/envoy/blob/97bc67a06a3908821c933b437f9268886383d210/ci/README.md?plain=1#L129-L141
> To leverage a bazel remote cache:
> * add bazel options like --remote_cache and/or --remote_cache_header to a .bazelrc file, such as user.bazelrc. For
> example
> `build:my-remote-cache --remote_cache=grpcs://remotecache.googleapis.com`
> `build:my-remote-cache --remote_cache_header=Authorization="Bearer <token>"`
> * specify the config in the BAZEL_BUILD_EXTRA_OPTIONS environment variable. Run
> `export BAZEL_BUILD_EXTRA_OPTIONS=--config=my-remote-cache

So in a prow job spec:

```
  spec:
    containers:
      - name: build-container
        image: ...
        args:
        - ./prow/proxy-presubmit.sh
        env:
        - name: BAZEL_BUILD_EXTRA_ARGS
          value: >-
            --config=remote-cache
```

**Fixes**: N/A

**Special notes for your reviewer**: This PR does not include .bazelrc changes, since this public repo's CI seems tuned for  GCP remote execution. In organizations not using GCP, the bazel remote cache is an easy way to reduce CI build times.

### DIFF
--- a/prow/proxy-common.inc
+++ b/prow/proxy-common.inc
@@ -31,7 +31,8 @@ GOPATH=/home/prow/go
 ROOT=/go/src
 
 # Configure available resources and disable IPv6 tests.
-export BAZEL_BUILD_ARGS="--verbose_failures --test_env=ENVOY_IP_TEST_VERSIONS=v4only --test_output=errors"
+BAZEL_BUILD_ARGS="${BAZEL_BUILD_EXTRA_ARGS:-} --verbose_failures --test_env=ENVOY_IP_TEST_VERSIONS=v4only --test_output=errors"
+export BAZEL_BUILD_ARGS="${BAZEL_BUILD_ARGS# }"
 
 # Override envoy.
 if [[ "${ENVOY_REPOSITORY:-}" && "${ENVOY_PREFIX:-}" ]]; then


keithmattix:
Seems reasonable.

/ok-to-test

zirain:
why not use `user.bazelrc`?

```
build --remote_cache=grpc://xxxx:9092
build --remote_timeout=3600
build --repo_env=https_proxy=http://xxxxx:7890
build --repo_env=http_proxy=http://xxxx:7890

build --build_tag_filters=-skip_on_windows

# build --override_repository=envoy=/Users/zirain/Codes/istio.io/envoy

build:macos --host_action_env=CXXFLAGS=-Wno-int-conversion
build:macos --action_env=CXXFLAGS=-Wno-int-conversion
build:macos --host_action_env=CFLAGS=-Wno-int-conversion
build:macos --action_env=CFLAGS=-Wno-int-conversion
```

kb000:
> why not use `user.bazelrc`?
> 
> ```
> build --remote_cache=grpc://xxxx:9092
> ```

This enables the remote cache on all builds, whether they're in CI or not.  This isn't the desired outcome.

zirain:
cc @kyessenov he's expert

kb000:
PTAL @noah8713

noah8713:
Thanks @kb000

kb000:
Hi @kyessenov @zirain please take another look and approve when you get a chance.

prow/proxy-common.inc
```diff
@@ -31,7 +31,8 @@ GOPATH=/home/prow/go
 ROOT=/go/src
 
 # Configure available resources and disable IPv6 tests.
-export BAZEL_BUILD_ARGS="--verbose_failures --test_env=ENVOY_IP_TEST_VERSIONS=v4only --test_output=errors"
+BAZEL_BUILD_ARGS="${BAZEL_BUILD_EXTRA_ARGS:-} --verbose_failures --test_env=ENVOY_IP_TEST_VERSIONS=v4only --test_output=errors"
```
kyessenov, line 34: Why not just append to BAZEL_BUILD_ARGS?

kb000: Are you asking me to use "+=" instead of interpolation? Bazel is finicky about whitespace between arguments and I'll still need to trim the string to keep it happy (using [`${...# } or ${...% }`](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html#:~:text=the%20last%20element.-,%24%7Bparameter%23word%7D,-%24%7Bparameter%23%23)).

kb000: If you're saying, assume `BAZEL_BUILD_ARGS` might have something interesting already in it, and expose that as the injection point: I think saying `EXTRA` in the variable name helps the user understand this isn't all that will be passed to bazel, it's what goes in above and beyond the defaults.

kyessenov (approved):
I'm not an expert in bash, so I trust you know what you are doing better than me :)

PR #5744, opened 2024-08-19
istio_build metric should not expiry

zirain:
fix: https://github.com/istio/istio/issues/52284


zirain:
cc @kyessenov

kyessenov:
Looks fine. Can we move `recordVersion` into the constructor and then pass the factory context as a parameter instead of a field?

zirain:
> Looks fine. Can we move `recordVersion` into the constructor and then pass the factory context as a parameter instead of a field?

done

kyessenov (approved):

PR #5786, opened 2024-09-11
Improve performance of AccessLog::Instance in Istio Stats Filter

danehans:
- Replaced manual tag insertion with `emplace_back()` for better performance and code clarity.
- Added missing include for `source/common/http/codes.h` to utilize `Http::CodeUtility::toString()`.
- Changed `grpc_response_status` from `absl::string_view` to `Stats::StatName` to ensure proper type usage in tags.
- Optimized tag handling by clearing the `tags_` vector at the start of each log call to avoid appending to existing entries.
- Replaced manual conversion of response flags with `StreamInfo::ResponseFlagUtils::toShortString()`.
- Ensured proper recording of metrics and custom metrics, and added a final report helper call for cleanup.

### DIFF
--- a/source/extensions/filters/http/istio_stats/istio_stats.cc
+++ b/source/extensions/filters/http/istio_stats/istio_stats.cc
@@ -23,6 +23,7 @@
 #include "extensions/common/metadata_object.h"
 #include "parser/parser.h"
 #include "source/common/grpc/common.h"
+#include "source/common/http/codes.h"
 #include "source/common/http/header_map_impl.h"
 #include "source/common/http/header_utility.h"
 #include "source/common/network/utility.h"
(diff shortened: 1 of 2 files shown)


source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -826,38 +827,42 @@ class IstioStatsFilter : public Http::PassThroughFilter,
 
   // AccessLog::Instance
   void log(const Formatter::HttpFormatterContext& log_context,
-           const StreamInfo::StreamInfo& info) override {
+          const StreamInfo::StreamInfo& info) override {
+    // Clear the tags vector to ensure we are not appending to existing tags from a previous log call.
```
kyessenov, line 831: Are you using the periodic access logger?

danehans:
@kyessenov thanks for the quick review. For additional context, we performed ambient waypoint performance benchmarking and [this flamegraph](https://flamegraph.com/share/5dafdc5c-5e6d-11ef-ab1b-b6898751f631) indicated higher-than-expected CPU utilization of Istio stats.

source/extensions/filters/http/istio_stats/istio_stats.cc
danehans: I am using a stock 1.23 Istio ambient install. I am new to the codebase so I can omit `tags_.clear();` if the logger guarantees a fresh state for each call.

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -826,38 +827,42 @@ class IstioStatsFilter : public Http::PassThroughFilter,
...
-      tags_.push_back({context_.request_protocol_, context_.http_});
-    }
+    // Evaluate the end stream override expressions for HTTP. This may change values for periodic metrics.
+    stream_.evaluate(info, request_headers, response_headers, response_trailers);
 
-    // TODO: copy Http::CodeStatsImpl version for status codes and flags.
-    tags_.push_back(
-        {context_.response_code_, pool_.add(absl::StrCat(info.responseCode().value_or(0)))});
+    // Set the request protocol tag.
+    tags_.emplace_back(context_.request_protocol_, is_grpc_ ? context_.grpc_ : context_.http_);
```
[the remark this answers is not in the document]
danehans: From my understanding, `emplace_back` constructs a new element at the end of the vector, directly in place, without needing to create a temporary object first. Isn't this more efficient than `push_back`?

kyessenov:
> @kyessenov thanks for the quick review. For additional context, we performed ambient waypoint performance benchmarking and [this flamegraph](https://flamegraph.com/share/5dafdc5c-5e6d-11ef-ab1b-b6898751f631) indicated higher-than-expected CPU utilization of Istio stats.

Thanks for sharing. Around 6% overhead seems in line with my observations. It used to be closer to 10-20% with a previous implementation, so I'm not sure how much further we can push it without changing its functionality.

source/extensions/filters/http/istio_stats/istio_stats.cc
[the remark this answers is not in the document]
kyessenov: push_back has a move variant which I think would be used here whenever possible. emplace_back is only better if you have a big object to pass and want to skip a copy, but these are all small structs.

source/extensions/filters/http/istio_stats/istio_stats.cc
kyessenov: log() is only called once normally. However, Envoy added a new periodic reporter that would call log() repeatedly (instead of a custom internal timer). We haven't integrated with that periodic reporting yet.

kyessenov:
It would help to have a micro benchmark that shows a significant performance improvement? It's not clear what the gains are.

danehans:
Micro benchmarks for [1.23.0](https://flamegraph.com/share/44fb9785-707c-11ef-8d9a-2e2d4d104be7) and [this](https://flamegraph.com/share/02d9bb6a-707d-11ef-8d9a-2e2d4d104be7) PR show little-to-no noticeable change in CPU usage related to the AccessLog::Instance for the Istio Stats Filter.

PR #5825, opened 2024-09-23
Refactor metadata exchange in preparation for upstreaming

kyessenov:
Change-Id: I1abef439650eebf6b9a8f26d078ac5ba05f0ba23

https://github.com/envoyproxy/envoy/issues/29681

Major refactor to prepare the code for upstream maintenance. The core changes are:

1. Remove Wasm/flatbuffers dependency and use the standard Envoy formatters for peer metadata access. This removes unnecessary round-trippers through flatbuffers on the datapath. In CEL, instead of `downstream_peer.name` or `filter_state[wasm.downstream_peer].name`, use `filter_state[downstream_peer].name`. This now works everywhere, including Envoy formatters, and does not need  Wasm code for lookup and generation. 
2. Deprecate `node` attribute that only works in Wasm, following the upstream [Envoy deprecation announcement](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#wasm-attributes). Use `xds.node` attribute instead.


keithmattix:
I think I'm generally in favor of this direction. Based on the code though, it looks like these changes will modify the shape/labels of telemetry; am I understanding that correctly?

kyessenov:
@keithmattix There is limited impact on the output telemetry:

1. app/version and canonical app/version are now combined, so if they were distinct, that changes. In general, this is a good thing as it is confusing to have two nomenclatures.

2. CEL expressions in the telemetry API are impacted. We are not loosing information but the syntax changes.

keithmattix:
@kyessenov in that case, can we make the merged fields + CEL changes alongside the current telemetry, deprecate the old versions for a release or two, and then make the hard cut. With telemetry API at v1, we don't need to break users

kyessenov:
> @kyessenov in that case, can we make the merged fields + CEL changes alongside the current telemetry, deprecate the old versions for a release or two, and then make the hard cut. With telemetry API at v1, we don't need to break users

I don't think you can have two parallel ways with the way this works internally. We never claimed the stability of "peer attributes" as long as I remember. I think a better way is to recommend a backfill CEL expression, e.g. `has(filter_state, "downstream_peer") ? <new way> : <old way>` which would work with both old and new versions.

keithmattix:
I thought as much but figured I'd ask. I still think this needs to land after branch cut though so users aren't surprised during upgrade

kyessenov:
> I thought as much but figured I'd ask. I still think this needs to land after branch cut though so users aren't surprised during upgrade

Can you elaborate? What branch cut? BTW how are you keeping CEL vocabulary stable across Envoy versions? That is not part of xDS contract and subject to change.

keithmattix:
I mean 1.24 branch cut; we should have at least one release where we notify but not remove. CEL is a bit of an outlier (I don't think the telemetry api CEL field got promoted) but generally, a user shouldn't have a v1 resource in one release that breaks in the next release with no advance notification. Any compatibility policy that breaks that experience is a broken policy IMO

howardjohn:
Do we *have* to do the change to the metrics labels or was it convenient? I
am pretty uncomfortable with that aspect of the change.

kyessenov:
> Do we *have* to do the change to the metrics labels or was it convenient? I am pretty uncomfortable with that aspect of the change.

No, we don't have to. I can undo it, and we can sort it out later. CEL changes (deprecation of `node`) were announced on Envoy pages a long time ago: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#wasm-attributes

keithmattix:
In that case, I'm more ok with the CEL changes. We can point to the upstream Envoy deprecation

kyessenov:
Undid the app/version change, so the impact is restricted to CEL and internal filter state naming. Once this goes upstream, CEL and filter state names are likely to become stable.

zirain:
we finnally get rid of flattbuffer?

kyessenov:
> we finnally get rid of flattbuffer?

Yes!

zirain (approved):
I think this is a good trend, but changes need to be mentioned in release notes(v1.24).

PR #5836, opened 2024-09-27
Fix the go toolchain version

jwendell:
It needs to be `x.y.z`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module istio.io/proxy
 
-go 1.22
+go 1.22.0
 
 require (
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

kyessenov (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module istio.io/proxy
 
-go 1.22
+go 1.22.0
```
zirain, line 3: should we bump to 1.23?

howardjohn: This is the minimum version -- there is no need to force 1.23

PR #5915, opened 2024-11-05
implement serializeAsProto

zirain:
xref: https://github.com/istio/istio/issues/53769


howardjohn (commented):
I don't have much context here but seems reasonable to me..

kyessenov (approved):

keithmattix (approved):

PR #5968, opened 2024-11-30
Fix tcp telemetry for service-oriented waypoints

keithmattix:
The metadata_exchange filter only looks at downstream metadata; this PR configures it to check upstream metadata and store it in filter state (with the appropriate key) if it exists. Note that workload oriented waypoints are still broken because those endpoints have no metadata. /cc @howardjohn on where to make the control plane adjustments there. 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: part of istio/istio#53593

Replaces #5963 for adding a compdb script; if you want to review #5963 on its own; I'll rebase this PR once that one is done.


source/extensions/filters/network/metadata_exchange/config/metadata_exchange.pb.html
zirain, line 1: reasonable change?

keithmattix: I think these are proto generator changes that @howardjohn or @craigbox have been working on

craigbox: That is from https://github.com/istio/tools/pull/3064.  There will be more coming from my recent PRs, which will remove some `<th>` and `<td>` and move things around.

zirain: better to keep it separated.

zirain (commented):
LGTM, can you rebase and just keep the core logic in this PR?

cc @howardjohn @kyessenov

zirain (approved):

PR #6028, opened 2024-12-28
Add metadata fallback for client sidecar reporters

keithmattix:
**What this PR does / why we need it**:  Companion to istio/istio#54383. Use MX fallback if there's no ssl context for the upstream connection (there won't be when we're doing CONNECT encap due to the extra internal listener in between).  Note that this doesn't help with the source principal; we may be able to set some filter state in the connect_encap listener and have stats read it before we flush, but I'd need to try it out.

### DIFF
--- a/source/extensions/filters/http/istio_stats/istio_stats.cc
+++ b/source/extensions/filters/http/istio_stats/istio_stats.cc
@@ -1113,7 +1113,7 @@ private:
       }
       if (peer_san.empty()) {
         std::optional<Istio::Common::WorkloadMetadataObject> endpoint_peer;
-        const auto* endpoint_object = peerInfo(config_->reporter(), filter_state);
+        const auto endpoint_object = peerInfo(config_->reporter(), filter_state);
         if (endpoint_object) {
           peer_san = endpoint_object->identity_;
         }
--- a/source/extensions/filters/http/istio_stats/istio_stats.cc
+++ b/source/extensions/filters/http/istio_stats/istio_stats.cc
@@ -1076,7 +1076,7 @@ private:
       }
     }
 
-    absl::string_view peer_san;
+    std::string peer_san;
     absl::string_view local_san;
     switch (config_->reporter()) {
     case Reporter::ServerSidecar:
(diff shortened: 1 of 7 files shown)


keithmattix:
Long term, I need to see if I can create a lasso of ssl contexts so we don't need WDS fallback for peer principal. I recall @kyessenov saying it's possible, and the way we get http status codes may be a clue

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -1074,9 +1075,18 @@ class IstioStatsFilter : public Http::PassThroughFilter,
       if (ssl_info && !ssl_info->uriSanPeerCertificate().empty()) {
         peer_san = ssl_info->uriSanPeerCertificate()[0];
       }
+      // This won't work for sidecar/ingress -> ambient becuase of the CONNECT
+      // tunnel.
       if (ssl_info && !ssl_info->uriSanLocalCertificate().empty()) {
         local_san = ssl_info->uriSanLocalCertificate()[0];
       }
+      if (peer_san.empty()) {
```
zirain, line 1083: move this to L1078 or L1076?

zirain:
is this covered by a test?

keithmattix:
> is this covered by a test?

Ah forgot to add one in the flurry of PRs; will do

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -1076,7 +1076,7 @@ class IstioStatsFilter : public Http::PassThroughFilter,
       }
     }
 
-    absl::string_view peer_san;
+    std::string peer_san;
```
zirain, line 1079: same thing borther me a lot in the past

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -1107,9 +1107,19 @@ class IstioStatsFilter : public Http::PassThroughFilter,
     case Reporter::ClientSidecar: {
       const Ssl::ConnectionInfoConstSharedPtr ssl_info =
           info.upstreamInfo() ? info.upstreamInfo()->upstreamSslConnection() : nullptr;
+      std::optional<Istio::Common::WorkloadMetadataObject> endpoint_peer;
       if (ssl_info && !ssl_info->uriSanPeerCertificate().empty()) {
         peer_san = ssl_info->uriSanPeerCertificate()[0];
       }
+      if (peer_san.empty()) {
+        auto endpoint_object = peerInfo(config_->reporter(), filter_state);
```
zirain, line 1115: can we reuse L1016?

keithmattix: Oh wow thought I responded to this. No we can't reuse because of scoping and the nested switches

zirain (approved):

PR #6030, opened 2024-12-29
MX support additional labels

zirain:
xref: https://github.com/istio/istio/issues/54000

### DIFF
--- a/extensions/common/metadata_object.cc
+++ b/extensions/common/metadata_object.cc
@@ -90,6 +90,9 @@ Envoy::ProtobufTypes::MessagePtr WorkloadMetadataObject::serializeAsProto() cons
   if (!app_version_.empty()) {
     (*message->mutable_fields())[AppVersionToken].set_string_value(app_version_);
   }
+  if (!identity_.empty()) {
+    (*message->mutable_fields())[IdentityToken].set_string_value(identity_);
+  }
 
   if (!labels_.empty()) {
     auto* labels = (*message->mutable_fields())[LabelsToken].mutable_struct_value();
--- a/extensions/common/metadata_object.cc
+++ b/extensions/common/metadata_object.cc
@@ -282,6 +282,8 @@ WorkloadMetadataObject::getField(absl::string_view field_name) const {
       return instance_name_;
     }
   }
+  // This is a workaround for FilterStateObjectWrapper.
+  // The field name is the label key.
   if (!labels_.empty()) {
     for (const auto& l : labels_) {
       if (l.first == std::string(field_name)) {
--- a/extensions/common/metadata_object.cc
+++ b/extensions/common/metadata_object.cc
@@ -62,11 +62,42 @@ absl::optional<absl::string_view> toSuffix(WorkloadType workload_type) {
 
 Envoy::ProtobufTypes::MessagePtr WorkloadMetadataObject::serializeAsProto() const {
   auto message = std::make_unique<Envoy::ProtobufWkt::Struct>();
-  auto& fields = *message->mutable_fields();
-  const auto parts = serializeAsPairs();
-  for (const auto& p : parts) {
-    fields[std::string(p.first)] = Envoy::ValueUtil::stringValue(std::string(p.second));
+  const auto suffix = toSuffix(workload_type_);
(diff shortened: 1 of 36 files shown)


extensions/common/metadata_object.cc
```diff
@@ -101,6 +101,11 @@ WorkloadMetadataObject::serializeAsPairs() const {
   if (!app_version_.empty()) {
     parts.push_back({AppVersionToken, app_version_});
   }
+  if (!labels_.empty()) {
+    for (const auto& l : labels_) {
+      parts.push_back({absl::string_view(l.first), absl::string_view(l.second)});
```
howardjohn, line 106: this makes it impossible to distinguish between a raw label and a well-known field. Since they can make a label with key like `cluster_name_` (for example)

source/extensions/filters/http/peer_metadata/config.proto
```diff
@@ -75,4 +75,13 @@ message Config {
 
   // True to enable sharing with the upstream.
   bool shared_with_upstream = 5;
+
+  // Additional labels to be added to the peer metadata.
+  // By default, following labels are added:
+  // - service.istio.io/canonical-name
```
howardjohn, line 81: nit: these are bad labels to add here since they are already part of the typed data. Maybe we should add a more realistic example?

zirain: I meant to say that these labels should not be added to additional labels, will update the doc to make it more clearer.

extensions/common/metadata_object.cc
zirain: yeah, need to bring back `CelStateHashable` instead of `FilterStateObjectWrapper` to support expr like `filter_state.upstream_peer.labels["role"]`.

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -122,13 +123,40 @@ bool peerInfoRead(Reporter reporter, const StreamInfo::FilterState& filter_state
...
   const auto& filter_state_key =
       reporter == Reporter::ServerSidecar || reporter == Reporter::ServerGateway
           ? Istio::Common::DownstreamPeer
           : Istio::Common::UpstreamPeer;
-  return filter_state.getDataReadOnly<Istio::Common::WorkloadMetadataObject>(filter_state_key);
+  // This's a workaround before FilterStateObject support operation like `.labels['role']`.
+  // The workaround is to use CelState to store the peer metadata.
+  // Rebuild the WorkloadMetadataObject from the CelState.
+  const auto* cel_state =
+      filter_state.getDataReadOnly<Envoy::Extensions::Filters::Common::Expr::CelState>(
```
keithmattix, line 136: Does this have control plane implications? Will calling filters have to specify a CEL formatter?

source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -122,13 +123,40 @@ bool peerInfoRead(Reporter reporter, const StreamInfo::FilterState& filter_state
...
+    return {};
+  }
+
+  ProtobufWkt::Struct obj;
+  if (!obj.ParseFromString(absl::string_view(cel_state->value()))) {
+    return {};
+  }
+
+  Istio::Common::WorkloadMetadataObject peer_info(
+      extractString(obj, Istio::Common::InstanceNameToken),
```
keithmattix, line 148: No labels?

keithmattix (commented):
Looks good at a high level; just a couple of questions

source/extensions/filters/http/istio_stats/istio_stats.cc
zirain: this is used before v1.24, so i think it wont.

source/extensions/filters/http/istio_stats/istio_stats.cc
zirain: labels won't be used in the next, unneed to rebuild labels here.

keithmattix (approved):
LGTM - thanks for the tests; they made me feel comfortable approving

PR #6066, opened 2025-01-14
Disable spammy logs in release

howardjohn:
The `debug` flag does not impact the result of the build, but it does turn on ~every logging option in bazel resulting in 200k lines of logs each run

### DIFF
--- a/scripts/release-binary.sh
+++ b/scripts/release-binary.sh
@@ -134,7 +134,7 @@ do
       fi
       ;;
     "debug")
-      CONFIG_PARAMS="--config=debug"
+      CONFIG_PARAMS=""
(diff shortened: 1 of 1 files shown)


scripts/release-binary.sh
```diff
@@ -134,7 +134,7 @@ do
       fi
       ;;
     "debug")
-      CONFIG_PARAMS="--config=debug"
+      CONFIG_PARAMS=""
```
keithmattix, line 137: Looks like something's not quite right; the binary doesn't exist where bazel expects. Try 
```suggestion
      CONFIG_PARAMS="-c dbg"
```

howardjohn: thanks seems to work now!

PR #6067, opened 2025-01-14
bump golang.org/x/net to v0.33.0

zmiklank:
to address CVE-2024-45338


keithmattix:
Ok-to-test

keithmattix (commented):
I'm pretty sure the only Go usage in this repo is the test runner, so I'm inclined to say this bulk doesn't apply

jwendell (approved):
@keithmattix even if it's only used in tests, there's no harm in update these deps and keep the repo up to date IMO...

keithmattix:
/ok-to test
(whoops)

keithmattix:
> @keithmattix even if it's only used in tests, there's no harm in update these deps and keep the repo up to date IMO...

That's fair - I just don't want someone to come back later and think we were affected by this vuln based on the description. This looks good to me if the CLA is signed and the description is modified a bit

zmiklank:
Hello, thank you for looking at the PR. I have modified the description and signed the CLA.

zirain (approved):

PR #6068, opened 2025-01-14
Fixes for waypoint stat generation

howardjohn:
This PR fixes a few issue in waypoint stat generation:

* Destination service namespace: before it hardcoded the waypoint namespace. Now it gets it from the cluster metadata (needed for cross-namespace waypoints)
* Destination cluster: before it hardcodes to the waypoint's cluster ,now it uses the destination
* Destination `***`: most of the fields filled in `""` when unset instead of `"unknown"`

These are all addressed and a test case is added

Fixes https://github.com/istio/istio/issues/54668


source/extensions/filters/http/istio_stats/istio_stats.cc
```diff
@@ -1206,10 +1219,10 @@ class IstioStatsFilter : public Http::PassThroughFilter,
         tags_.push_back({context_.destination_service_name_, service_host_name.empty()
                                                                  ? context_.canonical_name_
                                                                  : pool_.add(service_host_name)});
+        tags_.push_back({context_.destination_cluster_, context_.cluster_name_});
+        tags_.push_back({context_.destination_service_namespace_, context_.namespace_});
```
keithmattix, line 1223: This looks like it's for the client sidecar path not servergateway (waypoint)

howardjohn: Before the code was shared for the two. The waypoint part was wrong, so I moved the shared code (prev L1211) into the sidecar path and changed the waypoint code.

Also its ServerSidecar btw, its a double nested `switch` which is slightly confusing

keithmattix: Ah, I see - I misread it in the GitHub diff. Looks like a lot of tests failed though

howardjohn: accidentally re-ordered the lines in the move (which I think is a change that is a NOP except in tests, fwiw). Should be good now.

keithmattix (approved):
LGTM - thanks!

fjglira:
Hi @howardjohn, does this PR need to be cherry-picked to release-1.29 branch?

keithmattix:
No it's old

PR #6070, opened 2025-01-15
Drop protoc docs generation

howardjohn:
This is not used anywhere. The make target also doesn't work (due to
stackdriver removal) and no one complained which suggests no one has ran
it for a long time


zirain:
I recall stats is used here: https://github.com/istio/istio.io/blob/3209057b0f285a8e9523f68ebceabbde4a0c9d56/content/en/docs/reference/config/proxy_extensions/stats/index.html#L5

howardjohn:
Interesting, thanks! I looked at a couple of them and assumed none were, missed that one.

Do we want to? it doesn't really make sense to document an internal interface on our public user-facing docs

zirain:
I think we can remove it, I do believe we have provide first-class API(the telemetry API).

zirain (approved):
Feel free to remove the `DNM` after istio.io fixed.

PR #6107, opened 2025-02-06
reenable cryptomb

zirain:
fixes: https://github.com/istio/istio/issues/54984


keithmattix (approved):

.bazelrc
```diff
@@ -16,6 +16,8 @@ build:remote --remote_timeout=7200
 # Enable libc++ and C++20 by default.
 build:linux --config=libc++20
 
+build:linux --action_env=PATH=/usr/lib/llvm/bin:/usr/local/bin:/bin:/usr/bin:/opt/llvm/bin
```
phlax, line 19: @zirain just spotted the `/opt/llvm/bin` was left - probs best to remove that to avoid confusions

keithmattix: Ack - @zirain can you follow-up?

PR #6114, opened 2025-02-10
Enable envoy.filters.http.file_system_buffer

bacek:

Just add http filter file_system_buffer

Also fixes formatting on `gcp_authn`


zirain:
I'm not sure if this's a common requiement, can you raise this up in WG/TOC meeting?

cc @howardjohn @keithmattix

bacek:
> I'm not sure if this's a common requiement, can you raise this up in WG/TOC meeting?

I do need to fully decouple appserver from the slow clients. FileSystemBuffer on the gateway is the default solution. I can't give infinite amount of memory to the in memory buffer. Are the any other options?

keithmattix:
@bacek at what point in the request/connection lifecycle are you seeing the slow behavior? During listener filter execution? HTTP parsing?

zirain:
I recall that istio want make istio-proxy act as a readonly container as possible(correct me if I'm wrong), in that case, I think you need bring this up to WG meeting or get an approval from TOC.

cc @istio/technical-oversight-committee

bacek:
> I recall that istio want make istio-proxy act as a readonly container as possible(correct me if I'm wrong), in that case, I think you need bring this up to WG meeting or get an approval from TOC.

"Read only" as in "no state stored in the running pod" or "root fs inside pod is ro"? 

In the former case it's not true, because there are in memory buffers which will require graceful shut-down to fully flush. And integration with cloud provider to flush/drain on load balancer (NLB with AlbController on AWS in my case). 

For the latter I was using standard volume/volumeMounts feature to mount ephemeral volume into gateway for the FileSystemBuffer to use. So pod's root fs stays ro.

jwendell:
any decision on this?

zirain:
for any things like these, you could use envoy-contrib as an alternative istio-proxy(possibly starting from 1.30).

PR #6214, opened 2025-04-01
Automator: update envoy@ in istio/proxy@master

zirain:


zirain:
need https://github.com/istio/test-infra/pull/5633

dhawton:
Reopened for tests... they're passing. Will re-close and allow automator to update this overnight

PR #6514, opened 2025-08-28
Adding CEL Samplers #6513

Cam6349:
What this PR does / why we need it:

This PR will expose the CEL sampler for otel. We currently do not have a way to drop traces based on certain conditions in the service mesh. We would like to be able to make intelligent tracing decisions in the service mesh instead of having to make multiple changes to all micro services.

Special notes for your reviewer:


zirain:
/ok-to-test


please fix CLA

jwendell:
@Cam6349 do you still need this? If not, can you close it?

PR #6529, opened 2025-09-03
stats: utilize expiration

kyessenov:
Change-Id: I97c10eac08443566eca6f53dc73b86c2e51fa095

Replacing rotating scopes with the more accurate per-stat expiration implementation https://github.com/envoyproxy/envoy/pull/40395

### DIFF
--- a/source/extensions/filters/http/istio_stats/config.proto
+++ b/source/extensions/filters/http/istio_stats/config.proto
@@ -114,9 +114,11 @@ message PluginConfig {
 
   // Metric scope rotation interval. Set to 0 to disable the metric scope rotation.
   // Defaults to 0.
+  // DEPRECATED.
   google.protobuf.Duration rotation_interval = 11;
 
   // Metric expiry graceful deletion interval. No-op if the metric rotation is disabled.
   // Defaults to 5m. Must be >=1s.
+  // DEPRECATED.
   google.protobuf.Duration graceful_deletion_interval = 12;
 }
--- a/go.mod
+++ b/go.mod
@@ -5,12 +5,12 @@ go 1.24
 require (
 	github.com/cncf/xds/go v0.0.0-20250501225837-2ac532fd4443
 	github.com/envoyproxy/go-control-plane v0.13.5-0.20250816044120-872f08a0dcd7
-	github.com/envoyproxy/go-control-plane/envoy v1.32.4
+	github.com/envoyproxy/go-control-plane/envoy v1.32.5-0.20250902220953-4a410393a630
 	github.com/golang/protobuf v1.5.4
 	github.com/google/go-cmp v0.7.0
-	github.com/prometheus/client_model v0.6.1
+	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.46.0
-	go.opentelemetry.io/proto/otlp v1.1.0
+	go.opentelemetry.io/proto/otlp v1.7.1
 	go.starlark.net v0.0.0-20240123142251-f86470692795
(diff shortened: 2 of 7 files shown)


howardjohn (commented):
qq - the envoy PR seems to suggest it only works with otlp (or other delta protocols) while I think the old istio approach worked with Prometheus as well. Is the Prometheus behavior changed by this pr?

kyessenov:
The old behavior didn't work with prometheus either AFAICT, so it didn't change. The behavior is the same if you create all stats at once, and then wait some time.

howardjohn:
I believe it did which is why there was the timer - to make sure it was sufficiently large to ensure Prometheus had time to scrape? 

@zirain would know for sure

source/extensions/filters/http/istio_stats/config.proto
```diff
@@ -114,9 +114,11 @@ message PluginConfig {
 
   // Metric scope rotation interval. Set to 0 to disable the metric scope rotation.
   // Defaults to 0.
+  // DEPRECATED.
```
zirain, line 117: just for note, we need to remove all related stuffs(e.g. the env flags) from istio since it won't work anymore.

zirain:
> qq - the envoy PR seems to suggest it only works with otlp (or other delta protocols) while I think the old istio approach worked with Prometheus as well. Is the Prometheus behavior changed by this pr?

IIUC, this worked at the underlying mechanism of envoy. which means it works for both metric service and /stats/prometheus endpoint.

> The old behavior didn't work with prometheus either AFAICT, so it didn't change. The behavior is the same if you create all stats at once, and then wait some time.

this new behavior will work if you enable `stats_eviction` in bootstrap, is that right?

source/extensions/filters/http/istio_stats/config.proto
kyessenov: Yeah, that's a follow up. There's a difference in configuration since this is now a per-Envoy config, so maybe a pod annotation rather than telemetry API.

kyessenov:
> > qq - the envoy PR seems to suggest it only works with otlp (or other delta protocols) while I think the old istio approach worked with Prometheus as well. Is the Prometheus behavior changed by this pr?
> 
> IIUC, this worked at the underlying mechanism of envoy. which means it works for both metric service and /stats/prometheus endpoint.

It worked in the sense that prometheus would handle disappearing stats, even though that's not normal. The difference is that before a stat would disappear all the time, but now only when it's not being emitted.

> 
> > The old behavior didn't work with prometheus either AFAICT, so it didn't change. The behavior is the same if you create all stats at once, and then wait some time.
> 
> this new behavior will work if you enable `stats_eviction` in bootstrap, is that right?

Yes, it requires that field to turn on. It does nothing without it.

source/extensions/filters/http/istio_stats/config.proto
zirain: or just mentioned in custom bootstrap.

ringerc:
> this new behavior will work if you enable stats_eviction in bootstrap, is that right?

It sounds like that's what https://github.com/istio/api/pull/3562 and https://github.com/istio/istio/pull/57736 do, right?

zirain:
> > this new behavior will work if you enable stats_eviction in bootstrap, is that right?
> 
> It sounds like that's what [istio/api#3562](https://github.com/istio/api/pull/3562) and [istio/istio#57736](https://github.com/istio/istio/pull/57736) do, right?

bingo

ringerc:
Docs update in https://github.com/istio/istio.io/pull/17043

PR #6576, opened 2025-09-30
Fix the go toolchain version

jwendell:
It needs to be `x.y.z`.


jwendell:
Follows up https://github.com/istio/proxy/pull/5836

keithmattix (approved):

PR #6584, opened 2025-10-02
fix unknown workload type

kyessenov:
Change-Id: Iad070593f253333a1528d1537a5cebbf2d817841

https://github.com/istio/proxy/pull/6554 introduced a subtle bug where a property was printing something before but doesn't know. This broke some extensions on our side.

CC @jwendell


keithmattix (approved):

zirain:
can we add a test to guarantee this?

PR #6955, opened 2026-04-10
Enable dynamic modules envoy filter

freedomljc:
Similar to https://github.com/istio/proxy/pull/6919, enable dynamic modules envoy filter based on https://github.com/envoyproxy/envoy/blob/main/source/extensions/extensions_build_config.bzl

The change registers Envoy dynamic module extensions across the following categories:                     
   ```
  ┌─────────┬──────────────────────────────────────┐   
  │ Categor │              Extension               │ 
  │    y    │                                      │   
  ├─────────┼──────────────────────────────────────┤ 
  │ Access  │ envoy.access_loggers.dynamic_modules │ 
  │ loggers │                                      │ 
  ├─────────┼──────────────────────────────────────┤ 
  │ Cluster │ envoy.clusters.dynamic_modules       │ 
  │ s       │                                      │
  ├─────────┼──────────────────────────────────────┤   
  │ Bootstr │ envoy.bootstrap.dynamic_modules      │
  │ ap      │                                      │   
  ├─────────┼──────────────────────────────────────┤
  │ Input m │ envoy.matching.matchers.dynamic_modu │
  │ atchers │ les                                  │
  ├─────────┼──────────────────────────────────────┤   
  │ Matchin │ envoy.matching.inputs.dynamic_module │
  │ g       │ _data_input                          │   
  │ inputs  │                                      │
  ├─────────┼──────────────────────────────────────┤
  │ HTTP    │ envoy.filters.http.dynamic_modules   │
  │ filters │                                      │
  ├─────────┼──────────────────────────────────────┤
  │ Listene │ envoy.filters.listener.dynamic_modul │
  │ r       │ es                                   │
  │ filters │                                      │
  ├─────────┼──────────────────────────────────────┤
  │ Network │ envoy.filters.network.dynamic_module │
  │         │ s                                    │   
  │ filters │                                      │
  ├─────────┼──────────────────────────────────────┤   
  │ UDP     │ envoy.filters.udp_listener.dynamic_m │
  │ filters │ odules                               │
  ├─────────┼──────────────────────────────────────┤
  │ Tracers │ envoy.tracers.dynamic_modules        │
  ├─────────┼──────────────────────────────────────┤
  │ HTTP up │ envoy.upstreams.http.dynamic_modules │
  │ streams │                                      │   
  ├─────────┼──────────────────────────────────────┤
  │ TLS     │                                      │   
  │ cert    │ envoy.tls.cert_validator.dynamic_mod │
  │ validat │ ules                                 │
  │ ors     │                                      │
  ├─────────┼──────────────────────────────────────┤
  │ Load ba │                                      │
  │ lancing │ envoy.load_balancing_policies.dynami │   
  │  polici │ c_modules                            │
  │ es      │                                      │   
  └─────────┴──────────────────────────────────────┘
```


krinkinmu:
@freedomljc I presume the suggestion is that this will be configued via `EnvoyFilter`?

freedomljc:
> @freedomljc I presume the suggestion is that this will be configued via `EnvoyFilter`?

I'm leaning to the approach A mentioned in https://github.com/istio/istio/issues/59839#issuecomment-4257515442 - Incrementally add the `Dynamic Modules` support into Istio: To start with, I'd recommend the `EnvoyFilter` +  volume-mounted `.so` files

sschepens:
@zirain wdyt? I'm on-board with this, was actually thinking of opening up a PR myself, seeing pretty big WASM performance hits and want to migrate to dynamic modules.

Could we also get this into 1.30 before it releases?

krinkinmu:
@sschepens @zirain I'd like to approve the PR, but before that I wanted to followup with the security folks to see if they'd like to take a closer look before we starting building proxy with dynamic modules enabled. As far as upstream Envoy is concerned, dynamic modules are getting stable ABI in 1.38, or at least that was the plan, so the feature is definitely getting more mature.

@freedomljc regardless of how this PR goes, I think it makes sense to bring discussion about dynamic modules in Istio into a WG meeting just for visibility.

sschepens:
@krinkinmu I'm inclined to think that adding dynamic_modules support shouldn't be any more different than any other Envoy extension being enabled, but I think talking with security and bringing it up in WG are good ideas.

And yes, Envoy 1.38 is getting a stable ABI and it's the Envoy version we'll be using in 1.30, right? so this could probably be added to 1.30 release as well.

krinkinmu:
@sschepens dynamic modules might not be different in terms of implementing support for those in Istio (though, that beiung said, dynamic modules support very different types of extensions compared to proxy-wasm), but in terms of how dynamic modules work - they are nothing like WASM extensions.

One obvious difference is that WASM extensions operate inside a WASM sandbox (unless you opt into using nullvm) - dynamic modules do not provide that level of memory isolation. That's why I think it makes sense to for security folks to at least take a look at this before we merge it.

I'm not advocating for or against dynamic modules here, just highlighing the fact that isolation properties of WASM extensions and dynamic modules are quite different.

sschepens:
@krinkinmu yes totally agree, of course they are very different from WASM, and people should be warned about that, and even if we enable them, we should probably hold-off on adding an official API for them.

Let's get this to the security folk and in a WG :)

zirain:
@krinkinmu I had replid in the slack, I'm +1 on this.

But I also want to see a clear path about how Istio support in the future, will a new CRD be introuduced in the next?

krinkinmu:
> @krinkinmu I had replid in the slack, I'm +1 on this.
> 
> But I also want to see a clear path about how Istio support in the future, will a new CRD be introuduced in the next?

@zirain I think this will be one of the topics of discussion in the WG meeting today, so let's wait until then.

krinkinmu (approved):

krinkinmu:
@zirain filling you in on the discussion in the WG meeting:

TL;DR:
* We've got a green light for this PR specifically.
* We don't necessarily want to go the way of `EnvoyFilter` to configure dynamic modules, so some followup discussion will be needed
* We don't claim any support for the dynamic modules in Istio at the moment either
* There were some concerns about security of the dynamic modules, but nothing that should block this PR or to prevent us from exploring this path.

All-in-all, we can move forward with dynamic modules carefully.

jaellio (approved):

PR #6971, opened 2026-04-15
fix cel-cpp bazel name for envoy release/v1.37

PetrMc:
Envoy release/v1.37 uses com_google_cel_cpp, not cel-cpp (renamed on main). Fix BUILD to match the release branch naming. Ref https://github.com/istio/istio/issues/59425


zirain:
isn't 1.30 target to envoy 1.38(if branch cut not happen, it should be main)?

keithmattix (dismissed):

PetrMc:
> isn't 1.30 target to envoy 1.38(if branch cut not happen, it should be main)?

thanks @zirain we cut Istio 1.30 alpha with Envoy 1.37 as 1.38 wasn't available yet but it's going back from what master has.

PetrMc:
solving with `envoy/main` for now https://github.com/istio/test-infra/pull/5886#event-24552795714

PR #6997, opened 2026-04-28
Fix release binary script

Stevenjin8:

Same as https://github.com/istio/istio/pull/60020

### DIFF
--- a/scripts/release-binary.sh
+++ b/scripts/release-binary.sh
@@ -175,13 +175,16 @@ do
     gsutil cp "${BINARY_NAME}" "${SHA256_NAME}" "${DWP_NAME}" "${DST}/"
 
     R2_DST="${DST/gs:\/\//s3:\/\/}"
-    R2_ENDPOINT="https://${CF_ACCOUNT_ID}.r2.cloudflarestorage.com"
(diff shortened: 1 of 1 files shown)


scripts/release-binary.sh
```diff
@@ -175,13 +175,16 @@ do
     gsutil cp "${BINARY_NAME}" "${SHA256_NAME}" "${DWP_NAME}" "${DST}/"
 
     R2_DST="${DST/gs:\/\//s3:\/\/}"
-    R2_ENDPOINT="https://${CF_ACCOUNT_ID}.r2.cloudflarestorage.com"
+    ENDPOINT="$(echo "${CF_CREDENTIALS}" | jq -r '.endpoint' | tr -d '\n')"
```
krinkinmu, line 178: For my understanding where the `CF_CREDENTIALS` is coming from and do we already have the file with credentials there?

Stevenjin8: We define the requirement here:

https://github.com/istio/test-infra/blob/bd456c1a6fef8733912344b95c2b783a124b9a09/prow/config/jobs/.base.yaml#L135-L142

and we attach it to the job here

https://github.com/istio/test-infra/blob/bd456c1a6fef8733912344b95c2b783a124b9a09/prow/config/jobs/proxy.yaml#L57-L62

krinkinmu (approved):

PR #7047, opened 2026-05-26
Bump Envoy

jwendell:


envoy.bazelrc
```diff
@@ -235,8 +235,9 @@ common:aws-lc-fips --@envoy//bazel:crypto=@aws_lc//:crypto
 common:aws-lc-fips --@envoy//bazel:http3=False
 
 # OpenSSL
-common:openssl --//bazel:ssl=//compat/openssl:ssl
-common:openssl --//bazel:crypto=//compat/openssl:crypto
+common:openssl --@envoy//bazel:openssl=True
```
PetrMc, line 238: is there any change in the default behavoir because of this?

PetrMc (commented):
just worrying of any impact of ssl flag. 
is this also required in `master` ?

envoy.bazelrc
jwendell: Istio doesn't use this flag. Master already has this. Note that this update will happen automatically by the bot tomorrow. I didn't add anything special. I'm just doing this manually because we'd like to try this in 1.30 ASAP.

PetrMc: The automator just updates shas not changing flags isnt it?

PetrMc (approved):

envoy.bazelrc
jwendell: Automator also updates envoy.bazelrc (it's a copy of Envoy's .bazelrc)

jwendell: See https://github.com/istio/proxy/blob/release-1.30/scripts/update_envoy.sh#L63 - This is the script automator runs. I ran it manually to generate this PR.

PR #7083, opened 2026-06-07
fix: handle Bazel load statements in extension-check

immanuwell:
What this PR does / why we need it:
`tools/extension-check` blows up on normal Bazel config files because the embedded Starlark runner does not handle `load(...)`.

This strips top level `load(...)` statements before evaluation, returns regular errors instead of panics, and removes the README workaround. Pretty small fix, but it makes the tool usable out of the box.

Which issue this PR fixes:
N/A, this repo has issues disabled.

Special notes for your reviewer:
Repro before this patch:
`go run ./tools/extension-check --envoy-extensions-build-config ./bazel/extension_config/extensions_build_config.bzl --proxy-extensions-build-config ./bazel/extension_config/extensions_build_config.bzl --ignore-extensions ./tools/extension-check/wellknown-extensions`

Before:
`panic: load not implemented by this application`

After:
it loads the checked-in file and reaches the normal missing-extension output. so yeah, this is a real bug on a real file, not some weird edge case.

Checks:
`go test ./tools/extension-check`
`go test ./... -run '^$'`

### DIFF
--- a/tools/extension-check/main.go
+++ b/tools/extension-check/main.go
@@ -35,13 +35,13 @@ func main() {
 	flag.Parse()
 	envoyCoreExtensions, err := extensions(*envoyExtensionsBuildConfig, "EXTENSIONS")
 	if err != nil {
-		fmt.Printf("error: %v", err)
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
 		os.Exit(1)
 	}
 
 	proxyExtensions, err := extensions(*proxyExtensionsBuildConfig, "ENVOY_EXTENSIONS")
 	if err != nil {
-		fmt.Printf("error: %v", err)
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
 		os.Exit(1)
 	}
 
@@ -79,13 +79,18 @@ func main() {
 // The file is expected to be a starlark file that defines a global variable.
 // Depends on go.starlark.net/starlark.
 func extensions(filename, key string) (map[string]string, error) {
+	src, err := os.ReadFile(filename)
+	if err != nil {
+		return nil, err
+	}
+
 	thread := &starlark.Thread{
 		Name:  "extensions",
 		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
 	}
-	globals, err := starlark.ExecFile(thread, filename, nil, nil)
+	globals, err := starlark.ExecFile(thread, filename, stripLoadStatements(string(src)), nil)
 	if err != nil {
-		panic(err)
+		return nil, err
 	}
 
 	if v, ok := globals[key]; ok {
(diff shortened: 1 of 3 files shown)


jwendell (commented):
Since this PR was raised we removed the `load()` from our .bzl file. If you still want to pursue this, here are some review.

tools/extension-check/main.go
```diff
@@ -79,13 +79,18 @@ func main() {
...
+		return nil, err
+	}
+
 	thread := &starlark.Thread{
 		Name:  "extensions",
 		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
 	}
-	globals, err := starlark.ExecFile(thread, filename, nil, nil)
+	globals, err := starlark.ExecFile(thread, filename, stripLoadStatements(string(src)), nil)
 	if err != nil {
```
jwendell, line 92: ⚠️ **Merge conflict / API regression**

This PR uses `starlark.ExecFile` but master already migrated to `starlark.ExecFileOptions` in commit `1c3f7912`. The PR is in a CONFLICTING state and needs a rebase.

The correct resolution should use the new API:
```go
globals, err := starlark.ExecFileOptions(syntax.LegacyFileOptions(), thread, filename, stripLoadStatements(string(src)), nil)
```

tools/extension-check/main_test.go
```diff
@@ -0,0 +1,41 @@
+package main
```
jwendell, line 1: ⚠️ **Missing copyright banner**

This file is missing the required Apache 2.0 copyright header. `make lint` (`lint-copyright-banner`) will fail — it scans all `*.go` files for `Apache License, Version 2` and `Copyright`. See `main.go` for the standard 13-line header.

tools/extension-check/main.go
```diff
@@ -100,3 +105,33 @@ func extensions(filename, key string) (map[string]string, error) {
...
+		}
+
+		if skippingLoad {
+			parenDepth += strings.Count(line, "(")
+			parenDepth -= strings.Count(line, ")")
+			if parenDepth <= 0 {
+				skippingLoad = false
+				parenDepth = 0
+			}
+			continue
```
jwendell, line 130: 💡 **Line-number shift in error messages**

`stripLoadStatements` removes load-statement lines entirely, which shifts starlark error positions by the number of stripped lines. For example, a 4-line load block followed by an error on line 10 would be reported at line 6.

Consider replacing removed lines with empty strings instead of skipping them:
```go
// Instead of: continue
filtered = append(filtered, "")
```

tools/extension-check/main.go
```diff
@@ -100,3 +105,33 @@ func extensions(filename, key string) (map[string]string, error) {
 
 	return nil, errors.New("no extensions found")
 }
+
+// stripLoadStatements removes Bazel load() statements so build config files can
+// be evaluated without resolving external repositories.
+func stripLoadStatements(src string) string {
+	lines := strings.Split(src, "\n")
+	filtered := make([]string, 0, len(lines))
+	skippingLoad := false
```
jwendell, line 114: 💡 **Redundant state variable**

`skippingLoad` is always equivalent to `parenDepth > 0` at iteration boundaries. You could collapse both into just `parenDepth`, simplifying the logic and avoiding potential desync:
```go
if parenDepth == 0 && strings.HasPrefix(trimmed, "load(") {
    // enter skipping mode via parenDepth alone
}
```

tools/extension-check/main.go
```diff
@@ -100,3 +105,33 @@ func extensions(filename, key string) (map[string]string, error) {
 
 	return nil, errors.New("no extensions found")
 }
+
+// stripLoadStatements removes Bazel load() statements so build config files can
+// be evaluated without resolving external repositories.
+func stripLoadStatements(src string) string {
```
jwendell, line 111: 💭 **Consider AST-level filtering**

Text-level line removal is fragile against edge cases (parentheses in string literals, `load()`-like text in triple-quoted strings). The starlark library provides `syntax.FileOptions.Parse()` and `syntax.LoadStmt` — you could parse the file, filter out `LoadStmt` nodes from `File.Stmts`, and compile with `starlark.FileProgram`. This would handle all formatting edge cases correctly by construction and avoid the line-number shift issue above.

Not a blocker since the current approach works for the known `.bzl` files, but worth considering for robustness.

PR #7133, opened 2026-06-29
Update the Envoy's VERSION.txt when updating Envoy

jwendell:
So it's easier to track the Envoy version directly from this repo without having to dig into the commit SHA in WORKSPACE.


zirain:
you can get it from workspace, why this's a must?

jwendell:
@zirain I tried to explain in the commit message. This is not a must. It's a nice to have thing. To make our lives easier. Convenience e.g., non developers (people not familiar with the mechanism necessary to infer the proxy version from the commit).

krinkinmu:
I'm fine with having this for informational purposes, with the caveat that Envoy version is a bit coarse (e.g., it does not point to a specific commit, and the same vx.y.z-dev covers a range of possible commits in the upstream Envoy repo).

krinkinmu (approved):

zirain (approved):

PR #7182, opened 2026-07-28
Fix build

jwendell:
Replaces https://github.com/istio/proxy/pull/7180 and https://github.com/istio/proxy/pull/7179


jwendell:
cc @istio/wg-policies-and-telemetry-maintainers This is ready for review.

scripts/release-binary.sh
```diff
@@ -79,11 +79,11 @@ if [ "${DST}" == "none" ]; then
   DST=""
 fi
 
-# Make sure the release binaries are built on x86_64 Ubuntu 18.04 (Bionic)
```
krinkinmu, line 82: nit: it seems like this check lost its meaning with hermetic LLVM toolchain in Envoy, since we are not linking against the host glibc anymore, maybe drop it all together?

krinkinmu (approved):

PR #7189, opened 2026-07-29
Replace host OS check with post-build glibc version verification

jwendell:
The Ubuntu Noble check was a proxy for glibc compatibility that lost its meaning with the hermetic LLVM toolchain and sysroot. Replace it with a direct readelf-based check that verifies the built binary doesn't require a newer glibc than the sysroot provides (currently 2.28).


krinkinmu (approved):
Perfect, thank you!

PR #7199, opened 2026-08-03
Use remote cache

Stevenjin8:
This allows us to use a remote cache that is just a storage/cache endpoint by setting `BAZEL_BUILD_RBE_CACHE` to the url of the cache and `BAZEL_BUILD_RBE_INSTANCE=""`


Stevenjin8:
waiting for https://github.com/istio/test-infra/pull/5993

krinkinmu (approved):
For my understanding, we've been able to use remote cache, do I understand correctly that what this PR does is it allows to use remote cache without RBE?

PR #7210, opened 2026-08-05
Enable kill_request filter in the proxy build

jwendell:
The `kill_request` extension has `enabled_default = False` in its Envoy BUILD file, so listing it in `extensions_build_config.bzl` alone is not enough — it compiles as an empty target. Add the Bazel flag to `.bazelrc` so the filter is actually included in the binary.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -45,6 +45,9 @@ build:macos --define tcmalloc=disabled
 # Build with embedded V8-based WebAssembly runtime.
 build --define wasm=v8
 
+# Enable kill_request filter (disabled by default in Envoy via enabled_default = False).
+build --@envoy//source/extensions/filters/http/kill_request:enabled
+
 # Build Proxy-WASM plugins as native extensions.
 build --copt -DNULL_PLUGIN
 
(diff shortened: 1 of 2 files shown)


jwendell:
The issue here is: We think we are building with this extension, since it's in the `extensions_build_config.bzl` file, but we need this flag in addition. This PR fixes this mistake. Alternatively we could simply remove that extension, but I think the original intention was to include it.

zirain:
@kyessenov can you recall why it's disabled in the past?

jwendell:
That extension was added in Envoy disabled by default: https://github.com/envoyproxy/envoy/pull/14170 - A new mechanism to enable it was added after that: https://github.com/envoyproxy/envoy/pull/14240

In istio/proxy this extension was always included in our configuration since day 0, but it's not built by default unless a build flag is specified.

This PR adds this build flag.

My intention is to avoid confusion when reading the `extensions_build_config.bzl` file. One might think this extension is enabled when in fact it isn't.

Given it was never enabled in Istio, I think the best approach then is just remove it from `extensions_build_config.bzl`.

WDYT?

PR #7282, opened 2026-09-06
ci: use bazel mod

zirain:
xref: https://github.com/envoyproxy/envoy/pull/47217


zirain:
@krinkinmu @jwendell would like take review on this?

jwendell (commented):
Nice migration -- the `envoy_api` re-declaration and the `override_repo` for `envoy_build_config` are both non-obvious and well documented. Left inline notes; the two I'd really like resolved before merge are the `--override_repository` in `prow/proxy-common.inc` (Envoy-fork CI now builds against upstream API protos) and the lockfile handling in `update_envoy.sh`, since that runs unattended every day.

prow/proxy-common.inc
```diff
@@ -40,13 +40,13 @@ if [[ "${ENVOY_REPOSITORY:-}" && "${ENVOY_PREFIX:-}" ]]; then
   # The other path is preferred as it uses an API intended for programmatic access and allows fine-grained access token usage.
   TMP_DIR=$(mktemp -d -t envoy-XXXXXXXXXX)
   trap 'rm -rf ${TMP_DIR:?}' EXIT
-  ENVOY_SHA="${ENVOY_SHA:-$(grep -Pom1 "^ENVOY_SHA = \"\K[a-zA-Z0-9]{40}" "$WORKSPACE")}"
+  ENVOY_SHA="${ENVOY_SHA:-$(grep -Pom1 "^ENVOY_SHA = \"\K[a-zA-Z0-9]{40}" "$MODULE_BAZEL")}"
   BAZEL_BUILD_ARGS="${BAZEL_BUILD_ARGS} --override_repository=envoy=${TMP_DIR}/${ENVOY_PREFIX}-${ENVOY_SHA}"
   curl -nsSfL "${ENVOY_REPOSITORY}/archive/${ENVOY_SHA}.tar.gz" | tar -C "${TMP_DIR}" -xz
```
jwendell, line 51: Under WORKSPACE, `@envoy_api` was *derived* from `@envoy` by `envoy_api_binding()`, so this single override swapped both the Envoy source and its API protos. Now `envoy_api` is an independent module with its own `archive_override` (`MODULE.bazel:96-102`), which `--override_repository=envoy=` doesn't touch. Any `ENVOY_REPOSITORY`-driven job will build fork Envoy source against upstream `envoyproxy/envoy@ENVOY_SHA` protos -- a compile failure, or worse, silently wrong protos if the fork touches `api/`.

Separately, `--override_repository` is the wrong flag for a bzlmod *module*: module resolution still reads the pinned archive's `MODULE.bazel`, so the local checkout's dependency declarations are ignored. `--override_module` is the documented one.

Suggested for both branches (and for the local-dev instructions at `MODULE.bazel:78-80`):

```
BAZEL_BUILD_ARGS="${BAZEL_BUILD_ARGS} --override_module=envoy=${DIR} --override_module=envoy_api=${DIR}/api"
```

scripts/update_envoy.sh
```diff
@@ -52,15 +52,20 @@ SHA256=${SHAArr[0]}
...
 # Update ENVOY_SHA commit date
-sed -i "s/Commit date: .*/Commit date: ${DATE}/" "${WORKSPACE}"
+sed -i "s/Commit date: .*/Commit date: ${DATE}/" "${MODULE_BAZEL}"
 
-# Update the dependency in istio/proxy WORKSPACE
-sed -i 's/ENVOY_SHA = .*/ENVOY_SHA = "'"$LATEST_SHA"'"/' "${WORKSPACE}"
-sed -i 's/ENVOY_SHA256 = .*/ENVOY_SHA256 = "'"$SHA256"'"/' "${WORKSPACE}"
+# Update the dependency in istio/proxy MODULE.bazel
+sed -i 's/ENVOY_SHA = .*/ENVOY_SHA = "'"$LATEST_SHA"'"/' "${MODULE_BAZEL}"
+sed -i 's/ENVOY_SHA256 = .*/ENVOY_SHA256 = "'"$SHA256"'"/' "${MODULE_BAZEL}"
```
jwendell, line 59: `MODULE.bazel.lock` isn't regenerated here. It records `bzlTransitiveDigest`/`usagesDigest` for four Envoy-provided extensions (`@@envoy+//bazel:extensions.bzl%envoy_build_config_ext`, `%envoy_repo_extension`, `%envoy_toolchains_extension`, `%envoy_module_graph_extension`), all of which an Envoy SHA bump invalidates. Two consequences:

- `bazel/bazel_get_workspace_status:28` derives `BUILD_SCM_STATUS` from `git diff-index --quiet HEAD`. Bazel refreshes the lockfile during loading (default `--lockfile_mode=update`), i.e. *before* the workspace status command runs, so released binaries stamp `BUILD_SCM_STATUS=Modified` instead of `Clean`. This is a new regression -- there was no committed lock pre-bzlmod.
- Any job passing `--lockfile_mode=error` via `BAZEL_BUILD_EXTRA_ARGS` (`prow/proxy-common.inc:34`) will hard-fail on every bot PR.

Suggest adding a `bazel mod deps --lockfile_mode=update` step so the bot commits a refreshed lock.

scripts/update_envoy.sh
```diff
@@ -52,15 +52,20 @@ SHA256=${SHAArr[0]}
...
 curl -sSL "https://raw.githubusercontent.com/${ENVOY_ORG}/${ENVOY_REPO}/${LATEST_SHA}/.bazelversion" > .bazelversion
 curl -sSL "https://raw.githubusercontent.com/${ENVOY_ORG}/${ENVOY_REPO}/${LATEST_SHA}/.bazelrc" > envoy.bazelrc
 
 # Update VERSION.txt
 curl -sSL "https://raw.githubusercontent.com/${ENVOY_ORG}/${ENVOY_REPO}/${LATEST_SHA}/VERSION.txt" > ENVOY_VERSION.txt
+
+# Keep MODULE.bazel's `bazel_dep(name = "envoy", ...)` version aligned with
+# the version Envoy itself declares in its own MODULE.bazel (== VERSION.txt).
+ENVOY_MODULE_VERSION="$(cat ENVOY_VERSION.txt)"
+sed -i 's/bazel_dep(name = "envoy", version = ".*")/bazel_dep(name = "envoy", version = "'"${ENVOY_MODULE_VERSION}"'")/' "${MODULE_BAZEL}"
```
jwendell: This pattern can't match `bazel_dep(name = "envoy_api", version = ...)` at `MODULE.bazel:96` -- the literal `"envoy",` excludes it. After the next minor bump, line 81 will read e.g. `1.41.0-dev` while line 96 stays `1.40.0-dev`, even though both are pinned to the same archive and upstream keeps them in lockstep (`envoy@329dd51:MODULE.bazel:33` declares `envoy_api` at `1.40.0-dev`). Either the declared version matters, and the bot silently emits an inconsistent module file, or it doesn't, and this line is dead code that contradicts the comments at `:75-76` and `:95`.

```suggestion
sed -i -E 's/bazel_dep\(name = "(envoy|envoy_api)", version = "[^"]*"\)/bazel_dep(name = "\1", version = "'"${ENVOY_MODULE_VERSION}"'")/' "${MODULE_BAZEL}"
```

scripts/update_envoy.sh
```diff
@@ -52,15 +52,20 @@ SHA256=${SHAArr[0]}
...
 # Update .bazelversion and envoy.bazelrc
 curl -sSL "https://raw.githubusercontent.com/${ENVOY_ORG}/${ENVOY_REPO}/${LATEST_SHA}/.bazelversion" > .bazelversion
 curl -sSL "https://raw.githubusercontent.com/${ENVOY_ORG}/${ENVOY_REPO}/${LATEST_SHA}/.bazelrc" > envoy.bazelrc
 
 # Update VERSION.txt
 curl -sSL "https://raw.githubusercontent.com/${ENVOY_ORG}/${ENVOY_REPO}/${LATEST_SHA}/VERSION.txt" > ENVOY_VERSION.txt
+
+# Keep MODULE.bazel's `bazel_dep(name = "envoy", ...)` version aligned with
+# the version Envoy itself declares in its own MODULE.bazel (== VERSION.txt).
+ENVOY_MODULE_VERSION="$(cat ENVOY_VERSION.txt)"
```
jwendell: `curl -sSL` without `-f` exits 0 on a 404/5xx and writes the error body into `ENVOY_VERSION.txt`; line 71 then substitutes that body straight into `MODULE.bazel`. Previously a bad `ENVOY_VERSION.txt` was only cosmetic; now you get `version = "404: Not Found"`, breaking every build until someone reverts. A multi-line body is worse -- it makes the `sed` expression itself invalid (`unterminated 's' command`), and under `set -e` the script aborts partway through, leaving `MODULE.bazel` half-updated (new SHA/SHA256, stale version). Compare `prow/proxy-common.inc:45`, which correctly uses `curl -nsSfL`.

Suggest adding `-f` to these curls plus a sanity check before substituting:

```
grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+(-dev)?$' ENVOY_VERSION.txt || exit 1
```

scripts/release-binary.sh
```diff
@@ -74,8 +74,7 @@ if [ "${DST}" == "none" ]; then
   DST=""
 fi
 
-# Expected glibc version from the hermetic sysroot configured in WORKSPACE.
-EXPECTED_GLIBC=$(grep -oP 'glibc_version\s*=\s*"\K[^"]+' WORKSPACE)
+EXPECTED_GLIBC="2.28"
```
jwendell: This used to be derived from the single source of truth (`grep glibc_version WORKSPACE`). It's now a literal living in a different file from the actual pin at `MODULE.bazel:172`. Raise the sysroot pin and releases fail spuriously; lower it and the check happily passes binaries below the claimed floor -- which is exactly the drift this guard exists to catch. The value is still declaratively greppable:

```suggestion
EXPECTED_GLIBC="$(grep -oP 'glibc_version = "\K[^"]+' "${ROOT}/MODULE.bazel")"
```

.bazelrc
```diff
@@ -6,6 +6,17 @@
...
+# envoy_build_config (our //bazel:extension_config) is produced by an
+# Envoy-internal module extension that cannot be parameterized from
+# MODULE.bazel; override it the way Envoy's own docs recommend
+# (bazel/README.md: "Customize extension build config"). --override_repository
+# keys off the *canonical* repo name for an extension-generated repo (find it
+# with `bazel mod show_repo @envoy_build_config`), not the plain apparent
+# name -- the apparent name "envoy_build_config" is only what our own
+# MODULE.bazel's `use_repo(envoy_build_config_ext, "envoy_build_config")`
+# maps to it locally.
+common --override_repository=envoy++envoy_build_config_ext+envoy_build_config=%workspace%/bazel/extension_config
```
jwendell: `MODULE.bazel:136-139` already redirects `@envoy_build_config` via the supported `override_repo(envoy_build_config_ext, envoy_build_config = "envoy_build_config")`. Once that's in effect, nothing resolves to the extension-generated canonical repo `envoy++envoy_build_config_ext+envoy_build_config`, so this flag is a no-op -- while the comment above it asserts it's the mechanism that makes extension curation work.

That's a trap: someone removing the `override_repo` block (or Bazel renaming extension canonical repos) would believe this line still guards them, and the fallback is Envoy's `default_envoy_build_config`, i.e. a silently wrong extension set -- exactly the failure mode the comment warns about. Suggest deleting lines 9-18 and keeping only the `override_repo`.

scripts/update_envoy.sh
jwendell: Thanks -- the `bazel mod deps --lockfile_mode=update` step is the right fix, and I confirmed `bazel mod deps` is valid with no arguments (exits 0 on a scratch module).

The lockfile still needs regenerating **on this branch**, though: `MODULE.bazel.lock` is unchanged across `8f017cf9..a915abb2`, but those commits bumped `envoy_toolshed` to `0.4.14.envoy` (lock still records `0.4.13.envoy`), moved the bazel-registry pin to `dd48349c...` (lock still references only `b6e15c00...`), and bumped `ENVOY_SHA` to `e804587c...` (the four `@@envoy+` extension digests are unrefreshed).

So as it stands the PR merges in exactly the state this thread describes: Bazel refreshes the lock during loading under the default `--lockfile_mode=update`, the tree is dirty by the time the workspace status command runs, and released binaries get stamped `BUILD_SCM_STATUS=Modified`. Any job passing `--lockfile_mode=error` fails outright instead.

Could you run `bazel mod deps --lockfile_mode=update` and commit the resulting `MODULE.bazel.lock`? The manual bumps in the last two commits didn't go through the newly fixed script.

MODULE.bazel
```diff
@@ -0,0 +1,185 @@
...
+# sysroots under a distinct `name_prefix` (avoiding a repo-name collision
+# with envoy's own default-named ones), then `override_repo` envoy's
+# "sysroot_linux_amd64"/"sysroot_linux_arm64" to resolve to ours -- so
+# envoy's existing `llvm.sysroot(label = "@sysroot_linux_amd64//:sysroot")`
+# wiring picks up our pinned sysroot transparently, with no need to
+# duplicate envoy's llvm.toolchain/toolchain_root/cxx_cross_lib config.
+#
+# Note the real package is "sysroot", not "bazel/sysroot" as the
+# envoy_toolshed README (as of 0.4.13) still documents -- confirmed via
+# `bazel query "@sysroot_linux_amd64//:sysroot"` against the fetched repo.
```
jwendell: This is already stale -- line 56 is now `0.4.14.envoy`. Since it'''s a claim about an external README pinned to a specific version, it'''ll just go stale again on the next toolshed bump; suggest dropping it and keeping the code as the source of truth.

```suggestion
# duplicate envoy'''s llvm.toolchain/toolchain_root/cxx_cross_lib config.
```

scripts/update_envoy.sh
zirain: not sure what happen, but the command was stuck on my local compute.
Feel free to commit these change to my branch.

jwendell (approved):
Thank you!

krinkinmu (approved):

PR #4462, opened 2023-02-23
wasm: remove stats and attributegen from the tree

kyessenov:
After the switch to the native stats, we no longer need to maintain the v8 version of the same extension. However, we should continue exercising Wasm-related code in Envoy, to make sure ABI is backwards compatible.

This PR deletes the code for Wasm versions of attributegen and stats from the tree. It preserves the tests for those modules fixed at the version of the last SHA artifacts. We also enable TSan and Asan for more of the Wasm-related tests to gain more confidence in Wasm filter in Envoy.

PR #7300, opened 2026-09-21
test: check in attributegen.wasm instead of downloading it

jwendell:
`TestAttributeGen` relied on `env.EnsureWasmFiles` to fetch `attributegen.wasm` from storage.googleapis.com at test time, which breaks offline builds. The attributegen source was removed from this repo in #4462, so the artifact is frozen at a pinned 2023 SHA and is never rebuilt.

Check the module in under testdata/wasm/ (verified identical to the previously downloaded artifact) and drop the download helper.


kyessenov:
Yes, it's under 1MB so totally makes sense to avoid a download.

kyessenov (approved):

